Issue #19, opened 2017-06-05
Anonymous auth should use run-as-user instead of launch-as-user


dposada:
To be consistent with the other parts of the code

PR #20, opened 2017-06-05
Rename launch-as-user to run-as-user

shamsimam:
Fixes issue #19

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -151,7 +151,7 @@
                         :kind :anonymous
                         :anonymous {;; Custom implementations should specify a :factory-fn
                                     ;; that returns an instance of waiter.auth.authentication.Authenticator:
-                                    :factory-fn waiter.auth.authentication/anonymous-authenticator
+                                    :factory-fn waiter.auth.authentication/single-user-authenticator
                                     ;; the user account used to launch services
                                     :run-as-user "launch-username"}
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -153,7 +153,7 @@
                                     ;; that returns an instance of waiter.auth.authentication.Authenticator:
                                     :factory-fn waiter.auth.authentication/anonymous-authenticator
                                     ;; the user account used to launch services
-                                    :launch-as-user "launch-username"}
+                                    :run-as-user "launch-username"}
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/auth/authentication.clj
```diff
@@ -29,10 +29,10 @@
 
 ;; An anonymous request does not contain any authentication information.
 ;; This is equivalent to granting everyone access to the resource.
-;; The anonymous authenticator attaches the principal of launch-as-user to the request.
-;; In particular, this enables requests to launch processes as launch-as-user.
+;; The anonymous authenticator attaches the principal of run-as-user to the request.
+;; In particular, this enables requests to launch processes as run-as-user.
 ;; Use of this authentication mechanism is strongly discouraged for production use.
-(defrecord AnonymousAuthenticator [launch-as-user]
+(defrecord AnonymousAuthenticator [run-as-user]
```
sradack: Can we go with SingleUserAuthenticator?

shamsimam: But it is not authenticating anyone. It is accepting requests from anyone and making them behave as a request from a specific user.

shamsimam:
@sradack renamed to `SingleUserAuthenticator`

waiter/config-full.edn
```diff
@@ -151,9 +151,9 @@
                         :kind :anonymous
```
dposada: Maybe we should match cook and have the `:kind` be `:one-user` (https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/components.clj#L316)

waiter/config-full.edn
```diff
@@ -151,9 +151,9 @@
...
                         :anonymous {;; Custom implementations should specify a :factory-fn
                                     ;; that returns an instance of waiter.auth.authentication.Authenticator:
-                                    :factory-fn waiter.auth.authentication/anonymous-authenticator
+                                    :factory-fn waiter.auth.authentication/single-user-authenticator
```
dposada: Same comment as above, maybe `one-user-authenticator` to match cook

shamsimam:
@sradack any further feedback on this?

sradack:
LGTM

PR #27, opened 2017-06-14
Reduce stdout output while running unit tests

shamsimam:


waiter/test/waiter/test_helpers.clj
```diff
@@ -64,8 +64,7 @@
   (defmethod clojure.test/report :end-test-var [m]
     (let [test-name (full-test-name m)]
       (with-test-out
-        (println \tab (blue "FINISH:") test-name)
-        (println \tab "COUNTS:" (assoc @*report-counters* :name test-name)))
```
dposada: How about changing the "COUNTS" line from `println` to `log/debug`?

shamsimam: The counts are printed on stdout along with the `FINISH`. Do you want additional logging for the counts?

dposada: Looks good, I hadn't noticed the counts on the FINISH line

PR #28, opened 2017-06-14
Use make-inter-router-requests-sync-fn instead of handle-inter-rou…

shamsimam:
…ter-request-fn in token-reindex-handler-fn

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1067,12 +1067,13 @@
                                      (token/handle-refresh-token-request kv-store src-router-id request))
                                    request)))
    :token-reindex-handler-fn (pc/fnk [[:curator kv-store]
-                                      [:routines list-tokens-fn synchronize-fn]
-                                      handle-inter-router-request-fn handle-secure-request-fn]
+                                      [:routines list-tokens-fn make-inter-router-requests-sync-fn synchronize-fn]
+                                      handle-secure-request-fn]
                                (fn token-reindex-handler-fn [request]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/core.clj
```diff
@@ -1067,12 +1067,13 @@
...
    :token-reindex-handler-fn (pc/fnk [[:curator kv-store]
-                                      [:routines list-tokens-fn synchronize-fn]
-                                      handle-inter-router-request-fn handle-secure-request-fn]
+                                      [:routines list-tokens-fn make-inter-router-requests-sync-fn synchronize-fn]
+                                      handle-secure-request-fn]
                                (fn token-reindex-handler-fn [request]
                                  (handle-secure-request-fn
                                     (fn inner-token-handler-fn [request]
-                                      (token/handle-reindex-tokens-request synchronize-fn handle-inter-router-request-fn kv-store list-tokens-fn request))
+                                      (token/handle-reindex-tokens-request synchronize-fn make-inter-router-requests-sync-fn
```
dposada: What's the functional effect of this change?

shamsimam: `make-inter-router-requests-sync-fn` is the crrect function to call, it was a bug in the previous implementation.

dposada: What was the bug? Do we need a test?

dposada: Discussed in person - this re-index is only triggered manually, which is why this bug was not being triggered

PR #38, opened 2017-06-23
WIP: add support for buffering backend HTTP response

shamsimam:


shamsimam:
We ended up handling the buffer in `jet` instead as it gives much better performance.

PR #52, opened 2017-07-05
Read request body in request-info handler

pschorf:


kitchen/src/kitchen/core.clj
```diff
@@ -290,7 +290,9 @@
 
 (defn- request-info-handler
   "Returns the info received in the request."
-  [{:keys [headers] :as request}]
+  [{:keys [body headers] :as request}]
+  (if (and (not (nil? body)) (instance? InputStream body))
```
sradack: Can this line simply be `(instance? InputStream body)`?

pschorf: Done

PR #54, opened 2017-07-06
Remove spnego import from websockets + Fix content-type header issue

shamsimam:


shamsimam:
@pschorf approved it verbally :)

PR #55, opened 2017-07-06
Override permitted user when run as user is passed in headers

pschorf:


waiter/src/waiter/service_description.clj
```diff
@@ -490,6 +490,10 @@
           ; else rely on presence of x-waiter headers to set the run-as-user
           contains-waiter-header? (headers/contains-waiter-header waiter-headers on-the-fly-service-description-keys)
           user-service-description (cond-> sanitized-metadata-description
+                                           (and (contains? (:headers sources) "run-as-user")
```
shamsimam: Can we move this clause below and `or` with the `(and (not (contains? sanitized-metadata-description "permitted-user")) contains-waiter-header?)` clause

shamsimam: Nitpick: can we destructure `headers` from `sources`?

pschorf: The issue is that they're doing different things, the last case is setting the permitted-user to the current user when there is no permitted-user while this one is setting the permitted-user to the run-as-user (which may be different) unless it was specifically set in another header.

Did the destructuring.

pschorf: If you can think of a cleaner way to do it though, I'm open to it. This logic is a mess.

pschorf: Updated logic based on our discussion

waiter/src/waiter/service_description.clj
```diff
@@ -524,11 +524,12 @@
...
           ; else rely on presence of x-waiter headers to set the run-as-user
           contains-waiter-header? (headers/contains-waiter-header waiter-headers on-the-fly-service-description-keys)
+          contains-service-parameter-header? (headers/contains-waiter-header waiter-headers service-description-keys)
           user-service-description (cond-> sanitized-metadata-description
                                            (and (not (contains? sanitized-metadata-description "run-as-user")) contains-waiter-header?)
                                            ; can only set the run-as-user if some on-the-fly-service-description-keys waiter header was provided
                                            (assoc-run-as-requester-fields username)
-                                           (and (not (contains? sanitized-metadata-description "permitted-user")) contains-waiter-header?)
+                                           contains-service-parameter-header?
                                            ; can only set the permitted-user if some on-the-fly-service-description-keys waiter header was provided
```
shamsimam: Please update the comment

pschorf: Done

PR #61, opened 2017-07-07
Add support for JVM metrics reporting

shamsimam:
screen shot 2017-07-07 at 10 19 58 am

### DIFF
--- a/waiter/integration/waiter/metrics_output_test.clj
+++ b/waiter/integration/waiter/metrics_output_test.clj
@@ -36,8 +36,7 @@
   {(s/required-key "count") gauge-metric-schema})
 
 (def service-metrics-schema
-  {
-   (s/required-key "counters") {(s/required-key "request-counts") {(s/required-key "outstanding") schema/non-negative-num
+  {(s/required-key "counters") {(s/required-key "request-counts") {(s/required-key "outstanding") schema/non-negative-num
                                                                    (s/required-key "successful") schema/positive-num
                                                                    (s/required-key "total") schema/positive-num
                                                                    (s/required-key "streaming") schema/non-negative-num
@@ -65,8 +64,7 @@
    s/Str s/Any})
 
 (def jvm-metrics-schema
-  {
-   (s/required-key "attribute") s/Any
+  {(s/required-key "attribute") s/Any
    (s/required-key "file") s/Any
    (s/required-key "gc") s/Any
    (s/required-key "memory") s/Any
@@ -84,8 +82,7 @@
    s/Str s/Any})
 
 (def waiter-metrics-schema
-  {
-   (s/required-key "autoscaler") s/Any
+  {(s/required-key "autoscaler") s/Any
    (s/required-key "core") s/Any
    (s/required-key "gc") s/Any
    (s/required-key "requests") s/Any
@@ -120,11 +117,11 @@
                           metrics-json-response (make-request router-url "/metrics")
                           metrics-response (json/read-str (:body metrics-json-response))
                           service-metrics (get-in metrics-response ["services" service-id])]
-                      (log/info "Asserting jvm metrics output for" router-url)
+                      (log/info "asserting jvm metrics output for" router-url)
                       (assert-metrics-output (get metrics-response "jvm") jvm-metrics-schema)
-                      (log/info "Asserting service metrics output for" router-url)
+                      (log/info "asserting service metrics output for" router-url)
                       (assert-metrics-output service-metrics service-metrics-schema)
-                      (log/info "Asserting waiter metrics output for" router-url)
+                      (log/info "asserting waiter metrics output for" router-url)
                       (assert-metrics-output (get metrics-response "waiter") waiter-metrics-schema)))
                   (keys router->endpoint)))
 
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/main.clj
```diff
@@ -117,6 +118,7 @@
     (reify Thread$UncaughtExceptionHandler
       (uncaughtException [_ thread throwable]
         (log/error throwable (str (.getName thread) " threw exception: " (.getMessage throwable))))))
+  (jvm-metrics/instrument-jvm)
```
dposada: Does this have an effect on performance?

pschorf: 👍

shamsimam: I haven't measured the performance impact but wouldn't expect these to be high. The jvm metrics are all gauges, which means they only execute when we ask for the value.

shamsimam:
any more feedback on this?

dposada:
Should we add an integration test for this?

shamsimam:
@dposada Yeah, I'll add checking the presence of "jvm" entries.

waiter/src/waiter/main.clj
shamsimam: Inter-router metrics sync-ing is triggered via `waiter.metrics/get-core-metrics` which uses a `MetricFilter` that looks into service metrics only.

waiter/integration/waiter/metrics_output_test.clj
```diff
@@ -58,9 +64,37 @@
                               s/Str s/Any}
    s/Str s/Any})
 
+(def jvm-metrics-schema
+  {
+   (s/required-key "attribute") s/Any
```
dposada: Shouldn't this be on the previous line?

waiter/integration/waiter/metrics_output_test.clj
```diff
@@ -58,9 +64,37 @@
...
+                              (s/required-key "runnable") counted-gauge-metric-schema
+                              (s/required-key "terminated") counted-gauge-metric-schema
+                              (s/required-key "timed_waiting") counted-gauge-metric-schema
+                              (s/required-key "waiting") counted-gauge-metric-schema
+                              s/Str s/Any}
+   s/Str s/Any})
+
+(def waiter-metrics-schema
+  {
+   (s/required-key "autoscaler") s/Any
```
dposada: Shouldn't this be on the previous line?

waiter/integration/waiter/metrics_output_test.clj
```diff
@@ -83,11 +117,15 @@
...
-                      (log/info "Asserting /metrics output for" router-url)
-                      (assert-metrics-output app-metrics)))
+                          metrics-json-response (make-request router-url "/metrics")
+                          metrics-response (json/read-str (:body metrics-json-response))
+                          service-metrics (get-in metrics-response ["services" service-id])]
+                      (log/info "Asserting jvm metrics output for" router-url)
+                      (assert-metrics-output (get metrics-response "jvm") jvm-metrics-schema)
+                      (log/info "Asserting service metrics output for" router-url)
+                      (assert-metrics-output service-metrics service-metrics-schema)
+                      (log/info "Asserting waiter metrics output for" router-url)
```
dposada: Nitpick: lowercase first word of log statements

waiter/src/waiter/main.clj
dposada: So just calling `(jvm-metrics/instrument-jvm)` doesn't add any overhead?

waiter/src/waiter/handler.clj
```diff
@@ -191,11 +191,13 @@
   [request]
   (try
     (let [request-params (:params (ring-params/params-request request))
-          exclude-services (Boolean/parseBoolean (get request-params "exclude-services" "false"))
+          exclude-services (utils/request-flag request-params "exclude-services")
           service-id (get request-params "service-id" nil)
+          show-jvm-metrics (utils/request-flag request-params "show-jvm-metrics")
```
sradack: Should this be named `include-jvm-metrics` to be consistent with `exclude-services`?

waiter/src/waiter/main.clj
sradack: `instrument-jvm` creates gauges.  The overhead comes when the gauges are queried, at which point each gauge calls its backing function to return the current value.

sradack: And based my understanding, the only time we ask for the value of the jvm gauges is in the handler.  We do not evaluate them automatically on a regular basis.

shamsimam:
@dposada 

> So just calling (jvm-metrics/instrument-jvm) doesn't add any overhead?

`instrument-jvm` ends up creating gauges, the overhead with gauges only occurs when we try to extract their value (which we aren't during this call).

sradack:
Aside from @dposada's comments, LGTM

shamsimam:
@dposada responded to your comments. Passing build: https://travis-ci.org/twosigma/waiter/builds/252550413

shamsimam:
@pschorf any more feedback on this?

shamsimam:
Can we get this merged?

PR #71, opened 2017-07-12
Token delete and last-update-time metadata

shamsimam:


waiter/integration/waiter/token_request_test.clj
```diff
@@ -83,18 +83,25 @@
...
     token-response))
 
 (defn- list-tokens
   [waiter-url owner & {:keys [cookies] :or {cookies {}}}]
   (let [tokens-response (make-request waiter-url "/tokens" :cookies cookies)]
-    (log/debug "Retrieved tokens for owner " owner ":" (:body tokens-response))
+    (log/debug "retrieved tokens for owner " owner ":" (:body tokens-response))
     tokens-response))
 
+(defn parse-token-description
```
dposada: Missing docstring

waiter/integration/waiter/token_request_test.clj
```diff
@@ -113,23 +120,19 @@
...
-                                         (is false (str "Failed to parse token " body))))]
+                token-description (parse-token-description body)]
             (assert-response-status token-response 200)
-            (is (= "/custom-endpoint" (token-description "health-check-url")))
-            (is (= service-id-prefix (token-description "name")))
-            (is current-user (token-description "owner"))
-            (is (not-any? #(contains? token-description %) ["cmd" "cpus" "mem" "run-as-user" "permitted-user"])
-                (str "Unexpected content in token response: " body)))
+            (is (contains? token-description :last-update-time))
+            (is (= {:deleted false :health-check-url "/custom-endpoint" :name service-id-prefix :owner (retrieve-username)}
```
dposada: I would favor either one entry per line or separating entries with `,`

waiter/integration/waiter/token_request_test.clj
```diff
@@ -140,46 +143,62 @@
...
+            (is (not-any? (fn [token-entry] (= token (get token-entry "token"))) tokens)))))
+      (log/info "deleting the tokens in erase mode")
+      (doseq [token tokens-to-create]
+        (delete-token-and-assert waiter-url token :query-params {"erase" true}))
+      (log/info "ensuring tokens can no longer be retrieved on each router with show-deleted parameter after erase")
+      (doseq [token tokens-to-create]
+        (doseq [[_ router-url] (routers waiter-url)]
+          (let [{:keys [body] :as response} (get-token router-url token :cookies cookies :query-params {"show-deleted" true})]
+            (assert-response-status response 404)
+            (is (str/includes? (str body) "couldn't find token") (str body))))))))
```
dposada: This test is getting pretty big. Should we break it up into multiple smaller ones?

waiter/src/waiter/token.clj
```diff
@@ -67,12 +69,18 @@
...
   (defn delete-service-description-for-token
     "Delete a token from the KV"
-    [synchronize-fn kv-store token owner]
+    [clock synchronize-fn kv-store token owner & {:keys [erase] :or {erase false}}]
     (synchronize-fn
       token-lock
       (fn inner-delete-service-description-for-token []
-        (log/info "Deleting service description for token:" token)
-        (kv/delete kv-store token)
+        (log/info "Deleting service description for token:" token "erase mode:" erase)
```
dposada: Nitpick: lowercase first word

waiter/src/waiter/token.clj
```diff
@@ -235,19 +243,25 @@
...
+                      (throw (ex-info "Cannot change owner of token" {:existing-owner existing-service-description-owner :new-user owner :status 403})))
+                    (when-not (authz/run-as? entitlement-manager authenticated-user owner)
+                      (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user :owner owner :status 403}))))))
               ; Store the token
-              (store-service-description-for-token synchronize-fn kv-store token new-service-description-template {"owner" owner})
+              (let [new-token-metadata (merge {"deleted" false
+                                               "last-update-time" (.getMillis ^DateTime (clock))
+                                               "owner" owner}
+                                              new-token-metadata)]
+                (store-service-description-for-token synchronize-fn kv-store token new-service-description-template new-token-metadata))
```
dposada: Should we move the code for `:get`, `:post`, `:delete` into their own functions? This thing is getting pretty big

waiter/test/waiter/token_test.clj
```diff
@@ -69,31 +91,63 @@
...
+      (testing "delete:token-does-exist-authorized:erase-false"
+        (try
+          (kv/store kv-store token (assoc service-description1 "owner" "tu1"))
+          (is (not (nil? (kv/fetch kv-store token))))
+          (let [{:keys [status body]}
+                (run-handle-token-request
+                  kv-store waiter-hostname entitlement-manager make-peer-requests-fn nil
+                  {:authorization/user "tu1", :headers {"x-waiter-token" token}, :query-params {"erase" "false"}, :request-method :delete})]
+            (is (= 200 status))
+            (is (every? #(str/includes? body (str %)) [(str "\"delete\":\"" token "\""), "\"success\":true"]))
```
dposada: Can we parse the body and make assertions on the map?

waiter/test/waiter/token_test.clj
```diff
@@ -69,31 +91,63 @@
...
+      (testing "delete:token-does-exist-authorized:erase-true"
+        (try
+          (kv/store kv-store token (assoc service-description1 "owner" "tu1"))
+          (is (not (nil? (kv/fetch kv-store token))))
+          (let [{:keys [status body]}
+                (run-handle-token-request
+                  kv-store waiter-hostname entitlement-manager make-peer-requests-fn nil
+                  {:authorization/user "tu1", :headers {"x-waiter-token" token}, :query-params {"erase" "true"}, :request-method :delete})]
+            (is (= 200 status))
+            (is (every? #(str/includes? body (str %)) [(str "\"delete\":\"" token "\""), "\"success\":true"]))
```
dposada: Can we parse the body and make assertions on the map?

waiter/integration/waiter/token_request_test.clj
shamsimam: I thought the protocol was if we used keywords as keys and other data types as values, then commas were optional.

waiter/src/waiter/token.clj
shamsimam: I agree, but didn't want to do it as part of this PR. Will do it in a separate refactor PR.

waiter/test/waiter/token_test.clj
shamsimam: I'll do it in the refactor PR mentioned above, too many tests will be affected. Created: https://github.com/twosigma/waiter/issues/77

waiter/integration/waiter/token_request_test.clj
shamsimam: Added docstring.

waiter/integration/waiter/token_request_test.clj
shamsimam: I extracted a macro (`assert-get-token-response-200`) and separated the different stages with new lines.

waiter/src/waiter/token.clj
shamsimam: Done.

waiter/integration/waiter/token_request_test.clj
dposada: I thought it was simply commas if same line, no commas if newlines

waiter/test/waiter/token_test.clj
dposada: Can't we just `json/read-str` on `body`?

shamsimam: Let's do the refactor in another PR, other tests in this file will also need the same refactoring from json to map and changes to assertions.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -232,42 +275,96 @@
...
+        (testing "active-token"
+          (let [token (create-token-name waiter-url service-id-prefix)]
+            (try
+              (log/info "creating configuration using token" token)
+              (let [token-description {:health-check-url "/probe"
+                                       :name service-id-prefix
+                                       :owner (retrieve-username)
+                                       :run-as-user "foo-bar"
+                                       :token token
+                                       :update-mode "sync"}
```
sradack: I'm uneasy about a key `update-mode` inside something named `token-description`.

waiter/src/waiter/client_tools.clj
```diff
@@ -519,13 +519,14 @@
       (str/replace (str test-prefix service-name username (rand-int 3000000)) #"-" ""))))
 
 (defn delete-token-and-assert
-  [waiter-url token]
+  [waiter-url token & {:keys [query-params] :or {query-params {}}}]
```
sradack: What about erasing tokens created in tests that you haven't updated?  Should `erase` be on by default for unit tests?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -595,8 +693,9 @@
         (testing "token retrieval"
           (let [token-response (get-token waiter-url token)
                 response-body (-> token-response (:body) (json/read-str) (pc/keywordize-map))]
-            (is (= (assoc service-description :authentication "disabled" :owner current-user)
-                   response-body))))
+            (is (contains? response-body :last-update-time))
+            (is (= (assoc service-description :authentication "disabled" :deleted false :owner current-user)
```
sradack: Why is `deleted` in the response?  Are normal users going to start seeing `deleted`?  The only time it should show up is when `show-deleted` is `true` and the token is deleted.

waiter/src/waiter/authorization.clj
```diff
@@ -32,6 +32,23 @@
...
+
+(defn manage-token?
+  "Returns whether the auth-user is allowed to modify the specified token."
+  [entitlement-manager auth-user token token-metadata]
+  (authorized? entitlement-manager auth-user :manage (make-token-resource token token-metadata)))
+
+(defn sync-token?
+  "Returns whether the auth-user is allowed to sync the specified token."
+  [entitlement-manager auth-user token token-metadata]
+  (authorized? entitlement-manager auth-user :sync (make-token-resource token token-metadata)))
```
sradack: `:system` or `:admin` is a better name than `:sync`, leaving the option later to use this functionality for something else without the name being awkward.

waiter/src/waiter/token.clj
```diff
@@ -63,23 +65,29 @@
...
     "Delete a token from the KV"
-    [synchronize-fn kv-store token owner]
+    [clock synchronize-fn kv-store token owner & {:keys [erase] :or {erase false}}]
     (synchronize-fn
       token-lock
       (fn inner-delete-service-description-for-token []
-        (log/info "Deleting service description for token:" token)
-        (kv/delete kv-store token)
+        (log/info "deleting service description for token:" token "erase mode:" erase)
+        (if erase
```
sradack: I liked Datomic's term `excise`.  Or maybe  `hard-delete`.  `erase` is too similar to `delete`.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -232,42 +275,96 @@
...
+          (let [token (create-token-name waiter-url service-id-prefix)]
+            (try
+              (log/info "creating configuration using token" token)
+              (let [token-description {:deleted true
+                                       :health-check-url "/probe"
+                                       :name service-id-prefix
+                                       :owner (retrieve-username)
+                                       :run-as-user "foo-bar"
+                                       :token token
+                                       :update-mode "sync"}
```
sradack: Thoughts on having `update-mode` be a query string parameter instead of adding more to the post body?

waiter/test/waiter/token_test.clj
shamsimam: Replied above, will do as part of refactor outside this PR.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -232,42 +275,96 @@
                     (str headers))
                 (delete-service waiter-url service-id))))
           (finally
-            (delete-token-and-assert waiter-url token)))))))
+            (delete-token-and-assert waiter-url token :query-params {"erase" true})))))))
+
+(deftest ^:parallel ^:integration-fast test-token-sync-invalid-run-as-user
```
shamsimam: Done.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -232,42 +275,96 @@
...
+    (let [service-id-prefix (rand-name)]
+      (testing "token-syncing"
+        (testing "active-token"
+          (let [token (create-token-name waiter-url service-id-prefix)]
+            (try
+              (log/info "creating configuration using token" token)
+              (let [token-description {:health-check-url "/probe"
+                                       :name service-id-prefix
+                                       :owner (retrieve-username)
+                                       :run-as-user "foo-bar"
```
shamsimam: Done, used `i-do-not-exist-but-will-not-be-checked`

waiter/integration/waiter/token_request_test.clj
shamsimam: Passing it around as query parameter now.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -232,42 +275,96 @@
...
+                                       :token token
+                                       :update-mode "sync"}
+                    response (post-token waiter-url token-description)]
+                (assert-response-status response 200))
+              (log/info "created configuration using token" token)
+              (let [token-response (get-token waiter-url token)
+                    response-body (json/read-str (:body token-response))]
+                (is (contains? response-body "last-update-time"))
+                (is (= {"deleted" false, "health-check-url" "/probe", "name" service-id-prefix, "owner" (retrieve-username),
+                        "run-as-user" "foo-bar", "update-mode" "sync"}
```
shamsimam: Since it is no longer saved, it does not show up in the response.

waiter/integration/waiter/token_request_test.clj
shamsimam: Using query parameter now.

waiter/integration/waiter/token_request_test.clj
shamsimam: Done.

waiter/src/waiter/client_tools.clj
shamsimam: I'd rather the test explicitly excise the token as part of the test.

waiter/src/waiter/authorization.clj
shamsimam: Used `:system`

waiter/src/waiter/token.clj
shamsimam: renamed to `excise`

shamsimam:
@dposada  @sradack  Build passing, ready for another round of reviews.

waiter/src/waiter/client_tools.clj
sradack: What I mean is, we probably want to automatically excise all tokens in all other tests, instead of just soft-deleting them.

waiter/src/waiter/token.clj
```diff
@@ -159,40 +167,43 @@
...
-              (let [req-headers (:headers req)
-                    {:keys [token]} (sd/retrieve-token-from-service-description-or-hostname req-headers req-headers waiter-hostname)
-                    current-user (get req :authorization/user)]
+              (let [{:keys [token]} (sd/retrieve-token-from-service-description-or-hostname headers headers waiter-hostname)
+                    authenticated-user (get req :authorization/user)
+                    request-params (:query-params (ring-params/params-request req))
+                    excise (utils/request-flag request-params "excise")]
                 (if token
-                  (let [{:keys [service-description-template token-metadata]} (sd/token->token-description kv-store token)]
+                  (let [{:keys [service-description-template token-metadata]} (sd/token->token-description kv-store token :show-deleted excise)]
```
sradack: `:show-deleted` is awkward because the `token->token-description` function doesn't "show" anything.

waiter/src/waiter/token.clj
```diff
@@ -159,40 +167,43 @@
...
                     (if (and service-description-template (not-empty service-description-template))
                       (let [token-owner (get token-metadata "owner")]
-                        (when-not (can-run-as? current-user token-owner)
-                          (throw (ex-info "User not allowed to delete token"
-                                          {:existing-owner token-owner
-                                           :current-user current-user
-                                           :status 403})))
-                        (delete-service-description-for-token synchronize-fn kv-store token token-owner)
+                        (if excise
+                          (when-not (authz/sync-token? entitlement-manager authenticated-user token token-metadata)
```
sradack: Should `sync-token?` be renamed also.  Syncing is external to Waiter.  Waiter decides whether to allow "system" level access, i.e. ignoring normal permission scheme.

waiter/src/waiter/token.clj
```diff
@@ -235,19 +247,24 @@
...
-                  (throw (ex-info "Cannot run as user" {:authenticated-user authenticated-user :run-as-user run-as-user}))))
-              (let [existing-service-description-owner (get existing-token-metadata "owner")]
-                (if-not (str/blank? existing-service-description-owner)
-                  (when-not (can-run-as? authenticated-user existing-service-description-owner)
-                    (throw (ex-info "Cannot change owner of token"
-                                    {:existing-owner existing-service-description-owner
-                                     :new-user owner})))
-                  (when-not (can-run-as? authenticated-user owner)
-                    (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user :owner owner})))))
+              (when (= "sync" (get request-params "update-mode"))
```
sradack: `update-mode` `system`?  Not sure.  `sync` seems too specific.

waiter/src/waiter/token.clj
shamsimam: Do you like `administer-token?`?

waiter/src/waiter/token.clj
shamsimam: Changed to `include-deleted`

waiter/src/waiter/token.clj
shamsimam: CHanged to `system`

waiter/src/waiter/token.clj
sradack: I'm not in love with it but it's good enough.

waiter/src/waiter/client_tools.clj
shamsimam: Changed the default to `excise=true`.

waiter/src/waiter/token.clj
shamsimam: Done.

shamsimam:
@dposada any additional feedback on this?

shamsimam:
Rebased, ready for more feedback @dposada.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -213,40 +252,92 @@
           (finally
             (delete-token-and-assert waiter-url token)))))))
 
+(deftest ^:parallel ^:integration-fast test-token-sync-unaffected-by-run-as-user-permissions
```
dposada: I thought sync-ing was external to waiter

waiter/src/waiter/authorization.clj
```diff
@@ -32,6 +32,23 @@
...
+   :token token
+   :user owner})
+
+(defn manage-token?
+  "Returns whether the auth-user is allowed to modify the specified token."
+  [entitlement-manager auth-user token token-metadata]
+  (authorized? entitlement-manager auth-user :manage (make-token-resource token token-metadata)))
+
+(defn administer-token?
+  "Returns whether the auth-user is allowed to sync the specified token."
```
dposada: Should this be "allowed to admin"?

dposada:
Can we add some documentation explaining the new behavior that is introduced here?

waiter/bin/run-using-shell-scheduler.sh
```diff
@@ -13,4 +13,4 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 echo "Starting waiter..."
 cd ${DIR}/..
-WAITER_AUTH_RUN_AS_USER=$(id -un) lein run config-shell.edn
+WAITER_AUTH_RUN_AS_USER=$(id -un) lein do clean, compile, run config-shell.edn
```
shamsimam: Added it to `run-using-minimesos.sh`

waiter/integration/waiter/token_request_test.clj
```diff
@@ -79,22 +79,50 @@
...
 
+(defn- parse-token-description
+  "Parses a response as json and keywordizes the map."
+  [response-body]
+  (try
+    (-> response-body str json/read-str pc/keywordize-map)
+    (catch Exception _
+      (is false (str "Failed to parse token " response-body)))))
+
+(defmacro assert-get-token-response-200
```
shamsimam: Used `assert-token-response`

waiter/integration/waiter/token_request_test.clj
```diff
@@ -140,46 +164,68 @@
...
+                                    :cookies cookies
+                                    :query-params {"include-deleted" true})]
+            (assert-get-token-response-200 response service-id-prefix true))
           (let [{:keys [body] :as tokens-response} (list-tokens router-url current-user :cookies cookies)
                 tokens (json/read-str body)]
             (assert-response-status tokens-response 200)
-            (is (not-any? (fn [token-entry] (= token (get token-entry "token"))) tokens))))))))
+            (is (not-any? (fn [token-entry] (= token (get token-entry "token"))) tokens)))))
+
+      (log/info "deleting the tokens in excise mode")
```
shamsimam: @sradack liked the term `excise`.

With regards to the test, we want to ensure tokens can no longer be retrieved on each router after they have been excised (hard deleted).

waiter/integration/waiter/token_request_test.clj
shamsimam: Waiter still needs to add support for a sync call, which will be made by an external process. This test intends to verify that support works.

waiter/src/waiter/authorization.clj
shamsimam: It follows the same pattern as `manage-token?` and `run-as?` where the `can-` prefix is implied.

waiter/integration/waiter/token_request_test.clj
sradack: Excise is a term that Datomic uses for hard deletes.

shamsimam:
> Can we add some documentation explaining the new behavior that is introduced here?

Added documentation.

shamsimam:
@dposada ready for another round

waiter/integration/waiter/token_request_test.clj
dposada: Should we just use "hard delete"? My concern with excise is that it may not be obvious what it means. Do other systems also use excise?

waiter/src/waiter/authorization.clj
dposada: Sorry, my comment wasn't clear - I just meant in the docstring should "allowed to sync" become "allowed to admin" since the function is `administer-token?`?

waiter/src/waiter/token.clj
```diff
@@ -63,23 +65,29 @@
...
+    [clock synchronize-fn kv-store token owner & {:keys [excise] :or {excise false}}]
     (synchronize-fn
       token-lock
       (fn inner-delete-service-description-for-token []
-        (log/info "Deleting service description for token:" token)
-        (kv/delete kv-store token)
+        (log/info "deleting service description for token:" token "excise mode:" excise)
+        (if excise
+          (kv/delete kv-store token)
+          (when-let [existing-token-description (kv/fetch kv-store token)]
```
dposada: If there was no existing token, should we return an error?

waiter/integration/waiter/token_request_test.clj
shamsimam: I like excise, its definition is to surgically remove which is kind of we are doing with a hard deleted.

waiter/src/waiter/token.clj
```diff
@@ -63,23 +65,29 @@
...
+            (let [new-token-description (assoc existing-token-description
+                                          "deleted" true
+                                          "last-update-time" (.getMillis ^DateTime (clock)))]
```
dposada: If the existing token description happens to already have `"deleted" true`, should we still update the `"last-update-time"`?

waiter/src/waiter/authorization.clj
shamsimam: Done.

waiter/src/waiter/token.clj
```diff
@@ -63,23 +65,29 @@
...
   (defn delete-service-description-for-token
     "Delete a token from the KV"
-    [synchronize-fn kv-store token owner]
+    [clock synchronize-fn kv-store token owner & {:keys [excise] :or {excise false}}]
     (synchronize-fn
       token-lock
       (fn inner-delete-service-description-for-token []
-        (log/info "Deleting service description for token:" token)
-        (kv/delete kv-store token)
+        (log/info "deleting service description for token:" token "excise mode:" excise)
```
dposada: Maybe we should only log if actually end up changing something?

waiter/src/waiter/token.clj
```diff
@@ -159,40 +167,43 @@
...
+                        (delete-service-description-for-token clock synchronize-fn kv-store token token-owner :excise excise)
                         ; notify peers of token delete and ask them to refresh their caches
                         (make-peer-requests-fn "tokens/refresh"
-                                               :method :post
-                                               :body (json/write-str {:token token
-                                                                      :owner token-owner}))
-                        (utils/map->json-response {:delete token, :success true}))
+                                               :body (json/write-str {:owner token-owner :token token})
+                                               :method :post)
+                        (utils/map->json-response (cond-> {:delete token :success true}
```
dposada: Either separate lines or `,`-delimited

waiter/src/waiter/token.clj
```diff
@@ -159,40 +167,43 @@
...
+                                                          excise (assoc :excise true))))
```
dposada: Could we just do 

```clj
{:delete token 
 :success true
 :excise excise}
```

?

waiter/src/waiter/token.clj
```diff
@@ -235,19 +247,24 @@
...
+              (when (not= "admin" (get request-params "update-mode"))
+                (when (and run-as-user (not= "*" run-as-user))
+                  (when-not (authz/run-as? entitlement-manager authenticated-user run-as-user)
+                    (throw (ex-info "Cannot run as user" {:authenticated-user authenticated-user :run-as-user run-as-user :status 403}))))
+                (let [existing-service-description-owner (get existing-token-metadata "owner")]
+                  (if-not (str/blank? existing-service-description-owner)
+                    (when-not (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata)
+                      (throw (ex-info "Cannot change owner of token" {:existing-owner existing-service-description-owner :new-user owner :status 403})))
+                    (when-not (authz/run-as? entitlement-manager authenticated-user owner)
+                      (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user :owner owner :status 403}))))))
```
dposada: Could we change this to 

```clj
(case (get request-params "update-mode")
  nil
  ...
  
  "admin"
  ...
  
  (throw (ex-info "Invalid update-mode" request-params))
```

?

waiter/src/waiter/token.clj
```diff
@@ -235,19 +247,24 @@
...
-                (if-not (str/blank? existing-service-description-owner)
-                  (when-not (can-run-as? authenticated-user existing-service-description-owner)
-                    (throw (ex-info "Cannot change owner of token"
-                                    {:existing-owner existing-service-description-owner
-                                     :new-user owner})))
-                  (when-not (can-run-as? authenticated-user owner)
-                    (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user :owner owner})))))
+              (when (= "admin" (get request-params "update-mode"))
+                (when-not (authz/administer-token? entitlement-manager authenticated-user token new-token-metadata)
+                  (throw (ex-info "Cannot sync token" {:status 403 :token-metadata new-token-metadata :user authenticated-user}))))
```
dposada: Should this be "Cannot administer token"?

waiter/docs/rest-api.md
```diff
@@ -7,6 +7,6 @@ This API description is meant to be an overview of methods available in the REST
...
 |`/status`           |`GET`   |Check on the status of Waiter.|
-|`/token`            |`GET`   |Retrieve information about a token. Token should be passed in the `X-Waiter-Token` header.|
-|`/token`            |`POST`  |Define a token. Tokens can store all or part of the service description.|
-|`/token`            |`DELETE`|Remove a token. Token should be passed in the `X-Waiter-Token` header.|
+|`/token`            |`GET`   |Retrieve information about a token. Token should be passed in the `X-Waiter-Token` header. An extra `include-deleted=true` query parameter can be used to retrieve a soft deleted token.|
+|`/token`            |`POST`  |Define a token. Tokens can store all or part of the service description. An extra `update-mode=admin` query parameter can be used to trigger a forced update of a token by a user with admin privileges.|
+|`/token`            |`DELETE`|Remove a token. Token should be passed in the `X-Waiter-Token` header. An extra `excise=true` query parameter can be used to trigger a hard delete of the token.|
```
dposada: Should we add some context somewhere about why we are introducing soft vs hard deleting?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -71,9 +71,30 @@
...
+    (log/debug "retrieved tokens for owner " owner ":" (:body tokens-response))
     tokens-response))
 
+(defn- parse-token-description
+  "Parses a response as json and keywordizes the map."
+  [response-body]
+  (try
+    (-> response-body str json/read-str pc/keywordize-map)
+    (catch Exception _
+      (is false (str "Failed to parse token " response-body)))))
```
dposada: Should we put this function somewhere like `waiter.test-helpers`?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -82,33 +103,29 @@
...
-                (str "Unexpected content in token response: " body)))
+          (let [response (get-token router-url token :cookies cookies)]
+            (assert-token-response response service-id-prefix false))
           (let [{:keys [body] :as tokens-response} (list-tokens router-url current-user :cookies cookies)
                 tokens (json/read-str body)]
             (assert-response-status tokens-response 200)
             (is (some (fn [token-entry] (= token (get token-entry "token"))) tokens)))))
-      (log/info "deleting the tokens")
+
+      (log/info "deleting the tokens (not erasing)")
```
dposada: Here we use "erase". It's inconsistent, but I think it's more clear than "excise"

waiter/integration/waiter/token_request_test.clj
dposada: We keep on using the words "hard delete" to actually describe what we're doing (twice in this thread). There was also an occurrence of "erase" in a log statement somewhere else. 

Why not just use "hard delete"?

waiter/integration/waiter/token_request_test.clj
dposada: But we're not calling it sync-ing elsewhere, we're calling it "update-mode admin"

waiter/integration/waiter/token_request_test.clj
shamsimam: Used `hard-delete` instead of `excise`

waiter/integration/waiter/token_request_test.clj
shamsimam: Used `test-token-administer-unaffected-by-run-as-user-permissions`

waiter/src/waiter/token.clj
shamsimam: changed log to read `attempting to delete service...`.

The kv implementation does not currently notify whether a delete was a no-op.

waiter/src/waiter/token.clj
shamsimam: We can do it for the `soft-delete`, but then the behavior with the `hard-delete` becomes different due to the kv implementation. If we want to go down the path (which includes changes to the kv delete), I would prefer doing it as part of another PR.

waiter/src/waiter/token.clj
shamsimam: If `deleted` is already true, this path is not exercised. Added a unit test for it.

waiter/src/waiter/token.clj
shamsimam: Done.

waiter/src/waiter/token.clj
shamsimam: Done.

waiter/src/waiter/token.clj
shamsimam: Done

waiter/src/waiter/token.clj
shamsimam: Done.

waiter/docs/rest-api.md
shamsimam: We can add it in the token-syncer docs, I don't want to mention the token sysncer here.

waiter/integration/waiter/token_request_test.clj
shamsimam: We can move it later if another test wants to use it

waiter/integration/waiter/token_request_test.clj
shamsimam: used soft and hard delete in the logs

waiter/src/waiter/token.clj
dposada: Why would you have to change the `kv/delete`? Why can't you just have a higher-level `(if-let [existing-token-description (kv/fetch kv-store token)] ... ...)`?

waiter/docs/rest-api.md
dposada: Agreed, I was thinking something along the lines of "Tokens can either be soft-deleted or hard-deleted. If a token is soft-deleted, it can still be retrieved by ... whereas hard-deleted tokens are permanently removed from the database ..."

shamsimam: Done.

waiter/src/waiter/token.clj
shamsimam: Oops, my bad. There is already a 404 returned when the token cannot be found.

```
(let [{:keys [service-description-template token-metadata]} (sd/token->token-description kv-store token :include-deleted hard-delete)]
  (if (and service-description-template (not-empty service-description-template))
    ...
    (utils/map->json-response {:message (str "token " token " does not exist!")} :status 404)))
```

waiter/src/waiter/token.clj
```diff
@@ -235,24 +246,35 @@
...
-                (if-not (str/blank? existing-service-description-owner)
-                  (when-not (can-run-as? authenticated-user existing-service-description-owner)
-                    (throw (ex-info "Cannot change owner of token"
-                                    {:existing-owner existing-service-description-owner
-                                     :new-user owner})))
-                  (when-not (can-run-as? authenticated-user owner)
-                    (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user :owner owner})))))
+              (case (get request-params "update-mode")
+                "admin"
+                (do
```
dposada: I think we can remove the `do`

shamsimam:
Build is green, can we merge this?

waiter/src/waiter/token.clj
```diff
@@ -235,24 +246,34 @@
...
+                    (when-not (authz/run-as? entitlement-manager authenticated-user run-as-user)
+                      (throw (ex-info "Cannot run as user" {:authenticated-user authenticated-user, :run-as-user run-as-user, :status 403}))))
+                  (let [existing-service-description-owner (get existing-token-metadata "owner")]
+                    (if-not (str/blank? existing-service-description-owner)
+                      (when-not (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata)
+                        (throw (ex-info "Cannot change owner of token" {:existing-owner existing-service-description-owner, :new-user owner, :status 403})))
+                      (when-not (authz/run-as? entitlement-manager authenticated-user owner)
+                        (throw (ex-info "Cannot create token as user" {:authenticated-user authenticated-user, :owner owner, :status 403}))))))
+
+                (throw (ex-info "Invalid update-mode" {:mode (get request-params "update-mode"), :status 403})))
```
sradack: This should probably be a 400 to indicate the client made an erroneous request, not that it is forbidden.

waiter/src/waiter/token.clj
```diff
@@ -235,24 +246,34 @@
...
+
               ; Store the token
-              (store-service-description-for-token synchronize-fn kv-store token new-service-description-template {"owner" owner})
+              (let [new-token-metadata (merge {"last-update-time" (.getMillis ^DateTime (clock))
```
sradack: If `admin` mode, you need to accept `last-update-time` data and use that for the token so that you can perform a sync.

waiter/src/waiter/token.clj
shamsimam: Done.

waiter/src/waiter/token.clj
shamsimam: This snippet adds the `last-modified-time` only if it was not provided. I added a separate check to disallow a regular user from modifying the `last-update-time`.

shamsimam:
@sradack included last-update-time in admin mode assertions of integration tests

PR #74, opened 2017-07-12
Add exception logging to debug "Pop without matching push exception"

shamsimam:


waiter/src/waiter/state.clj
```diff
@@ -587,9 +585,12 @@
...
   (async/go
-    (async/<! (async/timeout blacklist-period-ms))
-    (cid/cinfo correlation-id "requesting instance" instance-id "to be unblacklisted")
-    (async/>! unblacklist-instance-chan {:instance-id instance-id})))
+    (try
+      (async/<! (async/timeout blacklist-period-ms))
+      (cid/cinfo correlation-id "requesting instance" instance-id "to be unblacklisted")
+      (async/>! unblacklist-instance-chan {:instance-id instance-id})
+      (catch Throwable th
+        (log/error th "unexpected error inside trigger-unblacklist-process")))))
```
dposada: Can we use `cid/cerror` here? Also, can we include the `instance-id` and `blacklist-period-ms` in the log message?

shamsimam: The push/pop error is due to the use of `cid`, so want to avoid that. I will include the the instance and blacklist period in the log.

shamsimam: Done.

shamsimam:
Passing build: https://travis-ci.org/twosigma/waiter/builds/253234513

shamsimam:
Can we merge this?

PR #81, opened 2017-07-14
Add logging for basic authentication results

shamsimam:


kitchen/src/kitchen/core.clj
```diff
@@ -413,68 +407,83 @@
...
-                      byte-data (byte-array (take num-bytes-int (cycle (range 103))))]
-                  (async/>! out byte-data))
-
-                :else
-                (async/>! out in-data))
-              (recur))))))))
-
-(defn basic-auth-middleware [username password handler]
-  (fn [{:keys [uri] :as req}]
+  (printlog request "Received websocket request:" request)
```
dposada: `request` is repeated twice + lowercase first word

kitchen/src/kitchen/core.clj
```diff
@@ -413,68 +407,83 @@
...
+  "Adds support for basic authentication when both the provided username and password are not nil.
+  /status urls always bypass basic auth check."
+  [username password handler]
+  (when (not (and username password))
+    (log/info "basic authentication is disabled since username or password is missing"))
+  (fn basic-auth-middleware-fn [{:keys [uri] :as request}]
     (cond
-      (not (and username password)) (handler req)
-      (= "/status" uri) (handler req)
+      (not (and username password)) (handler request)
```
dposada: Maybe it's cleaner to just move this `(handler request)` to the new `when` check above?

kitchen/src/kitchen/core.clj
```diff
@@ -413,68 +407,83 @@
...
-                      (= password p))))
-              req))))
+                 (let [result (and (= username u)
+                                   (= password p))]
+                   (printlog request "authenticating" u (if result "successful" "failed"))
+                   result)))
+              request))))
+
+(defn correlation-id-middleware
+  "Attached a x-cid header to the request if one is not already provided.
```
dposada: Nitpick: attaches is instead of attached

kitchen/test/kitchen/core_test.clj
```diff
@@ -223,4 +227,16 @@
...
 
     (testing "Successful authentication"
-      (is (= {:status 200} (handler {:uri "/handler", :headers {"authorization" auth}}))))))
\ No newline at end of file
+      (is (= {:status 200} (handler {:uri "/handler", :headers {"authorization" auth}}))))))
+
+(deftest test-correlation-id-middleware
+  (let [handler (fn test-correlation-id-middleware-helper
+                  [{:keys [headers request-method]}]
+                  (is (get headers "x-cid"))
```
dposada: Seems like these should be indented one more level

kitchen/src/kitchen/core.clj
```diff
@@ -511,9 +520,12 @@
           (when (pos? start-up-sleep-ms)
             (log/info "sleeping for" start-up-sleep-ms "ms before starting server")
             (Thread/sleep start-up-sleep-ms))
-          (let [partial-server-options {:ring-handler (basic-auth-middleware username password (params/wrap-params http-handler))
+          (let [partial-server-options {:ring-handler (->> (params/wrap-params http-handler)
+                                                           (basic-auth-middleware username password)
+                                                           correlation-id-middleware)
```
sradack: Nice!

kitchen/src/kitchen/core.clj
```diff
@@ -413,68 +407,83 @@
...
+              :else
+              (async/>! out in-data))
+            (recur)))))))
+
+(defn basic-auth-middleware
+  "Adds support for basic authentication when both the provided username and password are not nil.
+  /status urls always bypass basic auth check."
+  [username password handler]
+  (when (not (and username password))
+    (log/info "basic authentication is disabled since username or password is missing"))
```
sradack: I know this behavior wasn't introduced in this change but:
Why do we disable auth if username and password are missing?  Doesn't that mean if we ever stop passing auth info to kitchen it will be silently ignored?

shamsimam: I think the intent is to allow a non-basic version to run locally. We could switch to requiring a command-line flag to enable to "unauth" mode.

kitchen/src/kitchen/core.clj
shamsimam: Going to polish it (add `x-cid` in response in another PR) in kitchen and then change Waiter too. That way we can also ensure the presence of the `x-cid` header in all our responses.

kitchen/src/kitchen/core.clj
shamsimam: Done with lower-case.

The first `request` is used to extract the cid in `printlog`, the second to actually use in the log message.

kitchen/src/kitchen/core.clj
shamsimam: Moved.

kitchen/src/kitchen/core.clj

kitchen/test/kitchen/core_test.clj
shamsimam: reformatted entire file :)

shamsimam:
Passing build: https://travis-ci.org/twosigma/waiter/builds/254315615.

Ready for another pass.

kitchen/src/kitchen/core.clj
```diff
@@ -413,68 +407,85 @@
...
+
+              :else
+              (async/>! out in-data))
+            (recur)))))))
+
+(defn basic-auth-middleware
+  "Adds support for basic authentication when both the provided username and password are not nil.
+   /status urls always bypass basic auth check."
+  [username password handler]
+  (let [basic-auth-enabled (not (and username password))]
```
dposada: Shouldn't this be `basic-auth-disabled`?

shamsimam: removed the variable

kitchen/src/kitchen/core.clj
shamsimam: We can pursue that change in a separate PR if you like.

shamsimam:
@sradack any more feedback?

kitchen/src/kitchen/core.clj
sradack: https://github.com/twosigma/waiter/issues/85

Issue #80, opened 2017-07-14
/waiter-async endpoints should not propagate authorization header to backend


dposada:
They currently pass along whatever authorization header was provided by the client. This was discovered when some of our tests attempted to pass the authorization header on /waiter-async routes, and the header was forwarded to kitchen, which then failed to authenticate because it expects the `WAITER_USERNAME` and `WAITER_PASSWORD` it was configured with.

shamsimam:
This was resolved by the choice of removing hop-by-hop headers.

PR #82, opened 2017-07-15
Issue #80 - use common make-request helper function to benefit from standard http header cleanup

shamsimam:


waiter/src/waiter/handler.clj
```diff
@@ -42,22 +42,29 @@
   {:username (:authorization/user request)
    :principal (:authenticated-principal request)})
 
+(defn async-make-request-helper
```
dposada: Why is "async" in this function's name?

shamsimam: As it is a helper function for `async-make-http-request`. Happy to rename, any suggestions?

dposada: That's fine, I just didn't see anything async-specific in the function

PR #84, opened 2017-07-17
Update test-multiple-async-requests

shamsimam:
- introduce delete assertions
 - move assertions to test thread

### DIFF
--- a/waiter/integration/waiter/async_request_integration_test.clj
+++ b/waiter/integration/waiter/async_request_integration_test.clj
@@ -187,21 +187,20 @@
           request-processing-time-ms 60000
           num-threads 20
           num-requests-to-delete 10
-          status-locations
-          (parallelize-requests
-            num-threads 1
-            (fn []
-              (log/info "making kitchen request")
-              (let [async-request-headers
-                    (assoc request-headers :x-kitchen-delay-ms (int (* (max 0.5 (double (rand))) request-processing-time-ms))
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/async_request_integration_test.clj
```diff
@@ -187,21 +187,20 @@
...
+                            num-threads 1
+                            (fn []
+                              (log/info "making kitchen request")
+                              (let [async-request-headers
+                                    (assoc request-headers :x-kitchen-delay-ms (int (* (max 0.5 (double (rand))) request-processing-time-ms))
+                                                           :x-kitchen-store-async-response-ms 600000)]
+                                (make-kitchen-request waiter-url async-request-headers :http-method-fn http/get :body "" :path "/async/request")))
+                            :verbose true)
+          status-locations (for [{:keys [headers] :as response} async-responses]
+                             (let [status-location (get (pc/map-keys str/lower-case headers) "location")]
```
sradack: Does the client not automatically lowercase headers?

shamsimam: `clj-http` does not, e.g. we have other tests where we check headers such as `X-Waiter-Router-Id`

shamsimam:
@sradack Cursive messes up those assertions and pollutes the console with internal output of those assertions from `clojure.test`

PR #86, opened 2017-07-20
Add status-4xx handlers to simulate bad health checks

pschorf:


shamsimam:
Waiting on green build.

PR #91, opened 2017-08-01
Make the websocket message sizes configurable

shamsimam:


kitchen/src/kitchen/core.clj
```diff
@@ -448,9 +448,14 @@
               (let [num-chars-str (subs in-data (count "chars-"))
                     num-chars-int (Integer/parseInt num-chars-str)
                     chars (map char (range 65 91))
-                    string-data (->> (repeatedly #(rand-nth chars))
-                                     (take num-chars-int)
-                                     (reduce str))]
+                    seed-string (->> (repeatedly #(rand-nth chars))
```
shamsimam: Needed to speed-up this computation to avoid the integration test timing out.

shamsimam:
Assigned to @sradack to get some early feedback, I am still working on getting a green build.

kitchen/src/kitchen/core.clj
```diff
@@ -448,9 +448,14 @@
               (let [num-chars-str (subs in-data (count "chars-"))
                     num-chars-int (Integer/parseInt num-chars-str)
                     chars (map char (range 65 91))
```
sradack: This would have helped me:

`(range (-> \A int) (-> \Z int inc))`

kitchen/src/kitchen/core.clj
sradack: Can we factor this string generation out into a separate function?   It could certainly be faster but imagine this is fast enough.

waiter/config-full.edn
```diff
@@ -475,4 +475,10 @@
 
  ;; Waiter launches a loop that maintains the stacktraces for individual threads;
  ;; this value dictates how often (milliseconds) the stacktraces are refreshed:
- :thread-stack-state-refresh-interval-ms 600000}
+ :thread-stack-state-refresh-interval-ms 600000
+
+ ;; Waiter supports making websocket connections to backends.
+ ;; The maximum binary and text message sizes in the websocket request or response can be configured
+ ;; as follows, the values are in bytes:
+ :websocket-config {:ws-max-binary-message-size  137438953472
```
sradack: These are really big limits for a default config.  Should the default config be more reasonable?  Is there a reason for the numbers you chose that you can comment on?

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -30,6 +30,9 @@
   (-> request (.getCookies) (.add (HttpCookie. "x-waiter-auth" auth-cookie-value))))
 
 (let [websocket-client (WebSocketClient.)]
+  (doto (.getPolicy websocket-client)
+    (.setMaxBinaryMessageSize (* 1024 1024 128))
```
sradack: Comment on why 128M?

waiter/src/waiter/settings.clj
```diff
@@ -260,6 +262,8 @@
                                   "scale-up-factor" 0.1}
    :statsd :disabled
    :thread-stack-state-refresh-interval-ms 600000 ; 10 minutes
+   :websocket-config {:ws-max-binary-message-size  (* 1024 1024 128)
```
sradack: Comment on why 128M?

waiter/config-full.edn
shamsimam: Scott mentioned something around 32mb, so I set it a relatively higher 128 mb.

kitchen/src/kitchen/core.clj

waiter/config-full.edn
sradack: This 128GB not 128MB

shamsimam: Oops, changed to `41943040` (40mb)

waiter/integration/waiter/websocket_integration_test.clj
shamsimam: Changed to 40mb

waiter/src/waiter/settings.clj
shamsimam: Changed to 40mb

PR #92, opened 2017-08-01
Handle MessageTooLargeException websocket exceptions

shamsimam:


waiter/src/waiter/websocket.clj
```diff
@@ -297,24 +304,21 @@
         (timers/start-stop-time!
           stream
           (when-let [close-message (async/<! request-close-promise-chan)]
-            (let [[source close-code status-code close-message] close-message]
-              (log/info "websocket connections requested to be closed due to" close-code)
+            (let [[source close-code status-code-or-exception close-message] close-message]
+              (log/info "websocket connections requested to be closed due to" source close-code close-message)
               (counters/dec! requests-streaming)
               ;; explicitly close the client connection if backend triggered the close
-              (when (and (= :instance source) (integer? status-code))
```
sradack: Was this why we weren't getting an error in the logs?

shamsimam: No, it should still have been logged. This changes handles correctly setting the return code to an server error value.

shamsimam:
@sradack we can go ahead and merge this too. This PR handles MessageTooLargeException returned on the ctrl chan. Working on yet another PR for the gobbled MessageTooLargeException.

PR #93, opened 2017-08-01
Avoid classcastexception in correlation-id-middleware

shamsimam:


kitchen/src/kitchen/core.clj
```diff
@@ -502,7 +502,11 @@
         (let [response (handler request)]
           (if (map? response)
             (add-cid-into-response request response)
-            (async/go (add-cid-into-response request (async/<! response)))))))))
+            (async/go
+              (let [nested-response (async/<! response)]
+                (if (map? nested-response)
```
dposada: When would it not be a map?

shamsimam: Websocket responses return channels that lead to classcastexceptions

dposada: Should we add a comment explaining that?

shamsimam: Okay, will add

PR #94, opened 2017-08-02
Added integration test to verify MessageTooLargeException

shamsimam:
-- added logging in kitchen to track state of websocket array responses

### DIFF
--- a/kitchen/bin/run.sh
+++ b/kitchen/bin/run.sh
@@ -6,6 +6,24 @@
 #   run.sh
 #
 # Runs Kitchen, passing through any and all arguments provided.
+# This script has special handling for an argument if it is `--mem`.
+# When the argument is `--mem`, that value is extracted and used as
+# the Xmx flag for the java command while launching kitchen.
+# The default value for memory allocated to kitchen is 128M.
+
+KITCHEN_MEM="128M"
+
+i=0
+until [ "$((i=$i+1))" -gt "$#" ]
+do
+  case "$1" in
+    --mem)
+      KITCHEN_MEM="$2"; $((i=$i-1)); shift ;;
+    *)
+      set -- "$@" "$1"  ;;
+  esac;
+  shift;
+done
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 JAR=${DIR}/../target/uberjar/kitchen-0.1.0-SNAPSHOT-standalone.jar
@@ -17,4 +35,6 @@ if [ ! -f ${JAR} ]; then
     exit 1
 fi
 
-java -Xmx128M -jar ${JAR} "$@"
+JAVA_CMD="java -Xmx${KITCHEN_MEM} -jar ${JAR} "$@""
+echo ${JAVA_CMD}
+${JAVA_CMD}
--- a/kitchen/bin/run.sh
+++ b/kitchen/bin/run.sh
@@ -36,5 +36,5 @@ if [ ! -f ${JAR} ]; then
 fi
 
 JAVA_CMD="java -Xmx${KITCHEN_MEM} -jar ${JAR} "$@""
-echo ${JAVA_CMD}
+echo "Running ${JAVA_CMD}"
 ${JAVA_CMD}
--- a/waiter/integration/waiter/websocket_integration_test.clj
+++ b/waiter/integration/waiter/websocket_integration_test.clj
@@ -12,6 +12,7 @@
   (:require [clj-time.core :as t]
             [clojure.core.async :as async]
             [clojure.data.json :as json]
+            [clojure.string :as str]
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
             [clojure.walk :as walk]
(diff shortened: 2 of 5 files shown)


kitchen/bin/run.sh
```diff
@@ -6,6 +6,24 @@
...
+until [ "$((i=$i+1))" -gt "$#" ]
+do
+  case "$1" in
+    --mem)
+      KITCHEN_MEM="$2"; $((i=$i-1)); shift ;;
+    *)
+      set -- "$@" "$1"  ;;
+  esac;
+  shift;
+done
```
dposada: Do we need the `--mem` flag? Can we instead just increase the default?

shamsimam: I needed to use `1024M`, do we want to increase this value for all tests? This will potentially cause our tests to run slower as they wait to get allocated resources.

dposada: OK, what you did makes sense, thanks

kitchen/bin/run.sh
```diff
@@ -17,4 +35,6 @@ if [ ! -f ${JAR} ]; then
     exit 1
 fi
 
-java -Xmx128M -jar ${JAR} "$@"
+JAVA_CMD="java -Xmx${KITCHEN_MEM} -jar ${JAR} "$@""
+echo ${JAVA_CMD}
```
sradack: Can we throw a "Running" in this echo?

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -153,6 +160,62 @@
         (finally
           (delete-service waiter-url waiter-headers))))))
 
+(deftest ^:parallel ^:integration-slow test-fail-to-stream-large-responses
```
sradack: `test-message-rcvd-size-exceeds-max`

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -153,6 +160,62 @@
...
+  (testing-using-waiter-url
+    (let [^WebSocketClient websocket-client (websocket-client-factory)
+          waiter-settings (waiter-settings waiter-url)
+          {:keys [ws-max-binary-message-size ws-max-text-message-size]} (:websocket-config waiter-settings)
+          ws-max-binary-message-size' (+ 2048 ws-max-binary-message-size)
+          ws-max-text-message-size' (+ 2048 ws-max-text-message-size)
+          auth-cookie-value (auth-cookie waiter-url)
+          waiter-headers (-> (kitchen-request-headers)
+                             (assoc :x-waiter-mem 1024
```
sradack: This should be JVM heap size + overhead

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -153,6 +160,62 @@
         (finally
           (delete-service waiter-url waiter-headers))))))
 
+(deftest ^:parallel ^:integration-slow test-fail-to-stream-large-responses
+  (testing-using-waiter-url
+    (let [^WebSocketClient websocket-client (websocket-client-factory)
+          waiter-settings (waiter-settings waiter-url)
+          {:keys [ws-max-binary-message-size ws-max-text-message-size]} (:websocket-config waiter-settings)
+          ws-max-binary-message-size' (+ 2048 ws-max-binary-message-size)
```
sradack: What's the point of adding 2048 here, resetting the max, and then adding another 1024 when we send the message?  Why not use the configured max and just add the 1024 for the message without `update-max-message-sizes`?

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -153,6 +160,62 @@
...
+                  (async/>! out (str "bytes-" (+ ws-max-text-message-size 1024)))
+                  (let [backend-response (async/<! in)]
+                    (deliver backend-data-promise backend-response))
+                  (async/close! out)
+                  (let [ctrl-data (async/<! ctrl)]
+                    (deliver ctrl-data-promise ctrl-data))
+                  (deliver response-promise :done)))
+              {:middleware middleware})
+
+            (is (= :done (deref response-promise (-> 3 t/minutes t/in-millis) :timed-out)))
```
sradack: Why aren't these assertions in the go block, avoiding all the promise objects?

shamsimam: The assertions in the go blocks don't show up as counts accurately when I run `lein test`

waiter/integration/waiter/websocket_integration_test.clj
shamsimam: We need kitchen to support larger than Waiter's configured max (adding the 2048 to waiter config).
We then need kitchen to send a message larger than Waiter's configured max but smaller than what kitchen can handle (adding to 1024 (i.e. < 2048) to waiter config).

kitchen/bin/run.sh
shamsimam: Done.

waiter/integration/waiter/websocket_integration_test.clj
shamsimam: Used `test-message-size-received-from-backend-exceeds-supported-max`

waiter/integration/waiter/websocket_integration_test.clj
shamsimam: Used `1024` for the description and `(1024 - 64)` to launch kitchen.

waiter/integration/waiter/websocket_integration_test.clj
pschorf: Could you just avoid the go block and do blocking reads?

shamsimam: If everyone's on board, let's do that as a separate refactor then? There are other websocket integration tests that are also doing async reads.

PR #98, opened 2017-08-15
Catches misconfigured health checks

chzou:


waiter/src/waiter/scheduler.clj
```diff
@@ -395,50 +396,115 @@
...
+                                                                                (fn available [instance health-check-path]
+                                                                                  (available? instance health-check-path http-client))
+                                                                                service-id->service-description-fn))]
+                          (let [available-services (keys service->service-instances)
+                                available-service-ids (map :id available-services)]
+                            (log/debug "scheduler-syncer:" (count service->service-instances) "available services:" available-service-ids)
+                            (doseq [service available-services]
+                              (when (zero? (reduce + 0 (filter number? (vals (select-keys (:task-stats service) [:staged :running :healthy :unhealthy])))))
+                                (log/info "scheduler-syncer:" (:id service) "has no live instances!" (:task-stats service))))
+                            (loop [service-id->unhealthy-instances' {}
```
sradack: Consider one map, instead of separate maps (something like `id->health-check-context`)

waiter/src/waiter/scheduler.clj
```diff
@@ -395,50 +396,115 @@
...
+                                   instance-id->failed-health-checks' {}
+                                   scheduler-messages [[:update-available-apps {:available-apps available-service-ids :scheduler-sync-time request-apps-time}]]
+                                   [[service {:keys [active-instances failed-instances]}] & remaining] (seq service->service-instances)]
+                              (if service
+                                (let [request-instances-time (t/now)
+                                      service-id (:id service)
+                                      last-unhealthy-instances (get service-id->unhealthy-instances service-id)
+                                      service-instance-info (retrieve-instances-for-app service-id active-instances)
+                                      unhealthy-instances (:unhealthy-instances service-instance-info)
+                                      instance-id->instance-fn (fn [id instances] (first (filter #(= id (:id %)) instances)))
```
sradack: These functions can go away with optimized data structures.

waiter/src/waiter/scheduler.clj
```diff
@@ -395,50 +396,115 @@
...
+                                      instances->ids-fn (fn [instances] (set (map :id instances)))
+                                      tracked-failed-instances (reduce conj (get service-id->tracked-failed-instances service-id)
```
sradack: Track the failed instances separately from determining if there are existing failed instances.

waiter/src/waiter/scheduler.clj
```diff
@@ -395,50 +396,115 @@
...
+          (let [new-state
+                (async/alt!
+                  exit-chan
+                  ([message]
+                    (log/warn "Stopping scheduler-syncer")
+                    (when (not= :exit message)
+                      (throw (ex-info "Stopping scheduler-syncer" {:time (t/now), :reason message}))))
+
+                  state-query-chan
+                  ([{:keys [response-chan service-id]}]
```
sradack: Factor message handlers into separate functions.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,19 @@
                       :exit-code exit-value))
     instance))
 
+(defn- enforce-grace-period
+  "Kills processes for unhealthy instances exceeding their grace period"
+  [{:keys [:shell-scheduler/process started-at] :as instance} grace-period-secs port->reservation-atom port-grace-period-ms]
+  (if (and (unhealthy? instance) (.isAlive process))
+    (let [start (clj-time.coerce/from-string started-at)]
+      (if (and (t/before? start (t/now)) (>= (t/in-seconds (t/interval start (t/now))) grace-period-secs))
```
sradack: Factor out time calls.

sradack: Be explicit about checking if start is `nil`.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,19 @@
                       :exit-code exit-value))
     instance))
 
+(defn- enforce-grace-period
+  "Kills processes for unhealthy instances exceeding their grace period"
+  [{:keys [:shell-scheduler/process started-at] :as instance} grace-period-secs port->reservation-atom port-grace-period-ms]
+  (if (and (unhealthy? instance) (.isAlive process))
```
sradack: Remove `.isAlive`

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,19 @@
...
+    (let [start (clj-time.coerce/from-string started-at)]
```
sradack: Is this the same method that we use to parse other timestamps?

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,19 @@
...
+      (if (and (t/before? start (t/now)) (>= (t/in-seconds (t/interval start (t/now))) grace-period-secs))
+        (do (log/info "unhealthy instance exceeds its grace period, killing instance" {:instance instance :grace-period-secs grace-period-secs})
```
sradack: Add `start` and current time to map

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,19 @@
...
+            (kill-process! instance port->reservation-atom port-grace-period-ms)
+            (assoc instance :killed true                    ; not marking as failed because Waiter should detect this
```
sradack: No reason to not add the flag.

waiter/src/waiter/shell_scheduler.clj
chzou: changed to use `clj-time.format/parse` instead

waiter/src/waiter/scheduler.clj
```diff
@@ -391,54 +392,121 @@
...
+                                                                 (:unhealthy-instances service-instance-info)))
+                  not-in? (fn [id instances] (not (some #(= id (:id %)) instances)))
+                  instance-id->tracked-failed-instance' ((fnil into {}) instance-id->tracked-failed-instance
+                                                          (map #(vector % (get instance-id->unhealthy-instance %))
+                                                               (filter #(and (not-in? % active-instances)
+                                                                             (> (get instance-id->failed-health-check-count %) max-failed-health-checks))
+                                                                       (keys instance-id->unhealthy-instance))))
+                  failed-instances (into failed-instances
+                                         (map #(update-in (get instance-id->tracked-failed-instance' %) [:flags] conj :never-passed-health-checks)
+                                              (filter #(not-in? % failed-instances)
```
sradack: It appears you are checking for the existence an instance-id inside a sequence of instances.

waiter/src/waiter/scheduler.clj
```diff
@@ -391,54 +392,121 @@
...
+                  service-instance-info (retrieve-instances-for-app service-id active-instances)
+                  instance-id->unhealthy-instance' (into {} (map #(vector (:id %) %)
+                                                                 (:unhealthy-instances service-instance-info)))
+                  not-in? (fn [id instances] (not (some #(= id (:id %)) instances)))
+                  instance-id->tracked-failed-instance' ((fnil into {}) instance-id->tracked-failed-instance
+                                                          (map #(vector % (get instance-id->unhealthy-instance %))
+                                                               (filter #(and (not-in? % active-instances)
+                                                                             (> (get instance-id->failed-health-check-count %) max-failed-health-checks))
+                                                                       (keys instance-id->unhealthy-instance))))
+                  failed-instances (into failed-instances
```
sradack: It's better to add the `:never-passed-health-check` flag when you're adding the instance to the `instance-id->tracked-failed-instance`' map so you don't have to add it here.

waiter/src/waiter/scheduler.clj
```diff
@@ -391,54 +392,121 @@
...
+                  scheduler-messages' (if service-instance-info
+                                        ; Assume nil service-instance-info means there was a failure in invoking marathon
+                                        (conj scheduler-messages
+                                              [:update-app-instances
+                                               (assoc service-instance-info
+                                                 :service-id service-id
+                                                 :failed-instances failed-instances
+                                                 :scheduler-sync-time request-instances-time)])
+                                        scheduler-messages)
+                  instance-id->failed-health-check-count' (into {} (map #(vector % ((fnil inc 0) (get instance-id->failed-health-check-count %)))
```
sradack: Check out [`plumbing.core/map-from-keys`](https://github.com/plumatic/plumbing/blob/318af7798bb701607aaae8639d12829014941184/src/plumbing/core.cljx#L86).

waiter/config-full.edn
```diff
@@ -484,6 +481,15 @@
                            ;; The number of hosts on which health checks must fail in order to declare a deployment error:
                            :min-hosts 2}
 
+ ;; Sets tolerance level for Waiter regarding health checks
```
dposada: Not sure how helpful this comment is given that there are solid comments on each of the sub-elements

waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -15,9 +15,31 @@
...
+    (let [headers {:x-waiter-name (rand-name)
+                   :x-waiter-health-check-url "/status-402"}
+          {:keys [headers body] :as response} (make-request-with-debug-info headers #(make-kitchen-request waiter-url %))
+          service-id (get headers "x-waiter-service-id")]
+      (is (not (nil? service-id)))
+      (assert-response-status response 503)
+      (is (str/starts-with? body (str "Deployment error: " (-> (waiter-settings waiter-url) :messages :health-check-misconfigured))))
+      (delete-service waiter-url service-id)))
+
+  (testing-using-waiter-url
```
dposada: I don't think we need to repeat the `testing-using-waiter-url` call here, right?

waiter/src/waiter/core.clj
```diff
@@ -780,16 +780,16 @@
...
                                   [:state scheduler]]
                            (let [scheduler-state-chan (au/latest-chan)
                                  scheduler-state-mult-chan (async/mult scheduler-state-chan)
                                  http-client (http/client {:connect-timeout health-check-timeout-ms
                                                            :idle-timeout health-check-timeout-ms})]
-                             (scheduler/start-scheduler-syncer
-                               scheduler scheduler-state-chan scheduler-syncer-interval-secs
-                               service-id->service-description-fn scheduler/available? http-client)
-                             {:scheduler-state-mult-chan scheduler-state-mult-chan}))
+                             (merge (scheduler/start-scheduler-syncer
```
dposada: Can this be `assoc` instead of `merge`?

waiter/src/waiter/marathon.clj
```diff
@@ -116,6 +116,8 @@
                                         (assoc :message (str/trim message))
                                         (str/includes? (str message) "Memory limit exceeded:")
                                         (assoc :flags #{:memory-limit-exceeded})
+                                        (str/includes? (str message) "Task was killed since health check failed")
```
dposada: I would favor a newline between each of the condition/operation pairs

waiter/src/waiter/marathon.clj
```diff
@@ -209,7 +211,7 @@
                      :intervalSeconds 10
                      :portIndex 0
                      :timeoutSeconds 20
-                     :maxConsecutiveFailures 20}]
+                     :maxConsecutiveFailures 5}]
```
dposada: Why the change here?

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -309,6 +309,23 @@
...
     instance))
 
+(defn- enforce-grace-period
+  "Kills processes for unhealthy instances exceeding their grace period"
+  [{:keys [:shell-scheduler/process started-at] :as instance} grace-period-secs port->reservation-atom port-grace-period-ms]
+  (if (unhealthy? instance)
+    (let [start-time (f/parse (f/formatters :date-time) started-at)
+          current-time (t/now)]
+      (if (>= (t/in-seconds (t/interval start-time current-time)) grace-period-secs)
+        (do (log/info "unhealthy instance exceeds its grace period, killing instance"
```
dposada: Nitpick: "exceeded" vs "exceeds"?

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -355,10 +372,12 @@
       (loop [remaining-service-entries (vals id->service)
              id->service' {}]
         (if-let [{:keys [service id->instance] :as service-entry} (first remaining-service-entries)]
-          (let [{:strs [health-check-url]} (:service-description service)
+          (let [{:strs [health-check-url grace-period-secs]} (:service-description service)
+                _ (log/info "service-description" (:service-description service))
```
dposada: Should this really be info level?

waiter/test/waiter/service_description_test.clj
```diff
@@ -1218,6 +1218,7 @@
         (is false "Exception should have been thrown for invalid service description")
         (catch ExceptionInfo ex
           (let [friendly-message (get-in (ex-data ex) [:friendly-error-message :env])]
+            (println friendly-message)
```
dposada: Should we remove?

waiter/test/waiter/state_test.clj
```diff
@@ -868,6 +868,10 @@
                           :failed-instances [{:message "Memory limit exceeded:" :flags #{:memory-limit-exceeded}}
                                              {:message "Memory limit exceeded:" :flags #{:memory-limit-exceeded}}]
                           :expected :not-enough-memory}
+                         {:name "health-check-misconfigured" :healthy-instances [] :unhealthy-instances []
```
dposada: I would favor either one entry per line or commas between entries if on the same line

waiter/src/waiter/marathon.clj
chzou: Even after the grace period for an unhealthy instance passes, Marathon would wait `maxConsecutiveFailures` x `intervalSeconds` = 200 seconds (3 minutes & 20 seconds) before actually failing and killing the instance. Decreasing `maxConsecutiveFailures` allows us to report back to users more quickly on these instances, and users can just extend their `gracePeriodSeconds` to compensate if they actually need that extra time (which they probably should be doing anyways!). For reference, the default `maxConsecutiveFailures` set by Marathon is 3.

dposada: Got it. Is it worth adding a comment explaining some of this?

dposada: Did we add a comment for this?

kitchen/src/kitchen/core.clj
```diff
@@ -262,6 +262,13 @@
   {:status intended-status
    :body "Health check returned bad status"})
 
+(defn sleep-handler
+  "Sleeps for given number of milliseconds"
+  [sleep-ms]
+  (Thread/sleep sleep-ms)
+  {:status 400
```
dposada: Do we always want this to return 400? Aren't there scenarios where we'd want to sleep but get back, say, 200?

kitchen/src/kitchen/core.clj
```diff
@@ -405,6 +412,8 @@
                      "/kitchen-state" (state-handler request)
                      "/pi" (pi-handler request)
                      "/request-info" (request-info-handler request)
+                     "/sleep-3" (sleep-handler 3000)        ; TODO parse for number seconds instead
```
dposada: Are we going to do this TODO in this PR?

kitchen/src/kitchen/core.clj
chzou: ok made that a second parameter!

kitchen/src/kitchen/core.clj
chzou: fixed

kitchen/src/kitchen/core.clj
```diff
@@ -394,22 +403,24 @@
...
+                     (= "/async/status" uri) (async-status-handler request)
+                     (= "/chunked" uri) (chunked-handler request)
+                     (= "/die" uri) (die-handler request)
+                     (= "/environment" uri) (environment-handler request)
+                     (= "/gzip" uri) (gzip-handler request)
+                     (= "/kitchen-state" uri) (state-handler request)
+                     (= "/pi" uri) (pi-handler request)
+                     (= "/request-info" uri) (request-info-handler request)
+                     (= "/unchunked" uri) (unchunked-handler request)
+                     (str/starts-with? uri "/sleep-") (let [uri-params (map #(Integer/parseInt %) (rest (str/split uri #"-")))]
```
sradack: The parameter should be passed as part of the query string, not encoding in the path.

sradack: See `query-params` in `async-status-handler` for an example.

After the change, the requests look like this: `/sleep?sleep-ms=400` or something like that.

waiter/src/waiter/scheduler.clj
```diff
@@ -391,54 +401,125 @@
...
+            (let [request-instances-time (t/now)
+                  active-instance-ids (->> active-instances (map :id) set)
+                  {:keys [instance-id->unhealthy-instance
+                          instance-id->tracked-failed-instance
+                          instance-id->failed-health-check-count]} (get service-id->health-check-context id)
+                  {:keys [unhealthy-instances] :as service-instance-info} (retrieve-instances-for-app id active-instances)
+                  instance-id->tracked-failed-instance' ((fnil into {}) instance-id->tracked-failed-instance
+                                                          (keep (fn [[instance-id unhealthy-instance]]
+                                                                  (when (and (not (contains? active-instance-ids instance-id))
+                                                                             (pos? (compare (get instance-id->failed-health-check-count instance-id) max-failed-health-checks)))
```
sradack: Why not `>=`?  Does this do the right thing when the values are equal?

waiter/config-full.edn
```diff
@@ -484,6 +484,14 @@
                            ;; The number of hosts on which health checks must fail in order to declare a deployment error:
                            :min-hosts 2}
 
+ :health-check-config {
+                       ;; The HTTP connect timeout and idle timeout (milliseconds) for instance health checks:
+                       :health-check-timeout-ms 200
+
+                       ;; Number of allowed health check failures before unhealthy instances are
+                       ;; flagged as never-passed-health-checks upon failure
+                       :max-failed-health-checks 5}
```
sradack: The name seems off to me.  `min-failed-health-checks` seems equally appropriate, for example.

waiter/src/waiter/scheduler.clj
```diff
@@ -391,54 +401,125 @@
...
+                                                                    [instance-id (update-in unhealthy-instance [:flags] conj :never-passed-health-checks)]))
+                                                                instance-id->unhealthy-instance))
+                  all-failed-instances (vals (merge-with (fn [failed-instance tracked-instance]
```
sradack: Much cleaner.  Thank you.

waiter/src/waiter/state.clj
```diff
@@ -1015,11 +1015,15 @@
                                     (>= (count failed-instance-hosts) min-hosts)
                                     (apply = (map :flags failed-instances))
                                     (apply = (map :exit-code failed-instances)))
+          failed-instance-flags (-> failed-instances first :flags)
           has-unhealthy-instances (and (not-empty unhealthy-instances)
                                        (apply = (map :health-check-status unhealthy-instances)))
           unhealthy-status (-> unhealthy-instances first :health-check-status)]
       (cond
-        (and has-failed-instances (-> failed-instances first :flags :memory-limit-exceeded)) :not-enough-memory
```
sradack: It's a beautiful thing to see all these checks lined up cleanly like this.

waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -43,7 +43,7 @@
   (testing-using-waiter-url
     (let [headers {:x-waiter-name (rand-name)
                    ; health check endpoint sleeps for 300 seconds (= 5 minutes)
-                   :x-waiter-health-check-url "/sleep-300"}
+                   :x-waiter-health-check-url "/sleep-300-400"}
```
dposada: What are your thoughts on doing something like `/sleep/300/return-status/400` instead?

PR #100, opened 2017-08-24
Use only 127.0.0.[1-10].  Add loopback script for Macs.

sradack:


waiter/bin/add-loopback-ips-mac.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+# Usage: add-loopback-ips-mac.sh
+#
+# Examples:
+#   add-loopback-ips-mac.sh
```
dposada: Not sure how useful the Examples section is given that there are no arguments

PR #102, opened 2017-08-24
fixed cannot-connect test

chzou:


waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -30,8 +30,8 @@
 (deftest ^:parallel ^:integration-fast test-cannot-connect
   (testing-using-waiter-url
     (let [headers {:x-waiter-name (rand-name)
-                   ; listening on invalid port ($PORT0 --> 2020)
-                   :x-waiter-cmd (kitchen-cmd "-p 2020")}
+                   ; no kitchen to connect to
```
dposada: Maybe just "nothing to connect to"?

waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -30,8 +30,8 @@
...
+                   :x-waiter-cmd "sleep 3600"}
           {:keys [headers body] :as response} (make-request-with-debug-info headers #(make-kitchen-request waiter-url %))
```
dposada: It's a little confusing that we're not using kitchen but we're calling `make-kitchen-request`

PR #107, opened 2017-08-28
Improve error message when missing environment variables.

sradack:


waiter/src/waiter/settings.clj
```diff
@@ -135,8 +135,12 @@
                                 (s/required-key :leader-latch-relative-path) schema/non-empty-string
                                 (s/required-key :mutex-timeout-ms) schema/positive-int}})
 
-(defn env [name]
-  (System/getenv name))
+(defn env [var-name config-file-path]
+  (let [value (System/getenv var-name)]
+    (when-not value
+      (throw (ex-info (format "Environment variable '%s' referenced in config file '%s' is not set." var-name config-file-path)
+                      {:var-name var-name :config-file-path config-file-path})))
```
dposada: Are there no cases where we want to support the value being `nil`?

sradack: None that I saw.  

I could rename it to env-with-error but I'm not sure that's an improvement.  It's already the `env` function in the `config` namespace.

dposada: If that case isn't needed, then :+1: and the name is :+1:

PR #108, opened 2017-08-29
Remove fail early logic

sradack:
This code did two things incorrectly:
1.  We wait the entire `queue-timeout-ms` before attempting to fail early.
2. If you assume we didn't wait for `queue-timeout-ms`, we actually fail early only if we *don't* have a more than 3 failed instances.

Worse yet, it seems that if we *do* have many failed instances, we fail to ever respond with the 503.

Also, with changes from @chzou this is obsolete anyway.

### DIFF
--- a/waiter/src/waiter/service.clj
+++ b/waiter/src/waiter/service.clj
@@ -213,8 +213,7 @@
             (statsd/histo! metric-group "get_instance" nanos))
           (counters/inc! (metrics/service-counter service-id "request-counts" "waiting-for-available-instance"))
           (statsd/gauge-delta! metric-group "request_waiting_for_instance" +1)
-          (let [expiry-time (t/plus (t/now) (t/millis queue-timeout-ms))
-                failed-instances-at-start (counters/value (metrics/service-counter service-id "instance-counts" "failed"))]
+          (let [expiry-time (t/plus (t/now) (t/millis queue-timeout-ms))]
             (loop [iterations 1]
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/service.clj
```diff
@@ -226,39 +225,37 @@
...
-                                      :slots-available (counters/value (metrics/service-counter service-id "instance-counts" "slots-available"))
-                                      :slots-in-use (counters/value (metrics/service-counter service-id "instance-counts" "slots-in-use"))
-                                      :work-stealing-offers-received (counters/value (metrics/service-counter service-id "work-stealing" "received-from" "in-flight"))
-                                      :work-stealing-offers-sent (counters/value (metrics/service-counter service-id "work-stealing" "sent-to" "in-flight"))
-                                      :status 503})))
-                        (do
-                          (app-not-found-fn)
-                          (async/<! (async/timeout 1500))
-                          (recur (inc iterations)))))))))))
+                    (if-not (t/before? (t/now) expiry-time)
```
shamsimam: Should simplify to use `(if (t/after...))`?

shamsimam:
One minor comment, else LGTM

waiter/src/waiter/service.clj
sradack: That would not be true when they are the same.  It's subtle but the test should fail if they are equal so I'd rather leave it.

PR #109, opened 2017-08-29
Parse cookies directly.  Don't use cookie store.

sradack:
Our current method utilized a method of retrieving cookies that used `CookieStore` and wasn't thread-safe.  This change parses cookies directly from the response and doesn't utilize `CookieStore`.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -559,7 +559,7 @@
                                                 "service-id" service-id})]
                   (reset! cookies-atom cookies)
                   (is (= "Added cookie x-waiter-consent" body))
-                  (is (every? (fn [h] (some #(= h (:name %1)) cookies)) ["x-waiter-auth" "x-waiter-consent"]))
+                  (is (= (set (map :name cookies)) #{"x-waiter-auth" "x-waiter-consent"}))
                   (assert-response-status response 200)))
 
               (testing "auto run-as-user population on expected service-id"
@@ -573,7 +573,8 @@
                           {:keys [run-as-user permitted-user]} service-description]
                       (reset! service-id-atom service-id)
                       (is (= "Hello World" body))
-                      (is (every? (fn [h] (some #(= h (:name %1)) cookies)) ["x-waiter-auth" "x-waiter-consent"]))
+                      ; x-waiter-consent should not be re-emitted Waiter
+                      (is (= (set (map :name cookies)) #{"x-waiter-auth"}))
                       (is (= expected-service-id service-id))
                       (is (not (str/blank? permitted-user)))
                       (is (= run-as-user permitted-user))
@@ -608,7 +609,7 @@
                   (is (not= @cookies-atom cookies))
                   (reset! cookies-atom cookies)
                   (is (= "Added cookie x-waiter-consent" body))
-                  (is (every? (fn [h] (some #(= h (:name %1)) cookies)) ["x-waiter-auth" "x-waiter-consent"]))
+                  (is (= (set (map :name cookies)) #{"x-waiter-auth" "x-waiter-consent"}))
                   (assert-response-status response 200)))
 
               (testing "auto run-as-user population on approved token"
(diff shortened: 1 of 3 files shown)


waiter/integration/waiter/token_request_test.clj
```diff
@@ -573,7 +573,8 @@
                           {:keys [run-as-user permitted-user]} service-description]
                       (reset! service-id-atom service-id)
                       (is (= "Hello World" body))
-                      (is (every? (fn [h] (some #(= h (:name %1)) cookies)) ["x-waiter-auth" "x-waiter-consent"]))
+                      ; x-waiter-consent should not be re-emitted Waiter
+                      (is (= (set (map :name cookies)) #{"x-waiter-auth"}))
```
pschorf: Why did this change?

waiter/src/waiter/client_tools.clj
```diff
@@ -191,17 +190,27 @@
...
-    (cookies/add-cookie! cs (str HTTP-SCHEME waiter-url) cookie)))
-
 (defn strip-trailing-slash
   "If s ends with /, returns s with the / stripped"
   [s]
   (cond-> s (str/ends-with? s "/") (subs 0 (- (count s) 1))))
 
+(defn parse-set-cookie-string
+  "Parse the value of a single Set-Cookie header."
+  [string-value]
```
shamsimam: Should we just use a type hint `^String` and use the name `value`?

waiter/src/waiter/client_tools.clj
```diff
@@ -191,17 +190,27 @@
...
+  (when string-value
+    (let [cookie-list (java.net.HttpCookie/parse string-value)]
```
shamsimam: Can we import `java.net.HttpCookie` and then use `(HttpCookie/parse string-value)`?

waiter/src/waiter/client_tools.clj
```diff
@@ -191,17 +190,27 @@
...
+      (map (fn [^java.net.HttpCookie cookie]
+             {:name (.getName cookie)
+              :value (.getValue cookie)}) cookie-list))))
+
+(defn parse-cookies
```
shamsimam: Should parse-cookies handle exception while parsing the set-cookie value?

waiter/src/waiter/client_tools.clj
```diff
@@ -240,7 +249,7 @@
          (when verbose
            (log/info (get request-headers "x-cid") "response size:" (count (str response-body))))
          {:request-headers request-headers
-          :cookies (cookies/get-cookies (.getCookieStore client))
+          :cookies (parse-cookies (get headers "set-cookie"))
```
shamsimam: Is `[qbits.jet.client.cookies :as cookies]` an unused import and hence should be removed?

waiter/src/waiter/client_tools.clj
shamsimam: I think we should have unit tests for `parse-cookies`.

waiter/src/waiter/client_tools.clj
sradack: Done

waiter/src/waiter/client_tools.clj
sradack: Done

waiter/src/waiter/client_tools.clj
pschorf: I don't think we need to unit tests in client tools, since it's just used in tests anyway. But if @sradack wants to add them I'm not opposed

waiter/src/waiter/client_tools.clj
sradack: Done

waiter/src/waiter/client_tools.clj
sradack: What exception did you have in mind that should be handled locally?

sradack: Added unit tests.

shamsimam: My bad, I didn't realize this was only used in the tests and not while handling requests.

shamsimam:
Waiting on green build

sradack:
I want to see why the internal tests are failing before merging.

waiter/src/waiter/client_tools.clj
```diff
@@ -143,9 +142,8 @@
 (defn make-http-client
   "Instantiates and returns a new HttpClient with a cookie store"
   []
-  (let [client (http/client)
-        cookie-store (HttpCookieStore.)
-        _ (.setCookieStore ^HttpClient client cookie-store)]
+  (let [client (http/client)]
+    (.setCookieStore client (HttpCookieStore$Empty.))
```
dposada: I think we should modify the docstring to reflect the change

sradack: Done

sradack:
Internal integration test is failing.  I will fix it after this is merged.

PR #115, opened 2017-09-05
URL->host:port for WAITER_URI if passed in that way

sradack:
This has wasted far too much of our time.

### DIFF
--- a/waiter/src/waiter/client_tools.clj
+++ b/waiter/src/waiter/client_tools.clj
@@ -71,9 +71,19 @@
       (str username "." machine-name))
     "127.0.0.1"))
 
+(defn sanitize-waiter-url
+  "Be kind to the tester, extracting the host:port of the URL if passed in as
+  a URL instead of host:port."
+  [waiter-uri]
+  (if (and waiter-uri (str/starts-with? waiter-uri "http://"))
+    (-> (URI. waiter-uri)
+        (.getAuthority))
+    waiter-uri))
+
 (defn retrieve-waiter-url []
   {:post [%]}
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/client_tools.clj
```diff
@@ -71,9 +71,19 @@
       (str username "." machine-name))
     "127.0.0.1"))
 
+(defn sanitize-waiter-url
+  "Be kind to the tester, extracting the host:port of the URL if passed in as
+  a URL instead of host:port."
+  [waiter-uri]
+  (if (and waiter-uri (str/starts-with? waiter-uri "http://"))
+    (-> (URI. waiter-uri)
+        (.getAuthority))
```
dposada: How about

```clj
(-> waiter-uri URI. .getAuthority)
```

waiter/src/waiter/client_tools.clj
```diff
@@ -71,9 +71,19 @@
...
+    waiter-uri))
+
 (defn retrieve-waiter-url []
   {:post [%]}
-  (let [waiter-uri (System/getenv "WAITER_URI")]
+  (let [waiter-uri (-> (System/getenv "WAITER_URI")
+                       sanitize-waiter-url)]
```
dposada: How about 

```clj
(-> "WAITER_URI" System/getenv sanitize-waiter-url)
```

waiter/src/waiter/client_tools.clj
sradack: I like it.

waiter/src/waiter/client_tools.clj
sradack: I like it.

PR #117, opened 2017-09-08
Introduce TokenStore protocol

shamsimam:
This change will have a follow-up PR that introduces pluggability from the config using our Custom Extensions support.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -163,7 +163,6 @@
                   (cid/ensure-correlation-id nested-response get-request-cid)
                   nested-response)))))))))
 
-
 (defn- make-blacklist-request
   [make-inter-router-requests-fn blacklist-period-ms dest-router-id dest-endpoint {:keys [id] :as instance} reason]
   (log/info "peer communication requesting" dest-router-id "to blacklist" id "via endpoint" dest-endpoint)
@@ -690,6 +689,10 @@
    :token->token-description (pc/fnk [[:curator kv-store]]
                                (fn token->token-description [token]
                                  (sd/token->token-description kv-store token)))
+   :token-store (pc/fnk [[:curator kv-store]
+                         [:state clock]
+                         synchronize-fn]
+                  (token/->KeyValueTokenStore clock synchronize-fn kv-store))
    :validate-service-description-fn (pc/fnk [[:state service-description-builder]]
                                       (fn validate-service-description [service-description]
                                         (sd/validate service-description-builder service-description {})))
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/core.clj
```diff
@@ -830,7 +833,7 @@
                                 (async/tap state-chan-mult state-chan)
                                 (state/start-service-chan-maintainer
                                   {} instance-rpc-chan state-chan query-app-maintainer-chan start-service remove-service retrieve-channel)))
-   :state-query-chans (pc/fnk [[:state query-app-maintainer-chan scheduler]
+   :state-query-chans (pc/fnk [[:state query-app-maintainer-chan]
```
shamsimam: Minor cleanup of unused variable.

waiter/src/waiter/core.clj
```diff
@@ -163,7 +163,6 @@
                   (cid/ensure-correlation-id nested-response get-request-cid)
                   nested-response)))))))))
 
-
```
shamsimam: Removed extra line (unrelated to this PR)

waiter/src/waiter/token.clj
```diff
@@ -157,6 +169,25 @@
                   token-set (->> tokens (map :token) set)]
               (kv/store kv-store owner-key token-set))))))))
 
+;; A token store which is backed by a key-value store.
+(defrecord KeyValueTokenStore [clock synchronize-fn kv-store]
+
+  TokenStore
+
+  (fetch [_ token include-deleted?]
+    (sd/token->token-description kv-store token :include-deleted include-deleted?))
```
sradack: Inline `sd/token->token-description`

waiter/src/waiter/token.clj
```diff
@@ -157,6 +169,25 @@
...
+
+  (store [_ token service-description-template token-metadata]
+    ;; insert the last-update-time if it doesn't exist
+    (let [token-metadata' (merge {"last-update-time" (.getMillis ^DateTime (clock))} token-metadata)]
+      (store-service-description-for-token synchronize-fn kv-store token service-description-template token-metadata')))
```
sradack: Inline `store-service-description-for-token`

waiter/src/waiter/token.clj
```diff
@@ -157,6 +169,25 @@
...
+
+  (delete [this token hard-delete?]
+    (let [{:keys [token-metadata]} (fetch this token hard-delete?)]
+      (if-let [owner (get token-metadata "owner")]
+        (delete-service-description-for-token clock synchronize-fn kv-store token owner :hard-delete hard-delete?)
```
sradack: Inline `delete-service-description-for-token`

waiter/src/waiter/token.clj
shamsimam: Inlining requires moving the procol inside the `let` block that defines `token-lock`. Protocols don't appear to recognize variables defined in let blocks, I get the following error:
`Unable to resolve symbol: token-lock in this context, compiling:(waiter/token.clj:138:9)`.

I wrote this simple example and had the same error:
```
(defprotocol P (foo [this]))
(let [x 1]
  (defrecord R [] P (foo [_] (println x))))
=> 
CompilerException java.lang.RuntimeException: Unable to resolve symbol: x in this context, compiling:(/private/var/folders/wj/c9w0p0r93vl4rl815rspwxqw0000gn/T/form-init7112261126071510316.clj:3:30) 
```

waiter/src/waiter/token.clj
shamsimam: Same `token-lock` issue.

waiter/src/waiter/token.clj
shamsimam: `sd/token->token-description` and `sd/token->service-description-template` are closely related.

PR #125, opened 2017-09-09
Fix scaling in ShellScheduler.  Fix forced killing in MarathonScheduler.  Re-enable autoscaling tests.

sradack:


.travis.yml
```diff
@@ -12,6 +12,7 @@ env:
     - LEIN_GET='wget https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein'
     - LEIN_CHMOD='chmod +x ./lein'
     - LEIN_DOWN='lein downgrade 2.6.1'
+    - _JAVA_OPTIONS=''
```
dposada: Can we add a comment about why this is needed?

sradack:
This PR fails internally because `force-kill-after-ms` is set to 15 minutes.  That value needs to be adjusted, and once adjusted, this `test-scaling-unhealthy-app` should pass.

sradack:
RTM

waiter/integration/waiter/autoscaling_test.clj
```diff
@@ -15,55 +15,41 @@
...
-;   test-scaling-healthy-app Scaling healthy app
-;   Never scaled up to 5/2 task(s)
-;   expected: nil
-;     actual: nil
-(deftest ^:parallel ^:integration-slow ^:explicit test-scaling-healthy-app
+(deftest ^:parallel ^:integration-slow test-scaling-healthy-app
   (testing-using-waiter-url
-    (log/info (str "Scaling healthy app test (should take " (colored-time "~2 minutes") ")"))
     (let [custom-headers {:x-kitchen-delay-ms 5000
+                          :x-waiter-concurrency-level 3
```
shamsimam: Should we extract `concurrency-level` in a variable since it is being passed to `scaling-for-service-test`?

waiter/integration/waiter/autoscaling_test.clj
```diff
@@ -15,55 +15,41 @@
...
-;   FAIL in (test-scaling-unhealthy-app)
-;   Never scaled down to 1 task
-;   expected: nil
-;     actual: nil
-(deftest ^:parallel ^:integration-slow ^:explicit test-scaling-unhealthy-app
+(deftest ^:parallel ^:integration-slow test-scaling-unhealthy-app
   (testing-using-waiter-url
-    (log/info (str "Scaling unhealthy app test (should take " (colored-time "~2 minutes") ")"))
-    (let [custom-headers {:x-waiter-scale-up-factor 0.9
+    (let [custom-headers {:x-waiter-concurrency-level 3
```
shamsimam: Should we extract `concurrency-level` in a variable since it is being passed to `scaling-for-service-test`?

waiter/integration/waiter/autoscaling_test.clj
```diff
@@ -15,55 +15,41 @@
...
-        (future (dorun (pmap (fn [_] (dotimes [_ requests-per-thread] (request-fn))) (range threads))))
-        (expect-tasks-running service-id #(> % (/ threads 2)) (str "Never scaled up to " (/ threads 2) " task(s)"))
-        (expect-tasks-running service-id #(= % 1) "Never scaled down to 1 task")
-        (delete-service waiter-url service-id)))))
+    (let [continue-running (atom true)
+          first-request (request-fn)
+          service-id (retrieve-service-id waiter-url (:request-headers first-request))
+          count-instances (fn [] (num-instances waiter-url service-id))]
+      (is (wait-for #(= 1 (count-instances))) "First instance never started")
+      (future (dorun (pmap (fn [_] (while @continue-running (request-fn))) (range (* target-instances concurrency-level)))))
```
shamsimam: We can use `parallelize-requests` here with the `cenceled?` function passed in:

```(parallelize-requests 
  (* target-instances concurrency-level)
  1
  request-fn
  :canceled? (fn [] (not @continue-running))
  :wait-for-tasks false)```

File: waiter/integration/waiter/autoscaling_test.clj [hunk 5]
```diff
@@ -15,55 +15,41 @@
...
-            service-id (retrieve-service-id waiter-url (:request-headers first-request))]
-        (expect-tasks-running service-id #(= % 1) "Never saw initial 1 task running")
-        (future (dorun (pmap (fn [_] (dotimes [_ requests-per-thread] (request-fn))) (range threads))))
-        (expect-tasks-running service-id #(> % (/ threads 2)) (str "Never scaled up to " (/ threads 2) " task(s)"))
-        (expect-tasks-running service-id #(= % 1) "Never scaled down to 1 task")
-        (delete-service waiter-url service-id)))))
+    (let [continue-running (atom true)
+          first-request (request-fn)
+          service-id (retrieve-service-id waiter-url (:request-headers first-request))
+          count-instances (fn [] (num-instances waiter-url service-id))]
```
shamsimam: `num-instances` can still suffer from flakes if marathon gets busy. An alternate approach might be to take look at `[:autoscaler-state :target-instances]` from `/state/<service-id>` endpoint.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -223,13 +248,14 @@
     (if (contains? id->service service-id)
       (let [{:keys [id->instance]} (get id->service service-id)
             {:keys [:shell-scheduler/process] :as instance} (get id->instance instance-id)]
-        (if instance
+        (if (and instance (active? instance))
```
shamsimam: Why restrict to killing only active instances?

.travis.yml
shamsimam: This change got removed?

waiter/integration/waiter/autoscaling_test.clj
sradack: Honestly I think this is much clearer.

waiter/integration/waiter/autoscaling_test.clj
sradack: If this had been the test code, then the test would have passed with Marathon even though nothing was being scaled down.

waiter/integration/waiter/autoscaling_test.clj
sradack: I don't follow.  Can you state another way?

waiter/src/waiter/shell_scheduler.clj
sradack: We're in `kill-instance`, and something is only `active?` if it's not killed.

waiter/integration/waiter/autoscaling_test.clj
shamsimam: I think he issue is that Marathon might not scale up to `target-instances` if it is low on free resources.

waiter/integration/waiter/autoscaling_test.clj
shamsimam: Are the 3 in `x-waiter-concurrency-level 3` and second 3 in `scaling-for-service-test "Scaling healthy app" waiter-url 3 3` related?

waiter/src/waiter/shell_scheduler.clj
shamsimam: okay

waiter/integration/waiter/autoscaling_test.clj
sradack: That's one of the reasons I changed it to 3 instead of 5.  I'd like to avoid another future regression. I'm inclined to keep this a "full" integration test, including making sure Marathon scales all the way up and all the way down, with real tasks.

PR #129, opened 2017-09-25
Additional logging in the scheduler

shamsimam:


waiter/src/waiter/scheduler.clj
```diff
@@ -406,7 +406,12 @@
            service->service-instances' {}]
       (if-not service
         service->service-instances'
-        (let [active-instances (doall (map async/<!! active-instances))]
+        (let [start-time-ms (-> (t/now) .getMillis)
```
sradack: We have a timer called `app->available-tasks`.  It's not the greatest name, but why don't we just consult the timer data?

shamsimam: The main reason is the logging per service lets us get incremental logging updates for slow services vs the `app->available-tasks` timer which is for all services at once. I will switch to introducing a new metric per service and a new timer function like `start-stop-time!` to log when it is too slow.

shamsimam: I removed the code and added a timeout to the do-health-checks call and add a log if it takes too long.

waiter/src/waiter/scheduler.clj
```diff
@@ -401,15 +400,25 @@
...
-          (recur rest (assoc service->service-instances'
-                        service
-                        (assoc instances :active-instances active-instances))))))))
+  (let [timeout-threshold-mins 2
+        timeout-channel (async/timeout (-> timeout-threshold-mins t/minutes t/in-millis))
+        slowness-monitor-promise (promise)]
+    (async/go
+      (async/<! timeout-channel)
+      (when (not (realized? slowness-monitor-promise))
+        (log/info "scheduler-syncer: health checks are taking longer than" timeout-threshold-mins "minute(s) to complete!")))
```
sradack: This should be at least `warn`.

sradack: Nitpick but I'm not a fan of exclamation points in error messages.

shamsimam:
RTM

PR #132, opened 2017-09-26
Fix flaky unit tests

sradack:
Fixes the race between agent updating and asking for instances.


shamsimam:
Will wait for the build to go green.

PR #133, opened 2017-09-27
Refactoring - split token handler for delete, get, and post requests

shamsimam:


sradack:
Nice!

waiter/src/waiter/token.clj
```diff
@@ -168,144 +315,13 @@
...
-                                     :body (json/write-str {:token token, :owner owner}))
-              (utils/map->json-response {:message (str "Successfully created " token)
-                                         :service-description new-service-description-template}))
-            (catch Exception ex
-              (utils/exception->response ex req)))))
+    :delete (handle-token-delete-request clock synchronize-fn kv-store waiter-hostname entitlement-manager
+                                         make-peer-requests-fn req)
+    :get (handle-get-token-request kv-store waiter-hostname req)
+    :post (handle-post-token-request clock synchronize-fn kv-store waiter-hostname entitlement-manager
+                                     make-peer-requests-fn validate-service-description-fn req)))
```
sradack: `:else` throw status 405

shamsimam: I'll add it in a follow-up MR please.

PR #134, opened 2017-09-27
Optimistic concurrency control in token modification

shamsimam:
Handle concurrency issues in token update/delete.

Inspiration: http://fideloper.com/etags-and-optimistic-concurrency-control

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -13,6 +13,7 @@
             [clojure.string :as str]
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
+            [clojure.walk :as walk]
             [plumbing.core :as pc]
             [qbits.jet.client.http :as http]
             [waiter.client-tools :refer :all])
@@ -106,12 +107,31 @@
 
       (log/info "creating the tokens")
       (doseq [token tokens-to-create]
-        (let [response (post-token waiter-url {:health-check-url "/probe"
+        (let [response (post-token waiter-url {:health-check-url "/check"
                                                :name service-id-prefix
                                                :token token})]
           (assert-response-status response 200)))
 
-      (testing "token retrieval with etag"
+      (testing "update without etags"
+        (doseq [token tokens-to-create]
+          (let [response (post-token waiter-url {:health-check-url "/health"
+                                                 :name service-id-prefix
+                                                 :token token})]
+            (assert-response-status response 200))))
+
+      (testing "update with etags"
+        (doseq [token tokens-to-create]
+          (let [{:keys [body headers]} (get-token waiter-url token)
+                token-description (-> body
+                                      json/read-str
+                                      walk/keywordize-keys
+                                      (select-keys [:name])
+                                      (assoc :health-check-url "/probe"
+                                             :token token))
+                response (post-token waiter-url token-description :headers {"if-match" (get headers "etag")})]
+            (assert-response-status response 200))))
+
+      (testing "token retrieval - presence of etag header"
         (doseq [token tokens-to-create]
           (let [{:keys [body headers] :as response} (get-token waiter-url token :cookies cookies)]
             (assert-response-status response 200)
@@ -290,14 +310,17 @@
             (try
               (testing "successful admin mode update"
                 (log/info "creating configuration using token" token)
-                (let [token-description {:health-check-url "/probe-2"
-                                         :last-update-time last-update-time
-                                         :name service-id-prefix
-                                         :owner (retrieve-username)
-                                         :run-as-user "i-do-not-exist-but-will-not-be-checked"
-                                         :token token}
+                (let [{:keys [body headers]} (get-token waiter-url token)
+                      token-description (-> body
+                                            json/read-str
+                                            walk/keywordize-keys
+                                            (assoc :health-check-url "/probe-2"
+                                                   :last-update-time last-update-time
+                                                   :owner (retrieve-username)
+                                                   :run-as-user "i-do-not-exist-but-will-not-be-checked"
+                                                   :token token))
                       response (post-token waiter-url token-description
-                                           :headers (attach-token-etag waiter-url token {})
+                                           :headers {"if-match" (get headers "etag")}
                                            :query-params {"update-mode" "admin"})]
                   (assert-response-status response 200)
                   (log/info "created configuration using token" token))
@@ -348,15 +371,24 @@
                 token (create-token-name waiter-url service-id-prefix)]
             (try
               (log/info "creating configuration using token" token)
-              (let [token-description {:deleted true
-                                       :health-check-url "/probe"
-                                       :last-update-time last-update-time
+              (let [token-description {:health-check-url "/probe-3"
                                        :name service-id-prefix
                                        :owner (retrieve-username)
-                                       :run-as-user "foo-bar"
                                        :token token}
+                    response (post-token waiter-url token-description :headers {"host" token})]
+                (assert-response-status response 200))
+
+              (let [{:keys [body headers]} (get-token waiter-url token)
+                    token-description (-> body
+                                          json/read-str
+                                          walk/keywordize-keys
+                                          (assoc :deleted true
+                                                 :health-check-url "/probe"
+                                                 :last-update-time last-update-time
+                                                 :run-as-user "foo-bar"
+                                                 :token token))
                     response (post-token waiter-url token-description
-                                         :headers (attach-token-etag waiter-url token {})
+                                         :headers {"if-match" (get headers "etag")}
                                          :query-params {"update-mode" "admin"})]
                 (assert-response-status response 200))
               (log/info "created configuration using token" token)
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -152,6 +152,26 @@
             (assert-response-status tokens-response 200)
             (is (not-any? (fn [token-entry] (= token (get token-entry "token"))) tokens)))))
 
+      (testing "hard-delete without etag"
+        (doseq [token tokens-to-create]
+          (let [{:keys [body] :as response} (make-request waiter-url "/token"
+                                                          :headers {"host" token}
+                                                          :http-method-fn http/delete
+                                                          :query-params {"hard-delete" true})]
+            (assert-response-status response 400)
+            (is (str/includes? (str body) "Must specify if-match header for token hard deletes")))))
+
+      (testing "hard-delete with invalid etag"
+        (doseq [token tokens-to-create]
+          (let [{:keys [body] :as response} (make-request waiter-url "/token"
+                                                          :headers {"host" token
+                                                                    "if-match" "1010"}
+                                                          :http-method-fn http/delete
+                                                          :query-params {"hard-delete" true})]
+            (println body)
+            (assert-response-status response 412)
+            (is (str/includes? (str body) "Cannot modify stale token")))))
+
       (log/info "hard-deleting the tokens")
       (doseq [token tokens-to-create]
         (delete-token-and-assert waiter-url token))
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -295,6 +295,7 @@
                                        :run-as-user "i-do-not-exist-but-will-not-be-checked"
                                        :token token}
                     {:keys [body] :as response} (post-token waiter-url token-description
+                                                            :headers (attach-token-etag waiter-url token {})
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/client_tools.clj
```diff
@@ -574,7 +574,7 @@
   [waiter-url token & {:keys [hard-delete] :or {hard-delete true}}]
   (log/info "deleting token" token)
   (let [response (make-request waiter-url "/token"
-                               :headers {"host" token}
+                               :headers {"host" token, "if-match" (str (System/currentTimeMillis))}
```
sradack: Delete should not require an `If-Match` so this should be removed.

waiter/src/waiter/token.clj
```diff
@@ -69,18 +83,22 @@
 
   (defn delete-service-description-for-token
     "Delete a token from the KV"
-    [clock synchronize-fn kv-store token owner & {:keys [hard-delete] :or {hard-delete false}}]
+    [clock synchronize-fn kv-store token owner &
+     {:keys [hard-delete update-time] :or {hard-delete false, update-time Long/MAX_VALUE}}]
```
sradack: `if-unmodified-since` would be a better name.  `update-time` implies that is the timestamp that should be written for `last-update-time`.

waiter/src/waiter/client_tools.clj
shamsimam: Deletes require the `if-match`|`if-unmodified-since` header since they can also participate in the concurrency race.

sradack: Making them required would break everyone who is calling `POST` and `DELETE` that isn't the sync tool.  I'd like to see that our integration tests ensure that the headers are not required.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -267,7 +287,9 @@
                                        :owner (retrieve-username)
                                        :run-as-user "i-do-not-exist-but-will-not-be-checked"
                                        :token token}
-                    response (post-token waiter-url token-description :query-params {"update-mode" "admin"})]
+                    response (post-token waiter-url token-description
+                                         :headers {"if-unmodified-since" (str last-update-time)}
```
sradack: The timestamp should conform to the spec.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Unmodified-Since

sradack: There's lots of instances to fix.  I'm only annotating this one.

waiter/src/waiter/token.clj
```diff
@@ -170,17 +206,23 @@
...
+                                     :user authenticated-user}))))
                 (when-not (authz/manage-token? entitlement-manager authenticated-user token token-metadata)
                   (throw (ex-info "User not allowed to delete token"
                                   {:owner token-owner
                                    :status 403
                                    :user authenticated-user}))))
-              (delete-service-description-for-token clock synchronize-fn kv-store token token-owner :hard-delete hard-delete)
+              (delete-service-description-for-token clock synchronize-fn kv-store token token-owner
+                                                    :hard-delete hard-delete
+                                                    :if-unmodified-since (compute-last-modified clock headers))
```
sradack: `delete-service-description-for-token` should take a `long`, but it should be parsed from the header which conforms to the time spec.

waiter/src/waiter/token.clj
```diff
@@ -206,7 +248,8 @@
           ;;NB do not ever return the password to the user
           (do
             (log/info "successfully retrieved token " token)
-            (utils/map->json-response (merge service-description-template token-metadata)))
+            (utils/map->json-response (merge service-description-template token-metadata)
+                                      :headers {"last-modified" (-> token-metadata (get "last-update-time") str)}))
```
sradack: The timestamp in the `Last-Modifieid` header also needs to conform:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

waiter/src/waiter/token.clj

waiter/integration/waiter/token_request_test.clj

waiter/src/waiter/token.clj

waiter/src/waiter/client_tools.clj
shamsimam: It is only required when providing the hard-delete, changed the code to reflect this.

waiter/src/waiter/token.clj

shamsimam:
RTM

waiter/src/waiter/utils.clj
```diff
@@ -472,3 +473,17 @@
   "Parses the request header to determine if debug mode has been enabled."
   [request]
   (boolean (get-in request [:headers "x-waiter-debug"])))
+
+(let [last-modified-format "EEE, dd MMM yyyy HH:mm:ss zzz"]
```
shamsimam: @sradack  One issue with this format is that we can get only upto second granularity (instead of millisecond granularity we have in our timestamps). Do we want to make this sacrifice in accuracy?

shamsimam: "If sub-second accuracy is important, it may be more appropriate to use entity tags (ETag)."

https://stackoverflow.com/questions/18603154/http-if-modified-since-with-milliseconds

sradack:
If we’re going to use a standardized header we should follow the spec.

sradack:
Maybe ETags is better.  The sync algorithm is time-based so Last-Modified
appealed to me.

How do/would we handle syncing to a router that doesn’t already know about
the token?

shamsimam:
> How do/would we handle syncing to a router that doesn’t already know about the token?

For that router, sync-ing is equivalent to a token create. The timestamp check will be ignored.

sradack:
> For that router, sync-ing is equivalent to a token create. The timestamp check will be avoided.
But now you have a race again, no?

shamsimam:
RTM

waiter/src/waiter/client_tools.clj
```diff
@@ -570,11 +568,29 @@
          test-prefix (System/getenv "WAITER_TEST_PREFIX")]
      (str/replace (str test-prefix service-name username (rand-int 3000000)) #"-" ""))))
 
-(defn delete-token-and-assert
-  [waiter-url token & {:keys [hard-delete] :or {hard-delete true}}]
-  (log/info "deleting token" token)
+(defn token->etag
```
sradack: Typical client behavior:
 1.  Get token data
 2.  Inspect token data, making changes after having observed the token data
 3.  Post the modifications along with the ETag received in Step 1.

Having a function like `attach-token-etag` means we don't act like a client.  `attach-token-etag` feels like it's being used to simply blow away/force-write the token data intentionally avoiding the concurrency mechanism, and hence, I'm not a fan.

waiter/src/waiter/token.clj
```diff
@@ -23,6 +23,25 @@
...
+  (defn- token-metadata->etag
+    "Converts the token metadata to a tag"
+    [token-metadata]
+    (-> token-metadata
+        (get "last-update-time")
+        (or default-etag)
+        str)))
+
+(defn- validate-token-modification-based-on-etag
+  "Validates whether the token modification should be allowed on the update-time-etag (timestamp)."
```
sradack: What is an `update-time-etag`? The Etag should be opaque to this function.

waiter/src/waiter/token.clj
```diff
@@ -46,20 +65,24 @@
 
   (defn store-service-description-for-token
     "Store the token mapping of the service description template in the key-value store."
-    [synchronize-fn kv-store ^String token service-description-template token-metadata]
+    [synchronize-fn kv-store ^String token service-description-template token-metadata &
+     {:keys [hard-delete version-etag] :or {hard-delete false}}]
```
sradack: Why is `hard-delete` a parameter?

waiter/src/waiter/token.clj
```diff
@@ -157,6 +184,12 @@
                   token-set (->> tokens (map :token) set)]
               (kv/store kv-store owner-key token-set))))))))
 
+(defn compute-last-modified-etag
```
sradack: This function doesn't compute anything and the comment doesn't describe what it does.  Please just inline.

waiter/src/waiter/token.clj
```diff
@@ -321,7 +371,8 @@
                                          make-peer-requests-fn req)
     :get (handle-get-token-request kv-store waiter-hostname req)
     :post (handle-post-token-request clock synchronize-fn kv-store waiter-hostname entitlement-manager
-                                     make-peer-requests-fn validate-service-description-fn req)))
+                                     make-peer-requests-fn validate-service-description-fn req)
+    (utils/map->json-response {:unsupported-method request-method} :status 405)))
```
sradack: Please use a `try`, `throw` an exception instead, and use `exception->response` in the `catch` block.  Also, please create an integration test.

waiter/src/waiter/utils.clj
```diff
@@ -154,10 +154,10 @@
 
 (defn map->json-response
   "Convert the input data into a json response."
-  [data-map & {:keys [status] :or {status 200}}]
+  [data-map & {:keys [headers status] :or {headers {} status 200}}]
```
sradack: No longer needed after address my comments above.

waiter/test/waiter/token_test.clj
```diff
@@ -43,6 +44,12 @@
   (handle-token-request clock synchronize-fn kv-store waiter-hostname entitlement-manager make-peer-requests-fn
                         validate-service-description-fn request))
 
+(deftest test-compute-last-modified-etag
```
sradack: Please remove.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -291,7 +333,9 @@
                                        :owner (retrieve-username)
                                        :run-as-user "foo-bar"
                                        :token token}
-                    response (post-token waiter-url token-description :query-params {"update-mode" "admin"})]
+                    response (post-token waiter-url token-description
```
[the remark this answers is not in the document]
sradack: Please change the test to:
1. Create a token normally.
 2. Retrieve the token and ETag
3. Hard-delete the token with the ETag

waiter/integration/waiter/token_request_test.clj
```diff
@@ -145,6 +152,25 @@
             (assert-response-status tokens-response 200)
             (is (not-any? (fn [token-entry] (= token (get token-entry "token"))) tokens)))))
 
+      (testing "hard-delete without etag"
```
sradack: These tests belong alongside the delete inside the admin functionality test.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -267,7 +307,9 @@
                                        :owner (retrieve-username)
                                        :run-as-user "i-do-not-exist-but-will-not-be-checked"
                                        :token token}
-                    response (post-token waiter-url token-description :query-params {"update-mode" "admin"})]
```
sradack: We never create tokens in admin mode.  The test will be better when we structured it to:
1. Create a token normally
2. Get the token data and ETag
3. Update the token using the ETag

sradack: We should also test a token update with a stale ETag..

waiter/integration/waiter/token_request_test.clj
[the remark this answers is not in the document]
sradack: We should also test a hard-delete on a stale ETag.

waiter/src/waiter/client_tools.clj
shamsimam: Done.

shamsimam:
RTM

waiter/src/waiter/token.clj
```diff
@@ -157,152 +183,162 @@
...
       ; notify peers of token update
       (make-peer-requests-fn "tokens/refresh"
                              :method :post
                              :body (json/write-str {:token token, :owner owner}))
       (utils/map->json-response {:message (str "Successfully created " token)
-                                 :service-description new-service-description-template}))
-    (catch Exception ex
-      (utils/exception->response ex req))))
+                                 :service-description new-service-description-template}
+                                :headers {"etag" (token-metadata->etag new-token-metadata)}))))
```
sradack: Formatting

waiter/src/waiter/utils.clj
sradack: BTW I saw a pattern I liked.  Not that this needs to change, but:

<pre>
(-> (map->json-response {})
    (rr/header "x-header" "value")
    (rr/status 403))
</pre>
 
Instead of pushing everything we need into this function, just wrap it.  Had to get it off my chest.

PR #135, opened 2017-09-28
Safer way of calculating memory usage

sradack:
This PR avoids race that causes shell scheduler's unit tests to fail.

### Previous approach
Calculate the memory usage by summing usage of all processes in the subprocess' process group.

The race:
1. Start process (`setsid bash ...`)
2. Check memory usage of `pgid`
3. `setsid` hasn't changed the process group yet.

Since the process hasn't changed groups, the memory calculation includes all ancestor processes sharing the process group.  And since that number is much larger, the processes are killed due to using too much memory, and test assertions fail.

### New approach

Build a tree of processes, and when calculating memory, include strictly only the subprocess and its children.

### DIFF
--- a/waiter/integration/waiter/deployment_errors_test.clj
+++ b/waiter/integration/waiter/deployment_errors_test.clj
@@ -9,11 +9,21 @@
 ;;       actual or intended publication of such source code.
 ;;
 (ns waiter.deployment-errors-test
-  (:require [clojure.string :as str]
+  (:require [clojure.pprint :as pprint]
+            [clojure.string :as str]
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
+            [plumbing.core :as pc]
             [waiter.client-tools :refer :all]
             [waiter.settings :as settings]))
+
+(defn get-router->service-state
+  [waiter-url service-id]
+  (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")]
+    (->> (routers waiter-url)
+         (map (fn [[_ router-url]]
+                (service-state router-url service-id :cookies cookies)))
+         (pc/map-from-keys :router-id))))
 
 (deftest ^:parallel ^:integration-slow test-invalid-health-check-response
   (testing-using-waiter-url
@@ -23,10 +33,18 @@
                    :x-waiter-queue-timeout 600000}
           {:keys [headers body] :as response} (make-request-with-debug-info headers #(make-kitchen-request waiter-url %))
           service-id (get headers "x-waiter-service-id")]
-      (is (not (nil? service-id)))
-      (assert-response-status response 503)
-      (is (str/includes? body (str "Deployment error: " (-> (waiter-settings waiter-url) :messages :invalid-health-check-response))))
-      (delete-service waiter-url service-id))))
+      (try
+        (assert-response-status response 503)
+        (is service-id)
+        (is (get-router->service-state waiter-url service-id))
+        (is (str/includes? body (str "Deployment error: " (-> (waiter-settings waiter-url)
+                                                              :messages
+                                                              :invalid-health-check-response)))
+            (str "router->service-state:\n" (with-out-str
+                                              (pprint/pprint
+                                                (get-router->service-state waiter-url service-id)))))
+        (finally
+          (delete-service waiter-url service-id))))))
 
 (deftest ^:parallel ^:integration-slow test-cannot-connect
   (testing-using-waiter-url
@@ -36,10 +54,18 @@
                    :x-waiter-queue-timeout 600000}
           {:keys [headers body] :as response} (make-request-with-debug-info headers #(make-shell-request waiter-url %))
           service-id (get headers "x-waiter-service-id")]
-      (is (not (nil? service-id)))
-      (assert-response-status response 503)
-      (is (str/includes? body (str "Deployment error: " (-> (waiter-settings waiter-url) :messages :cannot-connect))))
-      (delete-service waiter-url service-id))))
+      (try
+        (assert-response-status response 503)
+        (is service-id)
+        (is (get-router->service-state waiter-url service-id))
+        (is (str/includes? body (str "Deployment error: " (-> (waiter-settings waiter-url)
+                                                              :messages
+                                                              :cannot-connect)))
+            (str "router->service-state:\n" (with-out-str
+                                              (pprint/pprint
+                                                (get-router->service-state waiter-url service-id)))))
+        (finally
+          (delete-service waiter-url service-id))))))
 
 (deftest ^:parallel ^:integration-slow test-health-check-timed-out
   (testing-using-waiter-url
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/shell_scheduler.clj
```diff
@@ -358,42 +358,62 @@
...
-    (let [ps-output (-> (sh/sh "ps" "-eo" "pid,pgid,rss") :out (str/split #"\n") rest)
-          pid->pgid (map #(-> % str/trim (str/split #"\s+") butlast
-                              ((fn [[a b]] [(Integer/parseInt a) (Integer/parseInt b)])))
-                         ps-output)
-          pgid->rss (apply merge-with +
-                           (map #(-> % str/trim (str/split #"\s+") rest
-                                     ((fn [[a b]] {(Integer/parseInt a) (Integer/parseInt b)})))
-                                ps-output))]
-      (into {} (for [[pid pgid] pid->pgid] [pid (get pgid->rss pgid)])))
+    (let [cols->entry (fn [[pid ppid rss]] {:pid (Integer/parseInt pid)
```
shamsimam: We can use `zipmap` here: `(fn [elements] (zipmap [: pid :ppid :rss] elements)`

waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -23,10 +33,18 @@
                    :x-waiter-queue-timeout 600000}
           {:keys [headers body] :as response} (make-request-with-debug-info headers #(make-kitchen-request waiter-url %))
           service-id (get headers "x-waiter-service-id")]
-      (is (not (nil? service-id)))
-      (assert-response-status response 503)
-      (is (str/includes? body (str "Deployment error: " (-> (waiter-settings waiter-url) :messages :invalid-health-check-response))))
-      (delete-service waiter-url service-id))))
+      (try
```
shamsimam: We should have an `assert-deployment-error` macro for this.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -358,42 +358,62 @@
...
-                                   (assoc :id->instance id->instance')
-                                   update-task-stats)]
-            (recur (rest remaining-service-entries) (assoc id->service' (:id service) service-entry')))
-          id->service')))))
+  (try
+    (timers/start-stop-time!
+      (metrics/waiter-timer "shell-scheduler" "update-health")
+      (let [pid->memory (get-pid->memory)
+            exit-codes-check #(associate-exit-codes % port->reservation-atom port-grace-period-ms)]
+        (loop [remaining-service-entries (vals id->service)
```
shamsimam: We can use `reduce` here.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -358,42 +358,62 @@
     instance))
 
 (defn- get-pid->memory
```
shamsimam: We should have unit tests for this by mocking the response to the `sh/sh ps` calls.

waiter/src/waiter/shell_scheduler.clj
```diff
@@ -345,7 +345,7 @@
   [{:keys [:shell-scheduler/process :shell-scheduler/pid] :as instance} mem pid->memory port->reservation-atom port-grace-period-ms]
   (if (and (active? instance) (.isAlive process))
     (let [memory-allocated (* mem 1000)
```
shamsimam: Should this be `1024`?

shamsimam: Also, we should include the units in the names, like `-mb` and `-kb` suffixes.

waiter/src/waiter/shell_scheduler.clj
sradack: Where were you planning on putting the `Integer/parseInt` calls?

waiter/integration/waiter/deployment_errors_test.clj
sradack: I couldn't get this to work:
<pre>
(defmacro assert-deployment-error
  [waiter-url {:keys [headers body] :as response} message-key]
  `(let [service-id# (get ~headers "x-waiter-service-id")    
         service-state# (get-router->service-state ~waiter-url service-id#)]
     (is service-id#)
     (is service-state#)
     (assert-response-status ~response 503)
     (is (str/includes? ~body (str "Deployment error: " (-> (waiter-settings ~waiter-url)
                                                            :messages
                                                            ~message-key)))
         (str "router->service-state:\n" (with-out-str (pprint/pprint service-state#))))))
</pre>

For some reason the `headers` and `body` were nil.  It worked fine if I avoided the destructuring and passed them as separate params.  If you can figure out, I'm happy to change it.  I'm not keen on factoring tests too much in the first place.

waiter/src/waiter/shell_scheduler.clj
sradack: Adding the units for memory doesn't flow like adding units to time.

Good: 
 - idle-timeout-mins
 - grace-period-secs

Meh: 
 - memory-used-kib
 - memory-allocated-gib

We use mem throughout the system to mean MB.  There are no exceptions except `pid->memory`.  We could make `pid->memory` return MB but it feels weird to me.  Instead, I added a comment to the `get-pid->memory` function.

waiter/src/waiter/shell_scheduler.clj
shamsimam: `(zipmap [:pid :ppid :rss] (map Integer/parseInt elements))` :)

waiter/integration/waiter/deployment_errors_test.clj
shamsimam: I think you can destructure in a let-block and use the `destructure` function. https://clojure.org/guides/destructuring#_macros

waiter/src/waiter/shell_scheduler.clj
sradack: The original is clearer and doesn't try to take abuse the coincidence that every value is an `int`.  If we ever need a string we have have to revert to something like this.

shamsimam:
This is ready to merge?

sradack:
RTM

waiter/src/waiter/client_tools.clj
```diff
@@ -818,3 +818,16 @@
                                      :query-params query-params)]
     (log/debug "retrieved token" token ":" (:body token-response))
     token-response))
+
+(defmacro with-service-cleanup
+  "Ensures a service is cleaned up."
```
shamsimam: Nitpick: 'Ensures' is not true as `delete-service` may fail.

waiter/src/waiter/client_tools.clj
```diff
@@ -818,3 +818,16 @@
                                      :query-params query-params)]
     (log/debug "retrieved token" token ":" (:body token-response))
     token-response))
+
+(defmacro with-service-cleanup
```
shamsimam: +1

shamsimam: Created: https://github.com/twosigma/waiter/issues/142

PR #139, opened 2017-10-03
Additional logs in autoscaler

shamsimam:


waiter/src/waiter/scaling.clj
```diff
@@ -49,8 +48,8 @@
         (loop [service-id->scaling-executor-chan initial-state]
           (recur
             (let [[data channel] (async/alts! [executor-multiplexer-chan query-chan] :priority true)]
-              (cond
-                (= channel executor-multiplexer-chan)
+              (condp = channel
```
sradack: How about `case`?

shamsimam: `autoscaler-goroutine` uses `condp =`, so I went off of that. Also, I believe `case` needs compile time constants.

PR #140, opened 2017-10-03
Additional logs and minor refactoring in autoscaler

shamsimam:


waiter/src/waiter/scaling.clj
```diff
@@ -453,21 +459,29 @@
...
+                                        (let [router-service-ids (set (keys service-id->router-state))
+                                              scheduler-service-ids (set (keys service-id->scheduler-state'))
+                                              scalable-service-ids (set/intersection router-service-ids scheduler-service-ids)
+                                              excluded-service-ids (-> (set/union router-service-ids scheduler-service-ids)
+                                                                       (set/difference scalable-service-ids))
+                                              scale-ticks (when previous-cycle-start-time
+                                                            (-> (difference-in-millis cycle-start-time previous-cycle-start-time)
+                                                                (/ 1000)
+                                                                int))]
+                                          (when (seq excluded-service-ids)
```
sradack: In what situations do we expect this to be true?

shamsimam: When the services loaded from the scheduler directly and Waiter's internal state of known services do not agree.

PR #141, opened 2017-10-03
Metrics-syncer changes

shamsimam:
The intent of this change is to include additional logging to help understand how/when inter-router requests fail. The major change is to ensure we are listening on the `ctrl` chan even if the `connect!` call fails and the registration callback is never invoked.

### DIFF
--- a/waiter/src/waiter/metrics_sync.clj
+++ b/waiter/src/waiter/metrics_sync.clj
@@ -19,11 +19,13 @@
             [metrics.meters :as meters]
             [metrics.timers :as timers]
             [plumbing.core :as pc]
+            [qbits.jet.servlet :as jet]
             [qbits.jet.client.websocket :as ws]
             [waiter.async-utils :as au]
             [waiter.correlation-id :as cid]
             [waiter.metrics :as metrics]
-            [waiter.utils :as utils]))
+            [waiter.utils :as utils])
+  (:import (qbits.jet.websocket WebSocket)))
 
 (defmacro with-catch
   [default-value & body]
@@ -78,14 +80,14 @@
       (do
         (cid/cinfo request-id "registering request from router" source-router-id)
         (counters/inc! (metrics/waiter-counter "metrics-syncer" (router-ws-key->name router-ws-key) (str source-router-id) "register"))
-        (if ctrl
+        (if (and ctrl (jet/chan? ctrl))
           (async/go
             (when-let [ctrl-data (async/<! ctrl)]
               (cid/cinfo request-id "deregistering request, data received on control channel is" ctrl-data)
               (send router-metrics-agent deregister-router-ws router-ws-key source-router-id request-id encrypt)))
           (cid/cinfo request-id "no ctrl-chan available to monitor request"))
         (assoc-in router-metrics-state [router-ws-key source-router-id]
-                  (select-keys ws-request [:in :ctrl :out :request-id :tag])))
+                  (select-keys ws-request [:in :initiated-promise :ctrl :out :request-id :time])))
       (do
         (log/warn "not registering request as it is missing request id")
         router-metrics-state))))
@@ -172,14 +174,15 @@
     router-metrics-state
     (let [time (utils/date-to-str (t/now))
           metrics-data {:router-metrics router-metrics, :source-router-id router-id, :time time}]
-      (doseq [[target-router-id {:keys [out request-id]}] (seq router-id->outgoing-ws)]
-        (let [encrypted-data (timers/start-stop-time!
-                               (metrics/waiter-timer "metrics-syncer" "encrypt" target-router-id)
-                               (encrypt metrics-data))]
-          (cid/cdebug request-id "sending" tag "metrics to" target-router-id "of size" (count router-metrics))
-          (meters/mark! (metrics/waiter-meter "metrics-syncer" "sent-rate" target-router-id))
-          (meters/mark! (metrics/waiter-meter "metrics-syncer" "sent-bytes" target-router-id) (.capacity encrypted-data))
-          (async/put! out encrypted-data)))
+      (doseq [[target-router-id {:keys [initiated-promise out request-id]}] (seq router-id->outgoing-ws)]
+        (when (and initiated-promise (realized? initiated-promise))
+          (let [encrypted-data (timers/start-stop-time!
+                                 (metrics/waiter-timer "metrics-syncer" "encrypt" target-router-id)
+                                 (encrypt metrics-data))]
+            (cid/cdebug request-id "sending" tag "metrics to" target-router-id "of size" (count router-metrics))
+            (meters/mark! (metrics/waiter-meter "metrics-syncer" "sent-rate" target-router-id))
+            (meters/mark! (metrics/waiter-meter "metrics-syncer" "sent-bytes" target-router-id) (.capacity encrypted-data))
+            (async/put! out encrypted-data))))
       (update-router-metrics router-metrics-state metrics-data))))
 
 (defn- cleanup-router-requests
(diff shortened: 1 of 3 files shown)


shamsimam:
@dposada this PR has had code changes since you last approved.

I added how we monitor failed websocket requests by moving the registration outside the callback in `ws/connect!`

shamsimam:
RTM

waiter/src/waiter/utils.clj
```diff
@@ -472,3 +473,8 @@
   "Parses the request header to determine if debug mode has been enabled."
   [request]
   (boolean (get-in request [:headers "x-waiter-debug"])))
+
+(defn keyset
```
sradack: Seems worse to me than just `(set (key m))` in all cases.  Should we also make `keyvec`, `valvec`, `valset`, `keylist`, `vallist`, etc?

waiter/src/waiter/metrics_sync.clj
```diff
@@ -206,31 +207,46 @@
...
+                                                    (cid/cinfo request-id "successfully connected to" router-id)
+                                                    (async/>!! out (encrypt {:dest-router-id router-id
+                                                                             :request-id request-id
+                                                                             :source-router-id my-router-id
+                                                                             :tag :initiated}))
+                                                    ;; trigger publishing of metrics
+                                                    (deliver initiated-promise :initiated))
+                                                             connect-options)]
+                ;; register outside connect! callback to handle messages on the ctrl channel
+                (let [ws-request {:ctrl (.ctrl socket)
```
sradack: Can we instead avoid adding a bad request to the agent state?

waiter/src/waiter/metrics_sync.clj
```diff
@@ -206,31 +207,46 @@
...
-                                                        :tag :initiate}))
-                               (send router-metrics-agent register-router-ws :router-id->outgoing-ws router-id
-                                     ws-request encrypt router-metrics-agent)))
-                           connect-options)))
-          (-> (preserve-metrics-from-routers router-metrics-state (set/union #{my-router-id} (set (keys router-id->http-endpoint))))
+              new-outgoing-router-ids (->> router-metrics-state
+                                           :router-id->outgoing-ws
+                                           utils/keyset
+                                           (set/difference known-router-ids))]
+          (when (not-empty new-outgoing-router-ids)
```
sradack: Idiomatic is `(when (seq new-outgoing-router-ids) ...)`, and avoids requiring a collection.

waiter/src/waiter/metrics_sync.clj
```diff
@@ -330,23 +346,22 @@
     (let [router-id->service-id->metrics (get-in @router-metrics-agent [:metrics :routers])
           service-ids (->> router-id->service-id->metrics
                            (vals)
```
sradack: Why didn't you make `valskeyset`?

waiter/src/waiter/metrics_sync.clj
```diff
@@ -71,22 +73,21 @@
...
-            (async/go
-              (when-let [ctrl-data (async/<! request-terminate-chan)]
-                (cid/cinfo request-id "received" ctrl-data "in control channel, deregistering request")
-                (send router-metrics-agent deregister-router-ws router-ws-key source-router-id request-id encrypt)))))
+        (if (and ctrl (jet/chan? ctrl))
+          (async/go
+            (when-let [ctrl-data (async/<! ctrl)]
+              (cid/cinfo request-id "deregistering request, data received on control channel is" ctrl-data)
+              (send router-metrics-agent deregister-router-ws router-ws-key source-router-id request-id encrypt)))
+          (cid/cinfo request-id "no ctrl-chan available to monitor request"))
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/metrics_sync.clj
shamsimam: I've changed the code to add a separate listener on the `ctrl` chan. The way the code is we can end up creating two listeners on the `ctrl` chan, but they won't affect correctness as `deregister` is called only when receiving non-nil data on the ctrl chan.

shamsimam:
RTM

waiter/src/waiter/utils.clj
shamsimam: Removed `utils/keyset`

waiter/src/waiter/metrics_sync.clj
```diff
@@ -68,27 +69,34 @@
...
-              (when-let [ctrl-data (async/<! request-terminate-chan)]
-                (cid/cinfo request-id "received" ctrl-data "in control channel, deregistering request")
-                (send router-metrics-agent deregister-router-ws router-ws-key source-router-id request-id encrypt)))))
-        (assoc-in router-metrics-state [router-ws-key source-router-id]
-                  (select-keys ws-request [:in :ctrl :out :request-id :tag])))
+        (cid/cinfo request-id "registering" (name router-ws-key) "request" router-id)
+        (counters/inc! (metrics/waiter-counter "metrics-syncer" (router-ws-key->name router-ws-key) router-id "register"))
+        (if ctrl
+          (do
+            (listen-on-ctrl-chan ctrl router-ws-key router-id request-id encrypt router-metrics-agent)
```
sradack: Why not limit the call to `listen-on-ctrl-chan` to `update-router-metrics-state`, since that will run in all cases?

waiter/src/waiter/metrics_sync.clj
```diff
@@ -206,31 +214,43 @@
...
+                                   (async/>!! out (encrypt {:dest-router-id router-id
+                                                            :request-id request-id
+                                                            :source-router-id my-router-id
+                                                            :tag :initiated}))
+                                   (let [ws-request (assoc ws-request :request-id request-id :time (t/now))]
+                                     (send router-metrics-agent register-router-ws :router-id->outgoing-ws router-id
+                                           ws-request encrypt router-metrics-agent)))
+                                 connect-options)]
+                ;; register outside connect! callback to handle messages on the ctrl channel
+                (let [ctrl (.ctrl socket)]
```
sradack: Double `let`.  Could be combined.

waiter/src/waiter/utils.clj
```diff
@@ -472,3 +472,4 @@
   "Parses the request header to determine if debug mode has been enabled."
   [request]
   (boolean (get-in request [:headers "x-waiter-debug"])))
+
```
sradack: Please revert whitespace change.

waiter/test/waiter/metrics_sync_test.clj
```diff
@@ -16,12 +16,30 @@
...
+  [m]
+  (-> m keys set))
+
+(defn- query-agent-state
+  "Queries the agent state and returns the result in the response-chan."
+  [agent-state response-chan]
+  (async/>!! response-chan agent-state)
+  agent-state)
+
+(defn- retrieve-agent-state
```
sradack: Check out `await`.
https://clojure.github.io/clojure/clojure.core-api.html#clojure.core/await

waiter/test/waiter/metrics_sync_test.clj
```diff
@@ -16,12 +16,30 @@
...
-  (:import (org.eclipse.jetty.websocket.client WebSocketClient)))
+  (:import (qbits.jet.websocket WebSocket)
+           (org.eclipse.jetty.websocket.client WebSocketClient)))
+
+(defn- keyset
+  "Returns the keys of the map as a set."
+  [m]
+  (-> m keys set))
+
+(defn- query-agent-state
```
sradack: Why not just use `@agent-state`?

Issue #143, opened 2017-10-03
Inter-router requests should use accept=application/json where appropriate


shamsimam:

dposada:
Is not doing this causing problems?

PR #145, opened 2017-10-03
Explicitly request json response in inter-router requests

shamsimam:
Fixes https://github.com/twosigma/waiter/issues/143

### DIFF
--- a/waiter/src/waiter/work_stealing.clj
+++ b/waiter/src/waiter/work_stealing.clj
@@ -205,21 +205,21 @@
             (async/go
               (let [response-result-promise (promise)
                     response->body (fn [{:keys [body error]}]
-                                       (if error
-                                         (throw error)
-                                         body))]
+                                     (if error
+                                       (throw error)
+                                       body))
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/work_stealing.clj
```diff
@@ -205,36 +205,37 @@
...
-                                                              (utils/map->json-response)
-                                                              :body)
-                                                    :method :post)
-                                                  (get target-router-id))
-                        response-result (if inter-router-response
+                  (let [{:keys [headers status] :as inter-router-response}
+                        (-> (make-inter-router-requests-fn "work-stealing"
+                                                           :acceptable-router? #(= target-router-id %)
+                                                           :body (-> reservation-parameters
+                                                                     (assoc :router-id router-id :service-id service-id)
```
dposada: Nitpick: I would favor each entry on its own line

shamsimam: Done.

shamsimam:
RTM

PR #149, opened 2017-10-05
Include marathon deployments info for conflict responses

shamsimam:
Includes marathon deployment messages when there are 409 conflicts.

### DIFF
--- a/waiter/src/waiter/marathon.clj
+++ b/waiter/src/waiter/marathon.clj
@@ -18,6 +18,7 @@
             [clojure.walk :as walk]
             [marathonclj.common :as mc]
             [marathonclj.rest.apps :as apps]
+            [marathonclj.rest.deployments :as md]
             [metrics.timers :as timers]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/marathon.clj
```diff
@@ -75,6 +76,29 @@
...
+  (try
+    (when-let [deployments (get-deployment-info response)]
+      (let [deployment-ids (->> deployments
+                                (map #(get % "id"))
+                                set)
+            all-marathon-deployments (md/get-deployments)]
+        (->> all-marathon-deployments
+             (filter (fn [{:strs [id]}] (contains? deployment-ids id)))
+             vec)))
+    (catch Exception _ (comment "gobble the exception"))))
```
sradack: Why not log it so we can know what the data was if parsing fails or something else happens?

shamsimam: Done.

shamsimam:
RTM

PR #155, opened 2017-10-10
Fix NPE in work-stealing

shamsimam:


waiter/src/waiter/work_stealing.clj
```diff
@@ -219,10 +219,10 @@
                                                                      (utils/map->json-response)
                                                                      :body)
                                                            :method :post)
-                            (get target-router-id))
+                            (get target-router-id)
+                            async/<!)
```
sradack: What are we doing if the request times out or connection fails?

sradack: Please make sure we are consistently checking for errors on all inter-router requests.

shamsimam: > What are we doing if the request times out or connection fails?

We throw an error which ends up reporting the work-stealing attempt as an error (see `response->body`).

> Please make sure we are consistently checking for errors on all inter-router requests.

That requires refactoring in another PR.

PR #156, opened 2017-10-10
Update config schema to support multiple hostnames

pschorf:


sradack:
We should support multiple hostnames before merging this.

sradack:
Let's not merge this until we support multiple hostnames in all cases where `hostname` is used.

pschorf:
Looking at the other ones, I think it's better to introduce a new config key for alternate hostnames and keep hostnames a single value

PR #157, opened 2017-10-10
Make /state endpoint asynchronous

shamsimam:


waiter/src/waiter/handler.clj
```diff
@@ -493,25 +492,34 @@
...
   (try
-    (let [timeout-ms 30000
+    (let [embed-param (-> request
+                          ring-params/params-request
+                          :params
+                          (get "embed"))
+          embed-pred (cond
+                       (string? embed-param) #{embed-param}
+                       (sequential? embed-param) (set embed-param)
+                       :else (constantly true))
```
sradack: How do I just get state?

waiter/src/waiter/handler.clj
```diff
@@ -493,25 +492,34 @@
...
+                                       (async/>!! scheduler-chan {:response-chan response-chan})
+                                       (async/alt!!
+                                         response-chan ([state] state)
+                                         (async/timeout timeout-ms) ([_] {:message "Query for scheduler state timed out"})
+                                         :priority true)))]
+      (-> (cond-> current-state
+                  (embed-pred "kv-store") (assoc :kv-store(kv/state kv-store))
+                  (embed-pred "leader") (assoc :leader (leader?-fn))
+                  (embed-pred "router-metrics-state") (assoc :router-metrics-state (router-metrics-state-fn))
+                  (embed-pred "scheduler") (assoc :scheduler (retrieve-scheduler-state))
```
sradack: Please update the tests to specify the embeds they need.

waiter/src/waiter/handler.clj
```diff
@@ -493,25 +492,34 @@
 
 (defn get-router-state
   "Outputs the state of the router as json."
-  [state-chan scheduler-chan router-metrics-state-fn kv-store leader?-fn scheduler request]
+  [state-chan scheduler-chan router-metrics-state-fn kv-store leader?-fn request]
   (try
-    (let [timeout-ms 30000
+    (let [embed-param (-> request
```
sradack: Let's make this method async so we don't block a thread.

waiter/src/waiter/handler.clj
```diff
@@ -493,25 +492,34 @@
...
-                              response-chan ([state] state)
-                              (async/timeout timeout-ms) ([_] {:message "Query for scheduler state timed out"})
-                              :priority true))]
-      (-> current-state
-          (assoc :leader (leader?-fn)
-                 :kv-store (kv/state kv-store)
-                 :router-metrics-state (router-metrics-state-fn)
-                 :scheduler scheduler-state
-                 :statsd (statsd/state))
+          retrieve-scheduler-state (fn retrieve-scheduler-state-fn []
```
sradack: If async, this can't be in a function.

waiter/src/waiter/handler.clj
shamsimam: state is always included by default. So passing `embed=state` will achieve that.

sradack: While `embed=state` works given the implementation, it's an awkward API. We don't "embed" state within state.  

We have an opportunity to clean things up, making `/state`, by default, return just state, and embed options add to the output.

Another opportunity is to create dedicated endpoints for the specific states, e.g.
`/state/scheduler`
`/state/statsd`
etc.

waiter/src/waiter/handler.clj

waiter/src/waiter/handler.clj
shamsimam: Converted to async.

waiter/src/waiter/handler.clj
shamsimam: Converted to async.

shamsimam:
RTM

shamsimam:
RTM, the scope of the PR has changed to make `/state` handler become asynchronous.

PR #158, opened 2017-10-10
Handle duplicates by avoiding #{} syntax

shamsimam:


DaoWen:
Agh! I typed up a review for this on the train, but my phone connection was spotty, so I was waiting until I got home to post it... I think we should use `(hash-set ...)` rather than `(set [...])`. I'll open a separate PR for that since this is already merged.

PR #159, opened 2017-10-11
Refactor for set idioms

DaoWen:
Refactoring several set expressions to:

1. Avoid building extra intermediate sets (or other data structures)
2. Use more efficient functions to perform the same operations

In most cases, I think the updated versions are faster, and at least as readable as the original versions. This also addresses my comment in #158.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -612,7 +612,8 @@
                                   (dissoc :run-as-user))
           waiter-port (.getPort (URL. (str "http://" waiter-url)))
           waiter-port (if (neg? waiter-port) 80 waiter-port)
-          host-header (str token ":" waiter-port)]
+          host-header (str token ":" waiter-port)
+          has-x-waiter-consent? (partial some #(= (:name %) "x-waiter-consent"))]
       (try
         (testing "token creation"
           (let [token-description (assoc service-description :token token)
@@ -657,7 +658,7 @@
                   (reset! cookies-atom cookies)
                   (is (= "Added cookie x-waiter-consent" body))
                   ; x-waiter-consent should be emitted Waiter
-                  (is (contains? (set (map :name cookies)) "x-waiter-consent"))
+                  (is (has-x-waiter-consent? cookies))
                   (assert-response-status response 200)))
 
               (testing "auto run-as-user population on expected service-id"
@@ -672,7 +673,7 @@
                       (reset! service-id-atom service-id)
                       (is (= "Hello World" body))
                       ; x-waiter-consent should not be re-emitted Waiter
-                      (is (not (contains? (set (map :name cookies)) "x-waiter-consent")))
+                      (is (not (has-x-waiter-consent? cookies)))
                       (is (= expected-service-id service-id))
                       (is (not (str/blank? permitted-user)))
                       (is (= run-as-user permitted-user))
@@ -707,7 +708,7 @@
                   (reset! cookies-atom cookies)
                   (is (= "Added cookie x-waiter-consent" body))
                   ; x-waiter-consent should be emitted Waiter
-                  (is (contains? (set (map :name cookies)) "x-waiter-consent"))
+                  (is (has-x-waiter-consent? cookies))
                   (assert-response-status response 200)))
 
               (testing "auto run-as-user population on approved token"
@@ -720,7 +721,7 @@
                       (reset! service-id-atom service-id)
                       (is (= "Hello World" body))
                       ; x-waiter-consent should not be re-emitted Waiter
-                      (is (not (contains? (set (map :name cookies)) "x-waiter-consent")))
+                      (is (not (has-x-waiter-consent? cookies)))
                       (is (not= previous-service-id service-id))
                       (assert-response-status response 200))
                     (finally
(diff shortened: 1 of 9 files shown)


shamsimam:
Can you please rebase?

shamsimam:
@DaoWen can you please rebase this PR?

DaoWen:
@shamsimam - I can rebase this, but I've been having issue getting the unit tests to run locally. Maybe you can help me with that tomorrow.

DaoWen:
@shamsimam - Rebased, and Travis is green.

waiter/src/waiter/scheduler.clj
```diff
@@ -362,12 +362,12 @@
   "Queries the scheduler and builds a list of healthy and unhealthy instances for the specified service-id."
   [service-id service-instances]
   (when service-instances
-    (let [healthy-instances (vec (filter :healthy? service-instances))
-          unhealthy-instances (vec (clojure.set/difference (set service-instances) (set healthy-instances)))]
+    (let [{healthy-instances true unhealthy-instances false}
+          (group-by (comp boolean :healthy?) service-instances)]
```
shamsimam: Nitpick: Can we move this to the previous line and have a comma as follows: `{healthy-instances true, unhealthy-instances false}`?

waiter/src/waiter/scaling.clj
```diff
@@ -429,8 +429,8 @@
                         state-chan
                         (let [{:keys [service-id->healthy-instances service-id->unhealthy-instances service-id->expired-instances]} args
                               existing-service-ids-set (set (keys service-id->router-state))
-                              service-ids-set (set/union (-> service-id->healthy-instances (keys) (set))
-                                                         (-> service-id->unhealthy-instances (keys) (set)))
+                              service-ids-set (set (concat (keys service-id->healthy-instances)
+                                                           (keys service-id->unhealthy-instances)))
```
shamsimam: How about `(into (-> service-id->healthy-instances keys set) (-> service-id->unhealthy-instances keys)` to help us avoid the intermediate lazy sequence.

waiter/src/waiter/core.clj
```diff
@@ -737,8 +737,8 @@
                          (let [local-router (InetAddress/getLocalHost)
                                waiter-router-hostname (.getCanonicalHostName local-router)
                                waiter-router-ip (.getHostAddress local-router)
-                               ;; use (set [...]) instead of #{...} below as there may be duplicate values
-                               hostnames (set/union waiter-hostnames (set [waiter-router-hostname waiter-router-ip]))]
+                               ;; use (hash-set ...) instead of #{...} below, as there may be duplicate values
+                               hostnames (apply hash-set waiter-router-hostname waiter-router-ip waiter-hostnames)]
```
shamsimam: How about `(into waiter-hostnames  [waiter-router-hostname waiter-router-ip])`? Then we no longer need the comment.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -657,7 +657,7 @@
                   (reset! cookies-atom cookies)
                   (is (= "Added cookie x-waiter-consent" body))
                   ; x-waiter-consent should be emitted Waiter
-                  (is (contains? (set (map :name cookies)) "x-waiter-consent"))
+                  (is (some #(= (:name %) "x-waiter-consent") cookies))
```
shamsimam: Nitpick, can we extract `#(= (:name %) "x-waiter-consent")` into a variable or helper function?

waiter/src/waiter/scheduler.clj
DaoWen: I'm OK with that. Do we have any guidelines on what to do about really long lines? I felt like this line was a bit too long, which is why I broke it onto two lines, but I don't really like doing that between let value–binding pairs.

waiter/src/waiter/core.clj
DaoWen: I like that better. This line was affected by the rebase. Previously, `waiter-hostnames` wasn't a set, so this made more sense.

waiter/src/waiter/scheduler.clj
shamsimam: We don't have a strict rule with long lines, but we just use our judgment to split lines. It didn't feel to me that this line would be too long.

waiter/src/waiter/core.clj
shamsimam: Okay.

waiter/src/waiter/scaling.clj
DaoWen: Using `concat` was faster than creating the intermediate sets and then doing a union, but I ran another microbenchmark and found that using `into` as you suggested results in performance as good as if not better than the `concat` version. That surprised me, since I assumed the `concat` version would perform better since `concat` itself is very low overhead. Apparently accessing the elements through the lazy sequence layer adds a non-trivial amount of overhead.

waiter/integration/waiter/token_request_test.clj
DaoWen: Done.

waiter/src/waiter/core.clj
DaoWen: I ended up changing it to `conj` instead.

shamsimam:
Waiting on green build

PR #161, opened 2017-10-11
Fix support for multiple hostnames

pschorf:


waiter/src/waiter/settings.clj
```diff
@@ -45,6 +45,7 @@
                                           (s/required-key :failed-check-threshold) schema/positive-int}
    (s/required-key :host) schema/non-empty-string
    (s/required-key :hostname) schema/non-empty-string
+   (s/optional-key :alternate-hostnames) [schema/non-empty-string]
```
shamsimam: Lexicographic order :(

waiter/test/waiter/core_test.clj
```diff
@@ -1108,7 +1108,8 @@
       (is (not (waiter-request?-fn {:headers {"host" "waiter-host-1"}})))))
 
   (testing "list hostname config"
-    (let [config {:settings {:hostname ["waiter-host-1" "waiter-host-2"]}}
+    (let [config {:settings {:hostname "waiter-host-1"
+                             :alternate-hostnames ["waiter-host-2"]}}
```
shamsimam: Lexicographic order :(

sradack: This is a case where I believe it's reasonable to have `:hostname` first.

waiter/src/waiter/settings.clj
pschorf: I think this should be next to hostname, and after it.

waiter/test/waiter/core_test.clj
pschorf: Same here

waiter/src/waiter/settings.clj
pschorf: But if it is going to hang up the review, I will just sort it.

waiter/test/waiter/token_test.clj
```diff
@@ -58,29 +58,29 @@
           service-description2 (clojure.walk/stringify-keys
                                  {:cmd "tc2", :cpus 2, :mem 400, :version "d1e2f3", :run-as-user "tu1", :permitted-user "tu3", :token token})
           service-id2 (sd/service-description->service-id service-id-prefix service-description2)
-          waiter-hostname "waiter-hostname.app.example.com"
+          waiter-hostnames #{"waiter-hostname.app.example.com"}
           handle-list-tokens-request (wrap-handler-json-response handle-list-tokens-request)]
 
       (testing "put:unsupported-request-method"
         (let [{:keys [status body]}
               (run-handle-token-request
```
shamsimam: `run-handle-token-request` parameter should be renamed to `waiter-hostnames` from `waiter-hostname` in the definition above.

shamsimam:
`prepend-waiter-url` is the only function that uses hostname now. Maybe we can do with just a single config entry?

waiter/test/waiter/token_test.clj
pschorf: Done

pschorf:
I agree, reverted back to the old config

waiter/src/waiter/core.clj
```diff
@@ -623,17 +627,18 @@
                                                  router-id async-request-store-atom make-http-request-fn instance-rpc-chan service-id metric-group
                                                  instance reason-map request-properties location response-headers)))
    :prepend-waiter-url (pc/fnk [[:settings port hostname]]
-                         (fn [endpoint-url]
-                           (if (str/blank? endpoint-url)
-                             endpoint-url
-                             (str "http://" hostname ":" port endpoint-url))))
+                         (let [hostname (if (sequential? hostname) (first hostname) hostname)]
```
shamsimam: The sequence can be empty, I think we need to enforce in the `settings` schema that the list may not be empty.

pschorf: Done

waiter/config-full.edn
```diff
@@ -109,7 +109,8 @@
  :port 9091
 
  ;; Set the DNS name for your Waiter cluster:
- :hostname "waiter.example.com"
```
sradack: The comment should indicate that a string or non-empty vector of strings are allowed, and that in the event of multiple hostnames, the first entry is the primary hostname.

pschorf: Done

PR #163, opened 2017-10-12
Introduce endpoints for /state/*

shamsimam:
Add `/state/***` endpoints for the different components. Modifies the integration tests to use these new endpoints.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -89,6 +89,12 @@
                      "settings" :display-settings-handler-fn
                      "sim" :sim-request-handler
                      "state" {"" :display-state-handler-fn
+                              "/kv-store" :display-kv-store-state-handler-fn
+                              "/leader" :display-leader-state-handler-fn
+                              "/maintainer" :display-maintainer-state-handler-fn
+                              "/router-metrics" :display-router-metrics-state-handler-fn
+                              "/scheduler" :display-scheduler-state-handler-fn
+                              "/statsd" :display-statsd-state-handler-fn
                               ["/" :service-id] :service-state-handler-fn}
                      "status" :status-handler-fn
                      "token" :token-handler-fn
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -491,21 +491,33 @@
...
-                          state-chan ([state-data] state-data)
-                          (async/timeout timeout-ms) ([_] {:message "Query for router state timed out"})
-                          :priority true)
-          scheduler-state (let [response-chan (async/promise-chan)]
-                            (async/>!! scheduler-chan {:response-chan response-chan})
-                            (async/alt!!
-                              response-chan ([state] state)
-                              (async/timeout timeout-ms) ([_] {:message "Query for scheduler state timed out"})
-                              :priority true))]
+          current-state (async/<!! (retrieve-maintainer-state state-chan timeout-ms))
```
sradack: Fix conflicts

waiter/src/waiter/core.clj
```diff
@@ -915,6 +921,49 @@
                                                              determine-priority-fn ws/process-response! ws/process-exception-in-request
                                                              ws/abort-request-callback-factory request))]
                                          (ws/request-handler password process-request-fn request)))))
+   :display-kv-store-state-handler-fn (pc/fnk [[:curator kv-store]
```
sradack: Can we remove `display-` from all these handler names?

shamsimam:
> We can handle fixing integration tests that use /state directly when we end up removing that functionality. 

This PR already handles some (if not all) of the calls to `/state` in the integration tests.

waiter/src/waiter/handler.clj

waiter/src/waiter/core.clj

shamsimam:
RTM

shamsimam:
Will be updating the `/state/leader` endpoint with the leader id.

shamsimam:
RTM

PR #164, opened 2017-10-12
OAuth support

sradack:
# Overview 

Add OAuth support for public-facing use cases.

The OAuth flow:
1. Unauthenticated requests are redirected to a page with a list of providers.
1. Client selects one provider from the list of configured providers (Google and GitHub are implemented.)
1. Client is redirected to the provider to approve Waiter's access to the client's credentials.
1. Provider redirects user back to Waiter with a token that can be used by Waiter to verify the user.
1. After authentication, client is redirected to original location, with path, query-string, and hash preserved.

OAuth requires an application to be registered with the provider.  Upon registering an application, the provider provides a client-id and a client-secret.  These values are configured via settings, where loading these values via environment variables is preferred for security.

Providers also require the URI redirecting back to the application to be specified when registering.  Some providers allow multiple callback URIs (e.g. Google), while others only one (e.g. GitHub).  Providers that allow multiple redirect URIs to be specified are preferred, since Waiter will likely be hosting multiple applications with different hostnames.  Each hostname hosted by Waiter will need to be registered.  Using a provider that supports only one redirect URI limits usage of the OAuth support to one hostname.

OAuth is, at its core, not a method of authenticating users, but it can be used for authentication if the provider has a way to look up a user's verified email address using the access_token provided.  Open ID Connect is a standardized flow for this, but while Google supports Open ID Connect, GitHub does not.

The user's email address is used as the principal for Waiter.  This implementation still expects a single `run-as-user` to be configured since we can't reliably map a principal to a username that makes sense to the scheduler.

If per-user instances are desired, a `ServiceDescriptionBuilder` can be used to add principal metadata to the service description, ensuring unique services per user.

Integration tests for OAuth would be difficult, given that interactive authorization by a user is required.

# Demo

After trying to visit `localhost:9091/settings`, the browser is redirected:

choose-provider

After choosing Google as the provider:

google-login

GitHub has its own login page.

After successfully logging in, browser is redirected back to `localhost:9091/settings`.

### DIFF
--- a/waiter/config-shell-oauth.edn
+++ b/waiter/config-shell-oauth.edn
@@ -1,0 +1,36 @@
+{
+ ; ---------- Cluster ----------
+
+ :zookeeper {
+             ;; Use an in-process ZK (not for production use):
+             :connect-string :in-process}
+
+ ; ---------- Network ----------
+
+ ;; Set the bind address to a specific IP:
+ :host "localhost"
+
+ ;; Set the port:
+ :port #config/env-int "WAITER_PORT"
+
+ ; ---------- Security ----------
+
+ :authenticator-config {:kind :oauth
+                        :oauth {:factory-fn waiter.auth.oauth/oauth-authenticator
+                                :providers {:kinds [:github :google]
+                                            :github {:factory-fn waiter.auth.oauth/map->GitHubOAuthProvider
+                                                     :client-id #config/env "GITHUB_OAUTH_CLIENT_ID"
+                                                     :client-secret #config/env "GITHUB_OAUTH_CLIENT_SECRET"}
+                                            :google {:factory-fn waiter.auth.oauth/map->GoogleOAuthProvider
+                                                     :client-id #config/env "GOOGLE_OAUTH_CLIENT_ID"
+                                                     :client-secret #config/env "GOOGLE_OAUTH_CLIENT_SECRET"}}
+                                :run-as-user #config/env "WAITER_AUTH_RUN_AS_USER"}}
+
+ ; ---------- Scheduling ----------
+
+ :scheduler-config {
+                    ;; :kind :shell simply schedules instances on your local machine (for testing purposes only):
+                    :kind :shell}
+
+ ; ---------- CORS ----------
+ :cors-config {:kind :allow-all}}
--- a/waiter/config-shell-oauth.edn
+++ b/waiter/config-shell-oauth.edn
@@ -18,10 +18,10 @@
  :authenticator-config {:kind :oauth
                         :oauth {:factory-fn waiter.auth.oauth/oauth-authenticator
                                 :providers {:kinds [:github :google]
-                                            :github {:factory-fn waiter.auth.oauth/map->GitHubOAuthProvider
+                                            :github {:factory-fn waiter.auth.oauth.github/map->GitHubOAuthProvider
                                                      :client-id #config/env "GITHUB_OAUTH_CLIENT_ID"
                                                      :client-secret #config/env "GITHUB_OAUTH_CLIENT_SECRET"}
-                                            :google {:factory-fn waiter.auth.oauth/map->GoogleOAuthProvider
+                                            :google {:factory-fn waiter.auth.oauth.google/map->GoogleOAuthProvider
                                                      :client-id #config/env "GOOGLE_OAUTH_CLIENT_ID"
                                                      :client-secret #config/env "GOOGLE_OAUTH_CLIENT_SECRET"}}
                                 :run-as-user #config/env "WAITER_AUTH_RUN_AS_USER"}}
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -349,13 +349,18 @@
   (testing-using-waiter-url
     (log/info "Basic waiter-auth test")
     (let [{:keys [status body headers]} (make-request waiter-url "/waiter-auth")
-          set-cookie (get headers "set-cookie")]
+          set-cookie (get headers "set-cookie")
+          parsed-json (-> (try (json/read-str body)
+                               (catch Exception _
+                                 (is false (str "Unable to parse JSON\n:" body))))
+                          (walk/keywordize-keys))]
       (is (= 200 status))
       (is (str/includes? set-cookie "x-waiter-auth="))
       (is (str/includes? set-cookie "Max-Age="))
       (is (str/includes? set-cookie "Path=/"))
       (is (str/includes? set-cookie "HttpOnly=true"))
-      (is (= (System/getProperty "user.name") (str body))))))
+      (is (= {:user (System/getProperty "user.name")
+              :principal (System/getProperty "user.name")} parsed-json)))))
 
 ; Marked explicit due to:
 ;   FAIL in (test-killed-instances)
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -350,9 +350,7 @@
     (log/info "Basic waiter-auth test")
     (let [{:keys [status body headers]} (make-request waiter-url "/waiter-auth")
           set-cookie (get headers "set-cookie")
-          parsed-json (-> (try (json/read-str body)
-                               (catch Exception _
-                                 (is false (str "Unable to parse JSON\n:" body))))
+          parsed-json (-> (try-parse-json body)
                           (walk/keywordize-keys))]
       (is (= 200 status))
       (is (str/includes? set-cookie "x-waiter-auth="))
@@ -513,9 +511,7 @@
         (is (str/includes? body "<html>"))))
     (testing "application/json explicit"
       (let [{:keys [body headers status]} (make-request waiter-url "/404" :headers {"accept" "application/json"})
-            {:strs [waiter-error]} (try (json/read-str body)
-                                        (catch Throwable _
-                                          (is false (str "Could not parse body that is supposed to be JSON:\n" body))))]
+            {:strs [waiter-error]} (try-parse-json body)]
         (is (= 404 status))
         (is (= "application/json" (get headers "content-type")))
         (is waiter-error (str "Could not find waiter-error element in body " body))
@@ -523,9 +519,7 @@
           (is (= 404 status)))))
     (testing "application/json implied by content-type"
       (let [{:keys [body headers status]} (make-request waiter-url "/404" :headers {"content-type" "application/json"})
-            {:strs [waiter-error]} (try (json/read-str body)
-                                        (catch Throwable _
-                                          (is false (str "Could not parse body that is supposed to be JSON:\n" body))))]
+            {:strs [waiter-error]} (try-parse-json body)]
         (is (= 404 status))
         (is (= "application/json" (get headers "content-type")))
         (is waiter-error (str "Could not find waiter-error element in body " body))
@@ -534,9 +528,7 @@
     (testing "support information included"
       (let [{:keys [body headers status]} (make-request waiter-url "/404" :headers {"accept" "application/json"})
             {:keys [messages support-info]} (waiter-settings waiter-url)
-            {:strs [waiter-error]} (try (json/read-str body)
-                                        (catch Throwable _
-                                          (is false (str "Could not parse body that is supposed to be JSON:\n" body))))]
+            {:strs [waiter-error]} (try-parse-json body)]
 
         (is (= 404 status))
         (is (= "application/json" (get headers "content-type")))
@@ -566,9 +558,7 @@
         (is (str/includes? body "Welcome to Waiter"))))
     (testing "accept application/json"
       (let [{:keys [body headers status]} (make-request waiter-url "/" :headers {"accept" "application/json"})
-            json-data (try (json/read-str body)
-                           (catch Exception _
-                             (is false ("Not json:\n" body))))]
+            json-data (try-parse-json body)]
         (is (= 200 status))
         (is (= "application/json" (get headers "content-type")))
         (is (= "Welcome to Waiter" (get json-data "message")))))
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -351,7 +351,7 @@
     (let [{:keys [status body headers]} (make-request waiter-url "/waiter-auth")
           set-cookie (get headers "set-cookie")
           parsed-json (-> (try-parse-json body)
-                          (walk/keywordize-keys))]
+                          walk/keywordize-keys)]
       (is (= 200 status))
       (is (str/includes? set-cookie "x-waiter-auth="))
       (is (str/includes? set-cookie "Max-Age="))
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -357,8 +357,8 @@
       (is (str/includes? set-cookie "Max-Age="))
       (is (str/includes? set-cookie "Path=/"))
       (is (str/includes? set-cookie "HttpOnly=true"))
-      (is (= {:user (System/getProperty "user.name")
-              :principal (System/getProperty "user.name")} parsed-json)))))
+      (is (= (System/getProperty "user.name") (:user parsed-json)))
+      (is (:principal parsed-json)))))
 
 ; Marked explicit due to:
 ;   FAIL in (test-killed-instances)
--- a/waiter/resources/web/oauth.html
+++ b/waiter/resources/web/oauth.html
@@ -31,10 +31,25 @@
     <div>
         <p>Choose a provider:</p>
         <% (doseq [[provider-key provider] providers] %>
-            <li><a href="/waiter-auth/oauth/<%= (name provider-key) %>/redirect/<%= path %><% (when query-string %><%= (str \? query-string) %><% ) %>"><%= (display-name provider) %></a></li>
+            <li><a class="redirect" href="/waiter-auth/oauth/<%= (name provider-key) %>/redirect/<%= path %><% (when query-string %><%= (str \? query-string) %><% ) %>"><%= (display-name provider) %></a></li>
         <% ) %>
     </div>
 </div>
+
+<script>
+if (location.hash) {
+    var redirectLinks = document.getElementsByClassName('redirect');
+    for (var i = 0; i < redirectLinks.length; i++) {
+        <% (when query-string %>
+            redirectLinks[i].href += "&_waiter_hash=" + encodeURIComponent(location.hash);
+        <% ) %>
+        <% (when-not query-string %>
+            redirectLinks[i].href += "?_waiter_hash=" + encodeURIComponent(location.hash);
+        <% ) %>
+
+    }
+}
+</script>
 
 </body>
 </html>
--- a/waiter/resources/web/oauth.html
+++ b/waiter/resources/web/oauth.html
@@ -41,10 +41,10 @@ if (location.hash) {
     var redirectLinks = document.getElementsByClassName('redirect');
     for (var i = 0; i < redirectLinks.length; i++) {
         <% (when query-string %>
-            redirectLinks[i].href += "&_waiter_hash=" + encodeURIComponent(location.hash);
+            redirectLinks[i].href += "&_<%= hash-query-param-name %>=" + encodeURIComponent(location.hash);
         <% ) %>
         <% (when-not query-string %>
-            redirectLinks[i].href += "?_waiter_hash=" + encodeURIComponent(location.hash);
+            redirectLinks[i].href += "?_<%= hash-query-param-name %>=" + encodeURIComponent(location.hash);
         <% ) %>
 
     }
--- a/waiter/resources/web/oauth.html
+++ b/waiter/resources/web/oauth.html
@@ -1,0 +1,40 @@
+<!doctype html>
+<html>
+<head>
+    <title>Login - Waiter</title>
+    <meta http-equiv="x-ua-compatible" content="IE=edge" />
+    <style type="text/css">
+   html {
+     font-family: 'Lato', 'Helvetica Neue', Arial, Helvetica, sans-serif;
+     color: rgba(0,0,0,.87);
+   }
+
+   .main-content {
+     margin: auto;
+     margin-top: 4em;
+     width: 800px;
+   }
+
+   .main-note {
+     color: #0e566c;
+     text-align: center;
+   }
+
+    </style>
+</head>
+<body>
+
+<div class="main-content">
+    <h2 class="main-note">
+        Login to Waiter
+    </h2>
+    <div>
+        <p>Choose a provider:</p>
+        <% (doseq [[provider-key provider] providers] %>
+            <li><a href="/waiter-auth/oauth/<%= (name provider-key) %>/redirect/<%= path %><% (when query-string %><%= (str \? query-string) %><% ) %>"><%= (display-name provider) %></a></li>
+        <% ) %>
+    </div>
+</div>
+
+</body>
+</html>
--- a/waiter/resources/web/oauth.html
+++ b/waiter/resources/web/oauth.html
@@ -40,13 +40,7 @@
 if (location.hash) {
     var redirectLinks = document.getElementsByClassName('redirect');
     for (var i = 0; i < redirectLinks.length; i++) {
-        <% (when query-string %>
-            redirectLinks[i].href += "&_<%= hash-query-param-name %>=" + encodeURIComponent(location.hash);
-        <% ) %>
-        <% (when-not query-string %>
-            redirectLinks[i].href += "?_<%= hash-query-param-name %>=" + encodeURIComponent(location.hash);
-        <% ) %>
-
+        redirectLinks[i].href += "<%= (if query-string \& \?) %>_<%= hash-query-param-name %>=" + encodeURIComponent(location.hash);
     }
 }
 </script>
--- a/waiter/src/waiter/auth/oauth.clj
+++ b/waiter/src/waiter/auth/oauth.clj
@@ -135,9 +135,9 @@
               (-> request
                   (auth/assoc-auth-in-request auth-principal)
                   handler-fn)
-              {:status 307
-               :headers {"location" (cond-> (str "/waiter-auth/oauth/providers" uri)
-                                      (not (str/blank? query-string)) (str \? query-string))}}))))
+              {:headers {"location" (cond-> (str "/waiter-auth/oauth/providers" uri)
+                                      (not (str/blank? query-string)) (str \? query-string))}
+               :status 307}))))
       (catch Exception ex
         (utils/exception->response ex request)))))
 
--- a/waiter/src/waiter/auth/oauth.clj
+++ b/waiter/src/waiter/auth/oauth.clj
@@ -10,21 +10,15 @@
 ;;
 (ns waiter.auth.oauth
   (:require [bidi.bidi :as bidi]
-            [clj-time.core :as t]
-            [clojure.core.async :as async]
-            [clojure.data.codec.base64 :as b64]
             [clojure.data.json :as json]
             [clojure.java.io :as io]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [comb.template :as template]
             [digest]
-            [full.async :as fa]
             [plumbing.core :as pc]
             [plumbing.graph :as graph]
             [qbits.jet.client.http :as http]
-            [ring.middleware.params :as ring-params]
-            [ring.util.codec :as ring-codec]
             [waiter.auth.authentication :as auth]
             [waiter.cookie-support :as cookie-support]
             [waiter.utils :as utils])
@@ -67,17 +61,6 @@
   (authenticate [this request]
     "Handles the response from the OAuth provider.  Returns an async response."))
 
-(defn google-discovery-document
-  "Gets Google's discovery document."
-  [http-client]
-  (fa/go-try
-    (let [discovery-uri "https://accounts.google.com/.well-known/openid-configuration"
-          {:keys [body]} (async/<! (http/get http-client discovery-uri {:headers {"accept" "application/json"}}))]
-      (try
-        (json/read-str (async/<! body))
-        (catch Exception e
-          (throw (ex-info "Couldn't retrieve Google discovery document" {:status 403})))))))
-
 (defn make-uri
   "Takes a URI and adds a formatted query-string from query parameters in a map."
   [uri query-params]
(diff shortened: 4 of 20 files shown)


waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+            [ring.middleware.params :as ring-params]
+            [ring.util.codec :as ring-codec]
+            [waiter.auth.authentication :as auth]
+            [waiter.cookie-support :as cookie-support]
+            [waiter.utils :as utils])
+  (:import java.security.SecureRandom
+           org.eclipse.jetty.util.HttpCookieStore$Empty
+           org.eclipse.jetty.util.UrlEncoded))
+
+(def secure-random (SecureRandom.))
```
shamsimam: Can this not be a field inside `OAuthAuthenticator` and initialized explicitly in `oauth-authenticator`?

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+  (auth-type [_]
+    :oauth)
+  (check-user [_ user _]
+    user)
+  (wrap-auth-handler [_ handler]
+    (wrap-oauth handler run-as-user password providers)))
+
+(defn oauth-authenticator
+  "Factory function for creating OAuthAuthenticator"
+  [{:keys [password providers run-as-user]}]
```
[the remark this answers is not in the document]
shamsimam: We should also perform validation on the input keys here e.g. via pre-conditions.

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,211 @@
...
+  (is (random-str)))
+
+(deftest test-provider-list-handler
+  (testing "get"
+    (let [{:keys [body headers]} (provider-list-handler
+                                   {:github (map->GitHubOAuthProvider {})
+                                    :google (map->GoogleOAuthProvider {})}
+                                   {:request-method :get})]
+      (is body)
+      (is (str/includes? body "GitHub"))
```
shamsimam: Instead of `GitHub` or `Google`, should we invoke `display-name` on the providers?

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,211 @@
...
+                                     :client-secret ""}
+                            :google {:factory-fn 'waiter.auth.oauth/map->GoogleOAuthProvider
+                                     :client-id ""
+                                     :client-secret ""}}
+                :run-as-user "user"
+                :host "host"
+                :port 8080}
+        authenticator (oauth-authenticator config)]
+    (is authenticator)
+    (is (= 2 (-> authenticator :providers count)))))
```
shamsimam: We should also validate the types of the provider instances here to ensure the `kinds` helper function worked as expected.

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,211 @@
...
+          (is location)
+          (is (= status 307)))))))
+
+(deftest test-make-uri
+  (is (= "http://test?a=1" (make-uri "http://test" {"a" "1"})))
+  (is (= "http://test?a=1&b=2" (make-uri "http://test" {"a" "1"
+                                                        "b" "2"})))
+  (is (= "http://test?message=hello+world" (make-uri "http://test" {"message" "hello world"}))))
+
+(deftest test-make-location
```
shamsimam: We should include a test where `_waiter_hash` is not the last parameter. We should also include a test where q query parameter is repeated, e.g. a=1&...&a=2.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+    "Handles a redirect to the OAuth provider.  Returns an async response.")
+  (authenticate [this request]
+    "Handles the response from the OAuth provider.  Returns an async response."))
+
+(defn google-discovery-document
+  "Gets Google's discovery document."
+  [http-client]
+  (fa/go-try
+    (let [discovery-uri "https://accounts.google.com/.well-known/openid-configuration"
+          {:keys [body]} (async/<! (http/get http-client discovery-uri {:headers {"accept" "application/json"}}))]
```
shamsimam: We should check for `error` here.

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,211 @@
...
+        (let [{{:strs [location]} :headers :keys [status] :as response}
+              (fa/<?? (authenticate provider {:headers {"cookie" cookie}
+                                              :query-string (str "code=code&state="
+                                                                 (-> state
+                                                                     json/write-str
+                                                                     UrlEncoded/encodeString))}))]
+          (is location)
+          (is (= status 307)))))))
+
+(deftest test-make-uri
```
shamsimam: We should also include a test where q query parameter is repeated, e.g. `{a [1 2]}`.

waiter/integration/waiter/basic_test.clj
```diff
@@ -351,13 +351,18 @@
...
+                               (catch Exception _
+                                 (is false (str "Unable to parse JSON\n:" body))))
+                          (walk/keywordize-keys))]
       (is (= 200 status))
       (is (str/includes? set-cookie "x-waiter-auth="))
       (is (str/includes? set-cookie "Max-Age="))
       (is (str/includes? set-cookie "Path=/"))
       (is (str/includes? set-cookie "HttpOnly=true"))
-      (is (= (System/getProperty "user.name") (str body))))))
+      (is (= {:user (System/getProperty "user.name")
```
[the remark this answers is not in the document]
sradack: They all seem to be internal, either from health checks or from other routers.  Would we rather have a new endpoint?  We should have a way so that the principal can be reported back.

waiter/src/waiter/auth/oauth.clj
sradack: It could.  Do we need a SecureRandom for each authenticator?

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+        (json/read-str (async/<! body))
+        (catch Exception e
+          (throw (ex-info "Couldn't retrieve Google discovery document" {:status 403})))))))
+
+(defn make-uri
+  "Takes a URI and adds a formatted query-string from query parameters in a map."
+  [uri query-params]
+  (str uri "?"
+       (->> query-params
+            (map (fn [[name value]]
```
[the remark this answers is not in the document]
sradack: No

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+                   (str name "=" (UrlEncoded/encodeString value))))
+            (str/join "&"))))
+
+(defn make-location
+  "Given a path and query-string, return the value for the Location header
+  after successful authentication.
+  Assumes the URL fragment (hash) is passed at the end of query-string under a special
+  query parameter HASH-QUERY-PARAM-NAME."
+  [{:keys [path query-string]}]
```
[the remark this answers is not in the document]
sradack: We should be as tolerant of malformed data.  We're modifying the application's query string, so we do well to treat it as opaquely as possible.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+                                "response_type" "code"
+                                "redirect_uri" (authenticate-uri-fn request)
+                                "scope" "openid email"
+                                "state" state})]
+        (-> {:status 307
+             :headers {"location" location}}
+            (cookie-support/add-encoded-cookie password OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
+  (authenticate [_ request]
+    (fa/go-try
+      (let [{:strs [code state]} (:params (ring-params/params-request request))
```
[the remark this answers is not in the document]
sradack: How would we validate more eagerly than we do?

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+            {:strs [token_endpoint]} (fa/<? (google-discovery-document http-client))
+            _ (when-not token_endpoint
+                (throw (ex-info "Couldn't get Google token endpoint" {:status 403})))
+            {:keys [body status]} (async/<! (http/post http-client token_endpoint
+                                                       {:form-params {"client_id" client-id
+                                                                      "client_secret" client-secret
+                                                                      "code" code
+                                                                      "grant_type" "authorization_code"
+                                                                      "redirect_uri" (authenticate-uri-fn request)}}))
+            _  (when-not (= status 200)
```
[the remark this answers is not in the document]
sradack: We expect the API to return 200. Anything else would imply the API has changed.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+                               {"client_id" client-id
+                                "scope" "user:email"
+                                "redirect_uri" (authenticate-uri-fn request)
+                                "state" state})]
+        (-> {:status 307
+             :headers {"location" location}}
+            (cookie-support/add-encoded-cookie password OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
+  (authenticate [_ request]
+    (fa/go-try
+      (let [{:strs [code state]} (:params (ring-params/params-request request))
```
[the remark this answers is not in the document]
sradack: Validate that they aren't nil?

waiter/src/waiter/auth/oauth.clj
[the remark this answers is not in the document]
sradack: The password is passed in `core.clj`.  It's the same password as the Waiter password.

waiter/src/waiter/core.clj
```diff
@@ -107,7 +107,7 @@
                                      :async-result-handler-fn
                                      ["/status/" :request-id "/" :router-id "/" :service-id "/" :host "/" :port "/" [#".+" :location]]
                                      :async-status-handler-fn}
-                     "waiter-auth" :waiter-auth-handler-fn
+                     "waiter-auth" {true :waiter-auth-handler-fn}
```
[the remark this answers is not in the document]
sradack: So we'd have 
<pre>
{"" :waiter-auth-handler-fn
 "/extensions" :waiter-auth-handler-fn}
</pre>
?

waiter/test/waiter/auth/oauth_test.clj
sradack: It's never used this way.

waiter/test/waiter/auth/oauth_test.clj
sradack: _waiter_hash will always be the last parameter.

waiter/src/waiter/core.clj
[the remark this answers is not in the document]
sradack: It seems good to me to allow the auth handler full control over the `/waiter-auth/*` path.

waiter/src/waiter/auth/oauth.clj
shamsimam: It allows the authenticator to manage how the random numbers are generated and also helps with testability.

waiter/src/waiter/auth/oauth.clj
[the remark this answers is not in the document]
shamsimam: Should it? A query parameter can be repeated in which case `value` will be a sequence.

waiter/src/waiter/core.clj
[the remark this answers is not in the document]
shamsimam: I was thinking more along the lines of `"/extensions" :waiter-auth-extensions-handler-fn`.

waiter/test/waiter/auth/oauth_test.clj
shamsimam: It is a query parameter, why must it appear last?

waiter/src/waiter/auth/oauth/github.clj
```diff
@@ -0,0 +1,88 @@
...
+        (-> {:status 307
+             :headers {"location" location}}
+            ;; We should store this cookie only long enough to provide a reasonable amount of time
+            ;; for the client to authenticate.
+            (cookie-support/add-encoded-cookie password oauth/OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
+  (authenticate [_ request]
+    (fa/go-try
+      (let [{:strs [code state]} (:params (ring-params/params-request request))
+            _ (when-not (and code state)
+                (throw (ex-info "Malformed request" {:status 400})))
```
shamsimam: It would be nice to know why the request is considered malformed, e.g. include `params` in the `ex-data`.

waiter/test/waiter/auth/oauth_test.clj
sradack: `oauth.html` puts it on the end of the query string.

waiter/src/waiter/auth/oauth.clj
sradack: I still don't follow why we should do it that way.

waiter/src/waiter/auth/oauth.clj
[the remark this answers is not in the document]
sradack: No calls to this function have a need for multiple values.

waiter/src/waiter/core.clj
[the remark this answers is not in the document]
sradack: Bind it to a no-op if there are no extensions?  I'm having trouble seeing the benefit.

waiter/src/waiter/auth/oauth/github.clj
sradack: Will do.

waiter/config-shell-oauth.edn
```diff
@@ -0,0 +1,36 @@
+{
```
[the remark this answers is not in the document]
shamsimam: Looks like this is done: `test-validate-shell-oauth-settings`

waiter/resources/web/oauth.html
```diff
@@ -0,0 +1,49 @@
+<!doctype html>
```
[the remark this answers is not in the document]
shamsimam: Description has a screenshot.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
+;;
```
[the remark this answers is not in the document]
shamsimam: Looks like this is done.

waiter/src/waiter/auth/oauth.clj
shamsimam: To avoid dependency on global state and make the snippets more testable if we ever need to control how the random strings are generated.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+  (:import java.security.SecureRandom
+           org.eclipse.jetty.util.HttpCookieStore$Empty
+           org.eclipse.jetty.util.UrlEncoded))
+
+(def secure-random (SecureRandom.))
+
+(def OAUTH-COOKIE-NAME "x-waiter-oauth")
+(def HASH-QUERY-PARAM-NAME "_waiter_hash")
+
+(defn make-http-client
```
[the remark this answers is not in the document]
shamsimam: Configuration is done. 

Maybe make this function private?

waiter/test/waiter/auth/oauth_test.clj
shamsimam: Can we clarify this in the docstring then?

waiter/src/waiter/auth/oauth.clj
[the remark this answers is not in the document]
shamsimam: Please clarify in the docstring that repeated query parameters are not supported.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,304 @@
...
+                (throw (ex-info "Can't get email from GitHub" {:status 403})))]
+        (-> {:status 307
+             :headers {"location" location}}
+            (auth/add-cached-auth password email))))))
+
+(defn provider-list-handler
+  "Responds with a login page listing available OAuth providers."
+  [providers {:keys [query-string request-method] {:keys [path]} :route-params}]
+  (case request-method
+    :get {:body (template/eval (slurp (io/resource "web/oauth.html"))
```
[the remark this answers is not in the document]
shamsimam: This has been done.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,166 @@
...
+    :oauth)
+  (check-user [_ user _]
+    user)
+  (wrap-auth-handler [_ handler]
+    (wrap-oauth handler run-as-user password providers)))
+
+(defn oauth-authenticator
+  "Factory function for creating OAuthAuthenticator"
+  [{:keys [password providers run-as-user]}]
+  {:pre [password providers run-as-user]}
```
shamsimam: The example provided in `config-shell-oauth.edn` does not have a password entry.

shamsimam: Should it be an error if providers is empty?

waiter/src/waiter/core.clj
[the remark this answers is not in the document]
shamsimam: As I mentioned earlier I wanted to avoid the handler counter clashes. I guess we can just have a single counter for all `/waiter-auth...` requests and then other counters inside oauth for the different endpoints.

waiter/src/waiter/core.clj
```diff
@@ -1233,15 +1249,18 @@
                              (fn waiter-auth-handler-fn [request]
                                (handle-secure-request-fn
                                  (fn inner-waiter-auth-handler-fn [request]
```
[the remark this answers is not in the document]
shamsimam: This is done.

waiter/test/waiter/auth/oauth_test.clj
shamsimam: Okay.

waiter/src/waiter/auth/oauth/github.clj
```diff
@@ -0,0 +1,89 @@
...
+             :headers {"location" location}}
+            ;; We should store this cookie only long enough to provide a reasonable amount of time
+            ;; for the client to authenticate.
+            (cookie-support/add-encoded-cookie password oauth/OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
+  (authenticate [_ request]
+    (fa/go-try
+      (let [{:strs [code state] :as params} (:params (ring-params/params-request request))
+            _ (when-not (and code state)
+                (throw (ex-info "Malformed request" {:status 400
+                                                     :params params})))
```
shamsimam: Nitpick: lexicographic order :(

waiter/src/waiter/auth/oauth/google.clj
```diff
@@ -64,9 +64,10 @@
             (cookie-support/add-encoded-cookie password oauth/OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
   (authenticate [_ request]
     (fa/go-try
-      (let [{:strs [code state]} (:params (ring-params/params-request request))
+      (let [{:strs [code state] :as params} (:params (ring-params/params-request request))
             _ (when-not (and code state)
-                (throw (ex-info "Malformed request" {:status 400})))
+                (throw (ex-info "Malformed request" {:status 400
+                                                     :params params})))
```
shamsimam: Nitpick: lexicographic order :(

waiter/integration/waiter/basic_test.clj
```diff
@@ -349,13 +349,16 @@
   (testing-using-waiter-url
     (log/info "Basic waiter-auth test")
     (let [{:keys [status body headers]} (make-request waiter-url "/waiter-auth")
-          set-cookie (get headers "set-cookie")]
+          set-cookie (get headers "set-cookie")
+          parsed-json (-> (try-parse-json body)
+                          (walk/keywordize-keys))]
```
shamsimam: Nitpick: lone function can be specified without surrounding parens.

waiter/resources/web/oauth.html
```diff
@@ -0,0 +1,49 @@
...
+<body>
+
+<div class="main-content">
+    <h2 class="main-note">
+        Login to Waiter
+    </h2>
+    <div>
+        <p>Choose a provider:</p>
+        <% (doseq [[provider-key provider] providers] %>
+            <li><a class="redirect" href="/waiter-auth/oauth/<%= (name provider-key) %>/redirect/<%= path %><% (when query-string %><%= (str \? query-string) %><% ) %>"><%= (display-name provider) %></a></li>
```
shamsimam: Query: why does `(display-name provider)` not need the namespace qualification?

waiter/src/waiter/auth/oauth.clj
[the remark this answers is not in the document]
shamsimam: Okay.

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,166 @@
...
+
+(defn wrap-oauth
+  "Middleware for handling OAuth."
+  [handler-fn run-as-user password providers]
+  (fn [{:keys [headers query-string uri] :as request}]
+    (try
+      (let [{:keys [handler route-params] :as match} (match-oauth-route uri)
+            request' (assoc request :route-params route-params)
+            provider-fn (fn [] (-> route-params :provider keyword providers))]
+        (case handler
```
shamsimam: Can we please include counters on these handlers so that they show up in waiter requests, e.g. `(counters/inc! (metrics/waiter-counter "requests" "oauth" (name handler)))`?

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,166 @@
...
+          ; else
+          (let [waiter-cookie (auth/get-auth-cookie-value (get headers "cookie"))
+                [auth-principal _ :as decoded-auth-cookie] (auth/decode-auth-cookie waiter-cookie password)]
+            (if (auth/decoded-auth-valid? decoded-auth-cookie)
+              (-> request
+                  (auth/assoc-auth-in-request auth-principal)
+                  handler-fn)
+              {:status 307
+               :headers {"location" (cond-> (str "/waiter-auth/oauth/providers" uri)
+                                      (not (str/blank? query-string)) (str \? query-string))}}))))
```
shamsimam: Nitpick: lexicographic ordering.

waiter/src/waiter/auth/oauth/github.clj
```diff
@@ -0,0 +1,89 @@
...
+                                            (throw (ex-info "Couldn't parse state from GitHub" {:status 403} e))))
+            _ (oauth/verify-oauth-cookie request token password)
+            {:keys [body error status]} (async/<! (http/post http-client "https://github.com/login/oauth/access_token"
+                                                             {:form-params {"client_id" client-id
+                                                                            "client_secret" client-secret
+                                                                            "code" code
+                                                                            "state" state}}))
+            _ (when error
+                (throw (ex-info "Error while communicating with GitHub" {:status 403} error)))
+            _ (when-not (= status 200)
```
shamsimam: I did not find where the [github doc](https://developer.github.com/apps/building-integrations/setting-up-and-registering-github-apps/identifying-users-for-github-apps/) says to expect only a 200 response for success.

waiter/src/waiter/auth/oauth/google.clj
```diff
@@ -0,0 +1,110 @@
...
+                                                 b64/decode
+                                                 String.
+                                                 json/read-str)
+                                             (catch Exception e
+                                               (throw (ex-info "Coudln't parse id token from Google"
+                                                               {:status 403} e))))
+            _ (when-not (and (not (str/blank? email)) email_verified)
+                (throw (ex-info "Can't get email from Google" {:status 403})))]
+        (-> {:status 307
+             :headers {"location" location}}
```
shamsimam: Nitpick: lexicographic order :(

waiter/src/waiter/client_tools.clj
```diff
@@ -50,6 +50,13 @@
 
 (defn colored-time [time-string] (yellow time-string))
 
+(defn try-parse-json
```
shamsimam: This should go away after a rebase :)

waiter/resources/web/oauth.html
sradack: In `oauth.clj`
 `:display-name (fn [provider] (display-name provider))`

waiter/src/waiter/auth/oauth.clj
sradack: The password is passed in via context:

<pre>
 :authenticator (pc/fnk [[:settings authenticator-config]
                           passwords]
                    (utils/create-component authenticator-config :context {:password (first passwords)}))
</pre>

waiter/src/waiter/auth/oauth/github.clj
sradack: Which additional status codes do we think would also represent a valid response?  Do we think `204 No Content` would be valid?

shamsimam: I would vote for checking 2XX  `(<= 200 status 299)`

shamsimam: At the very least we should log why the auth failed.

waiter/src/waiter/auth/oauth.clj
shamsimam: Is the resolution on this - not needed now/won't do?

waiter/src/waiter/auth/oauth.clj
```diff
@@ -0,0 +1,171 @@
...
+
+(defn- make-http-client
+  "Instantiates and returns a new HttpClient without a cookie store"
+  []
+  (let [client (http/client {:connect-timeout 5000
+                             :idle-timeout 5000})]
+    (.setCookieStore client (HttpCookieStore$Empty.))
+    client))
+
+(defn random-str
```
shamsimam: Query: why don't we use (multiple) `UUID`(s) as the random string?

waiter/src/waiter/auth/oauth/github.clj
```diff
@@ -0,0 +1,88 @@
...
+            location (oauth/make-uri "https://github.com/login/oauth/authorize"
+                                     {"client_id" client-id
+                                      "scope" "user:email"
+                                      "redirect_uri" (authenticate-uri-fn request)
+                                      "state" state})]
+        (-> {:headers {"location" location}
+             :status 307}
+            ;; We should store this cookie only long enough to provide a reasonable amount of time
+            ;; for the client to authenticate.
+            (cookie-support/add-encoded-cookie password oauth/OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
```
shamsimam: Should we extract `(-> 15 t/minutes t/in-seconds)` in a helper function? It is used in both providers and seems like something we should have in `oauth.clj`?

waiter/src/waiter/auth/oauth/google.clj
```diff
@@ -0,0 +1,109 @@
...
+                                      "nonce" (oauth/random-str)
+                                      "response_type" "code"
+                                      "redirect_uri" (authenticate-uri-fn request)
+                                      "scope" "openid email"
+                                      "state" state})]
+        (-> {:headers {"location" location}
+             :status 307}
+            ;; We should store this cookie only long enough to provide a reasonable amount of time
+            ;; for the client to authenticate.
+            (cookie-support/add-encoded-cookie password oauth/OAUTH-COOKIE-NAME token (-> 15 t/minutes t/in-seconds))))))
```
shamsimam: 15 minutes here again.

waiter/src/waiter/core.clj
```diff
@@ -1250,15 +1250,17 @@
...
-                                   {:body (str (:authorization/user request)), :status 200})
+                                   (utils/map->json-response
+                                     {:user (:authorization/user request)
+                                      :principal (:authenticated-principal request)}))
```
shamsimam: Nitpick: lexicographic order.

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,230 @@
...
+        cookie (str OAUTH-COOKIE-NAME "=" (-> (cs/encode-cookie token password)
+                                              UrlEncoded/encodeString))]
+    (testing "redirect"
+      (let [{{:strs [set-cookie location]} :headers :keys [status]} (fa/<?? (redirect provider {}))]
+        (is location)
+        (is (= status 307))
+        (is set-cookie)))
+    (testing "authenticate"
+      (with-redefs [random-str (fn [] token)
+                    http/post (fn [_ _ _]
```
shamsimam: Should we assert on the urls?

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,230 @@
...
+                                (async/go {:body (async/go "access_token=accesstoken&token_type=bearer")
+                                           :status 200}))
+                    http/get (fn [_ _ _]
```
shamsimam: Should we assert on the urls?

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,230 @@
...
+    (testing "already authenticated"
+      (let [cookie (str auth/AUTH-COOKIE-NAME "=" (-> (cs/encode-cookie [principal now] password)
+                                                      UrlEncoded/encodeString))
+            {{:strs [location]} :headers :keys [body status]}
+            (wrapped-handler {:uri "/requires-auth"
+                              :headers {"cookie" cookie}})]
+        (is (= 200 status))
+        (is (= body "handler"))))))
+
+(deftest test-github-provider
```
shamsimam: Should this move to `waiter/src/waiter/auth/oauth/github_test.clj`?

waiter/test/waiter/auth/oauth_test.clj
```diff
@@ -0,0 +1,230 @@
...
+  (is (= "/path?a=1" (make-location {:path "path"
+                                     :query-string "a=1"})))
+  (is (= "/path?a=1&b=2" (make-location {:path "path"
+                                         :query-string "a=1&b=2"})))
+  (is (= "/path#b=2" (make-location {:path "path"
+                                     :query-string "_waiter_hash=%23b%3D2"})))
+  (is (= "/path?a=1#b=2" (make-location {:path "path"
+                                         :query-string "a=1&_waiter_hash=%23b%3D2"}))))
+
+(deftest test-google-provider
```
shamsimam: Should this move to `waiter/src/waiter/auth/oauth/google_test.clj`?

shamsimam:
Please rebase.

PR #165, opened 2017-10-12
WIP: Add support for leadership gauge

shamsimam:
Adds support for reporting leadership as a codahale gauge metric.


dposada:
Can we delete the branch?

PR #166, opened 2017-10-13
Refactor: instance-rpc-chan accepts maps instead of vectors as messages

shamsimam:
The refactoring is in preparation for the last request time feature.

### DIFF
--- a/waiter/src/waiter/service.clj
+++ b/waiter/src/waiter/service.clj
@@ -43,15 +43,24 @@
 
 ;;; Service instance blacklisting, work-stealing, access and creation
 
+(defn instance-rpc-message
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/service.clj
```diff
@@ -43,15 +43,24 @@
 
 ;;; Service instance blacklisting, work-stealing, access and creation
 
+(defn instance-rpc-message
```
sradack: Let's not move from positionality to more positionality.  Why not have the maps be inline?

shamsimam: Done.

shamsimam:
RTM

PR #167, opened 2017-10-13
Last request time metric

shamsimam:
`last-request-time` are aggregated per service in a `last-request-times-agent`.
The agent values are publsihed to codahale counters every `last-request-times-publish-interval-ms` milliseconds (configurable via settings).
Publishing of `service-id->last-request-time` is triggered via a timer task instead of an agent action to avoid queueing delays.
The GC of published values are triggered via agent actions.

Screenshots of state output including `last-request-time`s:

screen shot 2017-11-02 at 10 57 40 pm

screen shot 2017-11-02 at 10 51 05 pm

screen shot 2017-11-02 at 9 22 55 pm

screen shot 2017-11-02 at 9 23 03 pm

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -204,7 +204,8 @@
                    :x-waiter-concurrency-level num-iteration-requests}
           {:keys [request-headers service-id] :as first-response}
           (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :http-method-fn http/get))]
-      (try
+      (with-service-cleanup
+        service-id
         (assert-response-status first-response 200)
         (dotimes [_ num-iteration-requests]
           (-> (make-kitchen-request waiter-url request-headers :http-method-fn http/get)
@@ -217,15 +218,13 @@
           (Thread/sleep last-request-publish-wait-time-ms)
           (let [service (service waiter-url service-id {})
                 service-metrics (-> (service-settings waiter-url service-id)
-                                     (get-in [:metrics :aggregate]))]
+                                    (get-in [:metrics :aggregate]))]
             (is service)
             (is (< current-last-request-time (get service "last-request-time"))
                 (str {:current-last-request-time current-last-request-time, :service service}))
             (is (= (get service "last-request-time")
                    (get-in service-metrics [:counters :last-request-time]))
-                (str {:service service, :service-metrics service-metrics}))))
-        (finally
-          (delete-service waiter-url service-id))))))
+                (str {:service service, :service-metrics service-metrics}))))))))
 
 (deftest ^:parallel ^:integration-fast test-list-apps
   (testing-using-waiter-url
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -202,11 +202,11 @@
                    :x-waiter-cmd (kitchen-cmd "-p $PORT0")}
           {:keys [headers request-headers service-id] :as first-response}
           (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :http-method-fn http/get))
-          canary-request-time-from-header (-> (get headers "x-waiter-request-timestamp" "0")
+          _ (assert-response-status first-response 200)
+          canary-request-time-from-header (-> (get headers "x-waiter-request-date")
                                               (utils/str-to-date utils/formatter-rfc822))]
       (with-service-cleanup
         service-id
-        (assert-response-status first-response 200)
         (is (pos? metrics-sync-interval-ms))
         (let [service-last-request-time (service-id->last-request-time waiter-url service-id)]
           (is (pos? (.getMillis canary-request-time-from-header)))
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -195,6 +195,7 @@
     (let [waiter-settings (waiter-settings waiter-url)
           last-request-times-publish-interval-ms (get-in waiter-settings [:last-request-times-publish-interval-ms])
           metrics-sync-interval-ms (get-in waiter-settings [:metrics-config :metrics-sync-interval-ms])
+          last-request-publish-wait-time-ms (+ last-request-times-publish-interval-ms metrics-sync-interval-ms)
           num-iteration-requests 20
           service-name (rand-name)
           headers {:x-kitchen-delay-ms 1000
@@ -208,16 +209,18 @@
         (dotimes [_ num-iteration-requests]
           (-> (make-kitchen-request waiter-url request-headers :http-method-fn http/get)
               (assert-response-status 200)))
-        (let [current-time-millis (System/currentTimeMillis)]
+        (Thread/sleep last-request-publish-wait-time-ms)
+        (let [current-last-request-time (-> (service-settings waiter-url service-id)
+                                            (get-in [:metrics :aggregate :counters :last-request-time]))]
           (make-kitchen-request waiter-url request-headers :http-method-fn http/get)
           ;; allow some time to elapse to allow inter-router metrics syncing
-          (Thread/sleep (+ last-request-times-publish-interval-ms metrics-sync-interval-ms))
+          (Thread/sleep last-request-publish-wait-time-ms)
           (let [service (service waiter-url service-id {})
                 service-metrics (-> (service-settings waiter-url service-id)
                                      (get-in [:metrics :aggregate]))]
             (is service)
-            (is (< current-time-millis (get service "last-request-time"))
-                (str {:current-time-millis current-time-millis, :service service}))
+            (is (< current-last-request-time (get service "last-request-time"))
+                (str {:current-last-request-time current-last-request-time, :service service}))
             (is (= (get service "last-request-time")
                    (get-in service-metrics [:counters :last-request-time]))
                 (str {:service service, :service-metrics service-metrics}))))
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -9,7 +9,8 @@
 ;;       actual or intended publication of such source code.
 ;;
 (ns waiter.basic-test
-  (:require [clojure.core.async :as async]
+  (:require [clj-time.core :as t]
+            [clojure.core.async :as async]
             [clojure.data.json :as json]
             [clojure.set :as set]
             [clojure.string :as str]
@@ -18,7 +19,8 @@
             [clojure.walk :as walk]
             [qbits.jet.client.http :as http]
             [waiter.client-tools :refer :all]
-            [waiter.service-description :as sd])
+            [waiter.service-description :as sd]
+            [waiter.utils :as utils])
   (:import java.io.ByteArrayInputStream))
 
 (deftest ^:parallel ^:integration-fast test-basic-functionality
@@ -201,20 +203,19 @@
           {:keys [headers request-headers service-id] :as first-response}
           (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :http-method-fn http/get))
           canary-request-time-from-header (-> (get headers "x-waiter-request-timestamp" "0")
-                                               Long/parseLong)]
+                                              (utils/str-to-date utils/formatter-rfc822))]
       (with-service-cleanup
         service-id
         (assert-response-status first-response 200)
         (is (pos? metrics-sync-interval-ms))
         (let [service-last-request-time (service-id->last-request-time waiter-url service-id)]
-          (is (pos? canary-request-time-from-header))
-          (is (pos? service-last-request-time))
-          (is (= canary-request-time-from-header service-last-request-time)))
-        (Thread/sleep 100)
+          (is (pos? (.getMillis canary-request-time-from-header)))
+          (is (pos? (.getMillis service-last-request-time)))
+          (is (zero? (t/in-seconds (t/interval canary-request-time-from-header service-last-request-time)))))
         (make-kitchen-request waiter-url request-headers :http-method-fn http/get)
         (let [service-last-request-time (service-id->last-request-time waiter-url service-id)]
-          (is (pos? service-last-request-time))
-          (is (< canary-request-time-from-header service-last-request-time)))))))
+          (is (pos? (.getMillis service-last-request-time)))
+          (is (t/before? canary-request-time-from-header service-last-request-time)))))))
 
 (deftest ^:parallel ^:integration-fast test-list-apps
   (testing-using-waiter-url
@@ -224,19 +225,19 @@
       (testing "without parameters"
         (let [service (service waiter-url service-id {})] ;; see my app as myself
           (is service)
-          (is (pos? (get service "last-request-time")))
+          (is (-> (get service "last-request-time") utils/str-to-date .getMillis pos?))
           (is (pos? (get-in service ["service-description" "cpus"])) service)))
 
       (testing "waiter user disabled" ;; see my app as myself
         (let [service (service waiter-url service-id {"force" "false"})]
           (is service)
-          (is (pos? (get service "last-request-time")))
+          (is (-> (get service "last-request-time") utils/str-to-date .getMillis pos?))
           (is (pos? (get-in service ["service-description" "cpus"])) service)))
 
       (testing "waiter user disabled and same user" ;; see my app as myself
         (let [service (service waiter-url service-id {"force" "false", "run-as-user" (retrieve-username)})]
           (is service)
-          (is (pos? (get service "last-request-time")))
+          (is (-> (get service "last-request-time") utils/str-to-date .getMillis pos?))
           (is (pos? (get-in service ["service-description" "cpus"])) service)))
 
       (testing "different run-as-user" ;; no such app
@@ -266,7 +267,7 @@
       (testing "list-apps-with-waiter-user-disabled-and-see-another-app" ;; can see another user's app
         (let [service (service waiter-url service-id {"force" "false", "run-as-user" current-user})]
           (is service)
-          (is (pos? (get service "last-request-time")))
+          (is (-> (get service "last-request-time") utils/str-to-date .getMillis pos?))
           (is (pos? (get-in service ["service-description" "cpus"])) service)))
       (delete-service waiter-url service-id))))
 
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -192,9 +192,13 @@
 
 (deftest ^:parallel ^:integration-fast test-last-request-time
   (testing-using-waiter-url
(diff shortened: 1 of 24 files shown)


waiter/src/waiter/core.clj
```diff
@@ -88,14 +88,16 @@
...
                      "sim" :sim-request-handler
-                     "state" {"" :state-all-handler-fn
-                              "/kv-store" :state-kv-store-handler-fn
-                              "/leader" :state-leader-handler-fn
-                              "/maintainer" :state-maintainer-handler-fn
-                              "/router-metrics" :state-router-metrics-handler-fn
-                              "/scheduler" :state-scheduler-handler-fn
-                              "/statsd" :state-statsd-handler-fn
-                              ["/" :service-id] :state-service-handler-fn}
+                     "state" (array-map ;; maintain key order, explicitly use array-map when size > 8
```
shamsimam: @dposada @pschorf @sradack this is where the unit tests for route matching came in handy!

DaoWen: I don't think it makes sense to use maps (whether array-map or hash-map) for a data structure where the order of entries is important. I think this is especially true since "maintaining key order" isn't part of the [array-map's public contract](https://clojure.github.io/clojure/clojure.core-api.html#clojure.core/array-map).

These route/pattern entries should *all* be vectors of pairs, not maps. I think that using vectors will make it much more intuitive to anyone reading the code that order matters, and eliminates the need for extra comments trying to clarify that fact. (I already tested this, and bidi is happy with a vector-of-vectors here instead of an array-map).

shamsimam: From the [clojure.org reference](https://clojure.org/reference/data_structures#ArrayMaps) 

> When doing code form manipulation it is often desirable to have a map which maintains key order. An array map is such a map - it is simply implemented as an array of key val key val…

waiter/src/waiter/core.clj
```diff
@@ -88,14 +88,15 @@
...
+                              ["/last-request-times" :state-last-request-times-handler-fn]
+                              ["/leader" :state-leader-handler-fn]
+                              ["/maintainer" :state-maintainer-handler-fn]
+                              ["/router-metrics" :state-router-metrics-handler-fn]
+                              ["/scheduler" :state-scheduler-handler-fn]
+                              ["/statsd" :state-statsd-handler-fn]
+                              [["/" :service-id] :state-service-handler-fn]]
                      "status" :status-handler-fn
                      "token" :token-handler-fn
                      "tokens" { "" :token-list-handler-fn
```
DaoWen: We should change **all** of these map literals to vectors-of-vectors so that we can ensure the order. That way when the `"apps"` entry above crosses the threshold from 8 to 9 somewhere down the road, we don't get the same weird behavior again. Since that threshold number (8 or 9 or whatever) is an implementation detail, we really shouldn't be relying on that anyway.

shamsimam: That can be part of another PR, I don't want to extend the scope of this PR

waiter/src/waiter/handler.clj
```diff
@@ -246,14 +246,16 @@
...
                                           :unhealthy-instances (count (get-in current-state [:service-id->unhealthy-instances service-id]))})
+              service-id->metrics (service-id->metrics-fn)
               include-effective-parameters? (utils/request-flag request-params "effective-parameters")
               response-data (map
                              (fn service-id->service-info [service-id]
                                (let [service-description (service-id->service-description-fn service-id :effective? false)]
                                  (cond->
-                                     {:service-id service-id
+                                     {:instance-counts (retrieve-instance-counts service-id)
+                                      :last-request-time (get-in service-id->metrics [service-id "last-request-time"])
```
sradack: Metrics are for external consumption.  We'd do well to move away using them to pass data around inside Waiter.  Would it be better to read from the agent instead?

waiter/src/waiter/handler.clj
```diff
@@ -536,6 +538,16 @@
...
       (utils/exception->response ex request))))
 
+(defn get-last-request-times-state
+  "Outputs the last-request-time state."
+  [router-id last-request-time-agent request]
+  (try
+    (-> {:router-id router-id
+         :state @last-request-time-agent}
+        (utils/map->streaming-json-response))
+    (catch Exception ex
```
sradack: Now that we have error handling middleware, we don't necessarily need to handle the exception here, especially given that we aren't adding decorating the exception with `:status`.

waiter/src/waiter/core.clj
```diff
@@ -454,6 +455,9 @@
    :http-client (pc/fnk [[:settings [:instance-request-properties connection-timeout-ms]]]
                   (http-client-factory connection-timeout-ms))
    :instance-rpc-chan (pc/fnk [] (async/chan 1024)) ; TODO move to service-chan-maintainer
+   :last-request-time-agent (pc/fnk [] (agent {:last-published {:service-id->last-request-time {}
+                                                                :time nil}
+                                               :pending {:service-id->last-request-time {}}}))
```
sradack: Why aren't we just keeping the latest request times of all services?  Sending only diffs seems more complex.  How do we handle populating a newly restarted router's request times?

waiter/src/waiter/metrics.clj
```diff
@@ -442,4 +458,73 @@
...
+            (< (.getMillis current-last-request-time)
+               (.getMillis candidate-last-request-time)))
+      (update-in agent-state [:pending :service-id->last-request-time]
+                 assoc service-id candidate-last-request-time)
+      agent-state)))
+
+(defn cleanup-last-request-times
+  "Removes stale entries in service-id->last-request-time based on entries provided in
+   published-service-id->last-request-time. Services who have more recent last request
+   times than those have been published will not be gc-ed."
```
sradack: "GC" doesn't seem like the right term to describe what is happening.  Am I understanding correctly that we are cleaning up services that have already been sent?

waiter/src/waiter/metrics.clj
```diff
@@ -442,4 +458,73 @@
...
+                      (and current-last-request-time
+                           (<= (.getMillis current-last-request-time)
+                               (.getMillis candidate-last-request-time)))
+                      (dissoc service-id))))
+          service-id->last-request-time
+          published-service-id->last-request-time)
+        num-items-cleaned (- (count service-id->last-request-time)
+                             (count service-id->last-request-time'))]
+    (when (pos? num-items-cleaned)
+      (cid/cinfo correlation-id "gc-ed" num-items-cleaned "services"))
```
sradack: Does this get printed every time any single service had a request but doesn't have a subsequent request within the publish time?

waiter/src/waiter/metrics.clj
```diff
@@ -442,4 +458,73 @@
...
+    (assoc agent-state
+      :last-published {:service-id->last-request-time published-service-id->last-request-time
+                       :time publish-time}
+      :pending {:service-id->last-request-time service-id->last-request-time'})))
+
+(defn publish-last-request-times!
```
sradack: Why aren't we `send`ing this to the agent, instead of `deref`ing the agent and then `send`ing the modification.  Shouldn't those actions be atomic?

shamsimam:
Reminder, fix this test:

```
lein parallel-test :only waiter.websocket-integration-test/test-last-request-time
FAIL in (test-last-request-time) (websocket_integration_test.clj:147)
test-last-request-time
{:data [1509024269233 1509024282897 1509024284142 1509024286441 1509024286441 1509024287511 1509024288592 1509024289656 1509024290719 1509024291780 1509024292844], :diffs (13664 1245 2299 0 1070 1081 1064 1063 1061 1064)}
expected: (every? (fn* [p1__62178#] (>= p1__62178# last-request-times-publish-interval-ms)) last-request-times-diff)
 actual: (not (every? #object[waiter.websocket_integration_test$fn__62179$fn__62180$fn__62181$fn__62364 0x60abc3aa "waiter.websocket_integration_test$fn__62179$fn__62180$fn__62181$fn__62364@60abc3aa"] (13664 1245 2299 0 1070 1081 1064 1063 1061 1064)))
	FINISH: waiter.websocket-integration-test/test-last-request-time {:test 3, :pass 7, :fail 1, :error 0}
	START: waiter.websocket-integration-test/test-request-stream-bytes-and-string
	FINISH: waiter.metrics-output-test/test-metrics-output {:test 4, :pass 16, :fail 1, :error 0}
```

waiter/integration/waiter/basic_test.clj
```diff
@@ -190,6 +190,42 @@
       (is (= {:foo "bar", :baz "quux", :begindate "null", :enddate "null", :timestamp "20160713201333949"} value))
       (delete-service waiter-url service-id))))
 
+(deftest ^:parallel ^:integration-fast test-last-request-time
+  (testing-using-waiter-url
+    (let [waiter-settings (waiter-settings waiter-url)
+          last-request-times-publish-interval-ms (get-in waiter-settings [:last-request-times-publish-interval-ms])
+          metrics-sync-interval-ms (get-in waiter-settings [:metrics-config :metrics-sync-interval-ms])
+          last-request-publish-wait-time-ms (+ last-request-times-publish-interval-ms metrics-sync-interval-ms)
+          num-iteration-requests 20
```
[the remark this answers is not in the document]
shamsimam: Changed to use just one request.

waiter/integration/waiter/basic_test.clj
```diff
@@ -190,6 +190,42 @@
...
+          (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :http-method-fn http/get))]
+      (with-service-cleanup
+        service-id
+        (assert-response-status first-response 200)
+        (dotimes [_ num-iteration-requests]
+          (-> (make-kitchen-request waiter-url request-headers :http-method-fn http/get)
+              (assert-response-status 200)))
+        (Thread/sleep last-request-publish-wait-time-ms)
+        (let [current-last-request-time (-> (service-settings waiter-url service-id)
+                                            (get-in [:metrics :aggregate :counters :last-request-time]))]
```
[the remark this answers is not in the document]
shamsimam: We pass along the information as a debug header now.

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -88,6 +88,65 @@
...
+          waiter-headers (assoc (kitchen-request-headers)
+                           "x-waiter-metric-group" "test-ws-support"
+                           "x-waiter-name" (rand-name))
+          service-id (retrieve-service-id waiter-url waiter-headers)
+          retrieve-last-request-time (fn []
+                                       (-> (service-settings waiter-url service-id)
+                                           (get-in [:metrics :aggregate :counters :last-request-time])))
+          last-time-trigger-chan (async/chan)
+          last-request-times-atom (atom [])
+          num-iterations 10]
```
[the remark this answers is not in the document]
shamsimam: The multiple iterations are there to ensure that the last request time from the multiple WebSocket frames is being reported.

waiter/integration/waiter/websocket_integration_test.clj
```diff
@@ -88,6 +88,65 @@
...
+                (async/close! out)
+                (async/close! last-time-trigger-chan)))
+            {:middleware (fn [_ ^UpgradeRequest request]
+                           (websocket/add-headers-to-upgrade-request! request waiter-headers)
+                           (add-auth-cookie request auth-cookie-value))})
+          (async/go-loop []
+            (if (async/<! last-time-trigger-chan)
+              (do
+                (log/info "retrieving last request time")
+                (swap! last-request-times-atom conj (retrieve-last-request-time))
```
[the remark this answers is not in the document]
shamsimam: Test has been rewritten. There is now an atom to track the connect time.

waiter/src/waiter/metrics.clj
shamsimam: Removed GC from the documentation and logging.

waiter/src/waiter/metrics.clj
shamsimam: This function has been removed.

shamsimam:
Addressed previous feedback, ready for more :)

waiter/src/waiter/metrics.clj
```diff
@@ -151,35 +152,39 @@
 
 (defn get-core-metrics
```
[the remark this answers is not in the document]
sradack: We're conflating metrics with state synchronization.  The `waiter.metrics` namespace is for Coda Hale metrics.  The `waiter.metrics-sync` is for syncing certain state between routers.  Now that we're syncing more than just "metrics" (ala Coda Hale), we should adjust things.   The conflation of the term "metric" is leading to ambiguity throughout.

waiter/src/waiter/metrics.clj
```diff
@@ -388,12 +398,30 @@
           (recur))
         (log/info "[transient-metrics-gc] stopping populating service-id->state-chan")))))
 
+(defn update-last-request-time
+  "Updates the last-request-time epoch time in the service-id->local-metrics to the maximum
+   of the current last request time and the provided last-request-datetime in epoch millis
+   for the specified service."
+  [service-id->local-metrics service-id ^DateTime candidate-last-request-time]
+  (let [current-last-request-time (get-in service-id->local-metrics [service-id "last-request-time"])
+        candidate-last-request-time-ms (.getMillis candidate-last-request-time)]
```
sradack: I don't think we should adopt long millis as our representation of time.  We use `clj-time` for other timestamps, like `last-update-time` in the scheduler.  Using millis is going to make inspecting state inspection harder for humans.

sradack: We should fix `last-update-time` in tokens to be consistent as well before it's too late.

waiter/integration/waiter/basic_test.clj
```diff
@@ -190,6 +190,32 @@
...
+  (testing-using-waiter-url
+    (let [waiter-settings (waiter-settings waiter-url)
+          metrics-sync-interval-ms (get-in waiter-settings [:metrics-config :metrics-sync-interval-ms])
+          service-name (rand-name)
+          headers {:x-kitchen-delay-ms (* 4 metrics-sync-interval-ms)
+                   :x-waiter-name service-name
+                   :x-waiter-cmd (kitchen-cmd "-p $PORT0")}
+          {:keys [headers request-headers service-id] :as first-response}
+          (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :http-method-fn http/get))
+          canary-request-time-from-header (-> (get headers "x-waiter-request-timestamp" "0")
```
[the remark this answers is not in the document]
sradack: This timestamp is going to be inconsistent with the timestamp shown on error messages.

shamsimam:
Build is green! Ready for next round of feedback.

waiter/src/waiter/handler.clj
```diff
@@ -512,13 +511,14 @@
...
-  [state-chan scheduler-chan router-metrics-state-fn kv-store leader?-fn request]
+  [state-chan scheduler-chan router-metrics-state-fn kv-store leader?-fn local-metrics-agent request]
   (async/go
     (try
       (let [timeout-ms 30000]
         (-> (async/<! (retrieve-maintainer-state state-chan timeout-ms))
             (assoc :kv-store (kv/state kv-store)
                    :leader (leader?-fn)
+                   :local-metrics @local-metrics-agent
                    :router-metrics-state (router-metrics-state-fn)
```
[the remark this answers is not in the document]
sradack: Sounds good.

waiter/src/waiter/metrics_sync.clj
```diff
@@ -343,11 +348,34 @@
         metrics-agent (agent initial-state)]
     metrics-agent))
 
+(defn- merge-router-metrics
```
sradack: How about factoring out something like `merge-by` (similar to`merge-with`) that also takes the key as an argument?

<pre>
(merge-by
  (fn [key v1 v2] ...)
  m1 m2)
</pre>

waiter/src/waiter/process_request.clj
```diff
@@ -496,7 +497,11 @@
           (try
             (confirm-live-connection-without-abort)
             (let [{:keys [service-id service-description] :as descriptor} (request->descriptor-fn request)
-                  {:strs [metric-group]} service-description]
+                  {:strs [metric-group]} service-description
+                  ^DateTime request-time (t/now)]
+              (->> (utils/date-to-str request-time utils/formatter-rfc822)
+                   (add-debug-header-into-response! "x-waiter-request-timestamp"))
```
sradack: Name suggestion: `x-waiter-request-date` (a la the HTTP `Date` http header) or `x-waiter-request-datetime` or `x-waiter-request-time`

waiter/src/waiter/handler.clj
```diff
@@ -222,46 +222,45 @@
...
+                                       {:healthy-instances (count (get-in current-state [:service-id->healthy-instances service-id]))
+                                        :unhealthy-instances (count (get-in current-state [:service-id->unhealthy-instances service-id]))})
+            service-id->metrics (service-id->metrics-fn)
+            include-effective-parameters? (utils/request-flag request-params "effective-parameters")
+            response-data (map
+                            (fn service-id->service-info [service-id]
+                              (let [service-description (service-id->service-description-fn service-id :effective? false)]
+                                (cond->
+                                  {:instance-counts (retrieve-instance-counts service-id)
+                                   :last-request-time (get-in service-id->metrics [service-id "last-request-time"])
```
sradack: Why is the time being output in a local time zone?  `(t/now)` should be in UTC.  Our API should report times in UTC.  Separate PR is OK but we should do it before the next release.

waiter/src/waiter/metrics.clj
```diff
@@ -388,12 +392,29 @@
           (recur))
         (log/info "[transient-metrics-gc] stopping populating service-id->state-chan")))))
 
+(defn update-last-request-time
```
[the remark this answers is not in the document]
shamsimam: Appended `-usage-metric` to name to clarify which metric we are talking about.

waiter/src/waiter/metrics.clj
```diff
@@ -388,12 +392,29 @@
...
+   of the current last request time and the provided last-request-datetime in epoch millis
+   for the specified service."
+  [service-id->local-metrics service-id ^DateTime candidate-last-request-time]
+  (let [current-last-request-time (get-in service-id->local-metrics [service-id "last-request-time"])]
+    (if (or (nil? current-last-request-time)
+            (t/after? candidate-last-request-time current-last-request-time))
+      (assoc-in service-id->local-metrics [service-id "last-request-time"] candidate-last-request-time)
+      service-id->local-metrics)))
+
+(defn cleanup-local-metrics
```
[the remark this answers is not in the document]
shamsimam: Appended `-usage-metrics` to name to clarify which metric we are talking about.

shamsimam:
Ready for next round of feedback.

shamsimam:
@DaoWen any more feedback on this?

PR #168, opened 2017-10-14
Implements error middleware

sradack:


waiter/src/waiter/core.clj
```diff
@@ -178,6 +178,22 @@
...
+(defn wrap-error-handling
+  "Catches any uncaught exceptions and returns an error response."
+  [handler]
+  (fn wrap-error-handling-fn [request]
+    (try
+      (let [response (handler request)]
+        (if (au/chan? response)
+          (async/go
+            (try
+              (<? response)
```
shamsimam: Nitpick: Is it better to avoid the `throw` (and the try-catch block) and instead do a read (`async/<!`) and do an if-else based on `(instance? Throwable result)`?

sradack: In a former branch that I never made into a PR, I did it that way.  When I wrote it again I found this to be more pleasing.  What I like about this approach is that `<?` can be used with `go-try` without ever needing to know the underlying implementation, whether they communicate using just the bare exception, the exception in a map, or some other representation.

sradack: I also liked that it mirrored the non-async version more closely.

PR #169, opened 2017-10-16
Ensure nil is not written to response-chan while retrieving state …

shamsimam:
…for invalid service-ids

### DIFF
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -513,7 +513,7 @@
                          (if service-id
                            (let [scheduler-state (service-id->state scheduler service-id)
                                  health-check-context (get current-state service-id)]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler.clj
```diff
@@ -513,7 +513,7 @@
                          (if service-id
                            (let [scheduler-state (service-id->state scheduler service-id)
                                  health-check-context (get current-state service-id)]
-                             (async/>! response-chan (merge scheduler-state health-check-context)))
+                             (async/>! response-chan (merge {} scheduler-state health-check-context)))
```
sradack: Would we be safer with an integration test that ensured Waiter didn't crash if we passed an invalid service id?

shamsimam: Will add.

shamsimam:
RTM

PR #171, opened 2017-10-17
scheduler package

shamsimam:


waiter/src/waiter/scheduler/scheduler.clj
```diff
@@ -8,7 +8,7 @@
 ;;       The copyright notice above does not evidence any
 ;;       actual or intended publication of such source code.
 ;;
-(ns waiter.scheduler
+(ns waiter.scheduler.scheduler
```
sradack: Why not leave everything you moved to `waiter.scheduler.scheduler` in `waiter.scheduler`?  Prior art:

https://github.com/clojure/clojure/tree/master/src/clj/clojure, `core.clj` and `core` folder

pschorf: I agree with @sradack

shamsimam: Undid move of `scheduler.clj`

waiter/src/waiter/scheduler/shell_scheduler.clj
```diff
@@ -8,7 +8,7 @@
 ;;       The copyright notice above does not evidence any
 ;;       actual or intended publication of such source code.
 ;;
-(ns waiter.shell-scheduler
+(ns waiter.scheduler.shell-scheduler
```
sradack: Now that we have `scheduler` in the namespace, the `-scheduler` in `shell-scheduler` seems redundant.  What do you think?

shamsimam: I'm ok renaming it to `shell.clj`. Or do you prefer a different name?

shamsimam: Renamed to `shell.clj`

shamsimam:
RTM

shamsimam:
Can we get this merged?

Issue #172, opened 2017-10-17
Router crashes on marathon ConnectException


pschorf:
A ConnectException thrown at https://github.com/twosigma/waiter/blob/master/waiter/src/waiter/scheduler.clj#L348 causes the router to crash. The fix is probably to extend https://github.com/twosigma/waiter/blob/master/waiter/src/waiter/scheduler.clj#L107 to support ConnectException in addition to checking for 5xx status codes.

shamsimam:
Closing as the issue has been resolved.

PR #173, opened 2017-10-17
Issue #172 - handle transient scheduler connection issues

shamsimam:


waiter/src/waiter/scheduler.clj
```diff
@@ -119,7 +119,10 @@
                                     (ss/try+
                                       {:success (f)}
                                       (catch #(contains? retry-status-codes (:status %)) e
-                                        (log/warn (str "Scheduler unavailable (Error code:" (:status e) ").") msg)
+                                        (log/warn (str "scheduler unavailable (error code:" (:status e) ").") msg)
+                                        (ss/throw+ e))
+                                      (catch ConnectException e
```
sradack: Should we handle TimeoutException?  Why shouldn't we handle any exception at all in the same manner?

sradack:
What are you planning to check inside `/state` to ensure test if Waiter can talk to Marathon?

waiter/src/waiter/scheduler.clj
shamsimam: Other exceptions might not be transient.

shamsimam:
> What are you planning to check inside /state to ensure test if Waiter can talk to Marathon?

This change can still cause the router to fail fatally if the connection error is not transient and persists beyond the retry period.

shamsimam:
I am working on a separate PR that ensures the router keeps running even if the scheduler connections fail and exposing a waiter codahale metric gauge that specifies the last time a router heard from the scheduler.

waiter/src/waiter/scheduler.clj
sradack: Can we find out what exception occurs on timeouts and handle that as well?

shamsimam: Okay

shamsimam:
RTM

PR #174, opened 2017-10-17
Continue chugging along even if scheduler updates fail

shamsimam:


waiter/src/waiter/scheduler.clj
```diff
@@ -491,12 +493,15 @@
...
   and sends the data to the router state maintainer."
-  [scheduler scheduler-state-chan scheduler-syncer-interval-secs
+  [clock scheduler scheduler-state-chan scheduler-syncer-interval-secs
    service-id->service-description-fn available? http-client failed-check-threshold]
   (log/info "Starting scheduler syncer")
   (let [exit-chan (async/chan 1)
         state-query-chan (async/chan 32)
-        timeout-chan (chime/chime-ch (utils/time-seq (t/now) (t/seconds scheduler-syncer-interval-secs)))]
+        timeout-chan (chime/chime-ch (utils/time-seq (t/now) (t/seconds scheduler-syncer-interval-secs)))
+        last-request-time-atom (atom 0)
```
sradack: Can we add this to `current-state` and avoid the complexity of gauges/atoms?

shamsimam:
RTM

PR #177, opened 2017-10-19
Handle channels in get-deployment-info

shamsimam:
-- gitignore only top-level scheduler directory


shamsimam:
RTM, build is green

PR #181, opened 2017-10-23
Add logging for instance slot distribution

shamsimam:
Also, reduce state tracker per request

### DIFF
--- a/waiter/src/waiter/state.clj
+++ b/waiter/src/waiter/state.clj
@@ -1002,13 +1002,14 @@
   "Computes the slot distribution of instances on a given router."
   [router-id router-id->http-endpoint service-id->service-description-fn service-id->healthy-instances]
   (let [service-id->router->instance->slots
-        (into {}
-              (map (fn [[service-id instances]]
-                     (let [{:strs [concurrency-level distribution-scheme]} (service-id->service-description-fn service-id)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/state.clj
```diff
@@ -1002,13 +1002,14 @@
...
-                           router-id->instance->slots (distribute-slots-using-consistent-hash-distribution
-                                                        (keys router-id->http-endpoint) instances md5-hash-function concurrency-level distribution-scheme)]
-                       [service-id router-id->instance->slots]))
-                   service-id->healthy-instances))]
+        (->> service-id->healthy-instances
+             (map (fn [[service-id instances]]
+                    (let [{:strs [concurrency-level distribution-scheme]} (service-id->service-description-fn service-id)
+                          router-id->instance->slots (distribute-slots-using-consistent-hash-distribution
+                                                       (keys router-id->http-endpoint) instances md5-hash-function concurrency-level distribution-scheme)]
+                      (log/info "instance distribution of" service-id "is" (pc/map-vals #(pc/map-keys :id %) router-id->instance->slots))
```
dposada: How often does this get logged?

shamsimam: The computation is triggered when instances for a service change. This happens when we scale or if instances die/become unhealthy.

PR #188, opened 2017-11-28
outputs token last-update-time in ISO-8601 format

shamsimam:


.travis.yml
```diff
@@ -1,3 +1,6 @@
+sudo: required
+dist: trusty
+group: deprecated-2017Q4
```
DaoWen: Why do we need the deprecated image?

shamsimam: The builds fail on the new image with `clojure.lang.Compiler$CompilerException: java.lang.ClassNotFoundException: org.sonatype.aether.util.version.GenericVersionScheme, compiling:(leiningen/voom.clj:1:1)`. I haven't looked deeper into this, but I suspect this is an issue with a Maven upgrade on the Travis worker.
https://travis-ci.org/twosigma/waiter/jobs/315702964

shamsimam:
@pschorf @DaoWen  RTM

PR #192, opened 2017-12-16
Avoid throwing exceptions in too many request in queue response

shamsimam:

- refactors exception->response to allow passing custom data maps
- refactors request preprocessors to use data->error-response instead of generating exceptions
- refactors process to run request preprocessors outside the async block

The pre-processors do not need to execute inside the go-blocks nor need the overhead of generating exceptions while responding. By saving on these overheads, we can improve the overall performance while rejecting requests.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -196,7 +196,7 @@
 
 (defn- wrap-exception
   "Includes metadata such as cid and status along with the exception."
-  [exception instance message status headers]
+  [message exception instance status headers]
   (ex-info message
            (merge (metrics/retrieve-local-stats-for-service (scheduler/instance->service-id instance))
                   {:instance instance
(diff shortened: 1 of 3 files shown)


dposada:
Do we have something that shows that this helped performance?

shamsimam:
@dposada 

> Do we have something that shows that this helped performance?

Nope, I am currently manually testing with artificial load to verify that this helped. Did you want an automated test for this?

dposada:
It might be nice to have some "proof" that this is better

dposada:
Any thoughts on "proof" that this is better?

shamsimam:
@dposada I have run the same stress test on Waiter for Max Queue length exceeded and Waiter is running fine.

The context for this PR has changed slightly as the template rendering has been identified as the main issue for slowness. This PR is now an incremental improvement where we can render the error response without throwing an exception.

PR #194, opened 2017-12-19
Uses the faster template/fn instead of template/eval

shamsimam:

- Uses the faster template/fn instead of template/eval

- avoids repeated calls to eval (which requires repeated compilation)

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -485,7 +485,7 @@
             (service/offer-instance! instance-rpc-chan service-id offer-params)
             (let [response-status (async/<! response-chan)]
               (utils/map->json-response (assoc (select-keys offer-params [:cid :request-id :router-id :service-id])
-                                               :response-status response-status))))))
+                                          :response-status response-status))))))
       (catch Exception ex
         (utils/exception->response ex request)))))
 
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -715,21 +724,29 @@
...
       (meters/mark! (metrics/waiter-meter "auto-run-as-requester" "form-error"))
       (utils/exception->response ex request))))
 
+(let [html-fn (template/fn [{:keys [cid host hostname message port support-info timestamp]}]
+                           (slurp (io/resource "web/welcome.html")))]
+  (defn- render-welcome-html "Renders welcome html" [context] (html-fn context)))
+
+(let [text-fn (template/fn [{:keys [cid host hostname message port support-info timestamp]}]
+                           (slurp (io/resource "web/welcome.txt")))]
+  (defn- render-welcome-text "Renders welcome text" [context] (text-fn context)))
```
dposada: These `defn`s where everything is on the same line look odd to me

shamsimam:
> Do we know if this helps overall performance?

I am going to test this today.

shamsimam:
Stress tested and this change looks good. I will rebase https://github.com/twosigma/waiter/pull/192 after this is merged to avoid generating exceptions while rendering the error messages.

PR #197, opened 2018-01-24
upgrades the lein voom version to work with latest version of lein (i…

shamsimam:

- Update the voom versions for the `waiter-patch` branches.


shamsimam:
@pschorf Build is now green

PR #198, opened 2018-01-29
Add log when scheduler reports fewer than min instances instances

pschorf:
- Adds a log message when the scheduler reports fewer than min-instances instances.

We believe that there's a bug where the marathon scheduler occasionally returns 0 instances even though it is attempting to scale to 1 instance. We'd like to verify this with a log message and then implement a fix for the bug.

### DIFF
--- a/waiter/src/waiter/scaling.clj
+++ b/waiter/src/waiter/scaling.clj
@@ -373,7 +373,7 @@
                             {:scheduler-state scheduler-state, :service-id service-id})
                   {:scale-to-instances instances :target-instances target-instances :scale-amount 0}))]
           (when (< instances (service-description "min-instances"))
-            (log/warn "Scheduler reported service had fewer instances than min-instances" {:service-id service-id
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scaling.clj
```diff
@@ -372,6 +372,10 @@
                   (log/info "no target instances available for service"
                             {:scheduler-state scheduler-state, :service-id service-id})
                   {:scale-to-instances instances :target-instances target-instances :scale-amount 0}))]
+          (when (< instances (service-description "min-instances"))
+            (log/warn "Scheduler reported service had fewer instances than min-instances" {:service-id service-id
```
shamsimam: Nitpick: I think we are trying to get all the log messages to start with lower-case

shamsimam: Nitpick: Can we move the start of the map to the next line and have them all in a single line,

pschorf: done

shamsimam:
Awaiting green build

PR #199, opened 2018-01-31
Catch exceptions thrown while making health check request

pschorf:
- Catch exceptions thrown while making health check requests

Otherwise, uncaught exceptions are killing the async thread handling the request.


DaoWen:
Are we waiting on something to merge this?

PR #201, opened 2018-02-01
avoids reliance on Marathon to provide the protocol for healthcheck requests

shamsimam:

- retrieve the instance protocol from the service description (guaranteed to have the backend-proto) instead of Marathon (as it can be missing sometimes)

- Ensure the healthchecks from Waiter can execute despite missing protocol reported from Marathon

### DIFF
--- a/waiter/test/waiter/scheduler/marathon_test.clj
+++ b/waiter/test/waiter/scheduler/marathon_test.clj
@@ -48,7 +48,7 @@
                                               :app {
                                                     :id "test-app-1234",
                                                     :instances 3,
-                                                    :healthChecks [{:path "/health", :portIndex 0, :protocol "HTTPS", :timeoutSeconds 10}],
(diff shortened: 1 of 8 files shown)


waiter/test/waiter/scheduler/marathon_test.clj
```diff
@@ -146,43 +147,43 @@
                           :marathon-response {
                                               :framework-id "F123445"
                                               :app {
-                                                    :id "test-app-1234",
```
pschorf: Why did all of these change?

shamsimam: Since I am passing a single `service-id->service-description` map but want different protocols in the tests, I ended up changing their service-ids so I could configure them.

shamsimam: I changed the tests to make `service-id->service-description` test case specific.

shamsimam:
@pschorf ready for another round of review.

pschorf:
Approved, just waiting on the build

PR #203, opened 2018-02-02
waiter: provides the resources in the service description as env variables to the instances

shamsimam:

- adds `WAITER_CPUS` and `WAITER_MEM_MB` environment variables

Given these environment variables, tasks can decide how much memory and threads they can allocate at runtime.


DaoWen:
Are these different from the environment variables provide by Marathon?
https://mesosphere.github.io/marathon/docs/task-environment-vars.html

shamsimam:
@DaoWen It allows Waiter to _own_ these variables and also allows users to avoid dependency on the underlying scheduler.

PR #205, opened 2018-02-05
waiter: support for numeric maximum value limits on service description parameters

shamsimam:

- Adds generic support for maximum value restrictions on service parameters.

Introducing limits allows us to fail requests early if the cluster cannot support the requested resources for an instance.

An example error response is shown below:
screen shot 2018-02-06 at 10 25 05 am

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -357,10 +357,10 @@
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
 
- ;; Service description parameters with numeric values, e.g. cpus, can be configured to have
- ;; maximum value limits. When not configured, these defaults will be used:
- :service-description-resource-limits {"cpus" 100
-                                       "mem" 1048576}
+ ;; Service description parameters with numeric values, e.g. cpus, can be configured
+ ;; to have maximum value limits.
+ :service-description-upper-limits {"cpus" 40
+                                    "mem" 32768} ;; mem is in megabytes
 
  ; ---------- Timeouts ----------
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -320,6 +320,11 @@
                                                 ;; waiter.service-description.ServiceDescriptionBuilder:
                                                 :factory-fn waiter.service-description/->DefaultServiceDescriptionBuilder}}
 
+ ;; Additional configurable constraints on description parameters.
+ ;; Only max is supported currently.
+ :service-description-constraints {"cpus" {:max 40}
+                                   "mem" {:max 32768}} ;; mem is in megabytes
+
  ;; The following service description parameters are required and
  ;; therefore don't have default values:
  ;;
@@ -356,11 +361,6 @@
                                 "scale-down-factor" 0.001
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
-
- ;; Service description parameters with numeric values, e.g. cpus, can be configured
- ;; to have maximum value limits.
- :service-description-upper-limits {"cpus" 40
-                                    "mem" 32768} ;; mem is in megabytes
 
  ; ---------- Timeouts ----------
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -357,6 +357,11 @@
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
 
+ ;; Service description parameters with numeric values, e.g. cpus, can be configured to have
+ ;; maximum value limits. When not configured, these defaults will be used:
+ :service-description-resource-limits {"cpus" 100
+                                       "mem" 1048576}
+
  ; ---------- Timeouts ----------
 
  ; Waiter maintains a list of recently "killed" and "erroneous" (e.g. an error occurred while streaming the response)
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -179,7 +179,7 @@
       (is (= 400 status))
       (is (str/includes? body "Command type fakecommand is not supported")))))
 
-(deftest ^:parallel ^:integration-fast test-basic-parameters-exceed-limits
+(deftest ^:parallel ^:integration-fast test-basic-parameters-violates-upper-limit-constraints
   (testing-using-waiter-url
     (let [constraints (setting waiter-url [:service-description-constraints])
           upper-limits (->> constraints
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -188,6 +188,7 @@
       (is (seq upper-limits))
       (doseq [[parameter upper-limit] (rest upper-limits)]
         (let [headers {:x-waiter-cmd "false"
+                       :x-waiter-cmd-type "shell"
                        :x-waiter-name (rand-name)
                        :x-waiter-version "1"
                        (keyword (str "x-waiter-" (name parameter))) (inc upper-limit)}
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -178,6 +178,23 @@
       (is (= 400 status))
       (is (str/includes? body "Command type fakecommand is not supported")))))
 
+(deftest ^:parallel ^:integration-fast test-basic-parameters-exceed-limits
+  (testing-using-waiter-url
+    (let [upper-limits (get (waiter-settings waiter-url) :service-description-upper-limits)]
+      (is (seq upper-limits))
+      (doseq [[parameter upper-limit] (rest upper-limits)]
+        (let [headers {:x-waiter-cmd "false"
+                       :x-waiter-name (rand-name)
+                       :x-waiter-version "1"
+                       (keyword (str "x-waiter-" (name parameter))) (inc upper-limit)}
+              {:keys [body status]} (make-light-request waiter-url headers)]
+          (is (= 400 status))
+          (is (not (str/includes? body "clojure")) body)
+          (is (every? #(str/includes? body %)
+                      ["The following fields exceed their allowed limits"
+                       (str (name parameter) " is " (inc upper-limit) " but the max allowed is " upper-limit)])
+              body))))))
+
 (deftest ^:parallel ^:integration-fast test-header-metadata
   (testing-using-waiter-url
     (let [headers {:x-waiter-name (rand-name)
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -17,6 +17,7 @@
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
             [clojure.walk :as walk]
+            [plumbing.core :as pc]
             [qbits.jet.client.http :as http]
             [waiter.client-tools :refer :all]
             [waiter.service-description :as sd]
@@ -180,7 +181,10 @@
 
 (deftest ^:parallel ^:integration-fast test-basic-parameters-exceed-limits
   (testing-using-waiter-url
-    (let [upper-limits (get (waiter-settings waiter-url) :service-description-upper-limits)]
+    (let [constraints (setting waiter-url [:service-description-constraints])
+          upper-limits (->> constraints
+                            (filter (fn [[_ constraint]] (contains? constraint :max)))
+                            (pc/map-vals :max))]
       (is (seq upper-limits))
       (doseq [[parameter upper-limit] (rest upper-limits)]
         (let [headers {:x-waiter-cmd "false"
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -17,7 +17,6 @@
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
             [clojure.walk :as walk]
-            [plumbing.core :as pc]
             [qbits.jet.client.http :as http]
             [waiter.client-tools :refer :all]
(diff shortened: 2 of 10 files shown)


waiter/src/waiter/settings.clj
```diff
@@ -288,11 +289,13 @@
                                   "scale-down-factor" 0.001
                                   "scale-factor" 1
                                   "scale-up-factor" 0.1}
+   :service-description-resource-limits {"cpus" 100
+                                         "mem" (* 1024 1024)}
```
dposada: These seem too high for the default maximums

waiter/config-full.edn
```diff
@@ -350,6 +350,11 @@
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
 
+ ;; Service description parameters with numeric values, e.g. cpus, can be configured to have
+ ;; maximum value limits. When not configured, these defaults will be used:
```
dposada: > When not configured, these defaults will be used

This sentence is confusing. It makes it sound like this section is for the defaults, but some other section is for configuring the maximums. I think we could just remove that sentence altogether, no?

waiter/config-full.edn
```diff
@@ -350,6 +350,11 @@
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
 
+ ;; Service description parameters with numeric values, e.g. cpus, can be configured to have
```
dposada: Does this apply to all parameters with numeric values?

waiter/config-full.edn
```diff
@@ -350,6 +350,11 @@
...
+ ;; maximum value limits. When not configured, these defaults will be used:
+ :service-description-resource-limits {"cpus" 100
```
dposada: Would `:service-description-maximums` be a better name?

waiter/config-full.edn
```diff
@@ -350,6 +350,11 @@
...
+                                       "mem" 1048576}
```
dposada: Should we clarify with a comment that this is in MB? Also, maybe a smaller amount for this example would be more appropriate

waiter/src/waiter/service_description.clj
```diff
@@ -276,38 +276,57 @@
...
+                                    " is "
+                                    (get service-description-to-use param)
+                                    " but allowed max is "
+                                    (-> issue (get param) .schema .pred-name (str/replace "limit-" ""))))
+              friendly-error-message (str "The following fields exceed their allowed limits: "
+                                          (str/join ", " (->> issue
+                                                              keys
+                                                              sort
+                                                              (map param->message))))]
+          (throw-error e issue friendly-error-message))))
```
dposada: Can you post an example of what this ends up looking like to the user?

waiter/src/waiter/settings.clj
```diff
@@ -115,6 +115,8 @@
                                                    (s/required-key "scale-factor") schema/positive-fraction-less-than-or-equal-to-1
                                                    (s/required-key "scale-up-factor") schema/positive-fraction-less-than-1
                                                    (s/required-key "scale-down-factor") schema/positive-fraction-less-than-1}
+   (s/required-key :service-description-resource-limits) {(s/required-key "cpus") schema/positive-int
+                                                          (s/required-key "mem") schema/positive-int}
```
dposada: Why are these int?

waiter/config-full.edn
shamsimam: Yes, I updated the schema in `settings.clj` to reflect this.

waiter/config-full.edn
shamsimam: Removed.

waiter/config-full.edn
shamsimam: Renamed to `service-description-upper-limits`. Okay with the name?

waiter/config-full.edn
shamsimam: Added.

waiter/src/waiter/settings.clj
shamsimam: Changed to `positive-num`

waiter/src/waiter/settings.clj
shamsimam: Changed to `32` and `128` Gb

waiter/src/waiter/service_description.clj
shamsimam: Added an example image in the description.

waiter/src/waiter/service_description.clj
```diff
@@ -276,38 +276,57 @@
...
                                              "run-as-user" "default-run-as-user"})
         service-description-to-use (merge default-valid-service-description service-description-template)
-        exception-message (utils/message :invalid-service-description)]
+        exception-message (utils/message :invalid-service-description)
+        throw-error (fn throw-error [e issue friendly-error-message]
+                      (sling/throw+ (cond-> {:type :service-description-error
+                                             :message exception-message
+                                             :service-description service-description-template
+                                             :status 400
+                                             :issue issue}
```
dposada: Do we really want end users to see the `:issue`? It seems unhelpful and potentially distracting

waiter/src/waiter/service_description.clj
```diff
@@ -276,38 +276,57 @@
...
+
+    (try
+      (s/validate upper-limits-schema service-description-to-use)
+      (catch Exception e
+        (let [issue (s/check upper-limits-schema service-description-to-use)
+              param->message (fn [param]
+                               (str param
+                                    " is "
+                                    (get service-description-to-use param)
+                                    " but allowed max is "
```
dposada: Maybe "**the** max allowed"?

waiter/src/waiter/service_description.clj
```diff
@@ -276,38 +276,57 @@
...
+                                    (-> issue (get param) .schema .pred-name (str/replace "limit-" ""))))
+              friendly-error-message (str "The following fields exceed their allowed limits: "
+                                          (str/join ", " (->> issue
```
dposada: What would you think of formatting with bullets, like:

```
The following fields exceed their allowed limits:
- cpus is 2 but the max allowed is 1
- mem is 4 but the max allowed is 3
```

waiter/src/waiter/service_description.clj
shamsimam: This is consistent with how we report existing service description errors. https://github.com/twosigma/waiter/pull/205/files/8c2d5e0f45754b98c26ac864e912dd28aa62412a#diff-c2eeea4ec2cf83e87d136f5e3f30c705L291

waiter/src/waiter/service_description.clj
shamsimam: Sounds good.

waiter/src/waiter/service_description.clj

shamsimam:
@dposada ready for another round of review.

waiter/config-full.edn
```diff
@@ -350,6 +350,11 @@
                                 "scale-factor" 1
                                 "scale-up-factor" 0.1}
 
+ ;; Service description parameters with numeric values, e.g. cpus, can be configured
+ ;; to have maximum value limits.
+ :service-description-upper-limits {"cpus" 40
```
sradack: It's nice to be general but numeric limits only apply to specific parameters.  We can't have a numeric limit on `health-check-url` for example.

I'd prefer `:service-constraints {:max-cpus 40 :max-mem 32768}` and handle these specifically.  Seems simple to me.  It feels awkward to use `schema` to enforce these configurable limits.

And if you're not a fan of that, then maybe: 

`:service-description-constraints {"cpu" {:max 40} "mem" {:max 32768}}` which leaves the door open to non-numeric limits, e.g. {"backend-proto" {:is "https"}}

waiter/src/waiter/service_description.clj
```diff
@@ -276,38 +276,56 @@
...
+          (throw-error e issue friendly-error-message))))
+
+    (try
+      (s/validate upper-limits-schema service-description-to-use)
+      (catch Exception e
+        (let [issue (s/check upper-limits-schema service-description-to-use)
+              issue->param->limit (fn [issue param]
+                                    (-> issue (get param) .schema .pred-name (str/replace "limit-" "")))
+              param->message (fn [param]
+                               (str "- " param " is " (get service-description-to-use param) " but the max allowed is "
```
sradack: The use of newlines assumes the user is expecting `text/plain`, which, when rendered in HTML, will look funny.

waiter/src/waiter/service_description.clj
dposada: OK. If we want to be consistent, that's fine, but I think we should reconsider having errors containing Clojure syntax exposed to the end user

waiter/config-full.edn
shamsimam: I ended up using `:service-description-constraints {"cpu" {:max 40} "mem" {:max 32768}}` and `schema`.

waiter/src/waiter/service_description.clj
shamsimam: Removed the newlines. 

@dposada do you agree with @sradack?

waiter/src/waiter/service_description.clj
shamsimam: We can perform the cleanup as part of another PR.

shamsimam:
@sradack any additional feedback on this?

shamsimam:
@sradack any additional feedback on this?

shamsimam:
@sradack ping

waiter/src/waiter/service_description.clj
```diff
@@ -340,7 +362,20 @@
...
 
   (validate [_ service-description args-map]
-    (validate-schema service-description (merge-with set/union args-map {:valid-cmd-types #{"shell"}}))))
+    (->> (merge-with set/union args-map {:valid-cmd-types #{"shell"}})
+         (validate-schema service-description upper-limits-schema))))
+
+(defn create-default-service-description-builder
+  "Returns a new DefaultServiceDescriptionBuilder which uses the specified resource limits."
+  [{:keys [constraints]}]
+  (let [upper-limits (->> constraints
```
sradack: Would it be better to pass the full constraints to the builder, instead of a projection of some of the constraints?  This logic could belong inside the builder.

waiter/integration/waiter/basic_test.clj
```diff
@@ -178,6 +179,27 @@
       (is (= 400 status))
       (is (str/includes? body "Command type fakecommand is not supported")))))
 
+(deftest ^:parallel ^:integration-fast test-basic-parameters-exceed-limits
```
sradack: Can we be consistent on naming limits vs. constraints?  Should this be called `test-basic-parameters-outside-constraints`

waiter/src/waiter/service_description.clj
shamsimam: We need to build it here as `DefaultServiceDescriptionBuilder` is a record. We can create the new constraints when we support them and pass them along.

waiter/integration/waiter/basic_test.clj
shamsimam: Renamed to `test-basic-parameters-violates-upper-limit-constraints`

sradack: `test-basic-parameters-violates-max-constraints`

waiter/src/waiter/service_description.clj
```diff
@@ -340,7 +362,20 @@
...
+                          (filter (fn [[_ constraint]] (contains? constraint :max)))
+                          (pc/map-vals :max))
+        upper-limits-schema (-> (->> upper-limits
+                                     (pc/map-keys s/optional-key)
+                                     (pc/map-vals (fn [v] (s/pred #(<= % v) (symbol (str "limit-" v))))))
+                                (assoc s/Str s/Any))]
+    (->DefaultServiceDescriptionBuilder upper-limits-schema)))
```
sradack: `max-constraints-schema`?

sradack: Let's be consistent.  `max` instead of `upper`, `constraint` instead of `limit` throughout

waiter/integration/waiter/token_request_test.clj
```diff
@@ -662,6 +662,22 @@
       (is (not (str/includes? body "clojure")) body)
       (is (str/includes? body "The following environment variable keys are reserved: HOME.") body))))
 
+(deftest ^:parallel ^:integration-fast test-token-parameters-exceed-limits
+  (testing-using-waiter-url
+    (let [constraints (setting waiter-url [:service-description-constraints])
+          upper-limits (->> constraints
```
sradack: I've seen it at least a couple of places.  Should we factor it out?

shamsimam:

waiter/src/waiter/service_description.clj

waiter/integration/waiter/token_request_test.clj
shamsimam: Factored out in `extract-max-constraints`

waiter/integration/waiter/token_request_test.clj
```diff
@@ -662,6 +662,22 @@
       (is (not (str/includes? body "clojure")) body)
       (is (str/includes? body "The following environment variable keys are reserved: HOME.") body))))
 
+(deftest ^:parallel ^:integration-fast test-token-parameters-exceed-limits
+  (testing-using-waiter-url
+    (let [constraints (setting waiter-url [:service-description-constraints])
+          max-constraints (->> constraints
```
sradack: `extract-max-constraints`?

shamsimam: Oops, fixed. Do you want to wait for the build?

PR #206, opened 2018-02-16
waiter: adds support for environment variables for on-the-fly requests

shamsimam:

- add support for configuring the environment in on-the-fly requests

Brings the on-the-fly support closer to tokens. Provides greater flexibility by increasing capability of using partial tokens with on-the-fly requests that can control the executed commands by changing the environment.

screen shot 2018-02-20 at 1 16 11 pm


shamsimam:
@sradack build is now green.

PR #210, opened 2018-02-23
Fix exception->response bug caused by shadowing ex-data

DaoWen:

Fix `exception->response` bug caused by shadowing the `ex-data` function in a local `let` binding.

We were using `ex-data` to reference two unique things (one a local data structure, and one a core clojure function) in the `exception->response` utility function, which obviously doesn't work. Choosing a different name for the local binding for the response data structure fixes the problem.

### DIFF
--- a/waiter/src/waiter/utils.clj
+++ b/waiter/src/waiter/utils.clj
@@ -275,13 +275,13 @@
   "Converts an exception into a ring response."
   [^Exception ex {:keys [] :as request}]
   (let [wrapped-ex (wrap-unhandled-exception ex)
-        {:keys [friendly-error-message message status] :as ex-data} (ex-data wrapped-ex)
+        {:keys [friendly-error-message message status] :as data} (ex-data wrapped-ex)
         message (or friendly-error-message message (.getMessage wrapped-ex))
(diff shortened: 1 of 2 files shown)


shamsimam:
And/Or maybe a unit test too.

waiter/src/waiter/utils.clj
```diff
@@ -275,13 +275,13 @@
   "Converts an exception into a ring response."
   [^Exception ex {:keys [] :as request}]
   (let [wrapped-ex (wrap-unhandled-exception ex)
-        {:keys [friendly-error-message message status] :as ex-data} (ex-data wrapped-ex)
+        {:keys [friendly-error-message message status] :as data} (ex-data wrapped-ex)
         message (or friendly-error-message message (.getMessage wrapped-ex))
         {:keys [headers suppress-logging]} (ex-data wrapped-ex)
```
shamsimam: We can just move this destructuring and merge it with the one two lines above.

DaoWen:
I updated the existing unit test to use a record rather than a map. I think this would be difficult to integration-test since the control path where I managed to trigger the bug was only being taken due to a bug in a new feature I was implementing. I confirmed that the unit test fails when those test updates are cherry-picked onto the current master branch.

PR #219, opened 2018-03-09
Creates wrap-descriptor

sradack:

* Moves `request->descriptor` out of `process` and into `wrap-descriptor`
* Introduces `wrap-context` as a general method "double wrapping" request/response pairs to pass context both upstream and downstream.

* Demonstrate a pattern for passing context upstream and downstream.
* First step toward factoring of `process` method.
* `descriptor` ultimately available for logging.
* Some unit tests related are simplified due to factoring of `process` method.
* Further simplifications can be made:
  * Examples
    * Tracking response codes by metric-group can now be done via middleware since `descriptor` is available
    * Moving `handle-too-many-requests`/`handle-suspended-service` outside of process

### DIFF
--- a/waiter/src/waiter/middleware.clj
+++ b/waiter/src/waiter/middleware.clj
@@ -1,0 +1,31 @@
+;;
+;;       Copyright (c) 2018 Two Sigma Investments, LP.
+;;       All Rights Reserved
+;;
+;;       THIS IS UNPUBLISHED PROPRIETARY SOURCE CODE OF
+;;       Two Sigma Investments, LP.
+;;
+;;       The copyright notice above does not evidence any
+;;       actual or intended publication of such source code.
+;;
+(ns waiter.middleware
+  (:require [clojure.core.async :as async]
+            [full.async :refer (<?? <? go-try)]
+            [waiter.async-utils :as au]
+            [waiter.utils :as utils]))
+
+(defn wrap-context
+  "Wraps a handler, ensuring the request and response (or exception) contains data in the context map."
+  [handler context]
+  (fn [request]
+    (try
+      (let [response (handler (merge request context))]
+        (if (au/chan? response)
+          (async/go
+            (try
+              (merge (<? response) context)
+              (catch Exception e
+                (utils/merge-exception e context))))
+          (merge response context)))
+      (catch Exception e
+        (throw (utils/merge-exception e context))))))
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -484,16 +484,15 @@
 
 (defn wrap-descriptor
   "Adds the descriptor to the request/response.
-   Redirects users in the case of missing user/run-as-requestor."
+  Redirects users in the case of missing user/run-as-requestor."
   [handler request->descriptor-fn]
   (fn [request]
-    (try
-      (let [descriptor (request->descriptor-fn request)
-            handler (-> handler
-                        (middleware/wrap-context {:descriptor descriptor}))]
-        (handler request))
-      (catch Exception e
-        (if (missing-run-as-user? e)
+    (let [[descriptor error] (utils/tryv (request->descriptor-fn request))]
+      (if-not error
+        (let [handler (-> handler
+                          (middleware/wrap-context {:descriptor descriptor}))]
+          (handler request))
+        (if (missing-run-as-user? error)
           (let [{:keys [query-string uri]} request
                 location (str "/waiter-consent" uri (when (not (str/blank? query-string)) (str "?" query-string)))]
             (counters/inc! (metrics/waiter-counter "auto-run-as-requester" "redirect"))
@@ -502,7 +501,7 @@
              :status 303})
           (do
             (meters/mark! (metrics/waiter-meter "core" "process-errors"))
-            (utils/exception->response e request)))))))
+            (utils/exception->response error request)))))))
 
 (let [process-timer (metrics/waiter-timer "core" "process")]
   (defn process
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -23,6 +23,7 @@
             [qbits.jet.client.http :as http]
             [qbits.jet.servlet :as servlet]
             [slingshot.slingshot :refer [try+]]
+            [try-let :refer [try-let]]
             [waiter.async-request :as async-req]
             [waiter.async-utils :as au]
             [waiter.auth.authentication :as auth]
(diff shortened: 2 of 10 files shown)


waiter/src/waiter/websocket.clj
```diff
@@ -358,10 +358,12 @@
 
 (defn process-exception-in-request
   "Processes exceptions thrown while processing a websocket request."
-  [track-process-error-metrics-fn {:keys [out] :as request} descriptor exception]
+  [track-process-error-metrics-fn {:keys [descriptor out] :as request} exception]
   (log/error exception "error in processing websocket request")
+  (log/info "descriptor is " descriptor)
```
shamsimam: Nitpick: Maybe remove this?

waiter/src/waiter/middleware.clj
```diff
@@ -0,0 +1,31 @@
...
+;;       actual or intended publication of such source code.
+;;
+(ns waiter.middleware
+  (:require [clojure.core.async :as async]
+            [full.async :refer (<?? <? go-try)]
+            [waiter.async-utils :as au]
+            [waiter.utils :as utils]))
+
+(defn wrap-context
+  "Wraps a handler, ensuring the request and response (or exception) contains data in the context map."
```
shamsimam: Why does `response` need to have this context, e.g. `descriptor`? Should something upstream rely on the presence of the descriptor?

waiter/src/waiter/middleware.clj
```diff
@@ -0,0 +1,31 @@
...
+  [handler context]
+  (fn [request]
+    (try
+      (let [response (handler (merge request context))]
+        (if (au/chan? response)
+          (async/go
+            (try
+              (merge (<? response) context)
```
shamsimam: Why are we using `merge` instead of `assoc` with a specific key?

waiter/src/waiter/process_request.clj
```diff
@@ -465,39 +466,51 @@
...
+  "Adds the descriptor to the request/response.
+   Redirects users in the case of missing user/run-as-requestor."
+  [handler request->descriptor-fn]
+  (fn [request]
+    (try
+      (let [descriptor (request->descriptor-fn request)
+            handler (-> handler
+                        (middleware/wrap-context {:descriptor descriptor}))]
+        (handler request))
+      (catch Exception e
```
shamsimam: Should the `try-catch` be scoped to the call to `request->descriptor-fn`?

waiter/src/waiter/middleware.clj
sradack: Not yet.

waiter/src/waiter/middleware.clj
sradack: Merge handles the case where I have a map or just a key/value pair.  That was my thought.  Is it worth an overload to replace
`(wrap-context {:descriptor descriptor})`
with
`(wrap-context :descriptor descriptor)`
I don't feel strongly that we need an overload.  

Additionally, I suspect this `merge` will eventually become a deep merge, in which case the signature doesn't change, but the `assoc`-style version is less useful.  We'd have to make an `assoc-in`-style version as well.

sradack:
@shamsimam Let me know what you think of the changes to `wrap-descriptor` and `tryv`.  I considered failjure, narrowing what is caught by the `catch` clause, etc.  I thought about doing this multiple ways, but ultimately decided this was the most straightforward.

waiter/src/waiter/utils.clj
```diff
@@ -544,3 +544,11 @@
   (if (instance? ExceptionInfo e)
     (ex-info (.getMessage e) (merge (ex-data e) m) (.getCause e))
     (ex-info (.getMessage e) m (.getCause e))))
+
+(defmacro tryv
+  "Like try, but [result exception]"
```
shamsimam: Why not return a map with `:result` and `:exception`? We don't seem to be using both slots of the vector at the same time.

waiter/src/waiter/process_request.clj
```diff
@@ -502,7 +501,7 @@
              :status 303})
           (do
             (meters/mark! (metrics/waiter-meter "core" "process-errors"))
-            (utils/exception->response e request)))))))
+            (utils/exception->response error request)))))))
```
shamsimam: The two versions are no longer equivalent as we will not mark `process-errors` if the `handler` call throws an exception.

waiter/src/waiter/process_request.clj
```diff
@@ -484,16 +484,15 @@
...
   [handler request->descriptor-fn]
   (fn [request]
-    (try
-      (let [descriptor (request->descriptor-fn request)
-            handler (-> handler
-                        (middleware/wrap-context {:descriptor descriptor}))]
-        (handler request))
-      (catch Exception e
-        (if (missing-run-as-user? e)
+    (let [[descriptor error] (utils/tryv (request->descriptor-fn request))]
```
shamsimam: I worry that `tryv` will lead to a lot of nested `(let [[... error] (tryv ...)] (if-not error ...))` blocks. Should we instead target a macro for that pattern where the macro would expect two arguments and behave similarly to an `if-let`?

sradack: `try-let` doesn't support destructuring.  Do you still want to proceed with `try-let`?

shamsimam: That is a `todo` item in `try-let`. We can have our own `try-let` macro if you like. My concern is with avoiding the extra if and I like `try-let` feature of avoiding the need for the vector.

sradack: https://github.com/rufoa/try-let/pull/1 Let's see.

DaoWen:
We talked about using an alternate `try-let` syntax yesterday. Here's what I came up with:

```clojure
(defrecord PassThru [caught-exception])

(defmacro try-let
  "Convenience macro for wrapping let binding expressions in a try/catch/finally block,
   while not catching exceptions thrown within the body of the let block.
   This construct adds optional :catch and :finally forms to the end of the let-binding vector.
   Each clause consists of a keyword-vector pair, similar to the :let clause in Clojure's `for` macro.
   The :catch vector must contain one or more forms,
   each of which should look like a `(catch ...)` clause, but without the `catch` keyword.
   The :finally vector's contents are simply wrapped in a `(finally ...)` block.
   At most one of each :catch and :finally may be provided,
   and if both are given, :catch must immedialy precede :finally.

   Example:

   (try-let [x (throw (RuntimeException. \"Hi\"))
             :catch [(IllegalArgumentException e (comment Cannot happen))
                     (Exception e (println \"Caught exception:\" (.getMessage e)))]
             :finally [(println \"doing cleanup\")
                       (comment Free some resources)]] x)"
  [bindings-vec & body]
  (assert (even? (count bindings-vec))
          "Odd number of forms in try-let binding vector. Binding-expression pairs are required.")
  (let [[binding-pairs catch+finally] (->> bindings-vec
                                           (partition 2)
                                           (split-with (comp not keyword? first)))
        bindings' (apply concat binding-pairs)
        clauses (->> catch+finally (apply concat) (apply array-map))
        {catches-vec :catch finally-body :finally} clauses
        clause-count (count catch+finally)]
    (assert (and (<= 1 clause-count 2)
                 (== (count clauses) clause-count)
                 (or (nil? catches-vec) (= :catch (ffirst catch+finally))))
            "Bindings vector must end with a :catch clause, or a :finally clause, or a :catch clause followed by a :finally clause.")
    `(let [try-result# (try
                         (let [~@bindings']
                           (try
                             ~@body
                             (catch java.lang.Throwable t#
                               (PassThru. t#))))
                         ~@(map (partial list* `catch) catches-vec)
                         ~@(when finally-body
                             `((finally ~@finally-body))))]
       (if (instance? PassThru try-result#)
         (throw (:caught-exception try-result#))
         try-result#))))
```

Sample usage (copied from the docstring):

```clojure
   (try-let [x (throw (RuntimeException. "Hi"))
             :catch [(IllegalArgumentException e (comment Cannot happen))
                     (Exception e (println "Caught exception:" (.getMessage e)))]
             :finally [(println "doing cleanup")
                       (comment Free some resources)]] x)
```

## Advantages of this version of the macro:
* There is very little additional overhead for this solution in the case that no exception is thrown in the `try-let` body. (The macro in the currently-used library has to create an auxiliary vector, and then it repeats the destructuring and binding logic in both the inner and outer scopes.)
* The `:catch`/`:finally` definitions go in the bindings list. That syntax feels idiomatic since it resembles `:when`/`:let` in the `for` macro, and it clearly denotes the intended scope of the `:catch` clauses.
* The macro definition is pretty straightforward, and trivially supports destructuring (since it just plops the binding forms unchanged into another `let` block).

## Disadvantages:
* Catches and re-throws exceptions coming from the `try-let` body. However, the overhead of *creating* the exception (especially constructing the stack trace) should completely dominate the catch/throw.
* Adds the `PassThru` record definition.
* _Some_ people might think having the `:catch` and `:finally` pieces in the binding list is "ugly"...

PR #229, opened 2018-03-14
ensures latest owner-key is looked up during token refresh

shamsimam:

- ensures latest owner-key is looked up during token refresh
- adds logging and sorting to token syncer to help detect control flow logic for token updates

A race in updating the token owner-key leads to repeated token updates.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -67,6 +67,7 @@
                                            {:owner->owner-key owner->owner-key})))
                          (or (get owner->owner-key owner)
                              (let [new-owner-key (new-owner-key)]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/token.clj
```diff
@@ -146,7 +149,7 @@
     (let [refreshed-token (kv/fetch kv-store token :refresh true)]
       (when owner
         ; NOTE: The token may still show up temporarily in the old owners list
-        (let [owner->owner-key (kv/fetch kv-store token-owners-key)
+        (let [owner->owner-key (kv/fetch kv-store token-owners-key :refresh true)
```
shamsimam: This is the key change in the PR

PR #231, opened 2018-03-15
Creates wrap-debug-info

sradack:

- Centralizes some debugging info in middleware

- x-waiter-debug applies to all requests, not just process
- Removes some responsibilities from process

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -43,6 +43,7 @@
             [waiter.metrics-sync :as metrics-sync]
             [waiter.password-store :as password-store]
             [waiter.process-request :as pr]
+            [waiter.ring-utils :as ru]
             [waiter.scaling :as scaling]
             [waiter.scheduler :as scheduler]
             [waiter.service :as service]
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/core.clj
```diff
@@ -163,14 +163,36 @@
...
+    [{:keys [request-id request-time router-id] :as request}]
+    (if (utils/request->debug-enabled? request)
+      (let [response (handler request)
+            add-headers (fn [response]
+                          (update response :headers
+                                  (fn [headers]
+                                    (-> headers
+                                        (assoc "X-Waiter-Request-Date" (utils/date-to-str request-time utils/formatter-rfc822))
+                                        (assoc "X-Waiter-Request-Id" request-id)
+                                        (assoc "X-Waiter-Router-Id" router-id)))))]
```
shamsimam: Nitpick: didn't we decide to use all lower-case in headers?

waiter/src/waiter/core.clj
```diff
@@ -163,14 +163,36 @@
...
+        (if (au/chan? response)
```
shamsimam: Should we create a helper function for this pattern in `middleware.clj` where we apply a function to the response?

PR #232, opened 2018-03-15
Creates wrap-response-status-metrics

sradack:

Create wrap-response-status-metrics middleware to handle updating metrics 

- Centralized middleware ensures we handle all cases and ensures we record the actual status we're returning.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -994,9 +994,9 @@
                                                                       determine-priority-fn process-response-fn pr/process-exception-in-http-request
                                                                       pr/abort-http-request-callback-factory local-usage-agent request))]
                            (-> inner-process-request-fn
-                               pr/wrap-response-status-metrics
                                pr/wrap-too-many-requests
                                pr/wrap-suspended-service
+                               pr/wrap-response-status-metrics
                                (pr/wrap-descriptor request->descriptor-fn)
                                wrap-secure-request-fn
                                wrap-auth-bypass-fn)))
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -994,6 +994,7 @@
                                                                       determine-priority-fn process-response-fn pr/process-exception-in-http-request
                                                                       pr/abort-http-request-callback-factory local-usage-agent request))]
                            (-> inner-process-request-fn
+                               pr/wrap-response-status-metrics
                                pr/wrap-too-many-requests
                                pr/wrap-suspended-service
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -405,10 +405,21 @@
...
+  [handler]
+  (fn wrap-response-status-metrics-fn
+    [{:keys [descriptor] :as request}]
+    (let [{:keys [service-id] {:strs [metric-group]} :service-description} descriptor
+          response (handler request)
+          update (fn [{:keys [status] :as response}]
+                   (counters/inc! (metrics/service-counter service-id "response-status" (str status)))
+                   (statsd/inc! metric-group (str "response_status_" status))
+                   response)]
+      (if (au/chan? response)
```
shamsimam: We should have a helper function for the if-else pattern.

sradack: I'm not sure this rises to the level of needed a helper.

sradack:
I wanted to keep the PR small and targeted.  I'm find to move the metrics in a separate PR.

sradack:
Rebased

waiter/src/waiter/process_request.clj
sradack: I'll create it in #231 and then rebase.

waiter/src/waiter/process_request.clj
```diff
@@ -405,10 +405,21 @@
           (async/close! body)
           (async/close! request-state-chan))))))
 
-(defn- track-response-status-metrics
-  [service-id {:strs [metric-group]} status]
-  (counters/inc! (metrics/service-counter service-id "response-status" (str status)))
-  (statsd/inc! metric-group (str "response_status_" status)))
+(defn wrap-response-status-metrics
```
shamsimam: Where is `wrap-response-status-metrics` wired in?

waiter/src/waiter/process_request.clj
shamsimam: Okay

waiter/src/waiter/core.clj
```diff
@@ -994,6 +994,7 @@
                                                                       determine-priority-fn process-response-fn pr/process-exception-in-http-request
                                                                       pr/abort-http-request-callback-factory local-usage-agent request))]
                            (-> inner-process-request-fn
+                               pr/wrap-response-status-metrics
```
shamsimam: Should this be after `pr/wrap-suspended-service`?

PR #240, opened 2018-03-17
Improves JSON error handling

sradack:

- Allow users to post JSON without setting `Content-Type` header
- Provides a JSON-specific error message when invalid JSON is `POST`ed
- Introduces Ring-style pattern for parsing JSON

- Alleviate support burden
- Code clean up

### DIFF
--- a/waiter/src/waiter/ring_utils.clj
+++ b/waiter/src/waiter/ring_utils.clj
@@ -10,6 +10,9 @@
 ;;
 (ns waiter.ring-utils
   (:require [clojure.core.async :as async]
+            [clojure.data.json :as json]
+            [ring.middleware.params :as ring-params]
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/ring_utils.clj
```diff
@@ -18,3 +21,17 @@
   (if (au/chan? response)
     (async/go (response-fn (async/<! response)))
     (response-fn response)))
+
+(defn json-request
+  "Tries to parse a request body as JSON, if error, throw 400."
+  [{:keys [body] {:strs [content-type]} :headers :as request}]
+  (try
+    (assoc request :body (-> body slurp (json/read-str)))
```
shamsimam: Nitpick: `(json/read-str)` can be `json/read-str`

PR #241, opened 2018-03-18
Refactors authentication

sradack:

- Authentication passes `:authorization/principal` and `:authorization/user` through response as well
- `merge-exception` -> `update-exception` (functional style, instead of assuming merge)
- `wrap-context` -> `wrap-merge` (simple rename), also creates `wrap-update` and `wrap-assoc` 
- `:authenticated-principal` -> `:authorization/principal`

- Make authentication information available downstream for logging and error pages.

### DIFF
--- a/waiter/src/waiter/auth/authentication.clj
+++ b/waiter/src/waiter/auth/authentication.clj
@@ -38,8 +38,8 @@
 (defn assoc-auth-params
   "Associate values for authenticated user in the request."
   ([request principal]
-   (assoc-auth-params request (first (str/split principal #"@" 2)) principal))
-  ([request user principal]
+   (assoc-auth-params request principal (first (str/split principal #"@" 2))))
+  ([request principal user]
    (assoc request
           :authorization/principal principal
           :authorization/user user)))
@@ -48,7 +48,7 @@
   "Invokes the given request-handler on the given request, adding the necessary
   auth headers on the way in, and the x-waiter-auth cookie on the way out."
   [handler request user principal password]
-  (let [assoc-auth-params-fn #(assoc-auth-params % user principal)
+  (let [assoc-auth-params-fn #(assoc-auth-params % principal user)
         handler' (middleware/wrap-update handler assoc-auth-params-fn)]
     (-> request
         handler'
--- a/waiter/src/waiter/auth/authentication.clj
+++ b/waiter/src/waiter/auth/authentication.clj
@@ -12,7 +12,8 @@
   (:require [clj-time.core :as t]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
-            [waiter.cookie-support :as cookie-support]))
+            [waiter.cookie-support :as cookie-support]
+            [waiter.middleware :as middleware]))
 
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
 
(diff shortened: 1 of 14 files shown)


waiter/src/waiter/middleware.clj
```diff
@@ -14,18 +14,18 @@
             [waiter.async-utils :as au]
             [waiter.utils :as utils]))
 
-(defn wrap-context
-  "Wraps a handler, ensuring the request and response (or exception) contains data in the context map."
-  [handler context]
+(defn wrap-update
+  "Wraps a handler, calling update on the request and response."
```
shamsimam: Nitpick: `update-fn` is called on the request, response and the `ex-data`.

waiter/src/waiter/middleware.clj
```diff
@@ -14,18 +14,18 @@
             [waiter.async-utils :as au]
             [waiter.utils :as utils]))
 
-(defn wrap-context
-  "Wraps a handler, ensuring the request and response (or exception) contains data in the context map."
-  [handler context]
+(defn wrap-update
```
shamsimam: Should we have `assoc` and `merge` versions which ends up using `update` under the covers? That makes the user code simpler to read instead of having to create the lambdas everywhere.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -28,21 +29,30 @@
...
 
 (defn- add-cached-auth
   [response password principal]
   (cookie-support/add-encoded-cookie response password AUTH-COOKIE-NAME [principal (System/currentTimeMillis)] 1))
 
+(defn assoc-auth-params
+  "Associate values for authenticated user in the request."
+  ([request principal]
+   (assoc-auth-params request (first (str/split principal #"@" 2)) principal))
+  ([request user principal]
```
shamsimam: Nitpick: we should make the new argument (`user`) the last in the overloaded version in this scenario.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -28,21 +29,30 @@
...
   auth headers on the way in, and the x-waiter-auth cookie on the way out."
-  [request-handler request user principal password]
-  (-> request
-      (assoc :authorization/user user
-             :authenticated-principal principal)
-      request-handler
-      (add-cached-auth password principal)))
+  [handler request user principal password]
+  (let [assoc-auth-params-fn #(assoc-auth-params % user principal)
+        handler' (middleware/wrap-update handler assoc-auth-params-fn)]
```
shamsimam: Why would `wrap-context` not work for this scenario? i.e. why did we need the `wrap-update` refactor?

waiter/src/waiter/cookie_support.clj
```diff
@@ -68,9 +69,7 @@
                                (string? existing-header) [existing-header set-cookie-header]
                                :else (conj existing-header set-cookie-header))]
               (assoc-in response [:headers "set-cookie"] new-header)))]
-    (if (map? response)
-      (add-cookie-into-response response)
-      (async/go (add-cookie-into-response (async/<! response))))))
+    (ru/update-response response add-cookie-into-response)))
```
shamsimam: 👍

waiter/src/waiter/auth/authentication.clj
sradack: `assoc-auth-params` doesn't build a separate map.  It `assoc`es onto an existing map.  We'd either need to create a `auth-params-map` or have a `wrap-update`.  I opted for the latter since `wrap-update` is more flexible than `wrap-merge`.  `wrap-merge` can be trivially implemented by `wrap-update`, but not vice versa.

waiter/src/waiter/middleware.clj
sradack: I thought about going down that path, but merge isn't used and assoc used only once.  I'll create `wrap-assoc`.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -28,21 +29,30 @@
...
-  [request-handler request user principal password]
-  (-> request
-      (assoc :authorization/user user
-             :authenticated-principal principal)
-      request-handler
-      (add-cached-auth password principal)))
+  [handler request user principal password]
+  (let [assoc-auth-params-fn #(assoc-auth-params % principal user)
+        handler' (middleware/wrap-update handler assoc-auth-params-fn)]
+    (-> request
```
shamsimam: I find `wrap-context` much more simpler to read and use than `wrap-update`.
This could have been:
```
auth-params-map (auth-params-map principal user)
handler' (middleware/wrap-context handler auth-params-map)
```

where:
```
(defn auth-params-map
  "Returns a map for authenticated user."
  ([principal]
   (auth-params-map principal (first (str/split principal #"@" 2))))
  ([principal user]
   {:authorization/principal principal
    :authorization/user user}))
```

waiter/src/waiter/handler.clj
```diff
@@ -39,7 +39,7 @@
   "Creates a map containing the username and principal from a request"
   [request]
   {:username (:authorization/user request)
-   :principal (:authenticated-principal request)})
+   :principal (:authorization/principal request)})
```
shamsimam: Nitpick: lexicographic order

waiter/src/waiter/process_request.clj
```diff
@@ -469,7 +469,7 @@
   (fn [request]
     (try-let [descriptor (request->descriptor-fn request)]
       (let [handler (-> handler
-                        (middleware/wrap-context {:descriptor descriptor}))]
+                        (middleware/wrap-assoc :descriptor descriptor))]
```
shamsimam: 👍

waiter/src/waiter/utils.clj
```diff
@@ -538,9 +538,14 @@
...
+(defn update-exception
+  "Updates an exception, regardless of whether it's an ExceptionInfo or just Exception."
+  [^Exception e update-fn]
   (if (instance? ExceptionInfo e)
-    (ex-info (.getMessage e) (merge (ex-data e) m) (.getCause e))
-    (ex-info (.getMessage e) m (.getCause e))))
+    (ex-info (.getMessage e) (update-fn (ex-data e)) (.getCause e))
+    (ex-info (.getMessage e) (update-fn {}) (.getCause e))))
+
+(defn assoc-exception
```
shamsimam: Is this used?

PR #242, opened 2018-03-18
Add details to error pages

sradack:
Based on #241 

- Adds `principal`, `service-id`, and `instance-id` to error pages.

- Help debugging when errors occur


waiter/src/waiter/utils.clj
```diff
@@ -206,30 +206,37 @@
...
     {:cid x-cid
      :details details
      :host host
+     :instance-id (:id instance)
      :message message
+     :principal principal
      :query-string query-string
      :request-method (-> (or request-method "") name str/upper-case)
      :status status
+     :service-id service-id
```
shamsimam: Nitpick: lexicographic order

PR #244, opened 2018-03-18
Disables Server header

sradack:

- Avoids adding `Server` header to all responses

- Avoid polluting proxied responses with the `Server` header
- Reduce bandwidth

### DIFF
--- a/waiter/src/waiter/main.clj
+++ b/waiter/src/waiter/main.clj
@@ -86,7 +86,8 @@
                                         :join? false
                                         :max-threads 250
                                         :port port
-                                        :request-header-size 32768})]
+                                        :request-header-size 32768
+                                        :send-server-version? false})]
                     (server/run-jetty options)))})
 
 (defn start-waiter [config-file]


waiter/src/waiter/main.clj
```diff
@@ -86,7 +86,8 @@
                                         :join? false
                                         :max-threads 250
                                         :port port
-                                        :request-header-size 32768})]
+                                        :request-header-size 32768
+                                        :send-server-version? false})]
```
shamsimam: Should we be adding the `via` header instead?

shamsimam: Should it be a separate PR?

sradack: I don't follow.  All this PR does is prevent Waiter for emitting it's own server version.

shamsimam: Probably a separate PR then. Since we are now losing information on Waiter headers, I was looking for the [`via` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via) to contain some information on the Waiter router that processed the request.

sradack: Ah.  The Server header was something about Jetty.  It doesn't mention Waiter specifically.  Agreed we could use something to indicate its from Waiter.  I closed the `Via` header PR, but should revive it?

shamsimam: Agreed, please revive. Created https://github.com/twosigma/waiter/issues/246

PR #245, opened 2018-03-19
Response headers atom

sradack:
It may be easier to follow this PR commit by commit.

- Removes `response-headers` atom

- Complect the code

### DIFF
--- a/waiter/src/waiter/async_request.clj
+++ b/waiter/src/waiter/async_request.clj
@@ -139,8 +139,9 @@
    The function assumes location begins with a slash.
    This method wires up the completion and status check callbacks for the monitoring system.
    It also modifies the status check endpoint in the response header."
-  [router-id async-request-store-atom make-http-request-fn instance-rpc-chan service-id metric-group {:keys [host port] :as instance}
-   {:keys [request-id] :as reason-map} request-properties location response-headers]
+  [router-id async-request-store-atom make-http-request-fn instance-rpc-chan response
+   service-id metric-group {:keys [host port] :as instance}
+   {:keys [request-id] :as reason-map} request-properties location]
   (let [correlation-id (cid/get-correlation-id)
         status-endpoint (scheduler/end-point-url instance location)
         _ (log/info "status endpoint for async request is" status-endpoint)
@@ -172,4 +173,5 @@
     (let [param-map {:host host, :location location, :port port, :request-id request-id, :router-id router-id, :service-id service-id}
           status-location (route-params->uri "/waiter-async/status/" param-map)]
       (log/info "updating status location to" status-location "from" location)
-      (swap! response-headers assoc "location" status-location))))
+      (-> response
+          (assoc-in [:headers "location"] status-location)))))
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -449,6 +449,16 @@
      (~handle-elapsed-nanos-fn elapsed#)
      out#))
 
+(defmacro with-timer
+  "Times the operation specified by body using the provided
+  timer and calls handle-elapsed-nanos-fn with the epased
+  time in nanoseconds. Returns [value of body, nanos]."
+  [timer & body]
+  `(let [^Timer$Context start# (metrics.timers/start ~timer)
+         out# (do ~@body)
+         elapsed# (metrics.timers/stop start#)]
+     {:out out# :elapsed elapsed#}))
+
 (defn stream-metric-map
   "Returns a map containing metrics used for reporting streaming metrics for a given service."
   [service-id]
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -452,7 +452,7 @@
 (defmacro with-timer
   "Times the operation specified by body using the provided
   timer and calls handle-elapsed-nanos-fn with the epased
-  time in nanoseconds. Returns [value of body, nanos]."
+  time in nanoseconds. Returns a map with :out and :elapsed keys."
   [timer & body]
   `(let [^Timer$Context start# (metrics.timers/start ~timer)
          out# (do ~@body)
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -284,17 +284,21 @@
 
 (defn inspect-for-202-async-request-response
   "Helper function that inspects the response and triggers async-request post processing."
-  [post-process-async-request-response-fn instance-request-properties service-id metric-group instance endpoint request
-   reason-map {:keys [headers status]} response-headers reservation-status-promise]
+  [{:keys [headers status] :as response} post-process-async-request-response-fn instance-request-properties
+   service-id metric-group instance endpoint request reason-map reservation-status-promise]
   (let [location-header (str (get headers "location"))
         location (async-req/normalize-location-header endpoint location-header)]
-    (when (= status 202)
+    (if (= status 202)
       (if (str/starts-with? location "/")
         (do
           (deliver reservation-status-promise :success-async) ;; backend is processing as an asynchronous request
-          (post-process-async-request-response-fn service-id metric-group instance (handler/make-auth-user-map request)
-                                                  reason-map instance-request-properties location response-headers))
-        (log/info "response status 202, not treating as an async request as location is" location)))))
+          (post-process-async-request-response-fn response service-id metric-group instance
+                                                  (handler/make-auth-user-map request) reason-map
+                                                  instance-request-properties location))
+        (do
+          (log/info "response status 202, not treating as an async request as location is" location)
+          response))
+      response)))
 
 (defn stream-http-response
   "Writes byte data to the resp-chan. If the body is a string, just writes the string.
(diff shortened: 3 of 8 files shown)


waiter/src/waiter/async_request.clj
```diff
@@ -172,4 +173,5 @@
     (let [param-map {:host host, :location location, :port port, :request-id request-id, :router-id router-id, :service-id service-id}
           status-location (route-params->uri "/waiter-async/status/" param-map)]
       (log/info "updating status location to" status-location "from" location)
-      (swap! response-headers assoc "location" status-location))))
\ No newline at end of file
+      (-> response
+          (assoc-in [:headers "location"] status-location)))))
```
shamsimam: Do we need to worry about the case where `location` might already be in headers but in a different case?

waiter/src/waiter/async_request.clj
```diff
@@ -139,8 +139,9 @@
    The function assumes location begins with a slash.
    This method wires up the completion and status check callbacks for the monitoring system.
    It also modifies the status check endpoint in the response header."
-  [router-id async-request-store-atom make-http-request-fn instance-rpc-chan service-id metric-group {:keys [host port] :as instance}
-   {:keys [request-id] :as reason-map} request-properties location response-headers]
+  [router-id async-request-store-atom make-http-request-fn instance-rpc-chan response
+   service-id metric-group {:keys [host port] :as instance}
+   {:keys [request-id] :as reason-map} request-properties location]
```
shamsimam: nitpick: `response` could be the last argument in the function as this is like a response handler

waiter/src/waiter/process_request.clj
```diff
@@ -544,40 +526,49 @@
                         ; request-state-chan should be explicitly closed after the request finishes processing
                         request-state-chan (async/tap control-mult (au/latest-chan) false)
                         queue-timeout-ms (:queue-timeout-ms instance-request-properties)
-                        instance (fa/<? (prepare-instance instance-rpc-chan service-id reason-map start-new-service-fn request-state-chan
-                                                          queue-timeout-ms reservation-status-promise metric-group
-                                                          add-debug-header-into-response!))]
+                        instance-timer (metrics/service-timer service-id "get-available-instance")
+                        timed-instance (metrics/with-timer
+                                         instance-timer
```
shamsimam: We can inline `instance-timer` and avoid introducing a new variable

waiter/src/waiter/process_request.clj
```diff
@@ -544,40 +526,49 @@
...
-                                  (metrics/service-timer service-id "backend-response")
-                                  (fn [nanos]
-                                    (add-debug-header-into-response! "X-Waiter-Backend-Response-ns" nanos)
-                                    (statsd/histo! metric-group "backend_response" nanos))
-                                  (async/<!
-                                    (make-request-fn instance request instance-request-properties
-                                                     passthrough-headers endpoint metric-group)))
+                                backend-timer (metrics/service-timer service-id "backend-response")
+                                timed-response (metrics/with-timer
+                                                 backend-timer
```
shamsimam: We can inline `backend-timer` and avoid introducing a new variable

waiter/src/waiter/process_request.clj
```diff
@@ -544,40 +526,49 @@
...
                                 confirm-live-connection-with-abort (confirm-live-connection-factory request-abort-callback)]
-                            (when error
-                              (throw-response-error error reservation-status-promise instance response-headers))
-                            (process-backend-response-fn local-usage-agent instance-request-properties descriptor instance request
-                                                         reason-map response-headers reservation-status-promise
-                                                         confirm-live-connection-with-abort request-state-chan response))
+                            (statsd/histo! metric-group "backend_response" response-elapsed)
+                            (-> (if error
+                                  (do
+                                    (async/close! request-state-chan)
```
shamsimam: Observation: unfortunate duplication of `(async/close! request-state-chan)`

waiter/src/waiter/process_request.clj
```diff
@@ -177,38 +177,20 @@
     (cond-> uri
             (not (str/blank? query-string)) (str "?" query-string))))
 
-(defn- wrap-exception
-  "Includes metadata such as cid and status along with the exception."
-  [message exception instance status headers]
-  (ex-info message
-           (merge (metrics/retrieve-local-stats-for-service (scheduler/instance->service-id instance))
```
shamsimam: Did we intentionally avoid merging the local stats for the service?

waiter/src/waiter/process_request.clj
```diff
@@ -544,40 +526,49 @@
...
+                                response-elapsed (:elapsed timed-response)
+                                {:keys [error] :as response} (:out timed-response)
                                 request-abort-callback (request-abort-callback-factory response)
                                 confirm-live-connection-with-abort (confirm-live-connection-factory request-abort-callback)]
-                            (when error
-                              (throw-response-error error reservation-status-promise instance response-headers))
-                            (process-backend-response-fn local-usage-agent instance-request-properties descriptor instance request
-                                                         reason-map response-headers reservation-status-promise
-                                                         confirm-live-connection-with-abort request-state-chan response))
+                            (statsd/histo! metric-group "backend_response" response-elapsed)
```
shamsimam: Observation: working with `response-elapsed` is no longer in spatially close locations, unlike the previous version.

waiter/src/waiter/process_request.clj
```diff
@@ -505,11 +488,10 @@
           confirm-live-connection-factory #(confirm-live-connection-factory control-mult reservation-status-promise %1)
           confirm-live-connection-without-abort (confirm-live-connection-factory nil)
           waiter-debug-enabled? (utils/request->debug-enabled? request)
-          ; update response headers eagerly to enable reporting these in case of failure
-          response-headers (atom {})
-          add-debug-header-into-response! (fn [name value]
-                                            (when waiter-debug-enabled?
-                                              (swap! response-headers assoc (str name) (str value))))]
+          add-debug-header (fn [response header value]
```
shamsimam: nitpick: rename to `assoc-debug-header`

waiter/src/waiter/process_request.clj
```diff
@@ -284,17 +266,21 @@
...
-    (when (= status 202)
+    (if (= status 202)
       (if (str/starts-with? location "/")
         (do
           (deliver reservation-status-promise :success-async) ;; backend is processing as an asynchronous request
-          (post-process-async-request-response-fn service-id metric-group instance (handler/make-auth-user-map request)
-                                                  reason-map instance-request-properties location response-headers))
-        (log/info "response status 202, not treating as an async request as location is" location)))))
+          (post-process-async-request-response-fn response service-id metric-group instance
+                                                  (handler/make-auth-user-map request) reason-map
```
shamsimam: `(handler/make-auth-user-map request)` has no status/body/headers. `post-process-async-request-response` calls this `response`. Should we interpret any map as a response? Should we call this something else?

waiter/src/waiter/metrics.clj
```diff
@@ -449,6 +449,16 @@
      (~handle-elapsed-nanos-fn elapsed#)
      out#))
 
+(defmacro with-timer
+  "Times the operation specified by body using the provided
+  timer and calls handle-elapsed-nanos-fn with the epased
+  time in nanoseconds. Returns [value of body, nanos]."
```
shamsimam: Returns a map with keys `:out` and `:elapsed`.

waiter/src/waiter/async_request.clj
sradack: `response` is first effectively, as we end up using this with `partial`, which works nicely with `->`.  Which response handlers have it at the end?

waiter/src/waiter/async_request.clj
sradack: Before this PRl, we were overriding existing headers when we `merge`d over headers with `@response-headers`.  This function is only called for `202` responses where `Location` starts with `/` (ensuring these conditions is handled in `inspect-for-202-async-request-response`.

waiter/src/waiter/process_request.clj
sradack: The stats are still included, but it is now handled in `handle-response-error`.

waiter/src/waiter/process_request.clj
sradack: We're calling `post-process-async-request-response-fn`, defined in core, and it throws this argument away.  Should we clean that up in this PR?

waiter/src/waiter/process_request.clj
sradack: Renamed

waiter/src/waiter/process_request.clj
sradack: Done

waiter/src/waiter/process_request.clj
sradack: Done.

waiter/src/waiter/metrics.clj
sradack: Fixed

waiter/src/waiter/async_request.clj
shamsimam: Request handlers tend to have it at the end that's why the nitpick to have response at the end for response handlers. Also, this is actually not even the response :)

waiter/src/waiter/async_request.clj
shamsimam: okay, `inspect-for-202-async-request-response` also uses lower-case `location`.

waiter/src/waiter/process_request.clj
shamsimam: My bad, missed the `metrics-map` in `handle-response-error`

waiter/src/waiter/process_request.clj
shamsimam: `post-process-async-request-response` seems to be using `response` https://github.com/twosigma/waiter/blob/888bbdf70f1c30755469f5dba5631b3e6e89e37e/waiter/src/waiter/async_request.clj#L176

sradack: I'm still not sure I follow.  `post-process-async-request-response` processes the `response` from the backend.  It needs to look at the `response` to determine if it needs to modify the location.  As an example, `stream-http-response` also has the `response` as the first argument.  Are you saying we still should have the `response` argument at the end?

shamsimam: The comment was whether to call this `response`.

PR #247, opened 2018-03-19
Support for Interstitial page

shamsimam:

- introduces support for interstitial page for services that opt in

Allows Waiter display a service loading page thus providing response to users while the service starts up.

### Implementation details:

- The interstitial redirects with a GET call and an extra query parameter to bypass the interstitial only for the first `interstitial-secs` seconds for a service
- The interstitial is not responsible for starting a service, the service start is triggered by the redirect call which bypasses the interstitial
- run-as-requester and on-the-fly requests do not trigger display of interstitials

### Screenshots:

**Interstitial page** rendered while the service is starting up.

interstitial-rendered-2

**Backend page** rendered after a get, note the extra query parameter to bypass interstitial

backend-rendered-2

**Interstitial State**:

interstitial-state

**Service interstitial metrics**:

service-metrics

**Waiter interstitial metrics**:

waiter-metrics


shamsimam:
@sradack this commit has the meat https://github.com/twosigma/waiter/pull/247/commits/9c9d85f635fe09d6ecb04b4feec2a3d07bce8cb4

waiter/src/waiter/process_request.clj
```diff
@@ -483,6 +482,57 @@
             (meters/mark! (metrics/waiter-meter "core" "process-errors"))
             (utils/exception->response e request)))))))
 
+(defn service-id->interstitial-promise
```
sradack: Should we create `waiter.interstitial` for these functions?

waiter/src/waiter/process_request.clj
```diff
@@ -483,6 +482,57 @@
...
+  [handler service-id->interstitial-promise-atom]
+  (fn wrap-interstitial-handler [{:keys [descriptor] :as request}]
+    (let [{:keys [service-description service-id]} descriptor
+          ;; trigger interstitial-promise irrespective of the bypass parameter
+          interstitial-promise (or (service-id->interstitial-promise service-id->interstitial-promise-atom service-id)
+                                   (let [{:strs [interstitial-secs]} service-description]
+                                     (start-service-interstitial! service-id->interstitial-promise-atom service-id interstitial-secs)))]
+      (if (or (comment TODO contains-bypass-interstitial-param)
+              (realized? interstitial-promise))
+        (handler request)
```
sradack: `start-new-app` is called in `handler`.  Are we planning on have interstitial start the app itself?  Are we going to have two places where apps might be started?

waiter/src/waiter/process_request.clj
```diff
@@ -483,6 +482,57 @@
...
+    (let [interstitial-timed-out-promise (promise)]
+      (->> (fn [service-id->interstitial-promise]
+             (log/info "creating interstitial promise for" service-id)
+             (if-not (contains? service-id->interstitial-promise service-id)
+               ;; TODO (shams) trigger call to start service in follow-up PR, so we can move it out of prepare instance
+               (assoc service-id->interstitial-promise service-id interstitial-timed-out-promise)
+               service-id->interstitial-promise))
+           (swap! service-id->interstitial-promise-atom))
+      ;; TODO (shams) trigger deliver as soon as we have a healthy instance for this service,
+      ;; TODO (shams) e.g. by running a separate go-block that listens to scheduler messages
```
sradack: We might already have a healthy instance (say in the event of a Waiter restart). Are we going to wait until we have a scheduler message?

waiter/src/waiter/process_request.clj
```diff
@@ -483,6 +482,57 @@
...
+      (if (pos? interstitial-secs)
+        (async/go
+          (-> interstitial-secs t/seconds t/in-millis async/timeout async/<!)
```
sradack: Did we decide that `interstitial-secs` is based upon the time we started the service?  Do we need to account for that here?

waiter/src/waiter/process_request.clj
shamsimam: We can work around that case by introducing a separate flag in the atom which will effectively monitor the interstitial state.

waiter/src/waiter/process_request.clj
shamsimam: Eventually this block will also include starting the service, so it will include the starting time as marathon delay + time to actually start service.

waiter/src/waiter/process_request.clj
shamsimam: The plan is to have interstitial be the only one that starts the service.

waiter/src/waiter/process_request.clj
shamsimam: I decided to not implement starting the service in this PR. Our current logic relies on an instance requesting for an instance to decide on starting a service, that logic is not easy to port when we eagerly respond with an interstitial but expect an outstanding service to trigger starting an instance.

shamsimam:
@sradack ready for another round of review

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+   It will return nil if the service is not currently mapped to a promise."
+  [interstitial-state-atom service-id]
+  (get-in @interstitial-state-atom [:service-id->interstitial-promise service-id]))
+
+(defn- resolve-promise!
+  "Resolves a promise to the specified value.
+   It must be called inside a critical section to avoid data races."
+  [interstitial-promise interstitial-resolution]
+  (deliver interstitial-promise interstitial-resolution)
+  (when (= interstitial-resolution (deref interstitial-promise))
```
sradack: Nitpick: Here we can use `@interstitial-promise`

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+    (counters/inc! (metrics/waiter-counter "interstitial" "promise" "resolved"))
```
sradack: How can the promise be resolved more than once?  I noticed the picture you posted in the PR description has a value of `3` for this.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+            bypass-interstitial? (str/ends-with? (str query-string) bypass-interstitial-param-name-value)]
+        (if (or bypass-interstitial? ;; bypass query parameter provided
+                (zero? interstitial-secs) ;; interstitial support has been disabled
+                (not (str/includes? (str accept) text-html)) ;; expecting html response
+                (not (interstitial-state-initialized? interstitial-state-atom))
+                (let [interstitial-promise (->> (partial process-interstitial-secs service-id interstitial-secs)
+                                                (start-service-interstitial! interstitial-state-atom service-id))]
+                  (or (nil? interstitial-promise)
+                      (realized? interstitial-promise))))
+          (->> (cond-> query-string
```
sradack: We're relying on the followup bypass request to actually start the service, correct?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+(defn process-interstitial-secs
+  "Resolves the interstitial-promise after a delay on interstitial-secs seconds."
+  [service-id interstitial-secs interstitial-promise]
+  (if (pos? interstitial-secs)
+    (async/go
+      (-> interstitial-secs t/seconds t/in-millis async/timeout async/<!)
+      (log/info "interstitial timed out for" service-id)
+      (resolve-promise! interstitial-promise :interstitial-time-out))
+    (do
+      (log/info service-id "opted out of interstitial")
```
sradack: Do we need to create promises for services that don't use interstitials?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+      (resolve-promise! interstitial-promise :interstitial-opt-out))))
+
+(defn- process-scheduler-messages
+  "Processes messages from the scheduler.
+   In particular, it resolves all promises for services which have healthy instances.
+   It also removes from the interstitial state any resolved promises for services which are no longer available."
+  [interstitial-state-atom service-id->service-description current-available-service-ids scheduler-messages]
+  (loop [[[message-type message-data] & remaining-scheduler-messages] scheduler-messages
+         scheduler-available-service-ids #{}]
+    (if message-type
```
sradack: Would it be clearer to extract the data you need from messages, and process based upon that data?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+            {:strs [interstitial-secs]} service-description
+            ;; the bypass interstitial should be the last query parameter
+            bypass-interstitial? (str/ends-with? (str query-string) bypass-interstitial-param-name-value)]
+        (if (or bypass-interstitial? ;; bypass query parameter provided
+                (zero? interstitial-secs) ;; interstitial support has been disabled
+                (not (str/includes? (str accept) text-html)) ;; expecting html response
+                (not (interstitial-state-initialized? interstitial-state-atom))
+                (let [interstitial-promise (->> (partial process-interstitial-secs service-id interstitial-secs)
+                                                (start-service-interstitial! interstitial-state-atom service-id))]
+                  (or (nil? interstitial-promise)
```
sradack: How can `interstitial-promise` be `nil`?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+                 interstitial-state))
+             (swap! interstitial-state-atom))
+        (if-let [interstitial-promise (service-id->interstitial-promise interstitial-state-atom service-id)]
+          (do
+            (when (identical? new-interstitial-promise interstitial-promise)
+              (counters/inc! (metrics/waiter-counter "interstitial" "promise" "total"))
+              (log/error "processing interstitial promise for" service-id)
+              (try
+                (process-interstitial-promise interstitial-promise)
+                (catch Exception e
```
sradack: What are we worried will throw an exception?

waiter/src/waiter/interstitial.clj
sradack: Ignore.  I realized later that there aren't independent counters per service.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+                  (or (nil? interstitial-promise)
+                      (realized? interstitial-promise))))
+          (->> (cond-> query-string
+                       bypass-interstitial? (subs 0
+                                                  (max 0
+                                                       (- (count query-string)
+                                                          (inc (count bypass-interstitial-param-name-value))))))
+               (assoc request :query-string)
+               handler)
+          (let [target-url (str (name (utils/request->scheme request)) "://" host uri "?"
```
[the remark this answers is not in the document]
shamsimam: Changed.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,238 @@
...
+                                (when (not (str/blank? query-string))
+                                  (str query-string "&"))
+                                ;; the bypass interstitial should be the last query parameter
+                                bypass-interstitial-param-name-value)]
+            (counters/inc! (metrics/service-counter service-id "request-counts" "interstitial"))
+            {:body (-> descriptor
+                       (assoc :target-url target-url)
+                       render-interstitial-template)
```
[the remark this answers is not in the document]
shamsimam: Done. managed to reuse `wrap-descriptor` :)

shamsimam:
@sradack ready for another round of review

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+   In particular, it resolves all promises for services which have healthy instances.
+   It also removes from the interstitial state any resolved promises for services which are no longer available."
+  [interstitial-state-atom service-id->service-description current-available-service-ids scheduler-messages]
+  (let [available-service-ids (->> scheduler-messages
+                                   (some (fn [[message-type message-data]]
+                                           (when (= :update-available-apps message-type)
+                                             (:available-apps message-data))))
+                                   set)
+        healthy-instance-service-ids (->> scheduler-messages
+                                          (map (fn [[message-type message-data]]
```
sradack: Nitpick:  We can use `keep`.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+          {:strs [interstitial-secs]} service-description
+          ;; the bypass interstitial should be the last query parameter
+          bypass-interstitial? (str/ends-with? (str query-string) bypass-interstitial-param-name-value)]
+      (if (or bypass-interstitial? ;; bypass query parameter provided
+              (zero? interstitial-secs) ;; interstitial support has been disabled
+              (not (str/includes? (str (get headers "accept")) "text/html")) ;; expecting html response
+              (headers/contains-waiter-header headers sd/service-description-keys) ;; on-the-fly request
+              (not (:initialized? @interstitial-state-atom))
+              (->> (fn [interstitial-promise]
+                     (store-service-description service-id service-description)
```
sradack: Why do we need to call `store-service-description`?  It doesn't feel like it should be a responsibility of `wrap-interstitial`.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+                     (install-interstitial-timeout! service-id interstitial-secs interstitial-promise))
```
sradack: Can `install-interstitial-timeout!` be moved inside `ensure-service-interstitial!`?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+  [handler interstitial-state-atom store-service-description]
+  (fn wrap-interstitial-handler [{:keys [descriptor headers query-string uri] :as request}]
+    (let [{:keys [service-description service-id]} descriptor
+          {:strs [interstitial-secs]} service-description
+          ;; the bypass interstitial should be the last query parameter
+          bypass-interstitial? (str/ends-with? (str query-string) bypass-interstitial-param-name-value)]
+      (if (or bypass-interstitial? ;; bypass query parameter provided
+              (zero? interstitial-secs) ;; interstitial support has been disabled
+              (not (str/includes? (str (get headers "accept")) "text/html")) ;; expecting html response
+              (headers/contains-waiter-header headers sd/service-description-keys) ;; on-the-fly request
```
sradack: Should we add `:on-the-fly?` to the descriptor?  We're already performing this check in `compute-service-description`.  Also we likely want to use `sd/on-the-fly-service-description-keys`.  

This will be called on every request for interstitial-enabled services. It might be preferable to do it only once inside `compute-service-description`.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+        removed-service-ids (remove-resolved-interstitial-promises! interstitial-state-atom service-ids-to-remove)]
+    (doseq [service-id available-service-ids]
+      (let [{:strs [interstitial-secs]} (service-id->service-description service-id)]
+        (when (pos? interstitial-secs)
+          (->> (partial install-interstitial-timeout! service-id interstitial-secs)
+               (ensure-service-interstitial! interstitial-state-atom service-id)))))
+    (doseq [service-id healthy-instance-service-ids]
+      (when-let [interstitial-promise (service-id->interstitial-promise @interstitial-state-atom service-id)]
+        (resolve-promise! service-id interstitial-promise :healthy-instance-found)))
+    (-> current-available-service-ids
```
sradack: Is the value of this thread ever different than the value of `available-service-ids`?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+  (fn wrap-interstitial-handler [{:keys [descriptor headers query-string uri] :as request}]
+    (let [{:keys [service-description service-id]} descriptor
+          {:strs [interstitial-secs]} service-description
+          ;; the bypass interstitial should be the last query parameter
+          bypass-interstitial? (str/ends-with? (str query-string) bypass-interstitial-param-name-value)]
+      (if (or bypass-interstitial? ;; bypass query parameter provided
+              (zero? interstitial-secs) ;; interstitial support has been disabled
+              (not (str/includes? (str (get headers "accept")) "text/html")) ;; expecting html response
+              (headers/contains-waiter-header headers sd/service-description-keys) ;; on-the-fly request
+              (not (:initialized? @interstitial-state-atom))
```
sradack: I still don't follow why I need to check `:initialized`.  If a promise doesn't already exist, won't it create a promise that will eventually be resolved?

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+              (->> (fn [interstitial-promise]
+                     (store-service-description service-id service-description)
+                     (install-interstitial-timeout! service-id interstitial-secs interstitial-promise))
+                   (ensure-service-interstitial! interstitial-state-atom service-id)
+                   realized?))
+        ;; continue processing down the handler chain
+        (->> (if bypass-interstitial?
+               (->> (count bypass-interstitial-param-name-value)
```
sradack: Nitpick: It might be nice to factor out a `fn` for this to keep things at a high level.

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+                    inc
+                    (- (count query-string))
+                    (max 0)
+                    (subs query-string 0))
+               query-string)
+             (assoc request :query-string)
+             handler)
+        ;; redirect to interstitial page
+        (let [location (str "/waiter-interstitial" uri (when (not (str/blank? query-string)) (str "?" query-string)))]
```
sradack: Nitpick: `when-not`

waiter/src/waiter/interstitial.clj
```diff
@@ -0,0 +1,237 @@
...
+        target-url (str "/" path "?"
+                        (when (not (str/blank? query-string))
+                          (str query-string "&"))
+                        ;; the bypass interstitial should be the last query parameter
+                        bypass-interstitial-param-name-value)]
+    {:body (render-interstitial-template
+             {:service-description service-description
+              :service-id service-id
+              :target-url target-url})
+     :headers {}
```
sradack: Nitpick: Remove `:headers`

waiter/src/waiter/interstitial.clj
shamsimam: Done.

waiter/src/waiter/interstitial.clj
shamsimam: Moved, now that I don't need to save the service description.

waiter/src/waiter/interstitial.clj
shamsimam: Done.

waiter/src/waiter/interstitial.clj
shamsimam: Done.

waiter/src/waiter/interstitial.clj
shamsimam: Done.

waiter/src/waiter/interstitial.clj
shamsimam: Added `:on-the-fly?`

waiter/src/waiter/interstitial.clj
sradack: Thank you.  That makes sense.

sradack: Good thinking/nice behavior.

PR #248, opened 2018-03-21
Creates request log

sradack:

- Creates a request log

- Having a request log is useful for debugging.

### DIFF
--- a/waiter/resources/log4j.properties
+++ b/waiter/resources/log4j.properties
@@ -19,12 +19,12 @@ log4j.appender.ErrorAppender.layout=org.apache.log4j.PatternLayout
 # CID will be replaced by the custom pattern layout configured in waiter.correlation-id/replace-pattern-layout-in-log4j-appenders
 log4j.appender.ErrorAppender.layout.ConversionPattern=%d{ISO8601} %-5p %c [%t] - [CID] %m%n
 
-log4j.category.waiter.request-log=INFO, requestlog
-log4j.additivity.waiter.request-log=false
+log4j.category.RequestLog=INFO, RequestLogAppender
+log4j.additivity.RequestLog=false
 
-log4j.appender.requestlog=org.apache.log4j.DailyRollingFileAppender
-log4j.appender.requestlog.Threshold=INFO
-log4j.appender.requestlog.File=log/${waiter.logFilePrefix}request.log
-log4j.appender.requestlog.DatePattern='.'yyyy-MM-dd
-log4j.appender.requestlog.layout=org.apache.log4j.PatternLayout
-log4j.appender.requestlog.layout.ConversionPattern=%m%n
+log4j.appender.RequestLogAppender=org.apache.log4j.DailyRollingFileAppender
+log4j.appender.RequestLogAppender.Threshold=INFO
+log4j.appender.RequestLogAppender.File=log/${waiter.logFilePrefix}request.log
+log4j.appender.RequestLogAppender.DatePattern='.'yyyy-MM-dd
+log4j.appender.RequestLogAppender.layout=org.apache.log4j.PatternLayout
+log4j.appender.RequestLogAppender.layout.ConversionPattern=%m%n
--- a/waiter/resources/log4j.properties
+++ b/waiter/resources/log4j.properties
@@ -18,3 +18,13 @@ log4j.appender.ErrorAppender.DatePattern='.'yyyy-MM-dd
 log4j.appender.ErrorAppender.layout=org.apache.log4j.PatternLayout
 # CID will be replaced by the custom pattern layout configured in waiter.correlation-id/replace-pattern-layout-in-log4j-appenders
 log4j.appender.ErrorAppender.layout.ConversionPattern=%d{ISO8601} %-5p %c [%t] - [CID] %m%n
+
+log4j.category.waiter.request-log=INFO, requestlog
+log4j.additivity.waiter.request-log=false
+
+log4j.appender.requestlog=org.apache.log4j.DailyRollingFileAppender
+log4j.appender.requestlog.Threshold=INFO
+log4j.appender.requestlog.File=log/${waiter.logFilePrefix}request.log
+log4j.appender.requestlog.DatePattern='.'yyyy-MM-dd
+log4j.appender.requestlog.layout=org.apache.log4j.PatternLayout
+log4j.appender.requestlog.layout.ConversionPattern=%m%n
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -555,7 +555,7 @@
                                   (process-backend-response-fn local-usage-agent instance-request-properties descriptor instance request
                                                                reason-map reservation-status-promise confirm-live-connection-with-abort
                                                                request-state-chan response))
-                                (assoc :backend-response-latency-ms (utils/nanos->millis response-elapsed))
+                                (assoc :backend-response-latency-ns response-elapsed)
                                 (assoc-debug-header "x-waiter-backend-response-ns" (str response-elapsed))))
                           (catch Exception e
                             (deliver reservation-status-promise :generic-error)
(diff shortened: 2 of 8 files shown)


waiter/resources/log4j.properties
```diff
@@ -18,3 +18,13 @@ log4j.appender.ErrorAppender.DatePattern='.'yyyy-MM-dd
 log4j.appender.ErrorAppender.layout=org.apache.log4j.PatternLayout
 # CID will be replaced by the custom pattern layout configured in waiter.correlation-id/replace-pattern-layout-in-log4j-appenders
 log4j.appender.ErrorAppender.layout.ConversionPattern=%d{ISO8601} %-5p %c [%t] - [CID] %m%n
+
+log4j.category.waiter.request-log=INFO, requestlog
```
shamsimam: we should have a comment in here about why this is configured to be separate.
It also means we cannot have any other log statements in `request-log.clj`

waiter/src/waiter/process_request.clj
```diff
@@ -555,13 +555,15 @@
...
+                                (assoc :backend-response-latency-ms (utils/nanos->millis response-elapsed))
                                 (assoc-debug-header "x-waiter-backend-response-ns" (str response-elapsed))))
                           (catch Exception e
                             (deliver reservation-status-promise :generic-error)
                             ; close request-state-chan to mark the request as finished
                             (async/close! request-state-chan)
                             (handle-process-exception e request)))
-                        (assoc :instance instance)
+                        (assoc :instance instance
+                               :get-instance-latency-ms (utils/nanos->millis instance-elapsed))
```
shamsimam: Nitpick: lexicographic order

waiter/src/waiter/request_log.clj
```diff
@@ -0,0 +1,106 @@
...
+            (if (au/chan? body)
+              (let [body' (async/chan 5)]
+                (async/go
+                  (let [response-body-size (async/<! (stream-countable-body! body body'))]
+                    (log-request! request (assoc response'
+                                                 :response-body-size response-body-size))))
+                (assoc response' :body body'))
+              (do
+                (log-request! request response')
+                response'))))
```
shamsimam: Why exclude the `response-body-size` here?

waiter/src/waiter/process_request.clj
```diff
@@ -555,13 +555,15 @@
                                   (process-backend-response-fn local-usage-agent instance-request-properties descriptor instance request
                                                                reason-map reservation-status-promise confirm-live-connection-with-abort
                                                                request-state-chan response))
+                                (assoc :backend-response-latency-ms (utils/nanos->millis response-elapsed))
```
shamsimam: Is there no path from which `response-elapsed` is defined to use here that an exception can be thrown. If not, the `try-catch` wrapping it can be much narrower.

waiter/src/waiter/request_log.clj
```diff
@@ -0,0 +1,106 @@
...
+      instance (assoc :instance-host (:host instance)
+                      :instance-id (:id instance)
+                      :instance-port (:port instance)
+                      :instance-proto (:protocol instance)
+                      :get-instance-latency-ms get-instance-latency-ms)
+      principal (assoc :principal principal)
+      response-body-size (assoc :response-body-size response-body-size)
+      total-latency-ms (assoc :total-latency-ms total-latency-ms))))
+
+(defn stream-countable-body!
```
shamsimam: `stream-http-response` has `(au/timed-offer! resp-chan buffer streaming-timeout-ms)`. Should that logic be repeated here?

waiter/src/waiter/utils.clj
```diff
@@ -551,3 +551,24 @@
...
+      int))
+
+(let [byte-array-type (Class/forName "[B")]
+  (defn byte-array?
+  "Tests if a value is a byte-array."
+  [val]
+    (when val
+      (instance? byte-array-type val))))
+
+(defn nanos->millis
```
shamsimam: Should we have a PR in `clojure.time` for this?

waiter/src/waiter/utils.clj
```diff
@@ -551,3 +551,24 @@
...
+  (-> value
+      Math/ceil
+      int))
+
+(let [byte-array-type (Class/forName "[B")]
+  (defn byte-array?
+  "Tests if a value is a byte-array."
+  [val]
+    (when val
+      (instance? byte-array-type val))))
```
shamsimam: Nitpick: this reads better as `(and val (instance? byte-array-type val))`

sradack: Removed this function

waiter/src/waiter/utils.clj
sradack: Removed this function.

waiter/src/waiter/request_log.clj
sradack: No longer counting bytes.

waiter/src/waiter/request_log.clj
sradack: Byte counting removed.

waiter/src/waiter/process_request.clj
sradack: Sorry, I don't follow.  Can you explain a bit more?

waiter/src/waiter/process_request.clj
```diff
@@ -555,13 +555,15 @@
...
+                                (assoc :backend-response-latency-ns response-elapsed)
                                 (assoc-debug-header "x-waiter-backend-response-ns" (str response-elapsed))))
                           (catch Exception e
                             (deliver reservation-status-promise :generic-error)
                             ; close request-state-chan to mark the request as finished
                             (async/close! request-state-chan)
                             (handle-process-exception e request)))
-                        (assoc :instance instance)
+                        (assoc :instance instance
+                               :get-instance-latency-ns instance-elapsed)
```
shamsimam: Nitpick: lexicographic order

waiter/src/waiter/request_log.clj
```diff
@@ -0,0 +1,81 @@
...
+            [metrics.timers :as timers]
+            [waiter.async-utils :as au]
+            [waiter.metrics :as metrics]
+            [waiter.ring-utils :as ru]
+            [waiter.utils :as utils]))
+
+(defn log
+  "Log log-data as JSON"
+  [log-data]
+  (log/log "RequestLog" :info nil (json/write-str log-data :escape-slash false)))
```
shamsimam: 👍

waiter/src/waiter/request_log.clj
```diff
@@ -0,0 +1,81 @@
...
+(defn request->context
+  "Convert a request into a context suitable for logging."
+  [{:keys [headers query-string remote-addr request-id request-method request-time uri] :as request}]
+  (let [{:strs [host x-cid x-forwarded-for]} headers]
+    (cond-> {:cid x-cid
+             :host host
+             :path uri
+             :remote-addr (or x-forwarded-for remote-addr)
+             :request-id request-id
+             :scheme (let [scheme (utils/request->scheme request)]
```
shamsimam: Can `scheme` be `nil`? Else we can call `name` on it directly. Else: `(some-> (utils/request->scheme request) name)`

waiter/src/waiter/process_request.clj
```diff
@@ -555,13 +555,15 @@
                                   (process-backend-response-fn local-usage-agent instance-request-properties descriptor instance request
                                                                reason-map reservation-status-promise confirm-live-connection-with-abort
                                                                request-state-chan response))
+                                (assoc :backend-response-latency-ns response-elapsed)
```
shamsimam: `response-elapsed` is created in https://github.com/twosigma/waiter/pull/248/files#diff-d77fcc9f68a157c1df598bfad7a0ccf4R546
Can an exception be thrown between then and this line? If so, this value will end up not being logged.

waiter/src/waiter/request_log.clj
sradack: I don't believe so.  If it is, then `name` is only called if it's a keyword.

sradack: Scheme can be a string if it's coming from haproxy via x-forwarded-proto

waiter/src/waiter/utils.clj
```diff
@@ -123,16 +123,16 @@
   ([^DateTime date-time formatter]
    (when date-time
      (f/unparse
-       (f/with-zone formatter (t/default-time-zone))
-       (.withZone date-time DateTimeZone/UTC)))))
```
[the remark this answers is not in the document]
sradack: Removed

waiter/src/waiter/request_log.clj
shamsimam:
```
(name "some-string") => "some-string"
(name :some-keyword) => "some-keyword"
```

shamsimam: Do you want to change or should I go ahead and merge?

PR #254, opened 2018-04-02
Provide service-description to the service factory

DaoWen:

Provide the `service-id->service-description-fn` mapping to the scheduler through the `create-component` context.

Other schedulers (e.g., Kubernetes) may need additional information from the service description in order to implement some of the Waiter `Scheduler` object API methods. We'd rather plumb this mapping through in one spot (to provide it to the scheduler's initialization function) than plumb it through to every scheduler method invocation.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -504,12 +504,6 @@
                              {:blacklist-backoff-base-time-ms blacklist-backoff-base-time-ms
                               :inter-kill-request-wait-time-ms inter-kill-request-wait-time-ms
                               :max-blacklist-time-ms max-blacklist-time-ms})
-   :scheduler (pc/fnk [[:settings scheduler-config]
-                       service-id-prefix]
-                (let [is-waiter-app?-fn
-                      (fn is-waiter-app? [^String service-id]
-                        (str/starts-with? service-id service-id-prefix))]
-                  (utils/create-component scheduler-config :context {:is-waiter-app?-fn is-waiter-app?-fn})))
    :scheduler-state-chan (pc/fnk [] (au/latest-chan))
    :service-description-builder (pc/fnk [[:settings service-description-builder-config service-description-constraints]]
                                   (when-let [unknown-keys (-> service-description-constraints
@@ -598,6 +592,24 @@
                          latch (LeaderLatch. curator leader-latch-path router-id)]
                      (.start latch)
                      latch))})
+
+(def scheduler
+  {:scheduler (pc/fnk [[:settings scheduler-config]
+                       [:state service-id-prefix]
+                       service-id->service-description-fn*]
+                (let [is-waiter-app?-fn (fn is-waiter-app? [^String service-id]
+                                          (str/starts-with? service-id service-id-prefix))]
+                  (utils/create-component scheduler-config :context {:is-waiter-app?-fn is-waiter-app?-fn
+                                                                     :service-id->service-description-fn service-id->service-description-fn*})))
+   ; This function is only included here for initializing the scheduler above.
+   ; Prefer accessing the non-starred version of this function through the routines map.
+   :service-id->service-description-fn* (pc/fnk [[:curator kv-store]
+                                                [:settings service-description-defaults metric-group-mappings]]
+                                          (fn service-id->service-description
+                                            [service-id & {:keys [effective?] :or {effective? true}}]
+                                            (sd/service-id->service-description
+                                              kv-store service-id service-description-defaults
+                                              metric-group-mappings :effective? effective?)))})
 
 (def routines
   {:allowed-to-manage-service?-fn (pc/fnk [[:curator kv-store]
@@ -734,14 +746,10 @@
                               (fn service-id->password [service-id]
                                 (log/debug "generating password for" service-id)
                                 (digest/md5 (str service-id (first passwords)))))
-   :service-id->service-description-fn (pc/fnk [[:curator kv-store]
-                                                [:settings service-description-defaults metric-group-mappings]]
-                                         (fn service-id->service-description
-                                           [service-id & {:keys [effective?] :or {effective? true}}]
-                                           (sd/service-id->service-description
-                                             kv-store service-id service-description-defaults
-                                             metric-group-mappings :effective? effective?)))
-   :start-new-service-fn (pc/fnk [[:state authenticator scheduler start-app-cache-atom task-threadpool]
+   :service-id->service-description-fn (pc/fnk [[:scheduler service-id->service-description-fn*]]
+                                         service-id->service-description-fn*)
+   :start-new-service-fn (pc/fnk [[:scheduler scheduler]
+                                  [:state authenticator start-app-cache-atom task-threadpool]
                                   service-id->password-fn store-service-description-fn]
                            (fn start-new-service [{:keys [service-id core-service-description] :as descriptor}]
                              (let [run-as-user (get-in descriptor [:service-description "run-as-user"])]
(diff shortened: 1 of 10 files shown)


shamsimam:
We had the same issue with `service-id->password-fn`. I'm evaluating why we chose not to do that same thing for it. If we end up merging this PR, we should do the same refactoring for `service-id->password-fn`.

waiter/src/waiter/core.clj
```diff
@@ -804,7 +813,8 @@
                      scaling/scale-app service-id->service-description-fn router-state-push-mult)))
    :autoscaling-multiplexer (pc/fnk [[:routines delegate-instance-kill-request-fn peers-acknowledged-blacklist-requests-fn]
                                      [:settings [:scaling inter-kill-request-wait-time-ms] blacklist-config]
-                                     [:state instance-rpc-chan scheduler]]
+                                     [:state instance-rpc-chan]
+                                     [:scheduler scheduler]]
```
shamsimam: Nitpick: lexicographic order on the keys (`:scheduler` before `:settings`).

waiter/src/waiter/main.clj
```diff
@@ -60,11 +60,12 @@
 
 (defn wire-app
   [settings]
-  {:settings (pc/fnk dummy-symbol-for-fnk-schema-logic :- settings/settings-schema [] settings)
-   :curator core/curator
-   :routines core/routines
```
shamsimam: Thank you for sorting!

waiter/src/waiter/core.clj
DaoWen: fixed

waiter/src/waiter/main.clj
DaoWen: I thought you'd like that 😄

DaoWen:
I removed the now-redundant parameters from `get-instances` and `get-apps->instances`. However, I think `service-id->password-fn` refactoring should be in a separate PR.

waiter/src/waiter/core.clj
```diff
@@ -868,14 +878,16 @@
                                  :maintainer-chans maintainer-chan}))
    :scheduler-broken-services-gc (pc/fnk [[:curator gc-state-reader-fn gc-state-writer-fn leader?-fn]
                                           [:settings scheduler-gc-config]
-                                          [:state clock scheduler]
+                                          [:state clock]
+                                          [:scheduler scheduler]
```
shamsimam: Nitpick: lexicographic order.

waiter/src/waiter/core.clj
```diff
@@ -868,14 +878,16 @@
...
                                           scheduler-maintainer]
                                    (let [scheduler-state-chan (async/tap (:scheduler-state-mult-chan scheduler-maintainer) (au/latest-chan))
                                          service-gc-go-routine (partial service-gc-go-routine gc-state-reader-fn gc-state-writer-fn leader?-fn clock)]
                                      (scheduler/scheduler-broken-services-gc scheduler scheduler-state-chan scheduler-gc-config service-gc-go-routine)))
    :scheduler-maintainer (pc/fnk [[:routines service-id->service-description-fn]
                                   [:settings [:health-check-config health-check-timeout-ms failed-check-threshold] scheduler-syncer-interval-secs]
-                                  [:state clock scheduler]]
+                                  [:state clock]
+                                  [:scheduler scheduler]]
```
shamsimam: Nitpick: lexicographic order.

waiter/src/waiter/core.clj
```diff
@@ -888,7 +900,8 @@
    :scheduler-services-gc (pc/fnk [[:curator gc-state-reader-fn gc-state-writer-fn leader?-fn]
                                    [:routines router-metrics-helpers service-id->service-description-fn]
                                    [:settings scheduler-gc-config]
-                                   [:state clock scheduler]
+                                   [:state clock]
+                                   [:scheduler scheduler]
```
shamsimam: Nitpick: lexicographic order.

waiter/src/waiter/core.clj
```diff
@@ -999,7 +1012,8 @@
                             :content-type "image/png"}))
    :kill-instance-handler-fn (pc/fnk [[:routines peers-acknowledged-blacklist-requests-fn]
                                       [:settings [:scaling inter-kill-request-wait-time-ms] blacklist-config]
-                                      [:state instance-rpc-chan scheduler]
+                                      [:state instance-rpc-chan]
+                                      [:scheduler scheduler]
```
shamsimam: Nitpick: lexicographic order.

waiter/src/waiter/core.clj
```diff
@@ -1039,7 +1053,8 @@
                                       router-metrics-agent metrics-sync-interval-ms bytes-encryptor bytes-decryptor request))))
    :service-handler-fn (pc/fnk [[:curator kv-store]
                                 [:routines allowed-to-manage-service?-fn generate-log-url-fn make-inter-router-requests-sync-fn]
-                                [:state router-id scheduler]
+                                [:state router-id]
+                                [:scheduler scheduler]
```
shamsimam: Nitpick: lexicographic order.

waiter/test/waiter/core_test.clj
```diff
@@ -391,8 +391,8 @@
                        :routines {:allowed-to-manage-service?-fn allowed-to-manage-service?
                                   :generate-log-url-fn nil
                                   :make-inter-router-requests-sync-fn nil}
-                       :state {:router-id "router-id"
-                               :scheduler (Object.)}
+                       :state {:router-id "router-id"}
+                       :scheduler {:scheduler (Object.)}
```
shamsimam: Nitpick: lexicographic order.

waiter/test/waiter/core_test.clj
```diff
@@ -463,8 +463,8 @@
                        :routines {:allowed-to-manage-service?-fn (constantly true)
                                   :generate-log-url-fn (partial handler/generate-log-url #(str "http://www.example.com" %))
                                   :make-inter-router-requests-sync-fn nil}
-                       :state {:router-id "router-id"
-                               :scheduler (Object.)}
+                       :state {:router-id "router-id"}
+                       :scheduler {:scheduler (Object.)}
```
shamsimam: Nitpick: lexicographic order.

waiter/test/waiter/scheduler/marathon_test.clj
```diff
@@ -583,7 +583,7 @@
         service-id "service-1"
         instance-id "service-1.A"
         make-marathon-scheduler #(->MarathonScheduler {} {} (fn [] nil) "/home/path/"
-                                                      (atom {}) %1 %2 (constantly true))
+                                                      (atom {}) %1 {} %2 (constantly true))
```
shamsimam: Can we please extract the map in a common variable and name it `service-id->service-description-fn`? Seeing the map is confusing when I was expecting to have a function there. We should be able to use the common variables in all the map occurrence below.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -268,11 +268,12 @@
 
 (defrecord MarathonScheduler [marathon-api mesos-api retrieve-framework-id-fn
                               home-path-prefix service-id->failed-instances-transient-store
-                              service-id->kill-info-store force-kill-after-ms is-waiter-app?-fn]
+                              service-id->kill-info-store service-id->service-description
```
shamsimam: Nice how the name just fits in and doesn't trigger downstream changes when removing the parameter :)

waiter/test/waiter/scheduler/marathon_test.clj
DaoWen: Switched to `(constantly nil)` as per our discussion.

shamsimam:
Let me know when it is ready for another pass.

DaoWen:
@shamsimam - I just checked the callers to `get-instances` and `get-apps->instances`. In each case, there's only one call site. The former passes the same `service-id->service-description-fn` to `get-instances` that we're now providing to the Scheduler record directly, so no change there. The latter looks up the service description in the kv-store, and then passes that to `get-apps->instances`. Since `service-id->service-description-fn` performs the same lookup, that call path won't introduce any new failures either. I don't think we need to add redundant logic to either of those methods in the marathon scheduler implemenation. (The shell scheduler was already ignoring the arguments that we removed, so no issues there.)

DaoWen:
@shamsimam - If you're happy with the patch, then I think this is good to merge.

DaoWen:
@shamsimam - It's green! (Finally!!!)

PR #255, opened 2018-04-02
Use Travis Docker VMs only when needed

DaoWen:

Update Travis YAML file to expand the test matrix like we did for Cook.

Not all of our tests require the docker service. Explicitly including the docker service only where we actually use it lets the simpler test suites run in a container rather than a VM. The Travis containers have faster start-up time, plus more CPUs and RAM allocated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,9 +54,9 @@ matrix:
     - env: "NAME='Waiter integration tests: shell-fast'"
       services: docker
       before_script: cd waiter && $LEIN_DEPS_CMD
-      script: ./bin/ci/run-integration-tests.sh parallel-test integration-fast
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,33 +1,62 @@
...
+
+    - env: "NAME='Waiter integration tests: shell-fast'"
+      services: docker
+      before_script: cd waiter && $LEIN_DEPS_CMD
+      script: ./bin/ci/run-integration-tests.sh parallel-test integration-fast
+
+    - env: "NAME='Waiter integration tests: shell-slow'"
+      services: docker
+      before_script: cd waiter && $LEIN_DEPS_CMD
+      script: ./bin/ci/run-integration-tests.sh parallel-test integration-slow
```
dposada: Shouldn't these last two be using `run-integration-tests-shell-scheduler.sh`?

PR #256, opened 2018-04-02
Force kill expired instances while scaling down

shamsimam:

- Never kill healthy instances in the presence of an expired instance
- Choose oldest expired instance while killing, an expired instance is chosen for killing only if it is idle or if it is servicing lingering requests
- Introduce `:lingering-request-threshold-ms` settings in `:instance-request-properties` to configure how long to wait before force killing an expired instance servicing lingering requests
- Always prefer oldest live healthy instances to youngest expired instances while servicing a request
- Refactor to pass config maps to `autoscaling-multiplexer` and `service-chan-responder`

When expired instances remain busy due to long-running requests, the autoscaler leads to jitters in scaling up (to replace an expired instance) and then scaling down the just created new instance. Instead, we want to kill any expired instance in favor of healthy idle instances while scaling down. This will prevent the jitter in the number of instances for a service that the autoscaler currently causes while scaling.


waiter/src/waiter/state.clj
```diff
@@ -46,6 +53,37 @@
...
+  (every? #(t/before? % earliest-request-threshold-time)
+          (->> request-id->use-reason-map
+               vals
+               (map :time))))
+
+(defn find-oldest-acceptable-expired-instance
+  "Retrieves the _oldest_ acceptable expired instance. An expired instance is acceptable if:
+   1. it passes the acceptable-instance-id? predicate, and
+   2. it is idle or all its requests are older than lingering-request-threshold-ms ms.
+   We do not explicitly check whether instances are idle as `find-killable-instance` should handle that scenario."
```
sradack: This function would never return an idle instance, so this comment may be confusing.

waiter/src/waiter/state.clj
```diff
@@ -31,10 +31,17 @@
 
 ;; Router state maintainers
 
-(defn- healthy? [status-tags]
+(defn- healthy?
```
sradack: This and `expired?` might make more sense accepting an instance's state instead of the `status-tags` map directly, since they tell you whether an instance is healthy, not whether a set of status tags is healthy.  Also it may improve readability where they are used.

waiter/src/waiter/state.clj
```diff
@@ -46,6 +53,37 @@
...
+   We do not explicitly check whether instances are idle as `find-killable-instance` should handle that scenario."
+  [instance-id->request-id->use-reason-map lingering-request-threshold-ms id->instance instance-id->state acceptable-instance-id?]
+  (some->> instance-id->state
+           (filter (fn [[_ {:keys [status-tags]}]] (expired? status-tags)))
+           seq ;; trigger short-circuit if there are no expired instances
+           (filter (fn [[instance-id _]] (acceptable-instance-id? instance-id)))
+           (filter (let [earliest-request-threshold-time (t/minus (t/now) (t/millis lingering-request-threshold-ms))]
+                     (fn [[instance-id _]]
+                       (-> (instance-id->request-id->use-reason-map instance-id)
+                           (only-lingering-requests?  earliest-request-threshold-time)))))
```
sradack: Extra space

waiter/src/waiter/state.clj
```diff
@@ -82,27 +115,44 @@
     (->> instance-id->state
          (filter (fn [[_ state]] (killable? state)))
          (filter (fn [[instance-id _]] (acceptable-instance-id? instance-id)))
+         ; if there are expired instances, only choose healthy instances that are blacklisted or expired
```
sradack: Would this preference (and the preference below it related to starting instances), be better expressed inside our comparator?

waiter/src/waiter/state.clj
```diff
@@ -325,24 +375,29 @@
...
-      {:current-state' (-> current-state 
-                           (assoc-in [:instance-id->request-id->use-reason-map instance-id request-id] reason-map)
-                           ; mark instance as locked if it is going to be killed.
-                           (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :locked)))
-       :response-chan resp-chan
-       :response instance-to-offer}
+  [{:keys [id->instance instance-id->request-id->use-reason-map instance-id->state] :as current-state}
+   update-status-tag-fn lingering-request-threshold-ms [{:keys [request-id] :as reason-map} resp-chan exclude-ids-set _]]
+  (let [acceptable-instance-id? #(not (contains? exclude-ids-set %))
+        ;; look for idle killable instance first before force killing an expired instance
```
sradack: Could `find-killable-instance` more naturally handle this preference, without an extra explicit call to `find-oldest-acceptable-expired-instance`?  It already maintains a priority over the order in which instances get killed.

waiter/src/waiter/state.clj
```diff
@@ -46,6 +53,37 @@
   (and (zero? slots-used)
        (not-any? #(contains? status-tags %) [:killed :locked])))
 
+(defn- find-max
+  "Uses the provided comparison function to find the max element in collection."
+  [comparison-fn collection]
+  (reduce #(if (pos? (comparison-fn %1 %2)) %1 %2) (first collection) (rest collection)))
+
+(defn- only-lingering-requests?
```
sradack: Is `only-lingering-request` an extension to the definition of what's `killable?`?  Would that be a more natural place to put this logic?

waiter/src/waiter/state.clj
```diff
@@ -46,6 +53,37 @@
...
+  "Retrurns truthy value if the instance is idle or processing expired requests based on the earliest-request-threshold-time"
+  [request-id->use-reason-map earliest-request-threshold-time]
+  (every? #(t/before? % earliest-request-threshold-time)
+          (->> request-id->use-reason-map
+               vals
+               (map :time))))
+
+(defn find-oldest-acceptable-expired-instance
```
sradack: This function feels a lot like `find-killable-instance`.  Could we merge the logic?  See my other comments for more color.

waiter/src/waiter/state.clj
```diff
@@ -46,6 +53,37 @@
   (and (zero? slots-used)
        (not-any? #(contains? status-tags %) [:killed :locked])))
 
+(defn- find-max
```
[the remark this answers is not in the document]
pschorf: If not, I think the two argument version of `reduce` would work

waiter/src/waiter/state.clj
pschorf: I don't think this would work, because you could only have the "replacement" instance in the list at this point (assuming the expired one was filtered out by `killable?`)

waiter/src/waiter/state.clj
```diff
@@ -80,43 +111,59 @@
         expired-instances? (some (fn [[_ {:keys [status-tags]}]] (contains? status-tags :expired)) instance-id->state)
         starting-instances? (some (fn [[_ {:keys [status-tags]}]] (contains? status-tags :starting)) instance-id->state)]
     (->> instance-id->state
-         (filter (fn [[_ state]] (killable? state)))
+         (filter (fn [[instance-id state]]
+                   (-> (instance-id->request-id->use-reason-map instance-id)
+                       (killable? earliest-request-threshold-time state))))
          (filter (fn [[instance-id _]] (acceptable-instance-id? instance-id)))
+         seq ;; trigger short-circuit if there are no expired instances
```
pschorf: Is this comment right? What is this `seq` doing?

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,75 @@
...
-  [id->instance instance-id->state acceptable-instance-id?]
-  (let [find-max (fn [comparison-fn collection]
-                   (reduce
-                     #(if (pos? (comparison-fn %1 %2)) %1 %2)
-                     (first collection)
-                     (rest collection)))
-        instance-id-state-pair->categorizer-vec (fn [[_ {:keys [status-tags]}]]
+  [id->instance instance-id->state acceptable-instance-id? instance-id->request-id->use-reason-map lingering-request-threshold-ms]
+  (let [earliest-request-threshold-time (t/minus (t/now) (t/millis lingering-request-threshold-ms))
+        extract-alphanumeric (fn [s] (str/replace (str s) #"[^A-Za-z0-9]" ""))
```
shamsimam: The requirement on the `started-at` is that it is a string that must be formatted to allow string comparisons. So I'm relying on the ordering of alphanumeric characters.

waiter/src/waiter/state.clj
```diff
@@ -325,24 +372,27 @@
...
-      {:current-state' (-> current-state 
-                           (assoc-in [:instance-id->request-id->use-reason-map instance-id request-id] reason-map)
-                           ; mark instance as locked if it is going to be killed.
-                           (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :locked)))
-       :response-chan resp-chan
-       :response instance-to-offer}
+  [{:keys [id->instance instance-id->request-id->use-reason-map instance-id->state] :as current-state}
+   update-status-tag-fn lingering-request-threshold-ms [{:keys [request-id] :as reason-map} resp-chan exclude-ids-set _]]
+  (let [acceptable-instance-id? #(not (contains? exclude-ids-set %))
+        ;; look for idle killable instance first before choosing an expired instance with lingering requests
```
pschorf: Remove this comment

waiter/src/waiter/state.clj
shamsimam: Ugh, changing it to `some->>` to ensure the comment stays right

waiter/src/waiter/state.clj
shamsimam: Done.

waiter/src/waiter/state.clj
```diff
@@ -79,44 +123,60 @@
...
+                                                             reverse
+                                                             seq)))]
+    (->> sorted-instance-ids
+         (filter acceptable-instance-id?)
+         (filter (fn [instance-id]
+                   (let [state (instance-id->state instance-id)]
+                     (and (not (nil? state))
+                          (slots-available? state)
+                          (healthy? state)))))
+         order-instance-ids-based-on-expired-state
```
sradack: Do we need `order-instance-ids-based-on-expired-state`?  Could `sorted-instance-ids` be already in the correct order?

waiter/src/waiter/state.clj
```diff
@@ -79,44 +123,60 @@
...
+         first
+         id->instance)))
 
 (defn- sort-instances
-  "Sorts two service instances, the comparison order specified by `instance-comparator`."
+  "Sorts two service instances, the comparison order is: expired, service-id, started-at, and finally id."
```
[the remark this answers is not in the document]
sradack: Changing this function also changes our instance distribution among routers?  Should we update how distribution works so that it doesn't depend on expiry?

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,88 @@
...
+           (not-any? #(contains? status-tags %) [:killed :locked]))
+      (and (expired? state)
+           (only-lingering-requests? request-id->use-reason-map earliest-request-threshold-time))))
+
+(defn- find-max
+  "Uses the provided comparison function to find the max element in collection."
+  [comparison-fn collection]
+  (reduce #(if (pos? (comparison-fn %1 %2)) %1 %2) (first collection) (rest collection)))
+
+(defn- instance->numeric-started-at
```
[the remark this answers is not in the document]
sradack: Do we need this function?  `sort-instances` assumes `started-at` is sortable without transformation.

waiter/src/waiter/state.clj
```diff
@@ -392,16 +454,23 @@
...
   (cid/with-correlation-id cid
     (log/info "attempt to blacklist" instance-id "which has"
               (count (get instance-id->request-id->use-reason-map instance-id)) "uses with state:"
               (get instance-id->state instance-id))
     ;; cannot blacklist if the instance is currently servicing a request
-    (let [response-code (if (and (contains? instance-id->request-id->use-reason-map instance-id)
-                                 (some #(= :serve-request (:reason %)) (vals (get instance-id->request-id->use-reason-map instance-id))))
-                          :in-use
-                          :blacklisted)]
+    (let [instance-not-allowed? (if (-> (instance-id->state instance-id) expired?)
```
sradack: Should we be calling `killable?`?

waiter/src/waiter/state.clj
shamsimam: Created `sort-instances-for-processing` to perform the sort for reserving instances. Killing instances has its own sort based on the categorizer.

waiter/src/waiter/state.clj
```diff
@@ -79,44 +123,60 @@
...
+         id->instance)))
 
 (defn- sort-instances
-  "Sorts two service instances, the comparison order specified by `instance-comparator`."
+  "Sorts two service instances, the comparison order is: expired, service-id, started-at, and finally id."
   [expired-instances coll]
-  (sort (instance-comparator expired-instances) coll))
+  (let [instance->feature-vector (memoize
+                                   (fn [{:keys [id service-id started-at]}]
+                                     [(contains? expired-instances id) service-id started-at id]))
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/state.clj
```diff
@@ -79,44 +123,60 @@
...
+                       (-> (instance-id->request-id->use-reason-map instance-id)
+                           (killable? earliest-request-threshold-time state))))
+             (filter (fn [[instance-id _]] (acceptable-instance-id? instance-id)))
+             seq ;; trigger short-circuit if there are no expired instances
+             ; if there are expired instances, only choose healthy instances that are blacklisted or expired
+             (filter (fn [[_ {:keys [status-tags]}]]
+                       (if (and expired-instances? (contains? status-tags :healthy))
+                         (some #(contains? status-tags %) [:blacklisted :expired])
+                         true)))
+             ; if there are expired and starting instances, only kill unhealthy instances that are not starting
```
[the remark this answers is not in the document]
shamsimam: Moved the logic to `killable?`

waiter/src/waiter/state.clj
```diff
@@ -79,44 +123,60 @@
...
-         (find-max instance-id-comparator)
-         (first) ; extract the instance-id
-         (get id->instance))))
+    (some->> instance-id->state
+             (filter (fn [[instance-id state]]
+                       (-> (instance-id->request-id->use-reason-map instance-id)
+                           (killable? earliest-request-threshold-time state))))
+             (filter (fn [[instance-id _]] (acceptable-instance-id? instance-id)))
+             seq ;; trigger short-circuit if there are no expired instances
+             ; if there are expired instances, only choose healthy instances that are blacklisted or expired
```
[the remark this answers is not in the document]
shamsimam: Moved the logic to `killable?`

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
-                   (reduce
-                     #(if (pos? (comparison-fn %1 %2)) %1 %2)
-                     (first collection)
-                     (rest collection)))
-        instance-id-state-pair->categorizer-vec (fn [[_ {:keys [status-tags]}]]
+  [id->instance instance-id->state acceptable-instance-id? instance-id->request-id->use-reason-map lingering-request-threshold-ms]
+  (let [earliest-request-threshold-time (t/minus (t/now) (t/millis lingering-request-threshold-ms))
+        instance-id-state-pair->categorizer-vec (fn [[instance-id {:keys [slots-used status-tags] :as state}]]
                                                   ; most important goes first
                                                   (vector
```
sradack: Nitpick: Use vector literal

waiter/src/waiter/state.clj
```diff
@@ -212,7 +254,7 @@
                                           expired-instance-ids))
             sorted-instance-ids (->> (concat healthy-instances unhealthy-instances expired-instances)
                                      (set)
-                                     (sort-instances expired-instance-ids)
+                                     (sort-instances-for-processing expired-instance-ids)
```
sradack: Do we want our instance distribution to be affected by expiry?

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
-       (not-any? #(contains? status-tags %) [:killed :locked])))
+  "Only the following instances can be killed:
+   - without slots used and not killed|locked, or
+   - expired and idle or processing lingering requests.
+   If there are expired instances, only choose healthy instances that are blacklisted or expired.
+   If there are expired and starting instances, only kill unhealthy instances that are not starting."
+  [request-id->use-reason-map earliest-request-threshold-time expired-instances? starting-instances?
+   {:keys [slots-used status-tags] :as state}]
+  (and (or (and (zero? slots-used)
+                (not-any? #(contains? status-tags %) [:killed :locked]))
```
sradack: Should we ever try killing a `:killed` or `:locked` instance?  Should this be a top-level condition?

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
+(defn- starting-instances?
+  "Returns truthy value if there is some instance in an starting state."
+  [instance-id->state]
+  (some (fn [[_ {:keys [status-tags]}]] (contains? status-tags :starting)) instance-id->state))
 
 (defn find-killable-instance
   "While killing instances choose using the following logic:
    if there are expired and starting instances, only kill unhealthy instances that are not starting;
    choose instances in following order:
    - choose amongst idle instances
```
sradack: The method documentation needs an overhaul given the new logic.

waiter/src/waiter/state.clj
shamsimam: I prefer `vector` as the elements are spread over multiple lines.

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
-        instance-id-state-pair->categorizer-vec (fn [[_ {:keys [status-tags]}]]
+  [id->instance instance-id->state acceptable-instance-id? instance-id->request-id->use-reason-map lingering-request-threshold-ms]
+  (let [earliest-request-threshold-time (t/minus (t/now) (t/millis lingering-request-threshold-ms))
+        instance-id-state-pair->categorizer-vec (fn [[instance-id {:keys [slots-used status-tags] :as state}]]
                                                   ; most important goes first
                                                   (vector
+                                                    (zero? slots-used)
+                                                    (contains? status-tags :expired)
                                                     (contains? status-tags :unhealthy)
                                                     (every? #(contains? status-tags %) [:healthy :blacklisted])
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
-                                                    (contains? status-tags :expired)
-                                                    (contains? status-tags :healthy)))
+                                                    (contains? status-tags :healthy)
```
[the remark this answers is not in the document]
shamsimam: No, removed.

waiter/src/waiter/state.clj
shamsimam: No and they are not currently.

waiter/src/waiter/state.clj
shamsimam: Moved to a top-level condition

waiter/src/waiter/state.clj
shamsimam: Done.

waiter/src/waiter/state.clj
```diff
@@ -31,44 +31,86 @@
...
+                                                    (cond-> (or (some-> instance-id id->instance :started-at .getMillis) 0)
```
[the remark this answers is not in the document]
shamsimam: Done.

shamsimam:
Find killable instance did not find any instance with the following input:
```
{waiter-service-testscalinghealthyapptravis1247626-9d673f29105c7cd7d565c8cf0846620b.682559af-3cb7-11e8-b8ab-0242ac110005 
 {:slots-assigned 3, :slots-used 0, :status-tags #{:healthy}}, 
 waiter-service-testscalinghealthyapptravis1247626-9d673f29105c7cd7d565c8cf0846620b.5d48488d-3cb7-11e8-b8ab-0242ac110005 
 {:slots-assigned 3, :slots-used 0, :status-tags #{:healthy}}, 
 waiter-service-testscalinghealthyapptravis1247626-9d673f29105c7cd7d565c8cf0846620b.6825cee0-3cb7-11e8-b8ab-0242ac110005 
 {:slots-assigned 0, :slots-used 0, :status-tags #{:blacklisted}}}
```.
I will debug and fix.

shamsimam:
@pschorf @sradack I added another commit after you guys last approved it. I ran this internally with an instance expiry of 3 minutes and it works like a charm :)

PR #257, opened 2018-04-03
[README] Set lein version requirement

dmed256:

- Set lein version requirement

Failed to install with `2.6.0`, `2.6.1`, and `2.7.0`

### DIFF
--- a/waiter/README.md
+++ b/waiter/README.md
@@ -7,7 +7,7 @@ Waiter uses [Marathon](https://mesosphere.github.io/marathon/) to schedule servi
 Prerequisites:
 
 * Java 8
-* [Leiningen](http://leiningen.org/) (at least v2.8.1)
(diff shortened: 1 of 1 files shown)


waiter/README.md
```diff
@@ -7,7 +7,7 @@ Waiter uses [Marathon](https://mesosphere.github.io/marathon/) to schedule servi
 Prerequisites:
 
 * Java 8
-* [Leiningen](http://leiningen.org/)
+* [Leiningen](http://leiningen.org/) (at least v2.8.1)
```
DaoWen: Maybe we should just specify "v2.8.1", getting rid of the "at least" part. We've had issues with `lein` plugins (e.g., `voom`) breaking with new versions of Leiningen, so I'd rather not imply that newer versions are also supported.

PR #258, opened 2018-04-05
Adds support for last-update-user metadata during token updates

shamsimam:

- add support for last-update-user metadata during token updates

Knowing who the last modifier of a token is is useful.

token-last-update-user

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -882,4 +882,4 @@
             (delete-service waiter-url service-id)))
 
         (finally
-          (delete-token-and-assert waiter-url token))))))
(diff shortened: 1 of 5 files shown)


waiter/integration/waiter/token_request_test.clj
```diff
@@ -869,4 +882,4 @@
             (delete-service waiter-url service-id)))
 
         (finally
-          (delete-token-and-assert waiter-url token))))))
+          #_(delete-token-and-assert waiter-url token))))))
```
sradack: Did we mean to remove this before committing?

shamsimam: Yes :(

PR #264, opened 2018-04-09
introduces waiter.util package

shamsimam:

- introduces waiter.util package

Improves code organization and simplifies the path to splitting `utils.clj`


DaoWen:
Please also move `client-tools` into this new package. `test-utils` seems like a more accurate name.

shamsimam:
@DaoWen Moved `client-tools` to `util` package.

waiter/src/waiter/util/utils.clj
```diff
@@ -8,7 +8,7 @@
 ;;       The copyright notice above does not evidence any
 ;;       actual or intended publication of such source code.
 ;;
-(ns waiter.utils
+(ns waiter.util.utils
```
DaoWen: Nitpick: Not a big fan of the name. It sounds like you're planning on splitting this up soon though, and I don't have a better suggestion right now...

PR #266, opened 2018-04-11
adds support for token history

shamsimam:

- add support for preserving token history
- the amount of history stored is configurable with `history-length` config

This helps track the change history (limited by the history length). It also provides support for enabling blue-green deployment where we can figure out the service-id of the previous version of the service for a given token.

screen shot 2018-04-12 at 1 03 29 am

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -527,7 +527,7 @@
 
  :token-config {
                 ;; Configures the amount of history we store for a token.
-                ;; When it is at least 1, it enables blue/green deployment of services.
+                ;; It must be a positive integer.
                 :history-length 5}
 
  ;; Waiter supports making websocket connections to backends.
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -525,6 +525,11 @@
                        ;; flagged as never-passed-health-checks upon failure
                        :failed-check-threshold 5}
 
+ :token-config {
+                ;; Configures the amount of history we store for a token.
+                ;; When it is at least 1, it enables blue/green deployment of services.
+                :history-length 5}
+
  ;; Waiter supports making websocket connections to backends.
  ;; The maximum binary and text message sizes in the websocket request or response can be configured
  ;; as follows, the values are in bytes:
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -113,7 +113,7 @@
 (def ^:const on-the-fly-service-description-keys (set/union service-description-keys #{"token"}))
 
 ; keys allowed in metadata for tokens, these need to be distinct from service description keys
-(def ^:const token-metadata-keys #{"deleted" "last-update-time" "last-update-user" "owner" "root"})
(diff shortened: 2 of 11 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -443,7 +443,9 @@
   [kv-store ^String token error-on-missing include-deleted]
   (let [{:strs [deleted run-as-user] :as token-data} (when token (kv/fetch kv-store token))
         token-data (when token-data ; populate token owner for backwards compatibility
-                     (update-in token-data ["owner"] (fn [current-owner] (or current-owner run-as-user))))]
+                     (cond-> token-data
```
sradack: Nitpick:
```clojure
(-> token-data
   (update "owner" (fn [owner] (or owner run-as-user)))
   (update "previous" (fn [previous] (or previous {}))))
```

waiter/src/waiter/token.clj
```diff
@@ -24,13 +24,26 @@
 (def ^:const ANY-USER "*")
 (def ^:const valid-token-re #"[a-zA-Z]([a-zA-Z0-9\-_$\.])+")
 
+(defn sanitize-history
+  "Limits the history length stored in the token-data."
+  [history-length token-data]
```
sradack: Should we have `token-data` since it's the most important argument?

waiter/src/waiter/token.clj
```diff
@@ -24,13 +24,26 @@
...
+  (utils/dissoc-in token-data (repeat history-length "previous")))
+
+(defn ensure-history
+  "Ensures a non-nil previous entry exists in `token-data`.
+   If one already was present when this function was called, returns `token-data` unmodified.
+   Else if assocs `(or previous {})` into `token-data`."
+  [previous token-data]
```
sradack: Should we have `token-data` since it's the most important argument?

waiter/config-full.edn
```diff
@@ -525,6 +525,11 @@
                        ;; flagged as never-passed-health-checks upon failure
                        :failed-check-threshold 5}
 
+ :token-config {
+                ;; Configures the amount of history we store for a token.
+                ;; When it is at least 1, it enables blue/green deployment of services.
```
sradack: Should we wait until blue/green is enabled to have this in the comments?

waiter/src/waiter/service_description.clj
shamsimam: I wanted to avoid creating and invoking that function. `update` also carries the connotation that we will be modifying the value using the existing value mapped against a key, which is not true in this case.

waiter/src/waiter/token.clj
shamsimam: Placing the argument second was motivated by the use of `->>` in the locations it is used.

waiter/src/waiter/token.clj
shamsimam: Placing the argument second was motivated by the use of `->>` in the locations it is used.

waiter/config-full.edn
shamsimam: Removed. Updated comment.

PR #270, opened 2018-04-12
handles invalid method on keyword during websocket close

shamsimam:

- handles invalid method on keyword during websocket close

Avoids exceptions while closing websocket connections during error or cookie timeouts.

### DIFF
--- a/waiter/src/waiter/websocket.clj
+++ b/waiter/src/waiter/websocket.clj
@@ -374,7 +374,7 @@
                         correlation-id
                         (if (integer? status-code-or-exception)
                           (close-client-session! request status-code-or-exception close-message)
-                          (let [ex-message (.getMessage status-code-or-exception)]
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/main.clj
```diff
@@ -133,7 +133,8 @@
   (Thread/setDefaultUncaughtExceptionHandler
     (reify Thread$UncaughtExceptionHandler
       (uncaughtException [_ thread throwable]
-        (log/error throwable (str (.getName thread) " threw exception: " (.getMessage throwable))))))
+        (->> (str (.getName thread) " triggered uncaught exception handler: " (.getMessage throwable))
```
sradack: Nitpick: threading enthusiasm

shamsimam: changed

PR #271, opened 2018-04-12
handles non-map responses in request-log

shamsimam:

- handles non-map responses in request-log

Websocket handlers are not required to return maps as responses. As a result the request-log handler may not assume that request is a map.

### DIFF
--- a/waiter/src/waiter/request_log.clj
+++ b/waiter/src/waiter/request_log.clj
@@ -9,13 +9,10 @@
 ;;       actual or intended publication of such source code.
 ;;
 (ns waiter.request-log
-  (:require [clj-time.core :as t]
-            [clojure.core.async :as async]
-            [clojure.data.json :as json]
+  (:require [clojure.data.json :as json]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [metrics.timers :as timers]
-            [waiter.async-utils :as au]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/request_log.clj
```diff
@@ -60,8 +57,12 @@
 
 (defn log-request!
   "Log a request"
-  [request response]
-  (log (merge (request->context request) (response->context response))))
+  [request response additional-context]
+  (log (merge (request->context request)
+              (when (map? response)
```
sradack: When is it not a map?  I was expecting the map to come back with all sorts of goodies.

shamsimam: Websocket handlers are not required to return a map. E.g. the inter-router metrics connections do not return a map. Another option is to create a separate request-log wrapper that only looks into the request map and use it for websocket handlers.

sradack: We should update the metrics handler to return a map so we can log which router authenticated and any other things we'd might like to log.  Would you rather do that in a separate PR?

shamsimam: Separate PR, we can merge that before merging this one.

shamsimam:
Not needed now that we have https://github.com/twosigma/waiter/pull/272

PR #278, opened 2018-04-16
Support authorization string in http-request helper options

DaoWen:

Add option for passing an `Authorization` header value to our `http-request` helper.

This was a nice way to provide an OAuth "Bearer" token to validate with the API server in the Kubernetes scheduler implementation (wip).

### DIFF
--- a/waiter/src/waiter/mesos/utils.clj
+++ b/waiter/src/waiter/mesos/utils.clj
@@ -23,18 +23,20 @@
    It performs a blocking read on the response and the response body.
    The body is assumed to be json and is parsed into a clojure data structure.
    If the status of the response in not 2XX, the response is thrown as an exception."
-  [http-client request-url & {:keys [accept body content-type headers query-string request-method
-                                     spnego-auth throw-exceptions]
+  [http-client request-url & {:keys [accept auth-str body content-type headers query-string
+                                     request-method spnego-auth throw-exceptions]
                               :or {spnego-auth false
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/mesos/utils.clj
```diff
@@ -23,18 +23,20 @@
...
    The body is assumed to be json and is parsed into a clojure data structure.
    If the status of the response in not 2XX, the response is thrown as an exception."
-  [http-client request-url & {:keys [accept body content-type headers query-string request-method
-                                     spnego-auth throw-exceptions]
+  [http-client request-url & {:keys [accept auth-str body content-type headers query-string
+                                     request-method spnego-auth throw-exceptions]
                               :or {spnego-auth false
                                    throw-exceptions true}}]
-  (let [request-map (cond-> {:as :string
+  (let [hs (cond-> headers
```
shamsimam: Why cannot the caller add the authorization header in the `headers` before invoking this function?

Nitpick: If we do go ahead with this change, I don't like the `hs` name, how about `headers'`?

DaoWen: I think I originally did it this way because I wanted to use the oath2 argument supported by http/request rather than associng it into the headers directly—but that didn't end up working correctly, so I patched this function to update the headers instead. You're right, this really doesn't need to be here. I'll move it into the Kubernetes code instead.

PR #279, opened 2018-04-16
Handle EOF exceptions in health-check connections

DaoWen:

Handle EOF exceptions in health-check connections.

I was getting failures with the shell scheduler in some of our integration tests because this exception wasn't being handled in the health-check logic.


shamsimam:
Please rebase.

PR #280, opened 2018-04-16
makes interstitials requests hit the same router

shamsimam:

- makes interstitials requests in the integration test hit the same router

Avoids flakiness in the test


sradack:
Why does sending requests to the same router make interstitials less flaky?

shamsimam:
https://github.com/twosigma/waiter/pull/280/files#diff-777b28cad6b2c2c5bca3dbc6beec28a7R759 relies on the interstital promise being already initialized on the router from a previous request (from one of the other threads). If the request in this thread ends up being the first request to hit that router, then it will be redirected to the interstitial and the test will fail.

PR #281, opened 2018-04-16
Two changes in waiter.state for readability

DaoWen:

- Inline the update function used by an `update-in`
- Refactor a `concat` expression to remove a confusing `cond->` and use `when` instead.

I was seriously confused by both of these expressions. I think the updated versions are more readable.


shamsimam:
Awaiting green build

PR #287, opened 2018-04-18
Allows token-syncer to update previously soft-deleted tokens

shamsimam:

- do not require etag while updating soft-deleted token
- requires etags for hard delete of active tokens

We would like the token syncer to be able to sync tokens.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -428,14 +428,6 @@
                   (assert-response-status response 412)
                   (is (str/includes? (str body) "Cannot modify stale token"))))
 
-              (testing "hard-delete without etag"
-                (let [{:keys [body] :as response} (make-request waiter-url "/token"
-                                                                :headers {"host" token}
-                                                                :http-method-fn http/delete
-                                                                :query-params {"hard-delete" true})]
-                  (assert-response-status response 400)
-                  (is (str/includes? (str body) "Must specify if-match header for token hard deletes"))))
-
               (testing "hard-delete with invalid etag"
                 (let [{:keys [body] :as response} (make-request waiter-url "/token"
                                                                 :headers {"host" token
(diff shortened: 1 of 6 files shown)


sradack:
Can we explain why this was a problem?  On the surface, it seems it might be correct that we would not stick soft-deleted tokens in the previous metadata.

shamsimam:
You're right, it seems correct that a soft-deleted token should not show up in the previous metadata. I've changed the syncer to handle updating soft-deleted tokens.

waiter/src/waiter/token.clj
```diff
@@ -236,15 +236,11 @@
           (let [token-owner (get token-metadata "owner")
                 version-hash (get headers "if-match")]
             (if hard-delete
-              (do
-                (when-not version-hash
```
sradack: Are we still ensuring that hard-deletes aren't happening on non-soft-deleted tokens?

shamsimam: Did we used to ensure it in the past? Afaik, any admin can hard-delete any token. 
The token syncer does hard deletes only if the token is soft-deleted on all clusters.

sradack: What allows us to remove this constraint?  Why don't we want `If-Match` on hard-deletes?

sradack: If we are hard deleting, require `If-Match` if not soft-deleted.

shamsimam:
Ready for review.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -428,14 +428,6 @@
                   (assert-response-status response 412)
                   (is (str/includes? (str body) "Cannot modify stale token"))))
 
-              (testing "hard-delete without etag"
```
sradack: Can we keep this test, or is there a change in behavior here?

shamsimam: Added back the integration test.

PR #292, opened 2018-04-20
Compute fallback service descriptor (token related fields)

shamsimam:

- extracts service-description-template computation in prepare-service-description-template-from-tokens
- adds support for computing the fallback service descriptor
- introduces assoc-if-absent to improve readability of the code
- order service-description schema keys

In preparation for the fallback service feature, we need to be able to compute the fallback service when tokens are used.


waiter/src/waiter/service_description.clj
```diff
@@ -747,12 +754,37 @@
                           (merge (error-message-map-fn passthrough-headers waiter-headers) (dissoc ex-data :message))
                           (:throwable &throw-context))))))))
 
+(defn retrieve-fallback-token
+  "Computes the most recently modified token from the token->token-data map."
+  [token->token-data]
+  (first (apply max-key (fn [[_ token-data]] (get-in token-data ["previous" "last-update-time"] 0)) token->token-data)))
```
sradack: Supporting fallbacks for multiple tokens has complications:

Case 1:
Let's assume we have a service that uses token A and B.  User updates A and B, in arbitrary order.
It seems arbitrary for us to explore the history of A, and not the history of B, or vice versa.

Case 2:
Let's assume user updates A to A'.
User sees that they didn't intend to update A, but really intended to update B.
User updates B to B', and then reverts A' to A.
When we find that [A B'] doesn't exist, we'll never explore for the running service [A B], instead looking for A'

Should we consider searching the space of all possible versions for any given time for running services?  Or should we consider not supporting multiple tokens for fallback?

waiter/src/waiter/service_description.clj
```diff
@@ -738,6 +741,10 @@
             (validate service-description-builder service-description {:allow-missing-required-fields? false}))
           {:core-service-description core-service-description
            :on-the-fly? contains-waiter-header?
+           :retrieve-fallback-service-description (fn []
```
sradack: What are the merits of this approach over repeated calls to `compute-service-description` directly?

waiter/src/waiter/service_description.clj
```diff
@@ -45,45 +45,45 @@
...
-   (s/optional-key "health-check-url") schema/non-empty-string
+   (s/optional-key "grace-period-secs") (s/both s/Int (s/pred #(<= 1 % (t/in-seconds (t/minutes 60))) 'at-most-60-minutes))
    (s/optional-key "health-check-interval-secs") (s/both s/Int (s/pred #(<= 5 % 60) 'between-5-seconds-and-1-minute))
    (s/optional-key "health-check-max-consecutive-failures") (s/both s/Int (s/pred #(<= 1 % 15) 'at-most-fifteen))
-   (s/optional-key "grace-period-secs") (s/both s/Int (s/pred #(<= 1 % (t/in-seconds (t/minutes 60))) 'at-most-60-minutes))
+   (s/optional-key "health-check-url") schema/non-empty-string
    (s/optional-key "idle-timeout-mins") (s/both s/Int (s/pred #(<= 1 % (t/in-minutes (t/days 30))) 'between-1-minute-and-30-days))
    (s/optional-key "interstitial-secs") (s/both s/Int (s/pred #(<= 0 % (t/in-seconds (t/minutes 60))) 'at-most-60-minutes))
    (s/optional-key "restart-backoff-factor") schema/positive-number-greater-than-or-equal-to-1
    ; auto-scaling related
```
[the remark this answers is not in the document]
shamsimam: This comment groups fields between the two comments `auto-scaling related` and `per-request related`. The fields inside that group have been ordered lexicographically.

waiter/src/waiter/service_description.clj
```diff
@@ -45,45 +45,45 @@
...
    (s/optional-key "scale-factor") schema/positive-fraction-less-than-or-equal-to-1
-   (s/optional-key "scale-up-factor") schema/positive-fraction-less-than-1
    (s/optional-key "scale-down-factor") schema/positive-fraction-less-than-1
-   (s/optional-key "jitter-threshold") schema/greater-than-or-equal-to-0-less-than-1
-   (s/optional-key "concurrency-level") (s/both s/Int (s/pred #(<= 1 % 10000) 'between-one-and-10000))
-   (s/optional-key "instance-expiry-mins") (s/constrained s/Int #(<= 0 %))
-   (s/optional-key "expired-instance-restart-rate") schema/positive-fraction-less-than-or-equal-to-1
+   (s/optional-key "scale-up-factor") schema/positive-fraction-less-than-1
    ; per-request related
    (s/optional-key "blacklist-on-503") s/Bool
```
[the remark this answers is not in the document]
shamsimam: The related parameters for `per-request related` (i.e. `blacklist-on-503` and `max-queue-length`) continue to be grouped together in this change of lexicographic ordering.

waiter/src/waiter/service_description.clj
shamsimam: Is your argument to inline the recursive call or to require the caller to make the explicit call to `compute-service-description`? I assume the latter as the first approach is a trivial change (and a matter of taste).

Remember that the user is actually calling `request->descriptor`. We could choose to offer a new function `descriptor->fallback-descriptor` or use this new method. With this approach we are generating and sending the function isntead of requiring the user to know about and invoke `descriptor->fallback-descriptor` directly. Also, some arguments such as `sources`, `waiter-headers`, `passthrough-headers`, ...) are either generated internally from other calls. With this approach, we do not require regeneration of the parameters (i.e. saving re-computation costs).

waiter/src/waiter/service_description.clj
shamsimam: > Should we consider searching the space of all possible versions for any given time for running services?

If we do, from your example, we would end up finding in the history in the following order using the token snapshots based on change timestamps: [A B'], [A' B'], [A' B], [A B]. So we will eventually end up finding the snapshot [A B] and its running service. We are bounded to `num-tokens` * `history-length` searches. We could even add a configuration to limit the lookup length. Also, if the user knows how the tokens will be looked up, they can change their behavior and decide to update A' to A before updating B to B'.

> Or should we consider not supporting multiple tokens for fallback?

That is one approach, but given that the other option is not that much harder, not sure why we would want to limit the search space.

waiter/src/waiter/service_description.clj
shamsimam: > In your example, does request->descriptor gain a new dependency on the state of running services?

It would be the job of `wrap-descriptor` or another filter that runs after `wrap-descriptor` to correct the descriptor based on running services. `request->descriptor` would remain as it is now, returning a descriptor based on the current request.

shamsimam:
Build is green.

PR #293, opened 2018-04-23
adds support for user-agent to the request log

shamsimam:

- adds support for user-agent to the request log

We would like to be able to determine the user-agent used in a request


dposada:

shamsimam:
Build is green.

PR #297, opened 2018-04-25
moves healthy-service-ids logic to update-scheduler-state

shamsimam:

- moves healthy-service-ids logic to update-scheduler-state
- return available-service-ids as a set in the scheduler messages

Moves the logic to determine services with healthy instances to the scheduler syncer. This centralizes the logic and avoids repetition in any of the consumers of the scheduler state.

It also return `available-service-ids` as a set in the scheduler messages as a couple of the consumers were converting it to a set. In addition `transient-metrics-state-producer` was expecting it to be a set, so that path should also work now.

### DIFF
--- a/waiter/src/waiter/interstitial.clj
+++ b/waiter/src/waiter/interstitial.clj
@@ -94,10 +94,10 @@
    In particular, it resolves all promises for services which have healthy instances.
    It also removes from the interstitial state any resolved promises for services which are no longer available."
   [interstitial-state-atom service-id->service-description current-available-service-ids scheduler-messages]
-  (let [summary-data (-> (fn [[message-type message-data]]
-                           (when (= :update-available-services message-type)
-                             message-data))
-                         (some scheduler-messages))
+  (let [summary-data (some (fn [[message-type message-data]]
+                             (when (= :update-available-services message-type)
+                               message-data))
+                           scheduler-messages)
         available-service-ids (-> summary-data :available-service-ids set)
         healthy-service-ids (-> summary-data :healthy-service-ids set)
         service-ids-to-remove (set/difference current-available-service-ids available-service-ids)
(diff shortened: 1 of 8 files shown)


waiter/src/waiter/interstitial.clj
```diff
@@ -94,29 +94,25 @@
...
-                                   (some (fn [[message-type message-data]]
-                                           (when (= :update-available-services message-type)
-                                             (:available-service-ids message-data))))
-                                   set)
-        healthy-instance-service-ids (->> scheduler-messages
-                                          (keep (fn [[message-type message-data]]
-                                                  (when (and (= :update-service-instances message-type)
-                                                             (seq (:healthy-instances message-data)))
-                                                    (:service-id message-data)))))
+  (let [summary-data (-> (fn [[message-type message-data]]
```
pschorf: Remove this threading macro

waiter/src/waiter/scheduler.clj
```diff
@@ -486,16 +489,29 @@
...
+              (recur
+                (assoc service-id->health-check-context'
+                  service-id {:instance-id->unhealthy-instance instance-id->unhealthy-instance'
+                              :instance-id->tracked-failed-instance instance-id->tracked-failed-instance'
+                              :instance-id->failed-health-check-count instance-id->failed-health-check-count'})
+                (cond-> healthy-service-ids
+                        (seq healthy-instances) (conj service-id))
+                scheduler-messages'
+                remaining))
+            (let [healthy-service-ids (->> scheduler-messages
```
pschorf: Remove this threading macro

waiter/src/waiter/interstitial.clj
shamsimam: Done, ended up improving the snippet quite a bit.

waiter/src/waiter/scheduler.clj
shamsimam: Removed. Actually, this snippet was not even needed. It also triggered my refactor of `available-service-ids` to a set.

shamsimam:
@pschorf ready for another review

PR #300, opened 2018-04-26
Simplifies text rendered on interstitial page

shamsimam:

- simplifies interstitial/display-interstitial-handler by reducing text that is rendered

This change allows us to choose any service specific to the token and does not promise the service id or the command that we will run.

Screenshot:

screen shot 2018-04-27 at 9 53 29 am

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -697,13 +697,13 @@
             (->> [(async/thread ;; check interstitial rendering
                     (let [request-headers (assoc request-headers "host" token)
                           {:keys [body] :as response}
-                          (make-request router-url (str "/waiter-interstitial/" service-id "/some-endpoint")
+                          (make-request router-url "/waiter-interstitial/some-endpoint"
                                         :cookies cookies
                                         :headers request-headers
                                         :query-params {"a" "b"})]
                       (assert-response-status response 200)
                       (is (str/includes? body (str "<title>Waiter - Interstitial for " service-id "</title>")))
-                      (is (str/includes? body "/some-endpoint?a=b&x-waiter-bypass-interstitial="))))
+                      (is (str/includes? body (str "/some-endpoint?a=b&x-waiter-bypass-interstitial=")))))
                   (async/thread ;; GET request inside the interstitial period, using DNS token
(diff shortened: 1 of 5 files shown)


shamsimam:
When we add support for fallback service, the initial redirected request and the interstitial page may not resolve to the same service descriptor due to asynchronous events being triggered internally. With the explicit use of service-id in the url, we can avoid this issue as we do not need to lookup the descriptor.

sradack:
I might be missing something:  The interstitial should only be displayed when no service (or fallback service) is running and available to handle the request.  They should get a non-interstitial response when the resolved service-id (or fallback service-id) is available.  If the previous statements are true, the interstitial seems to be a function of the token, and not of a particular service-id.

shamsimam:
Case-1: Assume service-v1, service-v2 and service-v3 are three versions of a token and service-v1, service-v2 are running.

user makes request to service-v3
fallback determines to redirect to service-v2, fallback starts service-v3
asynchronous state update causes service-v2 to be deleted
interstitial sees service-v2 is not running and redirects to interstitial
interstitial request initially resolved to service-v3
fallback determines to redirect to service-v1 (as v3 and v2 are not running)
interstitial rendered for service-v1

OR

Case-2: Assume service-v2 and service-v3 are three versions of a token and service-v2 is running.

user makes request to service-v3
fallback determines to redirect to service-v2, fallback starts service-v3
asynchronous state update causes service-v2 to be deleted
interstitial sees service-v2 is not running and redirects to interstitial
interstitial request initially resolved to service-v3
fallback determines no previous running service, stay on with service-v3
interstitial rendered for service-v3

Observations:
1. Interstitial can be triggered despite fallback logic.
2. In both cases, the interstitial was intended for service-v2 but ended up rendering interstitial for different services. 

We might choose to not care which interstitial is rendered as some service is being used to process the request. Or, we can be explicit about which service the interstitial was intended for and avoid the nondeterminism. This PR chooses the latter path.

shamsimam:
The asynchronous state updates can also happen after the interstitial is rendered which means the following bypass request can end up going to a service other than what is displayed on the interstitial page!

shamsimam:
I'm good with removing it.

shamsimam:
Ready for review.

PR #303, opened 2018-05-01
Pass through special characters in query strings to the backend

shamsimam:

- avoids passing the query-string as part of the endpoint in make-request and `BasicAuthentication$BasicResult`
- explicitly convert the query string into a map to pass into our library function for `make-request`
- adds `uri` and `query-string` to `/request-info` output in kitchen
- introduces `open-url-connection` helper function
- adds integration test to reproduce bad behavior with invalid characters in query strings

We would like to have Waiter be resilient to the presence of special characters in the query string.

### Previous log output

This was the error that was thrown before:
```
2018-05-01 11:26:03,271 ERROR waiter.process-request [async-dispatch-13] - [CID=4ff4388f3191-295e3b322333b8d6] error during process
java.net.URISyntaxException: Illegal character in query at index 41: http://127.0.0.4:10003/request-info?bad=a|b
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3111)
	at java.net.URI$Parser.parse(URI.java:3053)
	at java.net.URI.<init>(URI.java:588)
	at waiter.core$fn__37987$fnk37986_positional__37988$make_basic_auth_fn__37989.invoke(core.clj:681)
	at waiter.process_request$make_http_request.invokeStatic(process_request.clj:220)
	at waiter.process_request$make_http_request.invoke(process_request.clj:216)
...
```

and then later when we tried to pass in the query string in endpoint with our library:
```
2018-05-01 12:03:04,845 ERROR waiter.process-request [async-dispatch-50] - [CID=51f97bec24f6-36056137f5bdefc9] error during process
java.lang.IllegalArgumentException: Illegal character in query at index 41: http://127.0.0.5:10000/request-info?bad=a|b
	at java.net.URI.create(URI.java:852)
	at org.eclipse.jetty.client.HttpClient.newRequest(HttpClient.java:418)
	at qbits.jet.client.http$request.invokeStatic(http.clj:302)
	at qbits.jet.client.http$request.invoke(http.clj:275)
	at qbits.jet.client.http$get.invokeStatic(http.clj:398)
	at qbits.jet.client.http$get.invoke(http.clj:396)
	at waiter.process_request$make_http_request.invokeStatic(process_request.clj:227)
	at waiter.process_request$make_http_request.invoke(process_request.clj:219)
...
```

### Reference `ring-codec/form-decode` implementation:

```
(defn form-decode-str
  "Decode the supplied www-form-urlencoded string using the specified encoding,
  or UTF-8 by default."
  [^String encoded & [encoding]]
  (try
    (URLDecoder/decode encoded (or encoding "UTF-8"))
    (catch Exception _ nil)))

(defn form-decode
  "Decode the supplied www-form-urlencoded string using the specified encoding,
  or UTF-8 by default. If the encoded value is a string, a string is returned.
  If the encoded value is a map of parameters, a map is returned."
  [^String encoded & [encoding]]
  (if-not (.contains encoded "=")
    (form-decode-str encoded encoding)
    (reduce
     (fn [m param]
       (if-let [[k v] (str/split param #"=" 2)]
         (assoc-conj m (form-decode-str k encoding) (form-decode-str v encoding))
         m))
     {}
     (str/split encoded #"&"))))
```

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -22,7 +22,7 @@
             [waiter.util.client-tools :refer :all]
             [waiter.util.date-utils :as du])
   (:import (java.io ByteArrayInputStream)
-           (java.net HttpURLConnection)))
+           (java.net URLEncoder)))
 
 (deftest ^:parallel ^:integration-fast test-basic-functionality
   (testing-using-waiter-url
@@ -52,30 +52,26 @@
           (is (= "0" (get-in body-json ["headers" "content-length"])) (str body))
           (is (= "text/plain" (get-in body-json ["headers" "accept"])) (str body))))
 
-      (let [bad-query-value "~`!@$%^&*()_-+={}[]|:;'<>,.?"]
-        (testing "query-string with special characters"
-          (log/info "Basic test for query-string with special characters")
-          (let [{:keys [body] :as response} (make-kitchen-request
-                                              waiter-url
-                                              (assoc request-headers :accept "application/json")
-                                              :path "/request-info" ;; library does not allow us to use request-info?q=a|b
-                                              :query-params {"q" bad-query-value})]
-            (assert-response-status response 400)
-            (is (str/includes? body "Illegal character in query")))
+      (testing "query-string with special characters"
+        (log/info "Basic test for query-string with special characters")
+        (let [bad-query-string "q=~`!@$%^&*()_-+={}[]|:;'<>,.?&foo=%12jhsdf"
+              {:keys [body] :as response} (make-kitchen-request
+                                            waiter-url
+                                            (assoc request-headers :accept "application/json")
+                                            :path "/request-info"
+                                            :query-string bad-query-string)]
+          (assert-response-status response 200)
+          (is (= bad-query-string (get (json/read-str body) "query-string"))))
 
-          (log/info "using HttpURLConnection and query-string with special characters")
-          (try
-            (let [request-url (str "http://" waiter-url (str "/request-info?q=" bad-query-value))
-                  token (str "^SERVICE-ID#" service-id)
-                  ^HttpURLConnection url-connection (open-url-connection request-url :get {:x-waiter-token token})
-                  input-stream (.getInputStream url-connection)]
-              (try
-                (is (= 400 (.getResponseCode url-connection)))
-                (is (str/includes? (slurp input-stream) "Illegal character in query"))
-                (finally
-                  (.close input-stream))))
-            (catch Exception ex
-              (is false (.getMessage ex))))))
+        (log/info "Basic test for query-string with encoded characters")
+        (let [bad-query-string (str "q=" (URLEncoder/encode "~`!@$%^&*()_-+={}[]|:;'<>,.?&foo=%12jhsdf"))
+              {:keys [body] :as response} (make-kitchen-request
+                                            waiter-url
+                                            (assoc request-headers :accept "application/json")
+                                            :path "/request-info"
+                                            :query-string bad-query-string)]
+          (assert-response-status response 200)
+          (is (= bad-query-string (get (json/read-str body) "query-string")))))
 
       (testing "http methods"
         (log/info "Basic test for empty body in request")
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -22,8 +22,7 @@
             [waiter.util.client-tools :refer :all]
             [waiter.util.date-utils :as du])
   (:import (java.io ByteArrayInputStream)
-           (java.net HttpURLConnection)
-           (org.eclipse.jetty.util UrlEncoded)))
+           (java.net HttpURLConnection)))
 
 (deftest ^:parallel ^:integration-fast test-basic-functionality
   (testing-using-waiter-url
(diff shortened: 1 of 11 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -53,6 +54,35 @@
           (is (= "0" (get-in body-json ["headers" "content-length"])) (str body))
           (is (= "text/plain" (get-in body-json ["headers" "accept"])) (str body))))
 
+      (testing "query-string with special characters"
+        (log/info "Basic test for query-string with special characters")
+        (let [{:keys [body] :as response} (make-kitchen-request
```
shamsimam: `jet` encodes the query parameters, so we do not see the errors from this testing clause. Hence, we need to use the raw `HttpURLConnection` version below to create the issue.

waiter/src/waiter/process_request.clj
```diff
@@ -258,15 +260,23 @@
...
+          (not (str/blank? query-string)) (str "?" query-string)))
 
 (defn inspect-for-202-async-request-response
   "Helper function that inspects the response and triggers async-request post processing."
   [{:keys [headers status] :as response} post-process-async-request-response-fn instance-request-properties
-   service-id metric-group instance endpoint request reason-map reservation-status-promise]
+   service-id metric-group instance endpoint {:keys [query-string] :as request} reason-map reservation-status-promise]
   (let [location-header (str (get headers "location"))
-        location (async-req/normalize-location-header endpoint location-header)]
+        instance-endpoint (append-query-string endpoint query-string)
```
shamsimam: Preserve old behavior by appending the query string to the endpoint.

waiter/src/waiter/process_request.clj
```diff
@@ -229,11 +230,12 @@
        :fold-chunked-response? true
        :fold-chunked-response-buffer-size output-buffer-size
        :follow-redirects? false
-       :idle-timeout idle-timeout})))
+       :idle-timeout idle-timeout
+       :query-string query-params})))
```
sradack: Will this ultimately end up encoding `"?q=a|b"` into `"?q=a%7Cb"` when sending to the backend?  We should avoid changing the request.

A different approach might be add a new key `raw-query-string` to jet's `request` parameter map and use it to override `Request.getQuery()` on the object returned by `HttpClient.newRequest`.  This could avoid the encoding happening in `Request.param`, which is ultimately what jet currently passes `query-string` to.

sradack: Do we have to implement all of those?  I have a suspicion that only getQuery is used when rendering the bytes sent across the wire.

sradack: If it does use `getURL` or `getParams` to write the request, then agreed that this approach won't work.

shamsimam: All the `newRequest` methods eventually rely on a `URI` being passed. The implementation, `org.eclipse.jetty.client.HttpRequest`, constructor also expects a `URI` and is not easily extendible.

Creating a URI ends up encoding the query string:

```
(URI. "http", nil, "/path" "a=b|c", nil)
=> #object[java.net.URI 0x3180f50d "http:/path?a=b%7Cc"]
```

I'm not sure it is worth the effort to avoid encoding the query string.

waiter/src/waiter/process_request.clj
```diff
@@ -189,51 +188,31 @@
...
-   idle-timeout output-buffer-size]
+  [^HttpClient http-client make-basic-auth-fn request-method endpoint query-string headers body app-password
+   {:keys [username principal]} idle-timeout output-buffer-size]
   (let [auth (make-basic-auth-fn endpoint "waiter" app-password)
         headers (headers/assoc-auth-headers headers username principal)
-        http-method (http-method-fn request-method)]
-    (http-method
-      http-client endpoint
+        query-params (when-not (str/blank? query-string)
+                       (ring-codec/form-decode query-string))]
```
shamsimam: `form-decode` does not work for us as it ends up parsing the query string wrong:

```
(ring-codec/form-decode "q=~`!@$%^&*()_-+={}[]|:;'<>,.?")
=> {"q" nil, "*()_- " "{}[]|:;'<>,.?"}
```

How do you guys feel about reporting `400 Bad Request`as the query string has invalid syntax?

DaoWen: I'm thinking that simply catching the URISyntaxException and returning a 400 is looking like the best option.

waiter/src/waiter/process_request.clj
```diff
@@ -258,15 +234,23 @@
...
 
 (defn inspect-for-202-async-request-response
   "Helper function that inspects the response and triggers async-request post processing."
   [{:keys [headers status] :as response} post-process-async-request-response-fn instance-request-properties
-   service-id metric-group instance endpoint request reason-map reservation-status-promise]
+   service-id metric-group instance endpoint {:keys [query-string] :as request} reason-map reservation-status-promise]
   (let [location-header (str (get headers "location"))
-        location (async-req/normalize-location-header endpoint location-header)]
+        instance-endpoint (append-query-string endpoint query-string)
+        location (async-req/normalize-location-header instance-endpoint location-header)]
```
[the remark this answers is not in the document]
shamsimam: Good catch, thanks. Resolved.

kitchen/project.clj
```diff
@@ -10,7 +10,7 @@
 ;;
 (defproject kitchen "0.1.0-SNAPSHOT"
   :dependencies [^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "waiter-patch"}}
-                 [cc.qbits/jet "0.7.10-20180124_201515-g857338f"]
```
DaoWen: Does Kitchen need our custom-patched Jet, or could it use the official JAR?

shamsimam: At the very least, we use the websocket support which has had some changes.

waiter/src/waiter/process_request.clj
```diff
@@ -189,51 +187,29 @@
     (-> (ex-info message (assoc metrics-map :status status) error)
         (utils/exception->response request))))
 
-(defn http-method-fn
```
DaoWen: You're deleting these two functions here because we should have deleted them in #304, right?

shamsimam: The net effect is the same where we are not using the function but instead calling `http/request` directly. But this deletion is due to refactoring in `process_request/make-http-request` as part of this PR.

waiter/test/waiter/process_request_test.clj
```diff
@@ -307,47 +278,47 @@
...
                "http://www.example.com:1234/query/for/status"
                {:authorization/user "test-user"}
-               {:status 202, :headers {"location" "http://www.example2.com:1234/retrieve/result/location"}}))))
+               {:status 202 :headers {"location" "http://www.example2.com:1234/retrieve/result/location"}}))))
     (testing "202-absolute-url-different-port"
       (is (= {:result :not-async}
              (execute-inspect-for-202-async-request-response
                "http://www.example.com:1234/query/for/status"
                {:authorization/user "test-user"}
-               {:status 202, :headers {"location" "http://www.example.com:5678/retrieve/result/location"}}))))))
```
DaoWen: It would be easier to review this PR (handling the query string) if the unrelated refactoring changes (e.g., removing the commas here) were in a separate PR.

waiter/integration/waiter/basic_test.clj
```diff
@@ -52,6 +52,27 @@
           (is (= "0" (get-in body-json ["headers" "content-length"])) (str body))
           (is (= "text/plain" (get-in body-json ["headers" "accept"])) (str body))))
 
+      (testing "query-string with special characters"
```
DaoWen: We need to do these same tests of the bad query strings with WebSockets.

waiter/test/waiter/process_request_test.clj
shamsimam: Sorry, will try and avoid next time.

waiter/integration/waiter/basic_test.clj
shamsimam: I checked by trying to connect to kitchen directly, the upgrade request goes through fine, but the websocket request is closed when invalid characters are present (e.g. `/hello?q=~!@$%^&*()_-+={}[]|:;'<>,.?&foo=%12jhsdf`).

sradack: Works for me

waiter/src/waiter/async_request.clj
```diff
@@ -153,7 +153,7 @@
     ;; trigger execution of monitoring system
     (letfn [(make-get-request-fn []
               (counters/inc! (metrics/service-counter service-id "request-counts" "async-monitor"))
-              (let [request-stub {:body nil, :headers {}, :request-method :get}]
+              (let [request-stub {:body nil :headers {} :query-string query-string :request-method :get}]
```
sradack: Why is the `query-string` being passed?  Are we supposed to be checking what's in the `Location` header (which can include its own query-string) from the response?  Why should the `query-string` from the original request be involved into this process?

shamsimam: We extract the query-string from the location header now. Added unit test to verify this behavior.

waiter/integration/waiter/basic_test.clj
DaoWen: I was assuming it was going to work with just the current changes, and that all we needed was the tests. If you need to do more work to get the WebSockets working, then I'm fine with delaying that for a later PR.

shamsimam: It is hard to write a websocket test with a bad query string with jet/jetty. I get the following error:
```
java.lang.IllegalArgumentException: Illegal character in query at index 38: ws://127.0.0.1:9091/websocket-auth?q=~`!@$%^&*()_-+={}[]|:;'<>,.?&foo=%12jhsdf
```

However, I tried directly using a Chrome plugin and after the Jetty upgrade I get an error console (before it used to just fail in the backend and hang in the Chrome plugin):

screen shot 2018-05-08 at 2 48 53 pm

shamsimam:
@sradack any additional feedback on this?

PR #315, opened 2018-05-07
simplify /state endpoint

shamsimam:

- introduces helper functions to simplify retrieving state
- deletes unused destructuring clauses

Make it simpler to read the current code and to add future extensions to the state endpoint.

### Before

old-state

### After

new-state

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -554,16 +554,6 @@
       (catch Exception ex
         (utils/exception->response ex request)))))
 
-(defn get-kv-store-state
-  "Outputs the kv-store state."
-  [router-id kv-store request]
-  (try
-    (-> {:router-id router-id
-         :state (kv/state kv-store)}
-        (utils/map->streaming-json-response))
-    (catch Exception ex
-      (utils/exception->response ex request))))
-
 (defn- get-function-state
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -499,146 +499,98 @@
...
-                   :local-usage @local-usage-agent
-                   :router-metrics-state (router-metrics-state-fn)
-                   :scheduler (async/<! (retrieve-scheduler-state scheduler-chan timeout-ms))
-                   :statsd (statsd/state))
+      (let [routers (-> state-chan retrieve-channel-state async/<! :routers)
+            host (get-in request [:headers "host"])
+            scheme (some-> request utils/request->scheme name)
+            make-url (fn make-url [path]
+                       (str (when scheme (str scheme "://")) host "/state/" path))]
+        (-> {:details (->> ["fallback" "interstitial" "kv-store" "leader" "local-usage" "maintainer"
```
sradack: Should we just put these keys in the root of the map, instead of creating a `details` submap?

shamsimam: I like that they get clustered together. If put in the root level, they will be separated by `router-id` and `routers`

PR #322, opened 2018-05-14
deflakes test-blacklisted-instance-not-reserved

shamsimam:

- deflakes test-blacklisted-instance-not-reserved

It fixes how we compute the `max-backoff-exponent`.

### DIFF
--- a/waiter/integration/waiter/killed_instance_test.clj
+++ b/waiter/integration/waiter/killed_instance_test.clj
@@ -109,10 +109,11 @@
           (let [new-request-instance-ids-atom (atom #{})]
             (parallelize-requests
               10 1
-              #(let [{:keys [instance-id router-id]} (make-request-fn)]
-                 (if (t/before? (t/now) (du/str-to-date (router-id->blacklist-time router-id)))
-                   (swap! new-request-instance-ids-atom conj instance-id)
-                   (log/warn "request responded after blacklist period, not including instance" instance-id))))
+              (fn []
+                (let [{:keys [instance-id router-id]} (make-request-fn)]
+                  (if (t/before? (t/now) (du/str-to-date (router-id->blacklist-time router-id)))
+                    (swap! new-request-instance-ids-atom conj instance-id)
+                    (log/warn "request responded after blacklist period, not including instance" instance-id)))))
             (log/info "new-request-instance-ids:" @new-request-instance-ids-atom)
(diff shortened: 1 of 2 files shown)


shamsimam:
@pschorf any feedback on this?

waiter/src/waiter/state.clj
```diff
@@ -539,7 +539,11 @@
    initial-state]
   (when (some nil? (vals initial-state))
     (throw (ex-info "Initial state contains nil values!" initial-state)))
-  (let [max-backoff-exponent (int (max 1 (inc (/ (Math/log max-blacklist-time-ms) (Math/log blacklist-backoff-base-time-ms)))))
+  (let [max-backoff-exponent (->> (/ (Math/log max-blacklist-time-ms) (Math/log blacklist-backoff-base-time-ms))
+                                  Math/exp
```
pschorf: It doesn't look like this was in the original code, was it a bug that it was missing?

waiter/integration/waiter/killed_instance_test.clj
```diff
@@ -65,50 +43,73 @@
...
+                        (wait-for :interval 5 :timeout 10))))
+                (keys router-id->router-url))]
+          (doseq [[router-id _] router-id->router-url]
+            (is (router-id->blacklist-time router-id)
+                (str "instance not blacklisted " {:instance-id instance-id :router-id router-id})))
+
+          (let [new-request-instance-ids-atom (atom #{})]
+            (parallelize-requests
+              10 1
+              #(let [{:keys [instance-id router-id]} (make-request-fn)]
```
pschorf: I'd rather use `fn` here, I kept looking for the `%` inside the code.

waiter/src/waiter/state.clj
shamsimam: Yes, this was a bug that got fixed.

waiter/integration/waiter/killed_instance_test.clj
shamsimam: Okay, will do.

waiter/src/waiter/state.clj
pschorf: 👍

waiter/integration/waiter/killed_instance_test.clj
shamsimam: Done.

shamsimam:
Ready for review.

PR #325, opened 2018-05-15
deflakes test-simple-async-request and test-multiple-async-requests

shamsimam:

- deflakes test-simple-async-request and test-multiple-async-requests

### DIFF
--- a/waiter/integration/waiter/async_request_integration_test.clj
+++ b/waiter/integration/waiter/async_request_integration_test.clj
@@ -14,7 +14,8 @@
 ;; limitations under the License.
 ;;
 (ns waiter.async-request-integration-test
-  (:require [clojure.string :as str]
+  (:require [clj-time.core :as t]
+            [clojure.string :as str]
             [clojure.test :refer :all]
             [clojure.tools.logging :as log]
             [plumbing.core :as pc]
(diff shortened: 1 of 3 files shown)


waiter/integration/waiter/async_request_integration_test.clj
```diff
@@ -71,83 +72,87 @@
...
-;   actual: (not (zero? 1))
-(deftest ^:parallel ^:integration-fast ^:explicit test-simple-async-request
+(deftest ^:parallel ^:integration-fast test-simple-async-request
   (testing-using-waiter-url
-    (let [request-processing-time-ms 15000
+    (let [processing-time-secs 12
+          _ (log/info "making initial request with processing time of" processing-time-secs "seconds")
           {:keys [cookies request-headers response service-id status-location]}
-          (make-async-request waiter-url request-processing-time-ms)
+          (make-async-request waiter-url (* processing-time-secs 1000))
```
sradack: Could we use the`t/*` transformations?

waiter/src/waiter/process_request.clj
```diff
@@ -384,18 +384,18 @@
     (meters/mark! (metrics/service-meter service-id "response-status-rate" (str status)))
     (counters/inc! (metrics/service-counter service-id "request-counts" "waiting-to-stream"))
     (confirm-live-connection-with-abort)
-    (let [request-abort-callback (abort-http-request-callback-factory response)]
+    (let [modified-response (inspect-for-202-async-request-response
```
sradack: What do you think of `response'`?

sradack: Did this move because of promises?  You want `:success-async` to be the value received with `@reservation-status-promise`?

shamsimam: Yes.

waiter/integration/waiter/async_request_integration_test.clj
shamsimam: Done.

waiter/src/waiter/process_request.clj
shamsimam: Used `response`'

shamsimam: I now avoid passing the `reservation-status-promise` to `inspect-for-202-async-request-response`

PR #326, opened 2018-05-15
removes streaming-timeout integration tests and adds unit tests

shamsimam:

- removes test-streaming-timeout-on-default-settings and test-streaming-timeout-with-custom-settings in favor of unit test test-stream-http-response-configure-idle-timeout


dposada:
Needs rebase

shamsimam:
Rebased.

PR #327, opened 2018-05-15
Stores token source metadata in the service description

shamsimam:

We are adding support for the tokens (and their versions) used to create a service description in this PR. The field is stored as the `source-tokens` metadata field in the service description. While computing the service description, we auto-populate this new metadata field. The metadata field also needs to be populated during the waiter consent workflow.

source-tokens

As a result, we now have a distinction between service parameters and service description (includes service parameters and service metadata). This distinction also carries forward to tokens which only specify service parameter values.

### Summary of changes
- introduces `service-parameter-keys`, `service-non-override-keys` and `service-metadata-keys`
- distinguishes between token `service-description-template` and `service-parameter-template`

**Note**: A result of these changes is the refactoring to rename `service-description-keys` to `service-parameter-keys` and (in the context of tokens) `service-description-template` to `service-parameter-template`. Let me know if you prefer these renames to be provided in two separate PRs to minify the diff of this PR.

Our long-term goal is to be able to GC outdated services quicker. This PR enables that feature by providing the token information used to create a service. As a result of this change, on-the-fly services with the same service parameters as a tokenized service will be treated as two different services.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -497,18 +497,12 @@
     (when (or (not deleted) include-deleted)
       (select-keys token-data allowed-keys))))
 
-(defn token->token-hash
-  "Retrieves the token-data and converts it into a hash."
-  [kv-store ^String token & {:keys [error-on-missing] :or {error-on-missing true}}]
-  (-> (token->token-data kv-store token token-data-keys error-on-missing false)
-      token-data->token-hash))
-
 (defn token-data->token-description
   "Retrieves the token description for the given token when the raw kv data (merged value of service
    parameters and metadata) is provided.
    The token-description consists of the following keys: :service-description-template and :token-metadata"
   [config]
-  {:service-description-template (select-keys config service-parameter-keys)
+  {:service-parameter-template (select-keys config service-parameter-keys)
    :token-metadata (select-keys config token-metadata-keys)})
 
 (defn token->token-description
@@ -517,10 +511,19 @@
   (-> (token->token-data kv-store token token-data-keys false include-deleted)
       token-data->token-description))
 
-(defn token->service-description-template
-  "Retrieves the service description template for the given token."
+(defn token->service-parameter-template
+  "Retrieves the service description template for the given token containing only the service parameters."
   [kv-store ^String token & {:keys [error-on-missing] :or {error-on-missing true}}]
   (token->token-data kv-store token service-parameter-keys error-on-missing false))
+
+(defn token->service-description-template
+  "Retrieves the service description template for the given token including the service metadata values."
+  [kv-store ^String token & {:keys [error-on-missing] :or {error-on-missing true}}]
+  (let [token-data (token->token-data kv-store token token-data-keys error-on-missing false)
+        service-parameter-template (select-keys token-data service-parameter-keys)]
+    (cond-> service-parameter-template
+            (seq service-parameter-template)
+            (assoc "source-tokens" {token (token-data->token-hash token-data)}))))
 
 (defn token->token-metadata
   "Retrieves the token metadata for the given token."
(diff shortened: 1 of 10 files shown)


sradack:
Please rebase.  

A couple of things before I dive into the details:

1. Having the "E" in the version hash (what we also use for ETag) now seems even more awkward to me.  "E" standing for "Etag" is an HTTP concept, and has bled into our service descriptions.
2. Should `source-tokens` be a vector instead of a map?  Tokens are provided in an ordered list, and using a map means we lose that order.

shamsimam:
Rebased.

> Having the "E" in the version hash (what we also use for ETag) now seems even more awkward to me. "E" standing for "Etag" is an HTTP concept, and has bled into our service descriptions.

The `E-` is a prefix we use to quickly identify token hashes, since it is an internal implementation detail we can change it whenever we like.

> Should source-tokens be a vector instead of a map? Tokens are provided in an ordered list, and using a map means we lose that order.

I changed it to a vector.

waiter/src/waiter/service_description.clj
```diff
@@ -98,23 +98,30 @@
...
+(def ^:const service-non-override-keys
+  #{"allowed-params" "backend-proto" "cmd" "cmd-type" "cpus" "env" "health-check-url" "mem" "metadata"
+    "metric-group" "name" "permitted-user" "ports" "run-as-user" "version"})
+
+; keys used as parameters in the service description
+(def ^:const service-parameter-keys
+  (set/union service-override-keys service-non-override-keys))
+
+; keys allowed in service description metadata, these need to be distinct from service parameter keys
+(def ^:const service-metadata-keys #{"source-tokens"})
```
sradack: How are these different from the "other" metadata keys?

waiter/src/waiter/service_description.clj
```diff
@@ -498,10 +506,24 @@
   (-> (token->token-data kv-store token token-data-keys false include-deleted)
       token-data->token-description))
 
+(defn token->service-parameter-template
+  "Retrieves the service description template for the given token containing only the service parameters."
```
sradack: What's the difference between a service description template and a service parameter template?

waiter/src/waiter/service_description.clj
```diff
@@ -791,14 +817,9 @@
...
   [kv-store service-id service-description-defaults metric-group-mappings &
-   {:keys [refresh nil-on-missing? effective?] :or {refresh false nil-on-missing? true effective? true}}]
-  (let [core-service-description (fetch-core
-                                   kv-store
-                                   service-id
-                                   :refresh refresh
-                                   :nil-on-missing? nil-on-missing?)]
-    (cond-> core-service-description
-            effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id))))
+   {:keys [effective? nil-on-missing? refresh] :or {effective? true nil-on-missing? true refresh false}}]
```
sradack: Please move this to a separate PR.

waiter/src/waiter/token.clj
```diff
@@ -429,11 +429,12 @@
                                    (not (get existing-token-metadata "deleted")))
                             "updated "
                             "created ")]
-        (utils/map->json-response {:message (str "Successfully " creation-mode token)
-                                   :service-description new-service-description-template}
-                                  :headers {"etag" (-> {:service-description-template new-service-description-template
-                                                        :token-metadata new-token-metadata}
-                                                       token-description->token-hash)})))))
+        (utils/map->json-response
```
sradack: Please move this to a separate PR.

waiter/src/waiter/service_description.clj
```diff
@@ -746,7 +772,7 @@
         (throw (ex-info (utils/message :cannot-identify-service)
                         (error-message-map-fn passthrough-headers waiter-headers))))
       (sling/try+
-        (let [{:keys [service-id service-description core-service-description]}
```
sradack: Please move this to a separate PR.

waiter/src/waiter/service_description.clj
shamsimam: Currently all the fields we store in the service description are the service parameters, `surce-tokens` will be the first of its kind - a metadata entry that will be system generated, that we will not allow end users to configure.

waiter/src/waiter/service_description.clj
shamsimam: Service description == service parameters + system generated metadata

waiter/src/waiter/service_description.clj

waiter/src/waiter/service_description.clj

shamsimam:
Rebased.

PR #331, opened 2018-05-17
WIP: extends can-run-as support for permitted-user checks

shamsimam:

- extends can-run-as support for permitted-user checks

Makes the permitted-user check equivalent to run-as-user checks by using the `can-run-as` support for the entitlement manager.

### DIFF
--- a/waiter/src/waiter/authorization.clj
+++ b/waiter/src/waiter/authorization.clj
@@ -61,3 +61,9 @@
   [entitlement-manager auth-user run-as-user]
   (and auth-user run-as-user
        (authorized? entitlement-manager auth-user :run-as {:resource-type :credential, :user run-as-user})))
+
+(defn member-of?
+  "Helper function that checks the whether the auth-user is a member of the provided group."
+  [entitlement-manager auth-user group]
+  (and auth-user group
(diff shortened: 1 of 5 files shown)


shamsimam:
> This may have unintended consequences, e.g. perhaps a service owner would prefer a request to fail unless the request is made by the actual role account, instead of also allowing it from users that can become that role account. 

We ignore this concern while making `run-as-user` checks e.g. during token registration or while launching services - mainly for convenience of the end user. This PR causes the `permitted-user` check to have the same behavior.

> What is our motivation for this change?

Apart from bringing the check closer to how `run-as-user` checks works, we have at least one user who has requested for a more robust permitted user check, e.g. for groups or role accounts.

pschorf:
> more than one users who have requested for a more robust permitted user check, e.g. for groups or role accounts.

Maybe we should add a new field with support for more robust checks instead of changing the semantics of the current one?

shamsimam:
@pschorf  That would defeat the purpose of this PR which intends to bring how we treat `permitted-user` closer to how we treat `run-as-user`. We have made similar changes for `run-as-user` before with end user convenience in mind.

shamsimam:
We decided to not introduce additional burden of authorizing requests and delegate the responsibility to the backend instance.

PR #332, opened 2018-05-17
assigns a non-'other' metric group to a service (when possible)

shamsimam:

- assigns a non-'other' metric group to a service using the token name

We would like to more accurately identify the different services running on Waiter without defaulting them to `other`. With this PR we are going to use the first part of the token before a dot character as the default metric group name when it exists and satisfies the metric group schema.

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -134,6 +134,9 @@
           (let [response (make-request 24000)]
             (assert-response-status response 200))))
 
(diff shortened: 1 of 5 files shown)


shamsimam:
@dpo added integration tests :)

dpo:
Can I ask you guys to please check who you ping on this repo? It's not the first time I'm getting pinged instead of one of your developers. Thank you.

shamsimam:
My apologies to you dpo, I will try and be more careful henceforth.

PR #338, opened 2018-05-31
Scale down in the presence of pending deployments

shamsimam:

- adds support for forcefully scaling a service by deleting any outstanding deployments
- adds sync-deployment-maintainer to correct deployment state of services that may go out of sync due to failed scale operation that canceled a previous deployment

We would like to scale-down without waiting for previous scale-up deployments to complete. Avoiding the delay reduces the wasteful use of resources due to spin-up and almost immediate killing of instances.

### Example log from the deployment syncer:

```
2018-06-18 14:20:02,497 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529349972846] 0 service(s) have out-of-sync deployments: #{}
2018-06-18 14:20:02,497 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529349972846] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:20:17,517 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529349987836] 1 service(s) have out-of-sync deployments: #{s1475611-8391
5fcbf5fd559791b851e42c7e1f29}
2018-06-18 14:20:17,517 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529349987836] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:20:32,516 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350002838] 1 service(s) have out-of-sync deployments: #{s1475611-8391
5fcbf5fd559791b851e42c7e1f29}
2018-06-18 14:20:32,516 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350002838] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:20:47,559 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350017839] 2 service(s) have out-of-sync deployments: #{s1475611-8391
5fcbf5fd559791b851e42c7e1f29 s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:20:47,560 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350017839] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:21:02,537 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350032838] 2 service(s) have out-of-sync deployments: #{s1475611-8391
5fcbf5fd559791b851e42c7e1f29 s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:21:02,537 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350032838] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:21:17,502 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350062838] 3 service(s) have out-of-sync deployments: #{s1475611-83915fcbf5fd559791b851e42c7e1f29 s2354226-58faf6368292d9a758430337ea4e7c9f s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:21:17,502 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350062838] 1 service(s) qualify for sync deployments: #{s1475611-83915fcbf5fd559791b851e42c7e1f29}
2018-06-18 14:21:17,504 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350062838] triggering sync deployment {:service-id s1475611-83915fcbf5fd559791b851e42c7e1f29, :task-data {:instances-requested 12, :instances-scheduled 8}}
2018-06-18 14:21:32,498 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350077841] 2 service(s) have out-of-sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:21:32,498 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350077841] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:21:47,510 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350092837] 2 service(s) have out-of-sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:21:47,510 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350092837] 1 service(s) qualify for sync deployments: #{s2365382-db19481494bb9be94798510ad9d99fd0}
2018-06-18 14:21:47,510 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350092837] triggering sync deployment {:service-id s2365382-db19481494bb9be94798510ad9d99fd0, :task-data {:instances-requested 11, :instances-scheduled 8}}
2018-06-18 14:22:02,499 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350107841] 1 service(s) have out-of-sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f}
2018-06-18 14:22:02,499 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350107841] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:22:17,498 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350122839] 1 service(s) have out-of-sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f}
2018-06-18 14:22:17,498 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350122839] 0 service(s) qualify for sync deployments: #{}
2018-06-18 14:22:32,515 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350137838] 1 service(s) have out-of-sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f}
2018-06-18 14:22:32,515 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350137838] 1 service(s) qualify for sync deployments: #{s2354226-58faf6368292d9a758430337ea4e7c9f}
2018-06-18 14:22:32,515 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350137838] triggering sync deployment {:service-id s2354226-58faf6368292d9a758430337ea4e7c9f, :task-data {:instances-requested 10, :instances-scheduled 8}}
2018-06-18 14:22:47,497 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350152837] 0 service(s) have out-of-sync deployments: #{}
2018-06-18 14:22:47,497 INFO  [clj-agent-send-off-pool-1] - [CID=sync-deployment.1529350152837] 0 service(s) qualify for sync deployments: #{}
```

### Example screenshots from Marathon:

Deployment in flight:
screen shot 2018-06-11 at 3 09 09 pm
screen shot 2018-06-11 at 3 09 19 pm

Deployment forcefully cancelled so that instance and task counts no longer match:
screen shot 2018-06-11 at 3 09 48 pm

Deployment syncer _patches_ the counts so that they now agree:
screen shot 2018-06-11 at 3 10 35 pm

### Example screenshot for service state

service-out-of-sync-state

### DIFF
--- a/waiter/src/waiter/mesos/marathon.clj
+++ b/waiter/src/waiter/mesos/marathon.clj
@@ -60,9 +60,9 @@
 
 (defn get-apps
   "List all running apps including running and failed tasks."
-  [{:keys [http-client marathon-url spnego-auth]}]
+  [{:keys [http-client marathon-url spnego-auth]} query-params]
   (http-utils/http-request http-client (str marathon-url "/v2/apps")
-                           :query-string {"embed" ["apps.lastTaskFailure" "apps.tasks"]}
+                           :query-string query-params
                            :request-method :get
                            :spnego-auth spnego-auth))
 
--- a/waiter/src/waiter/mesos/marathon.clj
+++ b/waiter/src/waiter/mesos/marathon.clj
@@ -42,6 +42,15 @@
                            :request-method :delete
                            :spnego-auth spnego-auth))
 
+(defn delete-deployment
+  "Cancel the deployment with deployment-id.
+   No rollback deployment is created to revert the changes of deployment."
+  [{:keys [http-client marathon-url spnego-auth]} deployment-id]
+  (http-utils/http-request http-client (str marathon-url "/v2/deployments/" deployment-id)
+                           :query-string {"force" true}
+                           :request-method :delete
+                           :spnego-auth spnego-auth))
+
 (defn get-app
   "List the app specified by app-id."
   [{:keys [http-client marathon-url spnego-auth]} app-id]
--- a/waiter/src/waiter/scaling.clj
+++ b/waiter/src/waiter/scaling.clj
@@ -87,18 +87,20 @@
     {:executor-multiplexer-chan executor-multiplexer-chan
      :query-chan query-chan}))
 
-(defn- execute-scale-up-request
-  "Helper function to scale-up instances of a service."
-  [scheduler service-id scale-to-instances]
-  (try
-    (scheduler/suppress-transient-server-exceptions
-      "autoscaler"
-      (log/info "scaling service to" scale-to-instances "instances")
-      (scheduler/scale-app scheduler service-id scale-to-instances)
-      (counters/inc! (metrics/service-counter service-id "scaling" "scale-up" "success")))
-    (catch Exception e
-      (counters/inc! (metrics/service-counter service-id "scaling" "scale-up" "fail"))
-      (log/warn e "unexpected error when trying to scale" service-id "to" scale-to-instances "instances"))))
+(defn- execute-scale-service-request
+  "Helper function to scale instances of a service.
+   The force? flag can be used to detmerine whether we will make a best effort or a forced scale operation."
+  [scheduler service-id scale-to-instances force?]
+  (let [mode (if force? "scale-force" "scale-up")]
+    (try
+      (scheduler/suppress-transient-server-exceptions
+        "autoscaler"
+        (log/info mode "service to" scale-to-instances "instances")
+        (scheduler/scale-app scheduler service-id scale-to-instances force?)
+        (counters/inc! (metrics/service-counter service-id "scaling" mode "success")))
+      (catch Exception e
+        (counters/inc! (metrics/service-counter service-id "scaling" mode "fail"))
+        (log/warn e "unexpected error when trying to scale" service-id "to" scale-to-instances "instances")))))
 
 (defn- execute-scale-down-request
   "Helper function to scale-down instances of a service.
@@ -227,7 +229,7 @@
                             (do
                               (log/info "allowing previous scale operation to complete before scaling up again")
                               (counters/inc! (metrics/service-counter service-id "scaling" "scale-up" "ignore")))
-                            (execute-scale-up-request scheduler service-id scale-to-instances))
+                            (execute-scale-service-request scheduler service-id scale-to-instances false))
                           executor-state)
 
                         (pos? num-instances-to-kill)
@@ -243,9 +245,21 @@
                               (assoc executor-state :last-scale-down-time (t/now))
                               executor-state)
                             (do
-                              (log/debug "skipping scale-down as" inter-kill-request-wait-time-ms "ms has not elapsed since last scale down operation")
+                              (log/debug "skipping scale-down as" inter-kill-request-wait-time-ms
+                                         "ms has not elapsed since last scale down operation")
                               (counters/inc! (metrics/service-counter service-id "scaling" "scale-down" "ignore"))
                               executor-state)))
+
+                        (and scale-to-instances
+                             task-count
+                             (neg? scale-amount)
+                             (neg? (- task-count scale-to-instances)))
+                        (do
+                          (log/info "potential overshoot detected, triggering scale-force for service"
+                                    {:scale-to-instances scale-to-instances :task-count task-count})
+                          (counters/inc! (metrics/service-counter service-id "scaling" "scale-force" "total"))
+                          (execute-scale-service-request scheduler service-id scale-to-instances true)
+                          executor-state)
 
                         :else
                         (do
--- a/waiter/src/waiter/scaling.clj
+++ b/waiter/src/waiter/scaling.clj
@@ -250,10 +250,7 @@
                               (counters/inc! (metrics/service-counter service-id "scaling" "scale-down" "ignore"))
                               executor-state)))
 
-                        (and scale-to-instances
-                             task-count
-                             (neg? scale-amount)
-                             (neg? (- task-count scale-to-instances)))
+                        (and (neg? scale-amount) (< task-count scale-to-instances))
                         (do
                           (log/info "potential overshoot detected, triggering scale-force for service"
                                     {:scale-to-instances scale-to-instances :task-count task-count})
--- a/waiter/src/waiter/scheduler/marathon.clj
+++ b/waiter/src/waiter/scheduler/marathon.clj
@@ -470,39 +470,46 @@
     (catch Exception e
       (log/error e "unable to sync marathon deployment for" service-id))))
 
+(defn- compute-out-of-sync-state
+  "A helper function that computes the new out-of-sync-state and the trigger service-ids."
+  [marathon-scheduler current-time trigger-timeout]
+  (let [service-id->out-of-sync-state (retrieve-service-id->out-of-sync-state marathon-scheduler)
+        service-id->instance-count-map (retrieve-out-of-sync-apps marathon-scheduler)
+        out-of-sync-service-ids (-> service-id->instance-count-map keys set)
+        trigger-time (t/minus current-time trigger-timeout)
+        trigger-service-ids (->> service-id->out-of-sync-state
+                                 (filter (fn out-of-sync-trigger-predicate
+                                           [[service-id {:keys [data last-modified-time]}]]
+                                           (and (= data (service-id->instance-count-map service-id))
+                                                (t/before? last-modified-time trigger-time))))
+                                 (map first)
+                                 set)]
+    (log/info (count out-of-sync-service-ids) "service(s) have out-of-sync deployments:" out-of-sync-service-ids)
+    (log/info (count trigger-service-ids) "service(s) qualify for sync deployments:" trigger-service-ids)
+    {:service-id->instance-count-map service-id->instance-count-map
+     :service-id->out-of-sync-state (->> (set/difference out-of-sync-service-ids trigger-service-ids)
+                                         (pc/map-from-keys
+                                           (fn compute-out-of-sync-state [service-id]
+                                             (let [prev-state (service-id->out-of-sync-state service-id)
+                                                   prev-data (:data prev-state)
+                                                   curr-data (service-id->instance-count-map service-id)]
+                                               (if (= prev-data curr-data)
+                                                 prev-state
+                                                 {:data curr-data :last-modified-time current-time})))))
+     :trigger-service-ids trigger-service-ids}))
+
 (defn- process-out-of-sync-services!
   "Determines the out-of-sync services and triggers sync deployments on those services if needed.
    As a side-effect, it updates the service-id->out-of-sync-state in the scheduler."
-  [marathon-scheduler trigger-timeout]
+  [marathon-scheduler current-time  trigger-timeout]
   (try
-    (let [prev-service-id->out-of-sync-state (retrieve-service-id->out-of-sync-state marathon-scheduler)
-          service-id->instance-count-map (retrieve-out-of-sync-apps marathon-scheduler)
-          out-of-sync-service-ids (-> service-id->instance-count-map keys set)
-          current-time (t/now)
-          trigger-time (t/minus current-time trigger-timeout)
-          trigger-service-ids (->> prev-service-id->out-of-sync-state
-                                   (filter (fn out-of-sync-trigger-predicate
-                                             [[service-id {:keys [data last-modified-time]}]]
-                                             (and (= data (service-id->instance-count-map service-id))
-                                                  (t/before? last-modified-time trigger-time))))
-                                   (map first)
-                                   set)]
-      (log/info (count out-of-sync-service-ids) "service(s) have out-of-sync deployments:" out-of-sync-service-ids)
-      (log/info (count trigger-service-ids) "service(s) qualify for sync deployments:" trigger-service-ids)
+    (let [{:keys [service-id->instance-count-map service-id->out-of-sync-state trigger-service-ids]}
+          (compute-out-of-sync-state marathon-scheduler current-time trigger-timeout)]
       (doseq [service-id trigger-service-ids]
         (->> (service-id->instance-count-map service-id)
              (trigger-sync-deployment! marathon-scheduler service-id)))
       (log/info "updating local service-id->out-of-sync-state")
-      (->> (set/difference out-of-sync-service-ids trigger-service-ids)
-           (pc/map-from-keys
-             (fn compute-out-of-sync-state [service-id]
-               (let [prev-state (prev-service-id->out-of-sync-state service-id)
-                     prev-data (:data prev-state)
-                     curr-data (service-id->instance-count-map service-id)]
-                 (if (= prev-data curr-data)
-                   prev-state
-                   {:data curr-data :last-modified-time current-time}))))
-           (reset-service-id->out-of-sync-state! marathon-scheduler)))
+      (reset-service-id->out-of-sync-state! marathon-scheduler service-id->out-of-sync-state))
     (catch Exception e
       (log/error e "unable to process out-of-sync services"))))
 
@@ -534,14 +541,14 @@
             (recur current-state))
 
           timeout-chan
-          (do
+          (let [current-time (t/now)]
             (if-not (leader?-fn)
               (reset-service-id->out-of-sync-state! marathon-scheduler {})
               (cid/with-correlation-id
                 (str "sync-deployment." iteration)
-                (process-out-of-sync-services! marathon-scheduler trigger-timeout)))
+                (process-out-of-sync-services! marathon-scheduler current-time trigger-timeout)))
             (-> current-state
-                (assoc :iteration (inc iteration) :last-update-time (t/now))
+                (assoc :iteration (inc iteration) :last-update-time current-time)
                 recur)))))
     {:exit-chan exit-chan
      :query-chan query-chan}))
--- a/waiter/src/waiter/scheduler/marathon.clj
+++ b/waiter/src/waiter/scheduler/marathon.clj
@@ -521,7 +521,7 @@
              :service-id->out-of-sync-state {}))
          (send service-id->out-of-sync-state-store))))
 
-(defn start-sync-deployment-maintainer
+(defn- start-sync-deployment-maintainer
   "Launches the sync-deployment maintainer which triggers new deployments for services which have a mismatch
    in the counts for requested and scheduled instances."
   [leader?-fn service-id->out-of-sync-state-store marathon-scheduler {:keys [interval-ms timeout-cycles]}]
--- a/waiter/src/waiter/scheduler/marathon.clj
+++ b/waiter/src/waiter/scheduler/marathon.clj
@@ -507,30 +507,36 @@
       (assoc current-state :last-update-time current-time))))
 
 (defn trigger-sync-deployment-maintainer-iteration
-  "Triggers an agent send-off to run an individual iteration of the sync-deployment-maintainer."
-  [leader?-fn service-id->out-of-sync-state-store marathon-scheduler trigger-timeout]
+  "Runs an individual iteration of the sync-deployment-maintainer.
+   On non-leader routers it returns the reset state.
+   On the leader it processes out-of-sync services and returns the updated state."
+  [leader?-fn service-id->out-of-sync-state marathon-scheduler trigger-timeout]
   (if (leader?-fn)
     (let [current-time (t/now)]
       (cid/with-correlation-id
         (str "sync-deployment." (tc/to-long current-time))
-        (send-off service-id->out-of-sync-state-store
-                  process-out-of-sync-services! marathon-scheduler current-time trigger-timeout)))
-    (->> (fn reset-state-fn [current-state]
-           (assoc current-state
-             :last-update-time (t/now)
-             :service-id->out-of-sync-state {}))
-         (send service-id->out-of-sync-state-store))))
+        (process-out-of-sync-services! service-id->out-of-sync-state marathon-scheduler current-time trigger-timeout)))
+    (assoc service-id->out-of-sync-state
+      :last-update-time (t/now)
+      :service-id->out-of-sync-state {})))
 
 (defn- start-sync-deployment-maintainer
   "Launches the sync-deployment maintainer which triggers new deployments for services which have a mismatch
    in the counts for requested and scheduled instances."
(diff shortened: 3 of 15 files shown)


waiter/src/waiter/scaling.clj
```diff
@@ -243,10 +245,22 @@
                               (assoc executor-state :last-scale-down-time (t/now))
                               executor-state)
                             (do
-                              (log/debug "skipping scale-down as" inter-kill-request-wait-time-ms "ms has not elapsed since last scale down operation")
+                              (log/debug "skipping scale-down as" inter-kill-request-wait-time-ms
+                                         "ms has not elapsed since last scale down operation")
                               (counters/inc! (metrics/service-counter service-id "scaling" "scale-down" "ignore"))
                               executor-state)))
 
+                        (and scale-to-instances
```
sradack: If we  put `(neg? scale-amount)` first, it's impossible for `scale-to-instances` and `task-count` to be `nil` here and we can remove them from the condition.

waiter/src/waiter/scaling.clj
```diff
@@ -243,10 +245,22 @@
...
+                             task-count
+                             (neg? scale-amount)
+                             (neg? (- task-count scale-to-instances)))
```
sradack: Would this be more simply stated as `(< task-count scale-to-instances)`?

waiter/src/waiter/scaling.clj
shamsimam: Done.

waiter/src/waiter/scaling.clj
shamsimam: Done.

sradack:
Please rebase.

shamsimam:
@sradack Rebased.

waiter/src/waiter/mesos/marathon.clj
```diff
@@ -42,6 +42,15 @@
                             :request-method :delete
                             :spnego-auth spnego-auth))
 
+(defn delete-deployment
+  "Cancel the deployment with deployment-id.
+   No rollback deployment is created to revert the changes of deployment."
+  [{:keys [http-client marathon-url spnego-auth]} deployment-id]
+  (http-utils/http-request http-client (str marathon-url "/v2/deployments/" deployment-id)
+                           :query-string {"force" true}
```
sradack: Do we expect passing `force` here to be better than forgoing doing anything with deployments and instead passing `force` on the `update-app` call?  If so, why is that the case?

shamsimam: We want to control that an arbitrary instance does not get killed due to our update. If we did the force on the `update-app` its possible there is a race between the number of instances we knew about (the actual value can be higher during scale-up) and the number of instances we set the new value to during the update call. As a result, Marathon would then decide to arbitrarily kill an instance. We could have learnt about this instance from the scheduler syncer and sent it a request while Marathon chose to kill the instance.

sradack: When manually killing a deployment after a change in the number of instances (scale up operation), I observed that the `instances` field was unaffected.  Given that observation, is it possible to get in a bad state, whether we're scaling up or scaling down?  Have we observed the behavior of force deleting a deployment in both cases?  How does our current code + this PR handle those cases gracefully?

shamsimam: I reasoned the following scenario is possible if we do not explicitly force delete the outstanding deployment:
1. Scale-Up: Waiter asks a service to scale up to 10 instances.
2. Scale-Down: Waiter later decides to scale the service to 5 instances.
3. Scale-Down: We query marathon and find that the previous deployment is still in effect.
4. Scale-Down: We query the service state and hear back that it has 4 instances.
5. Marathon manages to scale the service to 6 instances after our previous query.
6. Syncer: Syncer hears about these 6 instances and assigns them to the router
7. Request: New client requests are assigned to these instances
8. Scale-Down: We decide to change the number of instances to 5 (we had last seen this value to be 4 in step-4, the data race we should try and protect against) using force mode
9. Marathon in actuality has already scaled the service to 6 instances so decides to arbitrarily kill an instance (that instance is potentially processing a request and hence causes that request to fail).

Do you agree that this scenario can happen?

This PR avoids the data race mentioned above by first force deleting the deployment to freeze the instances being created for the service. Hence when we subsequently query the number of instances (step-4) and when we set the new number of instances (step-8), we do not expect the number to change and avoid the data race. We would update the deployment to 6 instances instead of 5 instances since we take a max.

shamsimam: We spoke offline, the issue @sradack is raising is what happens if the call to `update-app` in `scale-app` fails after we have already deleted the current deployment. We verified on Marathon that deleting an existing deployment does not affect the `task-count` and causes the service to freeze in a state where the task counts (requested instances) does not match with scheduled instance count.

An approach to tackle this scenario (while still avoiding the data race issue I mentioned above) is:
1. Trigger deployment update on every scaling cycle if the requested and scheduled instance counts do not match up. This request will be ignored if a deployment is already in flight.
2. Run scaling for a service less frequently (than the current 1-second interval). E.g. if a service has been scaled this iteration, wait at least 60 seconds before attempting to scale again. 

In effect, we want to allow time for a previous deployment to complete and avoid making superfluous deployment update calls. The duplicate calls are necessary since we do not know if we are in the frozen state and they allow us to get out of the frozen state.

@sradack what do you think of this approach?

sradack: Since this seems localized to Marathon, it should be a Marathon-specific solution.  What do you think about setting up a loop in our Marathon code instead of inside the scaling logic?  Marathon supports embedding deployment information in the `get-app` response.  We could use that to search for the candidate apps that need to be fixed.

shamsimam: Sounds good.

shamsimam: Added support for `sync-deployment-maintainer`

shamsimam:
@sradack ready for review

waiter/src/waiter/mesos/marathon.clj
shamsimam: A race is still possible in setting the instance value with the call to `scale-app` from the auto-scaler and the sync-deployment go-blocks (it is expected to be rare). We can go the extra step and avoid the race altogether by introducing a service-level (implemented by using a hash to index into a bucket of locks) lock for scale read/write calls. Let me know what you think?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -408,14 +429,90 @@
...
+
+(defn- trigger-sync-deployments
+  "Triggers deployments for apps which are out of sync.
+   It returns the ids of the apps that were out of sync."
+  [marathon-scheduler]
+  (try
+    (let [apps (get-apps-with-deployments marathon-scheduler)
+          out-of-sync-apps (filter is-out-of-sync? apps)]
+      (log/info "found" (count out-of-sync-apps) "service(s) that require sync deployments")
+      (when (seq out-of-sync-apps)
```
sradack: Nitpick: We could avoid the `when`

waiter/test/waiter/mesos/marathon_test.clj
```diff
@@ -14,25 +14,30 @@
...
 (ns waiter.mesos.marathon-test
-  (:require [clojure.string :as str]
-            [clojure.test :refer :all]
+  (:require [clojure.test :refer :all]
             [waiter.mesos.marathon :refer :all]
             [waiter.util.http-utils :as http-utils]))
 
+(def ^:private http-client (Object.))
+
+(def ^:private marathon-url "http://www.marathon.com:1234")
```
sradack: This is a valid URL.  It would be best to avoid real URLs to prevent accidental traffic.

waiter/src/waiter/mesos/marathon.clj
sradack: Given the rarity of the case and the eventuality that we'll end up in a working state, I'm not sure implementing locking is worth the extra complexity.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -409,25 +565,36 @@
...
+        service-id->out-of-sync-state-store (atom {})
+        retrieve-framework-id-fn (memo/ttl #(retrieve-framework-id marathon-api) :ttl/threshold framework-id-ttl)
+        sync-deployment-maintainer-atom (atom nil)
+        marathon-scheduler (->MarathonScheduler
+                             marathon-api mesos-api retrieve-framework-id-fn home-path-prefix
+                             service-id->failed-instances-transient-store service-id->last-force-kill-store
+                             service-id->out-of-sync-state-store service-id->service-description-fn
+                             force-kill-after-ms is-waiter-app?-fn sync-deployment-maintainer-atom)
+        sync-deployment-maintainer (start-sync-deployment-maintainer leader?-fn marathon-scheduler sync-deployment)]
+    (reset! sync-deployment-maintainer-atom sync-deployment-maintainer)
```
shamsimam: We need this to be able to return the `sync-deployment-maintainer` channels to the caller via the `scheduler` instance.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -409,25 +565,36 @@
...
   (let [http-client (http-utils/http-client-factory http-options)
         marathon-api (marathon/api-factory http-client http-options url)
         mesos-api (mesos/api-factory http-client http-options mesos-slave-port slave-directory)
         service-id->failed-instances-transient-store (atom {})
         service-id->last-force-kill-store (atom {})
-        retrieve-framework-id-fn (memo/ttl #(retrieve-framework-id marathon-api) :ttl/threshold framework-id-ttl)]
-    (->MarathonScheduler marathon-api mesos-api retrieve-framework-id-fn home-path-prefix
-                         service-id->failed-instances-transient-store service-id->last-force-kill-store
-                         service-id->service-description-fn force-kill-after-ms is-waiter-app?-fn)))
+        service-id->out-of-sync-state-store (atom {})
```
shamsimam: Stores the `out-of-sync` state so that it can be queried via our endpoints.

shamsimam:
@sradack ready for another round of feedback.

Since the last review, I have added a loop to remember which services are out-of-sync and only trigger the 'fix' deployment after some timeout.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
+    (scheduler/scale-app marathon-scheduler service-id instances-scheduled false)
+    (catch Exception e
+      (log/error e "unable to sync marathon deployment for" service-id))))
+
+(defn- process-out-of-sync-services!
+  "Determines the out-of-sync services and triggers sync deployments on those services if needed.
+   As a side-effect, it updates the service-id->out-of-sync-state in the scheduler."
+  [marathon-scheduler trigger-timeout]
+  (try
+    (let [prev-service-id->out-of-sync-state (retrieve-service-id->out-of-sync-state marathon-scheduler)
```
sradack: Any reason to not name this simply `service-id->out-of-sync-state`?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
+          service-id->instance-count-map (retrieve-out-of-sync-apps marathon-scheduler)
+          out-of-sync-service-ids (-> service-id->instance-count-map keys set)
+          current-time (t/now)
```
sradack: What not use `last-update-time` from ` sync-deployment-maintainer`?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -276,19 +292,20 @@
 
 (defrecord MarathonScheduler [marathon-api mesos-api retrieve-framework-id-fn
                               home-path-prefix service-id->failed-instances-transient-store
-                              service-id->kill-info-store service-id->service-description
-                              force-kill-after-ms is-waiter-app?-fn]
+                              service-id->kill-info-store service-id->out-of-sync-state-store
```
sradack: Why doesn't `service-id->out-of-sync-state-store` belong in ` sync-deployment-maintainer`?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
+(defn- trigger-sync-deployment!
+  "Triggers deployment for out-of-sync service."
+  [marathon-scheduler service-id {:keys [instances-scheduled] :as task-data}]
+  (try
+    (log/info "triggering sync deployment" {:service-id service-id :task-data task-data})
+    (scheduler/scale-app marathon-scheduler service-id instances-scheduled false)
+    (catch Exception e
+      (log/error e "unable to sync marathon deployment for" service-id))))
+
+(defn- process-out-of-sync-services!
```
sradack: This function could be split into 2 parts:

f(out-of-sync-state) -> (out-of-sync-state', trigger-service-ids)
f!(trigger-service-ids)

I believe it makes sense to do so.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
+  [marathon-scheduler trigger-timeout]
+  (try
+    (let [prev-service-id->out-of-sync-state (retrieve-service-id->out-of-sync-state marathon-scheduler)
+          service-id->instance-count-map (retrieve-out-of-sync-apps marathon-scheduler)
+          out-of-sync-service-ids (-> service-id->instance-count-map keys set)
+          current-time (t/now)
+          trigger-time (t/minus current-time trigger-timeout)
+          trigger-service-ids (->> prev-service-id->out-of-sync-state
+                                   (filter (fn out-of-sync-trigger-predicate
+                                             [[service-id {:keys [data last-modified-time]}]]
```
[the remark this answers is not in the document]
sradack: Disregard.

shamsimam:
@sradack ready for another round of feedback

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
+  "Retrieves the apps with the deployment info embedded."
+  [{:keys [marathon-api is-waiter-app?-fn]}]
+  (get-apps marathon-api is-waiter-app?-fn {"embed" ["apps.deployments" "apps.tasks"]}))
+
+(defn retrieve-service-id->out-of-sync-state
+  "Retrieves the service-id->out-of-sync-state."
+  [{:keys [service-id->out-of-sync-state-store]}]
+  (deref service-id->out-of-sync-state-store))
+
+(defn- reset-service-id->out-of-sync-state!
```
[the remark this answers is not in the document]
DaoWen: If `process-out-of-sync-services!` returned the new value rather than doing the `reset!` inline, then `reset-service-id->out-of-sync-state!` would similarly be called in only one function, and it would probably be simpler to understand the inlined `reset!` rather than scrolling around to find this separate definition.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,143 @@
...
-     :service-id->kill-info-store @service-id->kill-info-store}))
+     :service-id->kill-info-store @service-id->kill-info-store
+     :service-id->out-of-sync-state-store @service-id->out-of-sync-state-store}))
+
+(defn- get-apps-with-deployments
+  "Retrieves the apps with the deployment info embedded."
+  [{:keys [marathon-api is-waiter-app?-fn]}]
+  (get-apps marathon-api is-waiter-app?-fn {"embed" ["apps.deployments" "apps.tasks"]}))
+
+(defn retrieve-service-id->out-of-sync-state
```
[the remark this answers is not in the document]
DaoWen: I'd agree with Steven for this function. This function is only called in once place. I think it would make more sense to destructure `{:keys [service-id->out-of-sync-state-store] :as marathon-scheduler}` in that function, and replace the current `(retrieve-service-id->out-of-sync-state marathon-scheduler)` call with `@service-id->out-of-sync-state-store`. That would be more consistent with the `@service-id->out-of-sync-state-store` on line 428 above.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +419,150 @@
...
+  (try
+    (let [{:keys [service-id->instance-count-map service-id->out-of-sync-state trigger-service-ids]}
+          (compute-out-of-sync-state marathon-scheduler current-time trigger-timeout)]
+      (doseq [service-id trigger-service-ids]
+        (->> (service-id->instance-count-map service-id)
+             (trigger-sync-deployment! marathon-scheduler service-id)))
+      (log/info "updating local service-id->out-of-sync-state")
+      (reset-service-id->out-of-sync-state! marathon-scheduler service-id->out-of-sync-state))
+    (catch Exception e
+      (log/error e "unable to process out-of-sync services"))))
```
DaoWen: Does it makes sense to leave the value of `service-id->out-of-sync-state` untouched in the case of an error? Or would it be better to reset it to some other value?

shamsimam:
@sradack ready for review. The go-block is gone and we are using agents now.

waiter/src/waiter/util/http_utils.clj
```diff
@@ -48,11 +48,10 @@
...
     (when (and throw-exceptions (not (<= 200 status 299)))
       (ss/throw+ response))
-    (-> response
-        :body
-        async/<!!
-        json/read-str
-        walk/keywordize-keys)))
+    (let [response-body (-> response :body async/<!!)]
+      (cond-> response-body
+        (and (string? response-body) (pos? (count response-body)))
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/scheduler/marathon.clj
shamsimam: Now that `service-id->out-of-sync-state-store` is an agent I do this, but just modify the time in the state.

waiter/test/waiter/scheduler/marathon_test.clj
```diff
@@ -884,19 +866,437 @@
...
+                      marathon/update-app (fn [in-marathon-api in-service-id descriptor]
+                                            (is (= marathon-api in-marathon-api))
+                                            (is (= service-id in-service-id))
+                                            (is (= {:cmd "tc" :cpus 2 :id service-id :instances instances :mem 4} descriptor))
+                                            (deliver updated-invoked-promise :invoked))]
+          (scheduler/scale-app marathon-scheduler service-id instances true)
+          (is (= :invoked (deref deleted-deployment-promise 0 :not-invoked)))
+          (is (= :invoked (deref updated-invoked-promise 0 :not-invoked))))))))
+
+(defmacro launch-sync-deployment-maintainer
```
sradack: We don't need to start the timer task for unit tests.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -409,25 +541,36 @@
...
+        retrieve-framework-id-fn (memo/ttl #(retrieve-framework-id marathon-api) :ttl/threshold framework-id-ttl)
+        sync-deployment-maintainer-atom (atom nil)
+        marathon-scheduler (->MarathonScheduler
+                             marathon-api mesos-api retrieve-framework-id-fn home-path-prefix
+                             service-id->failed-instances-transient-store service-id->last-force-kill-store
+                             service-id->out-of-sync-state-store service-id->service-description-fn
+                             force-kill-after-ms is-waiter-app?-fn sync-deployment-maintainer-atom)
+        sync-deployment-maintainer (start-sync-deployment-maintainer
+                                     leader?-fn service-id->out-of-sync-state-store marathon-scheduler sync-deployment)]
+    (reset! sync-deployment-maintainer-atom sync-deployment-maintainer)
```
sradack: Can we avoid this atom?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -395,11 +420,118 @@
...
+         (into {}))
+    (catch Exception e
+      (log/error e "unable to retrieve out-of-sync services"))))
+
+(defn- trigger-sync-deployment!
+  "Triggers deployment for out-of-sync service."
+  [marathon-scheduler service-id {:keys [instances-scheduled] :as task-data}]
+  (try
+    (log/info "triggering sync deployment" {:service-id service-id :task-data task-data})
+    (scheduler/scale-app marathon-scheduler service-id instances-scheduled false)
```
sradack: After what period of time do API calls to Marathon timeout?

shamsimam: The connection and socket timeouts are configured to 10 seconds.

waiter/src/waiter/scheduler/marathon.clj
shamsimam: How to turn off the timer task then? Or do you want to avoid that ability for now and only add it when we explicitly need to stop the timer task?

waiter/test/waiter/scheduler/marathon_test.clj
shamsimam: Done.

waiter/src/waiter/scheduler/marathon.clj
shamsimam: I've added support for avoiding concurrently processing two iterations of the maintainer.

PR #343, opened 2018-06-05
Lift current-test-name utility function

DaoWen:

Life the code for retrieving the current clojure.test test name to its own utility function.

I want to call this from some unit testing code that passes a cid through an async/channel.

### DIFF
--- a/waiter/src/waiter/util/client_tools.clj
+++ b/waiter/src/waiter/util/client_tools.clj
@@ -166,8 +166,7 @@
 (defn current-test-name
   "Get the name of the currently-running test."
   []
-  (str (or (-> *testing-vars* first meta :name)
-           "test-unknown-name")))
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/util/client_tools.clj
```diff
@@ -163,9 +163,15 @@
     (.setCookieStore client (HttpCookieStore$Empty.))
     client))
 
+(defn current-test-name
+  "Get the name of the currently-running test."
+  []
+  (str (or (-> *testing-vars* first meta :name)
+           "test-unknown-name")))
```
dposada: How about

```clj
(-> *testing-vars* first meta :name (or "test-unknown-name") str)
```

DaoWen: Fine with me!

PR #344, opened 2018-06-05
handles NPE while querying for service state before syncer update

shamsimam:

- handles NPE while querying for service state before syncer update

Avoid router crashes.

### DIFF
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -535,9 +535,8 @@
         state-query-chan (async/chan 32)]
     (async/go
       (try
-        (loop [{:keys [last-update-time service-id->health-check-context]
-                :or {service-id->health-check-context {}}
-                :as current-state} {}]
+        (loop [{:keys [last-update-time service-id->health-check-context] :as current-state}
+               {:service-id->health-check-context {}}]
           (when-let [next-state
                      (async/alt!
                        exit-chan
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/scheduler.clj
```diff
@@ -535,7 +535,9 @@
         state-query-chan (async/chan 32)]
     (async/go
       (try
-        (loop [{:keys [last-update-time service-id->health-check-context] :as current-state} {}]
+        (loop [{:keys [last-update-time service-id->health-check-context]
+                :or {service-id->health-check-context {}}
+                :as current-state} {}]
```
DaoWen: Rather than using `:or` to provide destructuring defaults, maybe we should provide a non-empty initial state.
```clojure
        (loop [{:keys [last-update-time service-id->health-check-context] :as current-state}
               {:last-update-time (t/now) :service-id->health-check-context {}}]
```

DaoWen: I think providing a non-empty initial value more clearly conveys the semantics here (those keys should never be missing on subsequent iterations). I also think it's good to provide the timestamp so that clients don't have to deal with a possible `nil` there. Using `(DateTime. 0)` might be more clear than `(t/now)` for conveying that the state hasn't been updated.

shamsimam: I initialized with `{:service-id->health-check-context {}}`. I left `last-update-time` as nil as none of the current usage triggers NPE with that value left as `nil` and it clearly indicates we haven't seen updates. Using a non-nil value then needs extra interpretation (is it the epoch date?) to decide whether the state was updated or not.

PR #346, opened 2018-06-07
avoids triggering expired instances logic in scaling when there are no expired instances

shamsimam:

- avoids triggering expired instances logic in scaling when there are no expired instances

We want to avoid jitters in the scaling algorithms, an example is shown in the image below:
waiter-scaling-jitters

When all target instances aren't healthy, the previous code used to incorrectly assume that it is trying to replace expired instances and compute incorrect scaling results that triggered the jitter in the chart above. The proposed fix avoids exploring this path if there are no expired instances.

### DIFF
--- a/waiter/src/waiter/scaling.clj
+++ b/waiter/src/waiter/scaling.clj
@@ -336,7 +336,8 @@
         ; since the instance killer will kill the expired instances
         scaling-down (< integer-delta 0)
         all-instances-are-healthy (= total-instances healthy-instances)
-        scale-to-instances' (if (and scaling-down all-instances-are-healthy)
+        scale-to-instances' (if (or (and scaling-down all-instances-are-healthy)
+                                    (zero? expired-instances))
                               scale-to-instances
                               (- (+ scale-to-instances expired-instances-to-replace) excess-instances))
         integer-delta' (int (- scale-to-instances' total-instances))]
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/scaling.clj
```diff
@@ -383,10 +384,11 @@
           (when-not (zero? scale-amount)
             (apply-scaling-fn service-id
                               {:outstanding-requests outstanding-requests
-                               :task-count task-count
-                               :total-instances instances
                                :scale-amount scale-amount
-                               :scale-to-instances scale-to-instances}))
+                               :scale-to-instances scale-to-instances
+                               :target-instances target-instances
```
shamsimam: We would like the target instances to get logged.

waiter/src/waiter/scaling.clj
```diff
@@ -336,7 +336,8 @@
         ; since the instance killer will kill the expired instances
         scaling-down (< integer-delta 0)
         all-instances-are-healthy (= total-instances healthy-instances)
-        scale-to-instances' (if (and scaling-down all-instances-are-healthy)
+        scale-to-instances' (if (or (and scaling-down all-instances-are-healthy)
+                                    (zero? expired-instances))
```
shamsimam: Avoid adjusting for expired instances if there are no expired instances.

waiter/test/waiter/scaling_test.clj
```diff
@@ -678,6 +678,9 @@
       (scales-like 2 3 2, fast-scaling 1 2 0.9 1 1))
     (testing "scale down with expired instances"
       (scales-like -4 1 1, fast-scaling 5 0 5.3 5 5))
+    (testing "scale down without expired instances"
+      (scales-like -2 8 7.919, (assoc default-scaling "scale-down-factor" 0.001 "scale-up-factor" 0.1) 10 0 7.927 9 0)
```
shamsimam: Previous code would have yielded:
`{:scale-to-instances 7, :target-instances 7.91907088006478, :scale-amount -3}`

shamsimam:
Updated the description.

PR #305, opened 2018-05-03
App instance startup metrics

DaoWen:
## Changes proposed in this PR

Add codahale timers for three metrics:
1. Time from a Waiter service's creation until the first instance appears ("schedule time").
2. Time from the first instance appearing to when it's healthy ("startup time").
3. Time from a Waiter service's creation until the first instance is healthy ("start-to-healthy time").

<span>#</span>3 is basically <span>#</span>1 + <span>#</span>2. Each of these metrics are tracked both per-service and overall.

## Why are we making these changes?

We want to have data on how long it takes to start up apps in production on our existing platforms (Mesos/Marathon) and future platforms (Kubernetes).

PR #348, opened 2018-06-11
Statsd reporting for service launch metrics

DaoWen:

Adds `statsd` reporting for a subset of the service instance launch metrics added in #305.

The current codahale metrics disappear when a Waiter router is restarted. We would like to persist some of the metrics so that we have historical data.

### DIFF
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -673,7 +673,7 @@
    2) the time from scheduling a new instance until it becomes healthy (the \"startup\" time)."
   [service-id->launch-tracker new-service-ids removed-service-ids service-id->healthy-instances
    service-id->unhealthy-instances service-id->instance-counts
-   service-id->service-description-fn waiter-schedule-timer]
+   service-id->service-description-fn leader? waiter-schedule-timer]
   (let [current-time (t/now)
         ;; Remove deleted services and add newly-discovered services
         service-id->launch-tracker'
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/scheduler.clj
```diff
@@ -727,7 +731,9 @@
         (doseq [start-time matched-start-times]
           (let [duration (metrics/duration-between start-time current-time)]
             (metrics/report-duration waiter-schedule-timer duration)
-            (metrics/report-duration service-schedule-timer duration)))
+            (metrics/report-duration service-schedule-timer duration)
+            (statsd/histo! metric-group "schedule_time_seconds"
+                           (t/in-seconds duration))))
```
sradack: Should be consistent with the other timer histograms?
- Our other timers are in millis.  
- For what it's worth, our other timers don't include the time unit in the name.

sradack: Should we add `startup-time` as well?

sradack: Should we only populate the statsd metrics from the leader?

sradack: Whoops, just noticed the WIP tag.

DaoWen: Thanks for the feedback.
* Renamed `schedule_time_seconds` → `schedule_time`
* Switched to milliseconds
* Added `startup_time` histo reporting
* Only reporting from leader

DaoWen:
@sradack - I've tested this patch in our staging environment, and the reporting looks good! I think this is ready to merge if you don't have any additional feedback.

PR #358, opened 2018-06-18
Add lightweight "kitchenette" for Travis

DaoWen:

- Add partial implementation of kitchen using bash+netcat: <i>kitchenette</i>.
- Update integration tests that need the full kitchen functionality to use `:full-kitchen true` when making kitchen requests.
- Add statically-linked `ncat` binary for easy installation during testing.

Launching kitchen can take over a minute in Travis's VMs and containers.
Using a lighter-weight bash+ncat based "kitchenette" should be faster.


dposada:
@DaoWen I love the idea. Can we use a name other than "kitchenette"? I think it's too close to "kitchen" and could get confusing. How about "canteen"?

DaoWen:
We decided not to merge this because the actual performance improvement we're seeing isn't as much as I was originally expecting. The additional complexity of having both Kitchen and Kitchenette doesn't seem to be offset by the Travis-CI test time gains.

PR #363, opened 2018-06-22
handles namespaced keys in json output

shamsimam:

- handles namespaced keys in json output

We would like to see the complete namespaced keyword string in JSON output to prevent automatically overriding matching keys.

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -114,6 +114,13 @@
     (log/debug "generate-secret-word" [src-id dest-id] "->" secret-word)
     secret-word))
 
+(defn- stringify-keys
+  [k]
+  (cond
+    (keyword? k) (str (.-sym k))
+    (nil? k) (throw (Exception. "JSON object properties may not be nil"))
+    :else (str k)))
+
 (defn stringify-elements
   [k v]
   (cond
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -114,6 +113,14 @@
     (log/debug "generate-secret-word" [src-id dest-id] "->" secret-word)
     secret-word))
 
+(defn- stringify-keys
+  [k]
+  (cond
+    (keyword? k) (str (.-sym k))
+    (instance? Named k) (name k)
```
DaoWen: Could you delete this so symbols also fall through to the default case using `str`? It seems strange to have our keywords namespaced but the symbols not namespaced.

waiter/src/waiter/util/utils.clj
```diff
@@ -27,8 +27,7 @@
             [taoensso.nippy.compression :as compression]
             [waiter.util.date-utils :as du])
   (:import clojure.core.async.impl.channels.ManyToManyChannel
-           clojure.lang.ExceptionInfo
-           clojure.lang.PersistentQueue
+           (clojure.lang ExceptionInfo Named PersistentQueue)
```
DaoWen: You can delete `Named` here if you remove the special case below.

waiter/src/waiter/util/utils.clj
shamsimam: Done.

waiter/src/waiter/util/utils.clj
shamsimam: Done.

PR #364, opened 2018-06-26
avoid setting the content-type by default

shamsimam:

- avoid setting the content-type by default

Waiter, which is acting as a proxy, should not insert a `content-type` header if the client did not provide one in the request.

## Related PRs

https://github.com/twosigma/jet/pull/10 which adds support for not providing a `content-type` based on the body.

### DIFF
--- a/kitchen/project.clj
+++ b/kitchen/project.clj
@@ -14,7 +14,7 @@
 ;; limitations under the License.
 ;;
 (defproject kitchen "0.1.0-SNAPSHOT"
-  :dependencies [^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "untyped-content-provider-default"}}
+  :dependencies [^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "waiter-patch"}}
                  [cc.qbits/jet "0.7.10-20180626_194651-g716c5e0"]
                  [log4j/log4j "1.2.17" :exclusions [javax.mail/mail
                                                     javax.jms/jms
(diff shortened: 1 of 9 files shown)


kitchen/project.clj
```diff
@@ -14,8 +14,8 @@
 ;; limitations under the License.
 ;;
 (defproject kitchen "0.1.0-SNAPSHOT"
-  :dependencies [^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "waiter-patch"}}
```
pschorf: Jet PR merged, can you change this back to waiter-patch?

token-syncer/project.clj
```diff
@@ -14,8 +14,8 @@
 ;; limitations under the License.
 ;;
 (defproject token-syncer "0.1.0-SNAPSHOT"
-  :dependencies [^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "waiter-patch"}}
```
pschorf: Change back to waiter-patch

waiter/project.clj
```diff
@@ -30,8 +30,8 @@
 
   :dependencies [[bidi "2.0.16"
                   :exclusions [prismatic/schema ring/ring-core]]
-                 ^{:voom {:repo "https://github.com/twosigma/jet.git" :branch "waiter-patch"}}
```
pschorf: Change to waiter-patch

kitchen/project.clj
shamsimam: Done.

token-syncer/project.clj
shamsimam: Done.

waiter/project.clj
shamsimam: Done.

PR #366, opened 2018-06-27
WIP: Cook scheduler

shamsimam:

- adds support for scheduling via Cook


wyegelwel:
This is so cool!

dposada:
@shamsimam I would favor separating the refactoring parts of this PR (e.g. moving functions from `marathon` to `mesos`) into a separate, preliminary PR

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,523 @@
...
+(let [instance-expiry-adjustment-mins 5]
+  (defn create-job-description
+    "Create the Cook job description for a service."
+    [service-id service-description service-id->password-fn home-path-prefix instance-priority]
+    (let [{:strs [backend-proto cmd cpus health-check-url instance-expiry-mins mem metadata name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID))
+          home-path (str home-path-prefix run-as-user)
+          environment (scheduler/environment service-id service-description service-id->password-fn home-path)
+          cook-backend-port (get metadata "cook-backend-port")
```
sradack: Why should these be in the metadata namespace?  Also is version a better place for this data since it's related to version?

sradack: What is `cook-backend-port` used for?  Does cook not automatically assign a port for the instance and set `$PORT0`?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,523 @@
...
+  "Create and start a new Cook job specified by the service-description."
+  [cook-api service-id service-description service-id->password-fn home-path-prefix instance-priority]
+  (log/info "creating a new job for" service-id)
+  (->> (create-job-description service-id service-description service-id->password-fn home-path-prefix instance-priority)
+       (post-jobs cook-api)))
+
+(defn determine-instance-priority
+  "Determines the instance priority based on allowed (sorted by highest first) priorities and unavailable priorities."
+  [allowed-priorities unavailable-priorities]
+  (or (some #(and (not (contains? unavailable-priorities %)) %) allowed-priorities)
```
sradack: Do we need `and` here?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,523 @@
...
+      (last allowed-priorities)))
+
+(defn launch-jobs
+  "Launch num-instances jobs specified by the service-description."
```
sradack: A comment would be useful describing how priorities are assigned.

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,523 @@
...
+
+;; Helper Methods
+
+(let [instance-expiry-adjustment-mins 5]
+  (defn create-job-description
+    "Create the Cook job description for a service."
+    [service-id service-description service-id->password-fn home-path-prefix instance-priority]
+    (let [{:strs [backend-proto cmd cpus health-check-url instance-expiry-mins mem metadata name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID))
```
sradack: TODO Use "less random" UUIDs for better Cook cache performance.

waiter/config-full.edn
```diff
@@ -228,6 +228,48 @@
...
+                           ;; Controls whether Cook impersonation should be enabled:
+                           :impersonate true
+
+                           ;; The port on which the mesos agents are listening.
+                           ;; This is an optional parameter.
+                           ;; When absent log directory and log url support will be disabled.
+                           :mesos-slave-port 5051
+
+                           ;; The time range to use while querying for running and waiting Cook jobs
+                           :search-interval-days 10
```
sradack: TODO Could this be based on `instance-expiry-mins`?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,523 @@
...
+
+(let [http-client (http-utils/http-client-factory {:conn-timeout 10000
+                                                   :socket-timeout 10000
+                                                   :spnego-auth false})
+      ;; TODO make this cache configurable
+      healthy-instance-cache (-> {}
+                                 (cache/fifo-cache-factory :threshold 5000)
+                                 (cache/ttl-cache-factory :ttl (-> 10 t/seconds t/in-millis))
+                                 atom)]
+  (defn- instance-healthy?
```
sradack: TODO Move health check out of critical path

waiter/config-full.edn
```diff
@@ -228,6 +228,48 @@
...
+                    ;; :kind :cook uses Cook (https://github.com/twosigma/Cook) for scheduling instances:
+                    :cook {
+                           ;; Custom implementations should specify a :factory-fn
+                           ;; that returns an instance of waiter.scheduler.ServiceScheduler:
+                           :factory-fn 'waiter.scheduler.cook/cook-scheduler
+
+                           ;; The interval (milliseconds) at which instance failed jobs will be checked:
+                           :failed-tracker-interval-ms 10000
+
+                           ;; Deprecated! The users who are allowed to launch services via the Cook Scheduler.
```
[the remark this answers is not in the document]
shamsimam: This is currently a gap in Cook-Waiter integration where Cook does not allow to search by metadata flags, instead, Cook requires the user being provided while searching jobs. When we add such support to Cook, we will not require explicitly stating the allowed users.

waiter/config-full.edn
shamsimam: This parameter affects the search query for all jobs of a user (regardless of service), whereas `instance-expiry-mins` is a service specific configuration.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Added TODO in code.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Added TODO in code.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Currently, `cook-backend-port` is additional information about the service that we are allowing the user to provide when they want to use hard-coded ports for their service. In the long run, I see us not needing to use `cook-backend-port` and this "feature" being unsupported.

shamsimam: Cook does assign ports, however, `cook-backend-port` is a stop-gap solution for scenarios where users are forced to use only certain ports for their services (e.g. Mesos agents being configured on hosts that make only certain ports public).

waiter/src/waiter/scheduler/cook.clj
shamsimam: The `and` clause is used to return the priority instead of a boolean. Added a comment to clarify this.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Added comment.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Moved the container configuration out of metadata by relying on `image-type::namespace/name:label` format of the `version` in the service description.

waiter/config-full.edn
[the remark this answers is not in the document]
sradack: It becomes deprecated when it's been replaced with something better.

waiter/config-full.edn
sradack: My point is that there shouldn't be anything older than `instance-expiry-mins`.

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,540 @@
...
+(let [instance-expiry-adjustment-mins 5]
+  (defn create-job-description
+    "Create the Cook job description for a service."
+    [service-id service-description service-id->password-fn home-path-prefix instance-priority]
+    (let [{:strs [backend-proto cmd cpus health-check-url instance-expiry-mins mem metadata name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID)) ;; TODO Use "less random" UUIDs for better Cook cache performance.
+          home-path (str home-path-prefix run-as-user)
+          environment (scheduler/environment service-id service-description service-id->password-fn home-path)
+          cook-backend-port (get metadata "cook-backend-port")
```
sradack: We talked about moving this into the config.

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,540 @@
...
+        (try
+          (when-not (pos? (Integer/parseInt cook-backend-port))
+            (throw (ex-info "cook-backend-port metadata parsed to a non-positive integer"
+                            {:cook-backend-port cook-backend-port})))
+          (catch ExceptionInfo ex
+            (throw ex))
+          (catch Exception ex
+            (throw (ex-info "cook-backend-port metadata must parse to a positive integer"
+                            {:cook-backend-port cook-backend-port} ex)))))
+      (when (seq container-type)
```
sradack: Is `container-type` expected to be a sequence?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,540 @@
...
+    (let [{:strs [backend-proto cmd cpus health-check-url instance-expiry-mins mem metadata name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID)) ;; TODO Use "less random" UUIDs for better Cook cache performance.
+          home-path (str home-path-prefix run-as-user)
+          environment (scheduler/environment service-id service-description service-id->password-fn home-path)
+          cook-backend-port (get metadata "cook-backend-port")
+          container-type (get metadata "container-type")
+          [image-namespace image-name-and-label] (str/split version #"/" 2)
+          [image-name image-label] (str/split (str image-name-and-label) #":" 2)
+          container-support-enabled? (and (seq container-type)
```
sradack: This strikes me as an odd wait to check for valid strings?  I'm in favor of checking either non-`nil` or being more prescriptive (e.g. making sure it's a string of a certain format)

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,540 @@
...
+  (defn create-job-description
+    "Create the Cook job description for a service."
+    [service-id service-description service-id->password-fn home-path-prefix instance-priority]
+    (let [{:strs [backend-proto cmd cpus health-check-url instance-expiry-mins mem metadata name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID)) ;; TODO Use "less random" UUIDs for better Cook cache performance.
+          home-path (str home-path-prefix run-as-user)
+          environment (scheduler/environment service-id service-description service-id->password-fn home-path)
+          cook-backend-port (get metadata "cook-backend-port")
+          container-type (get metadata "container-type")
```
sradack: Instead of using metadata, would it be better to use a new `cmd-type`?

shamsimam: Done.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Used a single pattern to match the version.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Avoided using `seq` to check blank strings.

waiter/src/waiter/scheduler/cook.clj
shamsimam: Done.

waiter/config-full.edn
shamsimam: We need a larger interval to search for failed instances for example.

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -0,0 +1,529 @@
...
+    "Create the Cook job description for a service."
+    [service-id service-description service-id->password-fn home-path-prefix instance-priority backend-port]
+    (let [{:strs [backend-proto cmd cmd-type cpus health-check-url instance-expiry-mins mem name ports
+                  run-as-user version]} service-description
+          job-uuid (str (UUID/randomUUID)) ;; TODO Use "less random" UUIDs for better Cook cache performance.
+          home-path (str home-path-prefix run-as-user)
+          environment (scheduler/environment service-id service-description service-id->password-fn home-path)
+          container-mode? (= "docker" cmd-type)
+          [_ image-namespace image-name image-label] (when container-mode?
+                                                       (re-matches #"(.*)/(.*):(.*)" version))
```
shamsimam: I intentionally avoid performing this regex check earlier, e.g. during service description creation or token registration, to avoid overheads.

shamsimam:
@sradack @dposada this is ready for review.

shamsimam:
@dposada any feedback on this?

PR #367, opened 2018-06-27
adds support for aggergated utilization metric

shamsimam:

- adds support for aggregating instance slots across routers

We would like to know the utilization of our instances at any given time. It can be computed from the slots assigned and available data.

instance-count-metrics

### DIFF
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -318,8 +318,12 @@
   [router->codahale-metrics]
   (assoc
     (apply merge-codahale-metrics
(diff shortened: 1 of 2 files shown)


sradack:
Why are we creating this metric?  It seems easily derivable from the existing metrics.

shamsimam:
We need to start exposing instance counts in our aggregated metrics, after that this metric will be easily derivable. I will make that change in this PR instead.

shamsimam:
We now aggregate slot information in the codahale metrics that enable computing utilization. Ready for another pass.

PR #370, opened 2018-06-28
adds missing `mesos-slave-port` and `slave-directory` config to config-full.edn

shamsimam:

- adds missing `mesos-slave-port` and `slave-directory` config to config-full.edn

Improves documentation of Marathon scheduler.

### DIFF
--- a/waiter/src/waiter/settings.clj
+++ b/waiter/src/waiter/settings.clj
@@ -276,6 +276,7 @@
                                                 :spnego-auth true}
                                  :force-kill-after-ms 60000
                                  :framework-id-ttl 900000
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/settings.clj
```diff
@@ -276,6 +276,7 @@
                                                 :spnego-auth true}
                                  :force-kill-after-ms 60000
                                  :framework-id-ttl 900000
+                                 :mesos-slave-port 5051
```
dposada: What was the effect of this not being present?

shamsimam: Done.

PR #376, opened 2018-07-05
adds support for utilization in the simulator output

shamsimam:

- adds support for utilization in the simulator output

We want to better analyze the scaling algorithm in the presence of additional metrics.

**Before:**

screen shot 2018-07-05 at 11 00 03 am

**After:**

screen shot 2018-07-05 at 3 33 56 pm

### DIFF
--- a/waiter/src/waiter/simulator.clj
+++ b/waiter/src/waiter/simulator.clj
@@ -53,6 +53,7 @@
   Stats:
     total-queue-time       The total amount of ticks that clients have been sitting in the queue.
     total-idle-server-time The total amount of ticks that servers have been sitting idle.
+    utilization            The percentage of servers that are in use during the current tick.
     scale-ups              The total number of scale up operations.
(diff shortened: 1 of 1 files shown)


shamsimam:
Documentation in this PR description or elsewhere?

shamsimam:
@dposada added documentation

waiter/src/waiter/simulator.clj
```diff
@@ -53,6 +53,7 @@
   Stats:
     total-queue-time       The total amount of ticks that clients have been sitting in the queue.
     total-idle-server-time The total amount of ticks that servers have been sitting idle.
+    utilization            The percentage of servers that are in use during the current tick.
```
dposada: servers or slots?

shamsimam: The sim doesn't support concurrency (it pre-dates concurrency support) and refers to servers elsewhere.

PR #379, opened 2018-07-07
uses keyword->str at all json write sites in utils

shamsimam:

- uses keyword->str at all json write sites in utils

We want to include the namespace prefix of keywords in all our json output.

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -114,10 +114,15 @@
     (log/debug "generate-secret-word" [src-id dest-id] "->" secret-word)
     secret-word))
 
+(defn- keyword->str
+  "Converts keyword to string including the namespace."
+  [k]
+  (str (.-sym k)))
+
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -157,7 +163,7 @@
      :body (fn [^ServletResponse resp]
              (let [writer (OutputStreamWriter. (.getOutputStream resp))]
                (try
-                 (json/write data-map writer :value-fn stringify-elements)
+                 (json/write data-map writer :key-fn stringify-keys :value-fn stringify-elements)
```
DaoWen: The tests are probably failing because `stringify-keys` doesn't exist.

DaoWen: Nope, it exists—it's just not in the patch. Looks like a random failure. I restarted, and if we get a green I'll go ahead and merge this.

PR #383, opened 2018-07-10
added flagging for services with a failed oom instance

mokshjawa:

- Flagging any services that have at least one instance failing due to being out of memory.

- As we attempt to add autosizing functionality, we want to be able to see if we are scaling an app down too much as to lead to an out of memory error.


waiter/src/waiter/state.clj
```diff
@@ -1164,6 +1169,13 @@
         (and has-failed-instances? (all-instances-flagged-with? :never-passed-health-checks)) :invalid-health-check-response
         (and has-unhealthy-instances? (= first-unhealthy-status 401)) :health-check-requires-authentication))))
 
+(defn get-instability-issue
```
shamsimam: Suggestion for name: `exhibits-oom-instability?`

waiter/src/waiter/state.clj
```diff
@@ -1256,7 +1271,11 @@
                                        deployment-error (get-deployment-error healthy-instances unhealthy-instances failed-instances deployment-error-config)
                                        service-id->deployment-error' (if deployment-error
                                                                        (assoc service-id->deployment-error service-id deployment-error)
-                                                                       (dissoc service-id->deployment-error service-id))]
+                                                                       (dissoc service-id->deployment-error service-id))
+                                       oom-instability (get-instability-issue failed-instances)
+                                       service-id->oom-instability' (if oom-instability
```
shamsimam: I think we should include the `oom-instability` in the `if` clause and maintain a set for OOM instability instead of `assoc`-ing `true` into a map. The new field could be called `oom-instable-service-ids`

waiter/src/waiter/state.clj
```diff
@@ -311,10 +311,10 @@
         (meters/mark! update-responder-state-meter)
         (assoc current-state
           :deployment-error deployment-error
-          :oom-instability oom-instability
           :id->instance id->instance'
           :instance-id->state instance-id->state'
           :instance-id->consecutive-failures instance-id->consecutive-failures'
+          :instability-issue instability-issue
```
shamsimam: Nitpick: lexicographic order

waiter/src/waiter/state.clj
```diff
@@ -1170,11 +1170,13 @@
         (and has-unhealthy-instances? (= first-unhealthy-status 401)) :health-check-requires-authentication))))
 
 (defn get-instability-issue
-  "Returns true if 1 or more instances failed due to OOM"
+  "Returns appropriate instability issue (only oom for now) for a service based on its instances, or nil if no such issues."
   [failed-instances]
   (when (not-empty failed-instances)
```
shamsimam: I found this easier to read:
```
(when (some #(contains? (:flags %) :memory-limit-exceeded) failed-instances-flags)
  :not-enough-memory)
```

We can use the `cond` version when we have more instability issues to support. At that stage, maybe we want to return a set of instability issues rather than a single one.

waiter/test/waiter/state_test.clj
```diff
@@ -1225,14 +1225,107 @@
...
+                                      :service-id->instability-issue {}})
 
                     state (async/<!! router-state-push-chan)
                     actual-state (dissoc state :iteration :service-id->healthy-instances :service-id->expired-instances :service-id->starting-instances
                                           :service-id->instance-counts :service-id->my-instance->slots :routers :time)]
                 (when (not= expected-state actual-state)
                   (clojure.pprint/pprint (clojure.data/diff expected-state actual-state)))
                 (is (= expected-state actual-state) (str (clojure.data/diff expected-state actual-state))))))
+          (async/>!! exit-chan :exit)))))
+  (testing "router-state-maintainer-deployment-errors-updated"
```
shamsimam: The label on the `testing` could be better.

shamsimam: Nitpick: add a new line before starting this `testing` clause.

waiter/src/waiter/state.clj
mokshjawa: Agreed! I was initially trying to make it extendable for other instability issues but this makes sense. Thanks.

waiter/src/waiter/state.clj
shamsimam: Comment no longer valid after the function was refactored.

pschorf:
Just needs an integration test and should be ready to go.

shamsimam:
Marked WIP as we are waiting on integration test.

mokshjawa:
Integration complete, all good on my end

kitchen/bin/kitchen
```diff
@@ -758,6 +758,13 @@ class Kitchen(HTTPWebSocketsHandler):
         elif path == '/unchunked':
             self.__set_response(itertools.cycle(lorem_ipsum), max_response_size)
 
+        elif path =='/oom-instability':
+            l = []
+            # time.sleep(30)
```
pschorf: Can you remove this comment?

waiter/integration/waiter/instability_test.clj
```diff
@@ -0,0 +1,29 @@
...
+(ns waiter.instability-test
+  (:require [clojure.test :refer :all]
+            [waiter.util.client-tools :refer :all]))
+
+(deftest ^:parallel ^:integration-fast ^:explicit test-oom-instability
+  (testing-using-waiter-url
+    (let [headers {:x-waiter-name (rand-name)
+                   :x-waiter-mem 100}
+          response (make-request-with-debug-info headers #(make-kitchen-request waiter-url % :path "/oom-instability"))]
+      (Thread/sleep 30000)
```
pschorf: You can use `wait-for` to avoid the `Thread/sleep`

waiter/test/waiter/state_test.clj
```diff
@@ -1206,13 +1224,108 @@
...
+                                      :service-id->instability-issue {}})
+
                     state (async/<!! router-state-push-chan)
                     actual-state (dissoc state :iteration :service-id->healthy-instances :service-id->expired-instances :service-id->starting-instances
                                           :service-id->instance-counts :service-id->my-instance->slots :routers :time)]
                 (when (not= expected-state actual-state)
                   (clojure.pprint/pprint (clojure.data/diff expected-state actual-state)))
                 (is (= expected-state actual-state) (str (clojure.data/diff expected-state actual-state))))))
+          (async/>!! exit-chan :exit)))))
+  (testing "router-state-maintainer-deployment-errors-updated"
```
pschorf: Is this related to your change? It seems to be about deployment errors.

mokshjawa: The function got copied over while rebasing/merging.

PR #385, opened 2018-07-12
Integration heavy/lite lein test selectors

DaoWen:

Add the `:resource-heavy` metadata key to integration tests that start multiple service instances.

Tests that start many service instances concurrently tend to be flaky when running on our Kubernetes back-end in Travis. We need to run these serially for our Kubernetes integration testing.


DaoWen:
@dposada - Green!

PR #390, opened 2018-07-29
[WIP] Instability Fallback

mokshjawa:

- Creating a fallback for instable services to restart them with additional memory allocated when they OOM

- Many services fail due to OOM and are restarted with the same config, which will just lead to another OOM eventually. Adjusting the mem request incrementally will help the service avoid OOMing in the future.


mokshjawa:
Working on unit & integration tests.

PR #391, opened 2018-07-30
retry create service after detecting old StopApplication conflict deployments

shamsimam:

- retry create service after detecting old StopApplication conflict deployments

Waiter should be resilient to Marathon issue. In particular, in this instance we want Waiter to be resilient to outdated `StopApplication` deployments which prevent Waiter from starting a new service.

### DIFF
--- a/waiter/src/waiter/scheduler/marathon.clj
+++ b/waiter/src/waiter/scheduler/marathon.clj
@@ -300,11 +300,10 @@
         conflict-handler-retry
         (fn conflict-handler-retry [error-data]
           (if-let [deployment (exception->stop-application-deployment error-data)]
-            (let [{:keys [id version]} deployment
-                  stop-application-timeout-ms (-> 5 t/minutes t/in-millis)]
+            (let [{:keys [id version]} deployment]
               (log/info "detected StopApplication deployment" {:deployment deployment :service-id service-id})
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -261,6 +261,60 @@
...
+                             error-data)
+                    "Exception starting new app"))
+        conflict-handler-retry
+        (fn conflict-handler-retry [error-data]
+          (if-let [deployment (exception->stop-application-deployment error-data)]
+            (let [{:keys [id version]} deployment
+                  stop-application-timeout-ms (-> 5 t/minutes t/in-millis)]
+              (log/info "detected StopApplication deployment" {:deployment deployment :service-id service-id})
+              (if (-> (du/str-to-date version formatter-marathon)
+                      (t/plus (t/millis stop-application-timeout-ms))
```
pschorf: What if you just use `(t/plus (t/minutes 5))` here, and kill the `stop-application-timeout-ms` variable?

shamsimam: Done, I was previously preparing to make it configurable ;)

shamsimam:
Build is green :)

PR #392, opened 2018-07-30
Deflake test-grace-period-with-tokens

DaoWen:

Kitchen only sleeps for 75% of the startup grace period.

Sleeping for the full grace period makes this test fail regularly since there's no room for variation in the launch time of the Kitchen service instance.

### DIFF
--- a/waiter/integration/waiter/request_timeout_test.clj
+++ b/waiter/integration/waiter/request_timeout_test.clj
@@ -198,10 +198,17 @@
                                       :cmd-type "shell"})]
           (is (= 200 status) (str "Did not get a 200 response. " body)))
         (log/info "Making request for" token)
-        (let [{:keys [status body service-id]} (make-request-with-debug-info
-                                                 {:x-waiter-token token}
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/request_timeout_test.clj
```diff
@@ -180,12 +180,13 @@
...
   (testing-using-waiter-url
     (let [grace-period (t/minutes 2)
+          startup-delay-ms (-> grace-period t/in-millis (* 0.75) long)
           token (rand-name)]
       (try
         (log/info "Creating token for" token)
         (let [{:keys [status body]}
               (post-token waiter-url {:cmd (kitchen-cmd (str "--port $PORT0 --start-up-sleep-ms "
-                                                             (t/in-millis grace-period)))
+                                                             startup-delay-ms))
```
shamsimam: Can we include an assertion on the canary request that it takes at least `startup-delay-ms` ms to come back with a response?

DaoWen: Done.

PR #393, opened 2018-07-30
test-grace-period-with-token better handling of non-200 response

DaoWen:

* Use the `assert-response-status` helper function.
* Patch the code from #392 to move the acquire-instance times after the status 200 check.
* Mark the test as `resource-heavy` since it ties up one of the Kubernetes CPUs for about 2 minutes.

`test-grace-period-with-token` wasn't gracefully handling non-200 responses (e.g., 503 from timeouts).

### DIFF
--- a/waiter/integration/waiter/request_timeout_test.clj
+++ b/waiter/integration/waiter/request_timeout_test.clj
@@ -177,14 +177,14 @@
           (is (str/blank? (:instance-id response))))
         (.await long-request-ended-latch)))))
 
-(deftest ^:parallel ^:integration-fast test-grace-period-with-tokens
+(deftest ^:parallel ^:integration-fast ^:resource-heavy test-grace-period-with-tokens
   (testing-using-waiter-url
     (let [grace-period (t/minutes 2)
           startup-delay-ms (-> grace-period t/in-millis (* 0.75) long)
           token (rand-name)]
       (try
         (log/info "Creating token for" token)
-        (let [{:keys [status body]}
+        (let [response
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/request_timeout_test.clj
```diff
@@ -196,21 +196,21 @@
                                       :name (str "test" token)
                                       :grace-period-secs (t/in-seconds grace-period)
                                       :cmd-type "shell"})]
-          (is (= 200 status) (str "Did not get a 200 response. " body)))
+          (assert-response-status response 200))
         (log/info "Making request for" token)
         (let [{:keys [status body service-id] :as response} (make-request-with-debug-info
```
shamsimam: `body` is now unused

waiter/integration/waiter/request_timeout_test.clj
```diff
@@ -196,21 +196,21 @@
...
-                                             :headers
-                                             (get "x-waiter-get-available-instance-ns")
-                                             Long/parseLong
-                                             (quot 1000000))] ; truncated nanos->millis
-          (is (= 200 status) (str "Did not get a 200 response. " body))
-          (is (<= startup-delay-ms instance-acquired-delay-ms)
-              (format "Healthy instance was found in just %dms (too short)" instance-acquired-delay-ms))
+                                                              #(make-request waiter-url "/secrun" :headers %))]
+          (assert-response-status response 200)
           (when (and (= 200 status) (can-query-for-grace-period? waiter-url))
```
shamsimam: `can-query-for-grace-period?` current implementation means we will not check the following assertions for kubernetes scheduler, is that intended?

DaoWen:
@shamsimam - any more feedback?

PR #396, opened 2018-08-03
 supports quanta-based scaling (while scaling-up)

shamsimam:

- adds service-id->service-description-fn parameter to service-scaling-multiplexer
- adds quanta-constraints to settings  
- adds quanta-constraints parameter to service-scaling-multiplexer
- supports quanta-based scaling (while scaling-up)

We would like our scale-up operations to be throttled to reduce the chances of overshoot (and waster) in presence of spiky traffic.

### DIFF
--- a/waiter/src/waiter/settings.clj
+++ b/waiter/src/waiter/settings.clj
@@ -86,7 +86,9 @@
    (s/required-key :router-syncer) {(s/required-key :delay-ms) schema/positive-int
                                     (s/required-key :interval-ms) schema/positive-int}
    (s/required-key :scaling) {(s/required-key :autoscaler-interval-ms) schema/positive-int
-                              (s/required-key :inter-kill-request-wait-time-ms) schema/positive-int}
+                              (s/required-key :inter-kill-request-wait-time-ms) schema/positive-int
+                              (s/required-key :quanta-constraints) {(s/required-key :cpus) schema/positive-int
+                                                                    (s/required-key :mem) schema/positive-int}}
    (s/required-key :scheduler-config) (s/constrained
                                         {:kind s/Keyword
                                          s/Keyword schema/require-symbol-factory-fn}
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/settings.clj
```diff
@@ -269,7 +271,9 @@
                    :interval-ms 1500}
    :scaling {:autoscaler-interval-ms 1000
              ; throttles the rate at which kill requests are sent to the scheduler
-             :inter-kill-request-wait-time-ms 1000}
+             :inter-kill-request-wait-time-ms 1000
+             :quanta-constraints {:cpus 64
+                                  :mem (* 512 1024)}}
```
shamsimam: We need to decide on sane defaults.

sradack:
Should we compare the graphs before merging?

shamsimam:
> Should we compare the graphs before merging?

Yes, I will generate and add them to this PR.

shamsimam:
Some initial results:

scaling-initial-results

I think we need to further improve out scaling algorithm to contraining scaling-up when there are no outstanding requests and there are already enough instances.

shamsimam:
Here are comparison results from the constrained and previous unconstrained scaling. With constrained scaling, we have lower idle time by sacrificing queue wait time.

The comparison results are for the same app with different startup times (5 units vs 30 units) with the following service settings:

service-settings

Here are results for different benchmark workloads:

Workload-1:

scaling-a-1
scaling-a-2

scaling-b-1
scaling-b-2

Workload-2:

scaling-c-1
scaling-c-2

scaling-d-1
scaling-d-2

Workload-3:

scaling-e-1
scaling-e-2

scaling-f-1
scaling-f-2

Workload-4:

scaling-g-1
scaling-g-2

scaling-h-1
scaling-h-2

shamsimam:
Rebased this branch after the sim ui changes got merged.

This is a comparison of the scaling algorithm on one benchmark workload which shows constrained scaling trades off queue time for utilization:

scaling-comparison-1

waiter/src/waiter/settings.clj
dposada: Is this something left to do in this PR?

shamsimam: @sradack thinks the current defaults are reasonable:

> Defaults seem reasonable enough.

shamsimam:
> Would it be possible to write an integration test for this?

I think it'd be hard to test reliable, we've already had flaky tests from the autoscaler in the past. I did, however, add unit tests for what the autoscaler should do in terms of throttling the scaling up.

shamsimam:
Another example where we have a long starting service with 100 percent utilization. As expected constrained scaling increases queued time in this scenrio without much savings on waste.

long-start-full-utilization-comparison

Issue #395, opened 2018-08-01
Split Authenticator to decouple orthogonal functions


DaoWen:
Our current authenticator interface handles two orthogonal checks:

1. HTTP client authentication (e.g., basic auth or kerberos)
2. Permission-checking for service creation (e.g., checking for prestashed kerberos tickets in the target compute cluster)

These should be decoupled to make reuse of the client authentication portion easier when we need to change the permission-checking portion.

shamsimam:
I think the `check-user` can become part of a `(validate-service service-id)` function in the `scheduler`.

PR #398, opened 2018-08-04
adds implementation of composite-scheduler

shamsimam:

- adds an implementation of composite-scheduler
- introduces scheduler into service-description

The composite scheduler allows supporting multiple schedulers in a single cluster. It works by uniquely resolving a scheduler responsible for a service and delegating any operations on that service to the resolved scheduler.

### Example scheduler metrics output:

screen shot 2018-08-30 at 12 37 19 pm


DaoWen:
Related to #395. We'll need to decouple the schedule-specific and general pieces of the authentication interface if we want to use this with a non-trivial configuration.

waiter/src/waiter/service_description.clj
```diff
@@ -79,6 +79,7 @@
    (s/optional-key "idle-timeout-mins") (s/both s/Int (s/pred #(<= 1 % (t/in-minutes (t/days 30))) 'between-1-minute-and-30-days))
    (s/optional-key "interstitial-secs") (s/both s/Int (s/pred #(<= 0 % (t/in-seconds (t/minutes 60))) 'at-most-60-minutes))
    (s/optional-key "restart-backoff-factor") schema/positive-number-greater-than-or-equal-to-1
+   (s/optional-key "scheduler") schema/non-empty-string
```
DaoWen: What happens when this is set, but the corresponding scheduler does not exist? Similarly, what happens when this is set, but we're not using a composite scheduler?

What happens when we *are* using a composite scheduler, but this field isn't specified? Do we give a default value? How does this affect the `service-id` computation?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,128 @@
...
+    (throw (ex-info "No :factory-fn specified" component-config))))
+
+(def component-schema
+  {s/Keyword {(s/required-key :factory-fn) s/Symbol
+              s/Keyword s/Any}})
+
+(defn initialize-component-schedulers
+  "Initializes individual component schedulers."
+  [{:keys [components] :as config}]
+  {:pre [(not-empty components)]}
```
DaoWen: `not-empty` → `seq`?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,128 @@
...
+      scheduler
+      (throw (ex-info "Invalid scheduler specified!"
+                      {:available-schedulers (-> scheduler-id->scheduler keys sort)
+                       :service-id service-id
+                       :specified-scheduler scheduler-id})))))
+
+(defn invoke-component-factory
+  "Creates a component based on the factory-fn specified in the component-config."
+  [context {:keys [factory-fn] :as component-config}]
+  (if factory-fn
```
DaoWen: Doesn't the `component-schema` validation make this condition superfluous?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,128 @@
...
+
+(defrecord CompositeScheduler [service-id->scheduler scheduler-id->scheduler]
+
+  scheduler/ServiceScheduler
+
+  (get-service->instances [_]
+    (->> (vals scheduler-id->scheduler)
+         (map #(future (scheduler/get-service->instances %)))
+         (map deref)
+         (apply merge)))
```
DaoWen: I think this is too lazy. I think we should be forcing these values to ensure that the queries go off to the scheduler immediately.

Also, why use `(->> ... (map #(future ...)) (map deref) ...)` rather than `(->> ... (pmap ...) ...)`?

```clojure
(->> (vals scheduler-id->scheduler)
     (pmap scheduler/get-service->instances)
     doall
     (reduce into {})))
```

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,128 @@
...
+
+  (get-services [_]
+    (->> (vals scheduler-id->scheduler)
+         (map #(future (scheduler/get-services %)))
+         (map deref)
+         (apply concat)))
```
DaoWen: I think this is too lazy. Same as `get-service->instances` above.

waiter/src/waiter/service_description.clj
shamsimam: > What happens when this is set, but the corresponding scheduler does not exist? 

An error is reported during service creation on the scheduler (i.e. `create-service-if-new`).

> Similarly, what happens when this is set, but we're not using a composite scheduler?

I will add a defense against this scenario and repsond to this comment. 

> What happens when we are using a composite scheduler, but this field isn't specified? Do we give a default value? 

Yes, the default is configurable in the composite scheduler configuration.

> How does this affect the service-id computation?

If the `scheduler` is not explicitly provided by the user, it is not included in the `service-id` computation.

waiter/src/waiter/scheduler/composite.clj
shamsimam: Yes, will remove.

waiter/src/waiter/scheduler/composite.clj
shamsimam: Sounds good.

waiter/src/waiter/service_description.clj
DaoWen: > If the scheduler is not explicitly provided by the user, it is not included in the service-id computation.

Does that mean you end up with two difference services depending on whether you explicitly set the `scheduler` or use the default?

DaoWen:
> There is a special case for when the default is changed in which case the old services will be moved to the new scheduler. This case may need to be handled manually.

Just to make sure I understand that correctly: If I restart waiter with a modified config that changes the default scheduler, then the next time I try to ping a service that was using the old default scheduler, a new equivalent service will be started on the new default scheduler. I could still ping the old instance by manually specifying the old scheduler as my target scheduler. Is that right?

shamsimam:
@DaoWen 

> If I restart waiter with a modified config that changes the default scheduler, then the next time I try to ping a service that was using the old default scheduler, a new equivalent service will be started on the new default scheduler. 

Yes.

> I could still ping the old instance by manually specifying the old scheduler as my target scheduler. Is that right?

No, as explicitly adding the old scheduler into the service description will change the service id.

The old service can still linger around and interfere with the new service created on a different scheduler but with the same service id. A fix is to filter out services we detect that are in this category in the output of `get-service->instances` and `get-services`.

shamsimam:
@DaoWen 

> The old service can still linger around and interfere with the new service created on a different scheduler but with the same service id. 
> A fix is to filter out services we detect that are in this category in the output of get-service-instances and get-services.

Added the check to detect, kill and avoid returning such services.

DaoWen:
> No, as explicitly adding the old scheduler into the service description will change the service id.

I think it would make more sense to users if the service-id stayed the same whether you leave the scheduler out (implicitly getting the default) or explicitly set the scheduler to the default value.

shamsimam:
> I think it would make more sense to users if the service-id stayed the same whether you leave the scheduler out (implicitly getting the default) or explicitly set the scheduler to the default value. Using setting the default via the service-description-defaults might work.

Agreed, done.

waiter/src/waiter/service_description.clj
shamsimam: Reverted publishing of ids. Explicitly specifying the `scheduler` or using the default will now map to the same service id if the values of the `scheduler` parameter are the same.

DaoWen: Is there a test to make sure we don't break this equivalence?

shamsimam: Ok, I'll add one.

shamsimam: I just verified that specifying a parameter same as the default value **ends up changing the `service-id`**. Iirc, we chose to go down this path earlier as we didn't want to end up changing a user's `service-id` just because we changed defaults.

I addinged integration tests that verifies the change in service-id even when we specify defaults.

waiter/config-composite.edn
```diff
@@ -0,0 +1,51 @@
...
+                                                     :health-check-timeout-ms 200
+                                                     :port-grace-period-ms 120000
+                                                     :port-range [10000 10999]
+                                                     :work-directory "scheduler"}
+                                             :shell-light {:factory-fn waiter.scheduler.shell/shell-scheduler
+                                                           :failed-instance-retry-interval-ms 5000
+                                                           :health-check-interval-ms 5000
+                                                           :health-check-timeout-ms 200
+                                                           :port-grace-period-ms 120000
+                                                           :port-range [10000 10999]
```
sradack: The port ranges shouldn't overlap.

waiter/config-composite.edn
```diff
@@ -0,0 +1,51 @@
...
+                    :kind :composite
+                    :composite {:factory-fn waiter.scheduler.composite/create-composite-scheduler
+                                :components {:shell {:factory-fn waiter.scheduler.shell/shell-scheduler
+                                                     :failed-instance-retry-interval-ms 5000
+                                                     :health-check-interval-ms 10000
+                                                     :health-check-timeout-ms 200
+                                                     :port-grace-period-ms 120000
+                                                     :port-range [10000 10999]
+                                                     :work-directory "scheduler"}
+                                             :shell-light {:factory-fn waiter.scheduler.shell/shell-scheduler
```
sradack: Why is it called `shell-light`?

waiter/config-full.edn
```diff
@@ -233,6 +233,25 @@
                     ;; :kind :marathon uses Marathon (https://mesosphere.github.io/marathon/) for scheduling instances:
                     :kind :marathon
 
+                    ;; :kind :composite uses a composite scheduler that in this example has two component
+                    ;; schedulers: kubernetes and marathon. The composite scheduler delegates to one of these
+                    ;; schedulers based on the "schedyler" service parameters.
```
sradack: "schedyler"

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,165 @@
...
+  (let [services (scheduler/get-services scheduler)
+        valid-service? #(-> % :id service-id->scheduler (= scheduler))
+        {valid-services true invalid-services false} (group-by valid-service? services)]
+    (when (seq invalid-services)
+      (->> invalid-services
+           (map :id)
+           (process-invalid-services scheduler)))
+    (into [] valid-services)))
+
+(defrecord CompositeScheduler [service-id->scheduler scheduler-id->scheduler]
```
sradack: We need to handle unavailability of component schedulers.

shamsimam: `service-id->scheduler` throws an exception if it can't find a scheduler for a `service-id`.

waiter/config-composite.edn
shamsimam: Thanks, fixed.

waiter/config-composite.edn
shamsimam: Changed to `shell-high` and `shell0low` to reflect the relative ports they are allocated.

waiter/bin/run-using-shell-scheduler.sh
```diff
@@ -21,7 +21,7 @@ if [ "${RECOMPILE}" != "0" ]; then
   lein do clean, compile
 fi
 
-WAITER_AUTH_RUN_AS_USER=$(id -un) lein run config-shell.edn
+WAITER_AUTH_RUN_AS_USER=$(id -un) lein run config-composite.edn
```
shamsimam: I have the shell scheduler tests use the composite scheduler as it is effectively running the shell scheduler, but also enables a single specific test (`waiter.basic-test/test-composite-scheduler-services`) for the composite scheduler.

shamsimam:
@sradack integration tests have been added (they replace the old shell-scheduler tests).

waiter/config-full.edn
```diff
@@ -497,7 +518,8 @@
                                 "restart-backoff-factor" 2
                                 "scale-down-factor" 0.001
                                 "scale-factor" 1
-                                "scale-up-factor" 0.1}
+                                "scale-up-factor" 0.1
+                                "scheduler" "default"}
```
sradack: Why do we need a default that specifies the scheduler to use is the default scheduler?

waiter/src/waiter/scheduler/composite.clj
sradack: What I mean is, what if one of the schedulers is down while the other is still available?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,165 @@
...
+
+(defn service-id->scheduler
+  "Resolves the scheduler for a given service-id using the scheduler defined in the description.
+   If the scheduler is not specified, the default scheduler is returned."
+  [service-id->service-description-fn scheduler-id->scheduler default-scheduler-id service-id]
+  (let [service-description (service-id->service-description-fn service-id)
+        scheduler-id (get service-description "scheduler" default-scheduler-id)]
+    (if-let [scheduler (scheduler-id->scheduler scheduler-id)]
+      scheduler
+      (throw (ex-info "No matching scheduler found!"
```
sradack: Does this error bubble up to response?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,165 @@
...
+  (let [context (dissoc config :components)]
+    (->> components
+         (pc/map-vals #(invoke-component-factory context %))
+         (pc/map-keys name))))
+
+(defn create-composite-scheduler
+  "Creates and starts composite scheduler with components using their respective factory functions."
+  [{:keys [service-description-defaults service-id->service-description-fn] :as config}]
+  (let [scheduler-id->scheduler (initialize-component-schedulers config)
+        default-scheduler-id (get service-description-defaults "scheduler")
```
sradack: Defaults shouldn't be read directly but instead read through the service-description.  It is elsewhere's responsibility to merge in the default if not specified by the user.

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,165 @@
...
+  (state [_]
+    (pc/map-vals scheduler/state scheduler-id->scheduler)))
+
+(defn service-id->scheduler
+  "Resolves the scheduler for a given service-id using the scheduler defined in the description.
+   If the scheduler is not specified, the default scheduler is returned."
+  [service-id->service-description-fn scheduler-id->scheduler default-scheduler-id service-id]
+  (let [service-description (service-id->service-description-fn service-id)
+        scheduler-id (get service-description "scheduler" default-scheduler-id)]
+    (if-let [scheduler (scheduler-id->scheduler scheduler-id)]
```
sradack: Shouldn't this always be non-`nil` if we have a default?

waiter/src/waiter/scheduler/composite.clj
sradack: What if we wanted the "default" scheduler to be based on something else in the `service-description`?

shamsimam:
@sradack ready for review

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,234 @@
...
+    {:aggregator (query-aggregator-state-fn)
+     :components (pc/map-vals scheduler/state scheduler-id->scheduler)}))
+
+(defn service-id->scheduler
+  "Resolves the scheduler for a given service-id using the scheduler defined in the description."
+  [service-id->service-description-fn scheduler-id->scheduler default-scheduler service-id]
+  (let [service-description (service-id->service-description-fn service-id)
+        default-scheduler-id (when default-scheduler (name default-scheduler))
+        scheduler-id (get service-description "scheduler" default-scheduler-id)]
+    (if-let [scheduler (scheduler-id->scheduler scheduler-id)]
```
sradack: Should we consider a local cache for this to reduce pressure on the kv cache?

waiter/src/waiter/scheduler/composite.clj
```diff
@@ -0,0 +1,234 @@
...
+                   instances-messages (->> (vals scheduler-id->messages')
+                                           (mapcat (fn [scheduler-messages]
+                                                     (filter (fn service-instances-filter [message]
+                                                               (= :update-service-instances (first message)))
+                                                             scheduler-messages)))
+                                           doall)
+                   composite-scheduler-messages (conj instances-messages
+                                                      [:update-available-services services-message])]
+               (log/info "sending" (-> services-message :available-service-ids count) "services along scheduler-state-chan")
+               (async/>! scheduler-state-chan composite-scheduler-messages)
```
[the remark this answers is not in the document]
shamsimam: The composite scheduler remembers the last message from any component scheduler. It then aggregates the message from all schedulers and sends it along `scheduler-state-chan`. As such, the latest message on the channels is representative of the last known state from all component schedulers and so it is okay to miss previous messages on `scheduler-state-chan`.

waiter/bin/run-using-shell-scheduler.sh
sradack: We should still have a vanilla shell scheduler config.

waiter/src/waiter/scheduler/composite.clj
sradack: We discussed this offline.  We'll address when this becomes an issue.

waiter/src/waiter/scheduler/composite.clj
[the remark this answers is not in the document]
sradack: I missed that all `:update-service-instances` messages are sent, not just the messages of the individual scheduler.

waiter/bin/run-using-shell-scheduler.sh
shamsimam: `config-shell.edn` continues to exist and undergoes basic validation in a unit test. Using `config-composite.edn` allows us to test the composite scheduler with one component shell scheduler.

PR #399, opened 2018-08-07
Use utils/map->json to preserve namespaces in k8s spec keys

DaoWen:

Use the `map->json` utility function to ensure that the namespace component of Clojure keywords are preserved when converting a ReplicaSet spec to JSON in the Kuberentes scheduler.

Annotation and label names in Kubernetes objects often use slashes to denote namespaces. The default behavior of the `clojure.data.json/write-str` function is to drop the namespace component of a keyword. We need to override that behavior so that the part before the slash isn't dropped in our JSON output.

### DIFF
--- a/waiter/test/waiter/scheduler/kubernetes_test.clj
+++ b/waiter/test/waiter/scheduler/kubernetes_test.clj
@@ -12,6 +12,7 @@
   (:require [clj-time.core :as t]
             [clojure.data]
             [clojure.pprint]
+            [clojure.string :as string]
             [clojure.test :refer :all]
             [clojure.walk :as walk]
             [plumbing.core :as pc]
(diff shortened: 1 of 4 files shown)


waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -762,6 +766,23 @@
                        (scheduler/create-service-if-new dummy-scheduler descriptor))]
           (is (= service actual)))))))
 
+(deftest test-keywords-in-replicaset-spec
+  (testing "namespaced keywords in annotation keys and values correctly converted"
+    (let [service-id "test-service-id"
+          service {:service-id service-id}
+          descriptor {:service-id service-id
+                      :service-description dummy-service-description}
```
shamsimam: Nitpick: lexicographic order

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -730,16 +743,7 @@
...
-                                          "cmd" "foo"
-                                          "cpus" 1.2
-                                          "grace-period-secs" 7
-                                          "health-check-interval-secs" 10
-                                          "health-check-max-consecutive-failures" 2
-                                          "mem" 1024
-                                          "min-instances" 1
-                                          "ports" 2
-                                          "run-as-user" "myself"}}
+                    :service-description dummy-service-description}
```
shamsimam: Nitpick: lexicographic order

shamsimam:
Awaiting green build

PR #408, opened 2018-08-16
Adds available service ids in router state maintainer

shamsimam:

- adds `all-available-service-ids` in router state maintainer
- adds support for `query-fn` in router-state-maintainer
- includes `router-id` in router-state-maintainer `query-state-fn` output
- removes `state-chan` from router-state-maintainer and consumers use `query-state-fn` instead

Two goals:
1. Avoid using go-blocks to read daemon states.
2. Enable a future where the router state is obtained from the maintainer instead of from the scheduler syncer.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1126,11 +1126,11 @@
                                      [:routines prepend-waiter-url router-metrics-helpers service-id->service-description-fn]
                                      [:state entitlement-manager]
                                      wrap-secure-request-fn]
-                              (let [state-chan (get-in router-state-maintainer [:maintainer-chans :state-chan])
+                              (let [query-state-fn (get-in router-state-maintainer [:maintainer-chans :query-state-fn])
                                     {:keys [service-id->metrics-fn]} router-metrics-helpers]
                                 (wrap-secure-request-fn
                                   (fn service-list-handler-fn [request]
-                                    (handler/list-services-handler entitlement-manager state-chan prepend-waiter-url
+                                    (handler/list-services-handler entitlement-manager query-state-fn prepend-waiter-url
                                                                    service-id->service-description-fn service-id->metrics-fn
                                                                    request)))))
    :service-override-handler-fn (pc/fnk [[:curator kv-store]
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/state.clj
```diff
@@ -1180,32 +1180,34 @@
...
   "Start the instance state maintainer.
-   Maintains the state of the router as well as the state of marathon
-   and the existence of other routers. Acts as the central access point
-   for modifying this data for the router."
+   Maintains the state of the router as well as the state of marathon and the existence of other routers.
+   Acts as the central access point for modifying this data for the router.
+   Exposes the state of the router via a `query-state-fn` no-args function that is returned."
   [scheduler-state-chan router-chan router-id exit-chan service-id->service-description-fn deployment-error-config]
-  (let [state-chan (async/chan)
+  (let [state-atom (atom {:all-available-service-ids #{}
```
sradack: What do you think about keeping the initialization inside the loop, and keeping the loop such that it doesn't ever read the atom?  That seems cleaner to me.  Then just `(reset! state-atom state)` (`state` is the loop var)

waiter/src/waiter/core.clj
```diff
@@ -1126,11 +1126,11 @@
                                      [:routines prepend-waiter-url router-metrics-helpers service-id->service-description-fn]
                                      [:state entitlement-manager]
                                      wrap-secure-request-fn]
-                              (let [state-chan (get-in router-state-maintainer [:maintainer-chans :state-chan])
+                              (let [query-state-fn (get-in router-state-maintainer [:maintainer-chans :query-state-fn])
```
sradack: Feels awkward to have `:query-state-fn` inside a map called `maintainer-chans`.  It's not a `chan` anymore.

waiter/src/waiter/state.clj
shamsimam: Initializing the atom as we do current avoids us having to initialize the atom to a `nil` value and worry about invocations that retrieve the state even before the go-block has had a chance to run and reset the value in the atom.

waiter/src/waiter/core.clj
shamsimam: Renamed `maintainer-chans` to `maintainer`.

waiter/src/waiter/state.clj
sradack: That seems like a problem with the `atom` approach as written, no?  Previously, did we wait for first available value?

shamsimam: Yes, it is a side-effect of going down the atom approach where we want the atom to reflect the daemon state (including the initial state). Previously we would read from the query channel and the go-block would respond only after it got started.

shamsimam: Btw, would you be happier if I created a variable `initial-state` in the `let` block wrapping the loop and used it in the loop initialization? Then the atom could be reset to this initial state outside the go-block. An example is here: https://github.com/twosigma/waiter/pull/408/commits/41d484af4e13ca6e006a7c08e847d4533995dd17#diff-4afda66f399219b670b4e625088d4eecR1187

PR #410, opened 2018-08-16
Statsd router state maintainer

shamsimam:

- uses messages from router-state-maintainer instead of scheduler-syncer for statsd metric group 
- integration tests for metric group gauges

Two goals:
1. Avoid using go-blocks to read daemon states.
2. Enable a future where the router state is obtained from the maintainer instead of from the scheduler syncer.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -589,7 +589,11 @@
 
           ;; Waiter aggregates metrics locally and publishes
           ;; to StatsD on the following interval (milliseconds):
-          :publish-interval-ms 10000}
+          :publish-interval-ms 10000
+
+          ;; Waiter aggregates metrics locally for instances and resource usage and
+          ;; maintains them locally on the following interval (milliseconds):
+          :sync-instances-interval-ms 5000}
 
  ;; Waiter allows a "metric group" string to be associated with a service so that related
  ;; services can be grouped together in the Waiter Stats environment. One or multiple
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -591,8 +591,8 @@
           ;; to StatsD on the following interval (milliseconds):
           :publish-interval-ms 10000
 
-          ;; Waiter aggregates metrics locally for instances and resource usage and
-          ;; maintains them locally on the following interval (milliseconds):
+          ;; Waiter polls the router state for instances and resource
+          ;; usage on the following interval (milliseconds):
           :sync-instances-interval-ms 5000}
 
  ;; Waiter allows a "metric group" string to be associated with a service so that related
--- a/waiter/integration/waiter/statsd_support_test.clj
+++ b/waiter/integration/waiter/statsd_support_test.clj
@@ -32,6 +32,27 @@
   [waiter-url]
   (not= "disabled" (:statsd (waiter-settings waiter-url))))
 
+(deftest ^:parallel ^:integration-fast test-statsd-instance-metrics-aggregation
+  (testing-using-waiter-url
+    (let [metric-group (rand-name "foo")
+          headers {:x-waiter-name (rand-name)
+                   :x-waiter-metric-group metric-group}
+          {:keys [status service-id] :as response}
+          (make-request-with-debug-info headers #(make-kitchen-request waiter-url %))]
+      (is (= 200 status))
+      (when (statsd-enabled? waiter-url)
+        (let [{:keys [metric-group]} (response->service-description waiter-url response)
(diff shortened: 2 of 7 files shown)


waiter/src/waiter/statsd.clj
```diff
@@ -348,22 +348,22 @@
...
                 (gauge! metric-group "cpus" cpus)
                 (gauge! metric-group "mem" mem))
-              counts-by-metric-group)))
+              metric-group->counts)))
 
-(defn process-update-service-instances-message
-  "Merges the current map of resources by metric group with data from a :update-service-instances scheduler message"
-  [counts-by-metric-group {:keys [service-id healthy-instances unhealthy-instances failed-instances] :as message-data}
-   service-id->service-description-fn]
+(defn process-service-instance-state
```
dposada: Maybe `merge-service-state`? `process` doesn't sound too much like something that takes in a map and returns a map. And the `instance` part seems superfluous...

waiter/src/waiter/statsd.clj
```diff
@@ -375,53 +375,51 @@
...
-            ([messages]
-              (process-scheduler-messages messages service-id->service-description-fn)
-              true))]
-      (if continue
-        (recur)
-        (log/info "Stopping scheduler-metrics-publisher")))))
+      (log/error e "error processing router state"))))
+
+(defn start-service-instance-metrics-publisher
+  "Go loop to continuously publish stats based on router state updates."
```
dposada: This is no longer explicitly a `go` loop

waiter/config-full.edn
```diff
@@ -589,7 +589,11 @@
 
           ;; Waiter aggregates metrics locally and publishes
           ;; to StatsD on the following interval (milliseconds):
-          :publish-interval-ms 10000}
+          :publish-interval-ms 10000
+
+          ;; Waiter aggregates metrics locally for instances and resource usage and
+          ;; maintains them locally on the following interval (milliseconds):
```
dposada: How about "Waiter polls the router state for instances and resource usage on the following interval (milliseconds):"?

waiter/integration/waiter/statsd_support_test.clj
```diff
@@ -32,6 +32,27 @@
...
+        (let [{:keys [metric-group]} (response->service-description waiter-url response)
+              metric-group-keyword (keyword metric-group)
+              {:keys [sync-instances-interval-ms]} (get (waiter-settings waiter-url) :statsd)]
+          (wait-for
+            (fn statsd-instance-metrics-predicate []
+              (let [metric-group-gauges (-> waiter-url statsd-state :state :gauge metric-group-keyword)]
+                (log/info metric-group "counts gauges:" metric-group-gauges)
+                (every? #(contains? metric-group-gauges %)
+                        [:cpus :instances.failed :instances.healthy :instances.unhealthy :mem])))
+            :interval 1 :timeout (* 2 (quot sync-instances-interval-ms 1000)))))
```
dposada: Nitpick: I would favor the `:timeout` entry on a separate line

dposada: Nitpick: How about `(-> sync-instances-interval-ms (quot 1000) (* 2))`?

waiter/src/waiter/statsd.clj
shamsimam: Done.

waiter/src/waiter/statsd.clj
shamsimam: Updated comment

waiter/config-full.edn
shamsimam: Changed.

waiter/integration/waiter/statsd_support_test.clj
shamsimam: Done.

shamsimam: Done.

waiter/config-full.edn
```diff
@@ -589,7 +589,11 @@
 
           ;; Waiter aggregates metrics locally and publishes
           ;; to StatsD on the following interval (milliseconds):
-          :publish-interval-ms 10000}
+          :publish-interval-ms 10000
+
+          ;; Waiter polls the router state for instances and resource
+          ;; usage on the following interval (milliseconds):
+          :sync-instances-interval-ms 5000}
```
sradack: Does this belong elsewhere?  Imagine we implement some other metrics system like New Relic to which we also publish instance metrics.

PR #415, opened 2018-08-17
Cook scheduler error logging

shamsimam:

- includes response body in thrown exception
- includes job uuid in log statement while creating cook job

Improves Cook error logging.

### DIFF
--- a/waiter/src/waiter/util/http_utils.clj
+++ b/waiter/src/waiter/util/http_utils.clj
@@ -45,7 +45,7 @@
         {:keys [error status] :as response} (async/<!! raw-response)]
     (when error
       (throw error))
-    (let [response (assoc response :body async/<!!)]
+    (let [response (update response :body async/<!!)]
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/util/http_utils.clj
```diff
@@ -45,15 +45,16 @@
...
+    (let [response (update response :body async/<!!)]
+      (when (and throw-exceptions (not (<= 200 status 299)))
+        (ss/throw+ response))
+      (let [{:keys [body]} response]
+        (try
+          (cond-> body
+            (not-empty body)
+            (-> json/read-str walk/keywordize-keys))
+          (catch Exception _
+            body))))))
```
dposada: Would it be possible to write an integration test for this?

shamsimam: Added a unit test :)

PR #417, opened 2018-08-19
adds support for naming schedulers

shamsimam:

- adds support for naming schedulers

Naming scheduler simplifies the path to debugging multiple schedulers with the ability to identify individual schedulers by name.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -613,11 +613,13 @@
                        service-id->password-fn*
                        service-id->service-description-fn*]
                 (let [is-waiter-app?-fn (fn is-waiter-app? [^String service-id]
-                                          (str/starts-with? service-id service-id-prefix))]
-                  (utils/create-component scheduler-config :context {:is-waiter-app?-fn is-waiter-app?-fn
-                                                                     :leader?-fn leader?-fn
-                                                                     :service-id->password-fn service-id->password-fn*
-                                                                     :service-id->service-description-fn service-id->service-description-fn*})))
+                                          (str/starts-with? service-id service-id-prefix))
+                      scheduler-context {:is-waiter-app?-fn is-waiter-app?-fn
+                                         :leader?-fn leader?-fn
+                                         :service-id->password-fn service-id->password-fn*
+                                         :service-id->service-description-fn service-id->service-description-fn*}]
+                  (-> (utils/create-component scheduler-config :context scheduler-context)
+                      (scheduler/attach-name (-> scheduler-config :kind name)))))
    ; This function is only included here for initializing the scheduler above.
    ; Prefer accessing the non-starred version of this function through the routines map.
    :service-id->password-fn* (pc/fnk [[:state passwords]]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -619,7 +619,7 @@
                                          :service-id->password-fn service-id->password-fn*
                                          :service-id->service-description-fn service-id->service-description-fn*}]
                   (-> (utils/create-component scheduler-config :context scheduler-context)
-                      (scheduler/attach-name (-> scheduler-config :kind name)))))
+                      (scheduler/attach-name (-> scheduler-config :kind utils/keyword->str)))))
    ; This function is only included here for initializing the scheduler above.
    ; Prefer accessing the non-starred version of this function through the routines map.
    :service-id->password-fn* (pc/fnk [[:state passwords]]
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/core.clj
```diff
@@ -613,11 +613,13 @@
...
-                                                                     :leader?-fn leader?-fn
-                                                                     :service-id->password-fn service-id->password-fn*
-                                                                     :service-id->service-description-fn service-id->service-description-fn*})))
+                                          (str/starts-with? service-id service-id-prefix))
+                      scheduler-context {:is-waiter-app?-fn is-waiter-app?-fn
+                                         :leader?-fn leader?-fn
+                                         :service-id->password-fn service-id->password-fn*
+                                         :service-id->service-description-fn service-id->service-description-fn*}]
+                  (-> (utils/create-component scheduler-config :context scheduler-context)
+                      (scheduler/attach-name (-> scheduler-config :kind name)))))
```
DaoWen: I think it would be better to use our `keyword->str` function here in place of `name` so that namespaces are preserved. That way we could name schedulers things like `marathon/production` or `kubernetes/prod/1` or something if we really wanted to.

waiter/src/waiter/scheduler.clj
```diff
@@ -73,6 +73,18 @@
 (defn make-ServiceInstance [value-map]
   (map->ServiceInstance (merge {:extra-ports [] :flags #{}} value-map)))
 
+(let [name-key :waiter/scheduler-name]
+  (defn attach-name
```
DaoWen: Why is this a separate function, invoked separately from the scheduler creation? Would it be better if `utils/create-component` attached the meta-data instead?

waiter/src/waiter/scheduler.clj
```diff
@@ -73,6 +73,18 @@
 (defn make-ServiceInstance [value-map]
   (map->ServiceInstance (merge {:extra-ports [] :flags #{}} value-map)))
 
+(let [name-key :waiter/scheduler-name]
```
DaoWen: Nitpick: This binding seems strange to me. In my experience, Clojure keywords are used like global constant names. If we renamed this keyword, we'd probably want to global find/replace all occurrences of the keyword anyway.

waiter/test/waiter/scheduler_test.clj
```diff
@@ -804,3 +804,12 @@
...
           (is (= expected-state-2 actual-state-2)))))))
+
+(deftest test-scheduler-naming
+  (let [scheduler (reify ServiceScheduler)]
+    (is (nil? (scheduler->name scheduler)))
+    (let [scheduler (attach-name scheduler "foo")]
+      (is (= "foo" (scheduler->name scheduler)))
+      (let [scheduler (attach-name scheduler "bar")]
+        (is (= "bar" (scheduler->name scheduler))))
+      (is (= "foo" (scheduler->name scheduler))))))
```
DaoWen: What's the point of these last 3 lines? Why would we ever want to do that? If the name is populated when you call `create-component` then we wouldn't ever see this, right?

shamsimam: These tests are there to verify that changing the metadata returns a new object and leaves the previous object untouched.

waiter/src/waiter/core.clj
shamsimam: Sounds good. Will change.

waiter/src/waiter/scheduler.clj
shamsimam: Do we want all components to get names? This means that usages of the `scheduler` now expect the scheduler to have been created with the `create-component` method and invoke a `component->name` function to get the name. Would/Should all schedulers be created as components?

My vote would be to leave the component naming restricted to scheduler instances and generify later for other components when we need it.

waiter/src/waiter/scheduler.clj
shamsimam: The binding extracts the common name reference in the two functions enclosed. We do not expect users to have to use this keyword directly and do not expect any other occurrences to be replaced when this keyword is renamed.

waiter/src/waiter/core.clj

waiter/src/waiter/scheduler.clj
DaoWen: Does this escape into JSON anywhere? If not, then I think it would be better to use `::scheduler-name` here instead. The unit tests would help you catch if for some reason you updated one but not the other. This extra `let` binding for a keyword looks so gross...

waiter/test/waiter/scheduler_test.clj
DaoWen: My question was: Why do we care if this is implemented as meta-data or some other way? It doesn't seem like we should care about that.

shamsimam: Okay, removed the superfluous assertions.

shamsimam:
Build is now green.

PR #419, opened 2018-08-21
sorts map and vector keyword entries in router-state-maintainer

shamsimam:

- sorts map and vector keyword entries in router-state-maintainer
- uses threading macro in do-health-checks

"We" like things ordered lexicographically :)

**Note**: [Hiding whitespaces in the diff](https://github.com/twosigma/waiter/pull/419/files?w=1) might make it easier to review.

### DIFF
--- a/waiter/src/waiter/state.clj
+++ b/waiter/src/waiter/state.clj
@@ -1185,170 +1185,171 @@
    Exposes the state of the router via a `query-state-fn` no-args function that is returned."
   [scheduler-state-chan router-chan router-id exit-chan service-id->service-description-fn deployment-error-config]
   (let [state-atom (atom {:all-available-service-ids #{}
-                          :service-id->healthy-instances {}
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/scheduler.clj
```diff
@@ -447,14 +447,15 @@
...
       (if-not service
         service->service-instances'
-        (let [active-instances (doall (map async/<!! active-instances))]
-          (recur rest (assoc service->service-instances'
-                        service
-                        (assoc instances :active-instances active-instances))))))))
+        (recur rest
+               (->> active-instances
+                    (map async/<!!)
+                    doall
```
DaoWen: How about `mapv` in place of `map` + `doall`?

shamsimam: Removing this change as it breaks the code.

shamsimam:
@DaoWen  Build is green

PR #421, opened 2018-08-22
adds support for filtering by token and version in /apps endpoint

shamsimam:

- adds support for filtering by token and version

Allows users to filter the running services by token and token version.

Example search for token:
screen shot 2018-08-22 at 10 14 52 am

Example search for token version:
screen shot 2018-08-22 at 10 15 17 am

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -231,7 +231,10 @@
 (defn- str->filter-fn
   "Returns a name-filtering function given a user-provided name filter string"
   [name]
-  (let [pattern (re-pattern (str/replace (str name) #"\*+" ".*"))]
+  (let [pattern (-> (str name)
+                    (str/replace #"\." "\\\\.")
+                    (str/replace #"\*+" ".*")
+                    re-pattern)]
     #(re-matches pattern %)))
 
 (defn list-services-handler
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -232,22 +232,31 @@
   "Retrieves the list of services viewable by the currently logged in user.
    A service is viewable by the run-as-user or a waiter super-user."
   [entitlement-manager query-state-fn prepend-waiter-url service-id->service-description-fn service-id->metrics-fn request]
-  (let [current-state (query-state-fn)]
-    (let [request-params (-> request ru/query-params-request :query-params)
+  (let [{:keys [all-available-service-ids service-id->healthy-instances service-id->unhealthy-instances]} (query-state-fn)]
+    (let [{:strs [run-as-user token token-version] :as request-params} (-> request ru/query-params-request :query-params)
           auth-user (get request :authorization/user)
-          run-as-user-param (get request-params "run-as-user")
+          run-as-user-predicate (fn [{:keys [service-description service-id]}]
+                                  (if run-as-user
+                                    (some-> (get service-description "run-as-user")
+                                            (= run-as-user))
+                                    (authz/manage-service? entitlement-manager auth-user service-id service-description)))
+          source-token-predicate-factory (fn [entry-key target-value]
+                                           (if target-value
+                                             (fn [{{:strs [source-tokens]} :service-description}]
+                                               (->> source-tokens (map #(get % entry-key)) (some #(= target-value %))))
+                                             (constantly true)))
+          token-predicate (source-token-predicate-factory "token" token)
+          token-version-predicate (source-token-predicate-factory "version" token-version)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -232,22 +232,31 @@
...
-    (let [request-params (-> request ru/query-params-request :query-params)
+  (let [{:keys [all-available-service-ids service-id->healthy-instances service-id->unhealthy-instances]} (query-state-fn)]
+    (let [{:strs [run-as-user token token-version] :as request-params} (-> request ru/query-params-request :query-params)
           auth-user (get request :authorization/user)
-          run-as-user-param (get request-params "run-as-user")
+          run-as-user-predicate (fn [{:keys [service-description service-id]}]
+                                  (if run-as-user
+                                    (some-> (get service-description "run-as-user")
+                                            (= run-as-user))
+                                    (authz/manage-service? entitlement-manager auth-user service-id service-description)))
```
dposada: If I understand this correctly, if you don't specify a run-as-user param, then you see all the services you can manage, but if you do specify the run-as-user param, then you see all the services with that run-as-user. If so, doesn't that mean that I could see services I can't manage simply by specifying the run-as-user param? Is that problematic?

waiter/src/waiter/handler.clj
```diff
@@ -232,22 +232,31 @@
...
-                                        (= run-as-user run-as-user-param)
-                                        (authz/manage-service? entitlement-manager auth-user % service-description))))
-                              (->> (concat (keys (:service-id->healthy-instances current-state))
-                                           (keys (:service-id->unhealthy-instances current-state)))
-                                   (apply sorted-set)))
+                              (fn [service-id]
+                                (let [entry-map {:service-description (service-id->service-description-fn service-id :effective? false)
+                                                 :service-id service-id}
+                                      predicates [run-as-user-predicate token-predicate token-version-predicate]]
+                                  (reduce #(and %1 (%2 entry-map)) true predicates)))
```
dposada: It feels like once we have to use `%2`, we should switch to a `fn` with named args for readability

waiter/src/waiter/handler.clj
```diff
@@ -232,22 +232,31 @@
...
+          run-as-user-predicate (fn [{:keys [service-description service-id]}]
+                                  (if run-as-user
+                                    (some-> (get service-description "run-as-user")
+                                            (= run-as-user))
+                                    (authz/manage-service? entitlement-manager auth-user service-id service-description)))
+          source-token-predicate-factory (fn [entry-key target-value]
+                                           (if target-value
+                                             (fn [{{:strs [source-tokens]} :service-description}]
+                                               (->> source-tokens (map #(get % entry-key)) (some #(= target-value %))))
+                                             (constantly true)))
```
dposada: I think it would be clearer to only call the predicate if the param was passed.

waiter/src/waiter/handler.clj
```diff
@@ -232,22 +232,31 @@
...
-          run-as-user-param (get request-params "run-as-user")
+          run-as-user-predicate (fn [{:keys [service-description service-id]}]
+                                  (if run-as-user
+                                    (some-> (get service-description "run-as-user")
+                                            (= run-as-user))
+                                    (authz/manage-service? entitlement-manager auth-user service-id service-description)))
+          source-token-predicate-factory (fn [entry-key target-value]
+                                           (if target-value
+                                             (fn [{{:strs [source-tokens]} :service-description}]
+                                               (->> source-tokens (map #(get % entry-key)) (some #(= target-value %))))
```
dposada: It seems useful to be able to filter with wildcards on the token. For example, `*dpo*` to give me all tokens containing `dpo`.

dposada: Cook allows this kind of filtering on job name

waiter/src/waiter/handler.clj
```diff
@@ -228,26 +228,40 @@
...
+                                       (or (str/blank? token)
+                                           (let [filter-fn (str->filter-fn token)]
+                                             (->> source-tokens
+                                                  (map #(get % "token"))
+                                                  (some filter-fn))))
+                                       (or (str/blank? token-version)
+                                           (let [filter-fn (str->filter-fn token-version)]
+                                             (->> source-tokens
+                                                  (map #(get % "version"))
+                                                  (some filter-fn)))))))
```
dposada: Would this be cleaner with `cond->>` and `filter`, kind of like this:

https://github.com/twosigma/Cook/blob/master/scheduler/src/cook/mesos/util.clj#L448-L451

PR #424, opened 2018-08-24
Revert "adds support for filtering by token and version in /apps endp…

shamsimam:
…oint (#421)"

This reverts commit 48ddb0f

-


dposada:
Why are we reverting?

dposada:
Can we delete the branch?

PR #426, opened 2018-08-26
Makes launching the syncer the responsibility of the scheduler

shamsimam:

- uses previously define `state.scheduler-state-chan` from the plumbing core graph in scheduler syncer and router state maintainer
- removes `scheduler-state-mult-chan` in scheduler syncer
- introduces `syncer-state-atom` to store scheduler syncer state
- retrieves service scheduler state from `syncer-state-atom`
- introduces the `PollableServiceScheduler` protocol to allow scheduler to do polling scheduler syncer
- creates `start-scheduler-syncer-fn` to allow each scheduler to manage its own scheduler syncer

We would like to be able for schedulers to choose whether they send updates via polling or via event-based push notifications from the underlying scheduler. This changes created a separate protocol (`PollableServiceScheduler`) to allow scheduler implementations to choose the behavior they like and explicitly start the syncer process. In the current implementation, the syncer state is sent back to the scheduler via state updates stored in an atom.

Future scheduler implementation can choose to implement other strategies to notify of scheduler state by writing to the provided `scheduler-state-chan`.


waiter/src/waiter/core.clj
```diff
@@ -611,16 +611,21 @@
...
                        service-id->password-fn*
-                       service-id->service-description-fn*]
+                       service-id->service-description-fn*
+                       start-scheduler-syncer-fn]
                 (let [is-waiter-app?-fn (fn is-waiter-app? [^String service-id]
                                           (str/starts-with? service-id service-id-prefix))
                       scheduler-context {:is-waiter-app?-fn is-waiter-app?-fn
                                          :leader?-fn leader?-fn
+                                         :scheduler-state-chan scheduler-state-chan
+                                         ;; TODO scheduler-syncer-interval-secs should be inside the scheduler's config
```
shamsimam: I did not move the `scheduler-syncer-interval-secs` inside each individual scheduler config yet as I wanted to get a clearance on this PR first. Once this PR is in, I will make a follow-up PR with that change.

waiter/src/waiter/core.clj
```diff
@@ -943,19 +962,6 @@
                                    (let [{{:keys [query-state-fn]} :maintainer} router-state-maintainer
                                          service-gc-go-routine (partial service-gc-go-routine gc-state-reader-fn gc-state-writer-fn leader?-fn clock)]
                                      (scheduler/scheduler-broken-services-gc service-gc-go-routine query-state-fn scheduler scheduler-gc-config)))
-   :scheduler-maintainer (pc/fnk [[:routines service-id->service-description-fn]
```
shamsimam: We no longer need a separate entry for the scheduler maintainer. This also ensures that no one is reading the scheduler syncer state.

waiter/src/waiter/scheduler.clj
```diff
@@ -86,12 +85,6 @@
 
 (defprotocol ServiceScheduler
 
-  (get-service->instances [this]
```
shamsimam: This method no longer needs to be part of the scheduler interface. Instead, it moves to a separate interface which allows the scheduler to express interest in being used as a 'pollable' scheduler.

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -493,18 +497,25 @@
...
   (let [http-client (http-utils/http-client-factory http-options)
         cook-api {:http-client http-client
                   :impersonate impersonate
                   :slave-port mesos-slave-port
                   :spnego-auth (:spnego-auth http-options false)
                   :url url}
         service-id->failed-instances-transient-store (atom {})
-        scheduler (create-cook-scheduler config cook-api service-id->failed-instances-transient-store)]
+        syncer-state-atom (atom {})
+        retrieve-syncer-state-fn (partial scheduler/retrieve-syncer-state)
```
shamsimam: Wraps the access to the atom and hence limits the writers of the atom. We cannot initialize the atom inside the syncer as there is a circular dependency between the syncer and the scheduler. Using the atom allows us to break the circular dependency.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -557,7 +562,8 @@
   [{:keys [home-path-prefix http-options force-kill-after-ms framework-id-ttl mesos-slave-port
            slave-directory sync-deployment url
            ;; functions provided in the context
-           is-waiter-app?-fn leader?-fn service-id->password-fn service-id->service-description-fn]}]
+           is-waiter-app?-fn leader?-fn scheduler-state-chan scheduler-syncer-interval-secs service-id->password-fn
```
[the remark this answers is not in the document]
sradack: Sure, that's true, but will `start-scheduler-syncer-fn` every be used without passing it `scheduler-state-chan`?

waiter/src/waiter/scheduler.clj
```diff
@@ -86,12 +85,6 @@
...
-    "Returns a map of scheduler/Service records -> map of scheduler/ServiceInstance records.
-     The nested map has the following keys: :active-instances and :failed-instances.
-     The active-instances should not be assumed to be healthy (or live).
-     The failed-instances are guaranteed to be dead.")
-
   (get-services [this]
```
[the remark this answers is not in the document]
sradack: OK sure, seems better to create `PollableServiceScheduler` in that separate PR as well since we're unclear.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -577,13 +590,16 @@
         service-id->out-of-sync-state-store (atom {})
         retrieve-framework-id-fn (memo/ttl #(retrieve-framework-id marathon-api) :ttl/threshold framework-id-ttl)
         sync-deployment-maintainer-atom (atom nil)
+        syncer-state-atom (atom {})
```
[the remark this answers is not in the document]
sradack: What's the circular dependency?

waiter/src/waiter/scheduler.clj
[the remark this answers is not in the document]
shamsimam: I removed the `PollableServiceScheduler` protocol, extracted `get-service->instances` out of the scheduler and pass it in explicitly to the syncer.

waiter/src/waiter/scheduler/marathon.clj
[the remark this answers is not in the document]
shamsimam: Yes, e.g. a composite scheduler may choose to send in its own `scheduler-state-chan` instances to component schedulers and not use the one it received in the context.

waiter/src/waiter/scheduler/marathon.clj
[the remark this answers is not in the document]
shamsimam: It no longer exists after extracting `get-service->instances`

shamsimam:
@sradack ready for another round of feedback

waiter/src/waiter/scheduler.clj
```diff
@@ -528,6 +518,15 @@
         (log/info (timing-message-fn) "and found no active services")
         {:service-id->health-check-context service-id->health-check-context}))))
 
+(defn retrieve-syncer-state
+  "Retrieves the scheduler and syncer state either for the entire scheduler or when provided for a specific service."
+  ([syncer-state-atom]
+   @syncer-state-atom)
```
sradack: Why wouldn't the caller simply deref the atom itself?

waiter/src/waiter/scheduler.clj
```diff
@@ -555,12 +562,10 @@
 
                        state-query-chan
```
sradack: Do we still need a `state-query-chan` now that there is a `retrieve-syncer-state-fn`?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -471,16 +474,18 @@
       (mesos/retrieve-directory-content-from-host cook-api host log-directory)))
 
   (service-id->state [_ service-id]
-    {:failed-instances (service-id->failed-instances service-id->failed-instances-transient-store service-id)})
+    (-> (retrieve-syncer-state-fn service-id)
```
sradack: Should the syncer state be in a key like `:syncer` instead of the top level map?

shamsimam: Eagerly made the change, If you disagree, will revert.

shamsimam: @sradack ready for review

waiter/src/waiter/scheduler.clj
```diff
@@ -536,57 +522,68 @@
...
+                                (retrieve-syncer-state current-state))
+                              (async/>! response-chan))
                          current-state)
 
                        timeout-chan
                        ([]
                          (try
                            (timers/start-stop-time!
-                             (metrics/waiter-timer "state" "scheduler-sync")
+                             (metrics/waiter-timer "state" "scheduler-sync" scheduler-name)
```
sradack: Does `scheduler-name` belong before `scheduler-sync`?

shamsimam: Yes, changed. We now use `"core" "scheduler" scheduler-name` to qualify scheduler metrics.

PR #429, opened 2018-08-29
renders user metadata keys by default in token output

shamsimam:

- renders user metadata keys by default in token output

Rendering user modifiable metadata entries by default in token output avoids confusion, e.g. when using fallback support. One side-effect of this change is that the `owner` entry is always displayed, even if the user did not provide it explicitly, as it is generated and stored when not provided.

**Example output:**
screen shot 2018-08-29 at 1 14 50 pm


shamsimam:
Build is green :)

PR #430, opened 2018-08-29
Update Kitchen SSL certificate environment variables

DaoWen:

- Changes the environment variables used for providing SSL certificates files
  + `KEYSTORE` → `CERT_PATH`
  + Add optional separate `KEY_PATH` variable
  + `KEYSTORE_PASSWORD` → `KEY_PASSWORD`
- Adds a Kitchen unit test for Kitchen using HTTPS
- Adds self-signed certificate generation to the Travis-CI setup script

Since Python does not have built-in support for Java keystores, it's more accurate to use environment variable names corresponding to separate certificate, private key, and password settings. We previously called the combined certificate+private-key PEM file a "keystore", but now we use the same terms as the Python function handling the SSL-related files.

### DIFF
--- a/kitchen/tests/kitchen/conftest.py
+++ b/kitchen/tests/kitchen/conftest.py
@@ -21,14 +21,17 @@ def _find_free_port(hostname, start_port=8000, attempts=1000):
 
 
 class KitchenServer():
-    def __init__(self):
+    def __init__(self, ssl=False):
+        self.scheme = 'https' if ssl else 'http'
         self.kitchen_path = os.getenv('KITCHEN_PATH', './bin/kitchen')
         self.hostname = os.getenv('KITCHEN_HOSTNAME', 'localhost')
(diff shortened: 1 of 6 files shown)


kitchen/tests/kitchen/conftest.py
```diff
@@ -37,22 +40,29 @@ def __init__(self):
...
     def kill(self):
         if self.__server_process:
             self.__server_process.terminate()
             logging.info(f'Kitchen server has been killed')
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session")
 def kitchen_server(request):
     """Manages an instance of the Kitchen test app server."""
     server = KitchenServer()
```
scrosby: Explicitly test ssl=False (instead of the relying on the default False). Ensures that we continue to test HTTP even if we change the default. (Or, get rid of the default ssl=False and be explicit everywhere.)

PR #432, opened 2018-08-30
adds kill messages for shell scheduler instances

shamsimam:

- adds kill messages for shell scheduler instances

It helps while debugging to determine why an instance was killed.


dposada:
How/where are the messages actually consumed?

Issue #204, opened 2018-02-02
Using `binding` inside an `async/go` block is not generally safe


DaoWen:
Consider the following example<sup>[1](#foot1)</sup>:

```clojure
(def ^:dynamic *foo* :original)

(defn -main [& _]
  (let [c (clojure.core.async/to-chan [42])]
    (clojure.core.async/go
      (binding [*foo* :rebound]
        (clojure.core.async/<! c)
        (println "done."))))

  (Thread/sleep 5000))
```

The `go` macro transforms its body to create a *state machine* that can be paused/resumed on *parking* operations. That state machine is then executed on a worker thread pool. Since the `<!` operation in the above example causes the state machine to *park* (i.e., yield), the worker thread that entered the `binding` block and *pushed* the new `*foo*` value onto its dynamic-bindings stack might not be the same thread that resumes the state machine (after the `<!` operation completes), which means that the dynamic-bindings *pop* operation at the end of the `binding` block can fail loudly:

```
Exception in thread "async-dispatch-3" java.lang.IllegalStateException: Pop without matching push
```

This is a [known issue](https://dev.clojure.org/jira/browse/ASYNC-170). The issue description says that it only applies to ahead-of-time-compiled code (e.g., apps run from an UberJAR)—but it's not clear to me why it wouldn't also apply to non-AOT-compiled code.

We've seen these errors cause Waiter to crash in our production environment. We should try our best to avoid using `binding` blocks within `async/go` blocks. If we can't avoid it, then we must ensure that there are no *parking* actions within the `binding` block.

-----

<a name="foot1">[1]:</a> Example adapted from the proof-of-concept in https://dev.clojure.org/jira/browse/ASYNC-170.

DaoWen:
These all look problematic:

```
./waiter/async_request.clj
046   (async/go
047     (cid/with-correlation-id
056                 [message trigger-chan] (async/alts! [exit-chan timeout-chan] :priority true)

046   (async/go
047     (cid/with-correlation-id
063               (let [{:keys [body headers error status]} (async/<! (make-http-request))]

./waiter/scaling.clj
046     (async/go
054                   (cid/with-correlation-id
065                           (async/>! executor-chan scaling-data)

106   (async/go
107     (cid/with-correlation-id
136                         (when response-chan (async/>! response-chan kill-result))

./waiter/service.clj
226   (async/go
227     (cid/with-correlation-id
278                         (async/<! (async/timeout 1500))
```

[...]

PR #436, opened 2018-09-05
avoids switching execution to another thread inside a cid binding block

shamsimam:

- avoids switching execution to another thread inside a cid binding block

We would like to avoid fatal crashes of the Waiter router due to exceptions in the autoscaler routine.


shamsimam:
Closing for now to debug the push/pop exception further (#204 ).

PR #439, opened 2018-09-06
adds additional info for mismatch in bindings

shamsimam:

- adds additional info for mismatch in bindings

We want to make it easier to debug the binding switches.

### DIFF
--- a/waiter/src/waiter/correlation_id.clj
+++ b/waiter/src/waiter/correlation_id.clj
@@ -41,23 +41,28 @@
 (def default-correlation-id "UNKNOWN")
 (def ^:dynamic dynamic-correlation-id default-correlation-id)
 
-(defmacro with-correlation-id
-  "Executes the body with the specified value of correlation-id."
-  [correlation-id & body]
-  `(binding [dynamic-correlation-id ~correlation-id]
-     (let [start-thread-id# (.getId (Thread/currentThread))
-           result# (do ~@body)
-           end-thread-id# (.getId (Thread/currentThread))]
-       (when (not= start-thread-id# end-thread-id#)
-         (log/warn "with-correlation-id binding executed on different threads"
-                   {:correlation-id ~correlation-id
-                    :thread-ids {:end end-thread-id# :start start-thread-id#}}))
-       result#)))
-
 (defn get-correlation-id
   "Retrieve the value of the current correlation-id."
   []
   dynamic-correlation-id)
+
+(defmacro with-correlation-id
+  "Executes the body with the specified value of correlation-id."
+  [correlation-id & body]
+  `(binding [dynamic-correlation-id ~correlation-id]
+     (let [correlation-id# ~correlation-id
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/correlation_id.clj
```diff
@@ -41,24 +41,28 @@
...
+       (when (not= (.getId start-thread#) (.getId end-thread#))
          (log/warn "with-correlation-id binding executed on different threads"
-                   {:correlation-id ~correlation-id
-                    :thread-ids {:end end-thread-id# :start start-thread-id#}}))
+                   {:end {:correlation-id (get-correlation-id)
+                          :thread-id (.getId end-thread#)
+                          :thread-name (.getName end-thread#)}
+                    :start {:correlation-id ~correlation-id
+                            :thread-id (.getId end-thread#)
+                            :thread-name (.getName end-thread#)}}))
```
DaoWen: These should both be `start-thread`.

waiter/src/waiter/correlation_id.clj
```diff
@@ -41,24 +41,28 @@
...
-       (when (not= start-thread-id# end-thread-id#)
+           end-thread# (Thread/currentThread)]
+       (when (not= (.getId start-thread#) (.getId end-thread#))
          (log/warn "with-correlation-id binding executed on different threads"
-                   {:correlation-id ~correlation-id
-                    :thread-ids {:end end-thread-id# :start start-thread-id#}}))
+                   {:end {:correlation-id (get-correlation-id)
+                          :thread-id (.getId end-thread#)
+                          :thread-name (.getName end-thread#)}
+                    :start {:correlation-id ~correlation-id
```
DaoWen: I'd prefer it if we created a binding `(let [correlation-id# ~correlation-id] ...)` at the top level (outside the `binding` block) so that we don't reuse the `correlation-id` argument (which might be an expression rather than a value).

shamsimam: Done.

waiter/src/waiter/correlation_id.clj
shamsimam: Oops, fixed.

waiter/src/waiter/correlation_id.clj
```diff
@@ -41,24 +41,29 @@
...
+  "Retrieve the value of the current correlation-id."
+  []
+  dynamic-correlation-id)
+
 (defmacro with-correlation-id
   "Executes the body with the specified value of correlation-id."
   [correlation-id & body]
   `(binding [dynamic-correlation-id ~correlation-id]
-     (let [start-thread-id# (.getId (Thread/currentThread))
+     (let [correlation-id# ~correlation-id
```
DaoWen: I was suggesting this:
```
(let [correlation-id# ~correlation-id
   (binding [dynamic-correlation-id correlation-id#]
```
That way `~correlation-id` is only evaluated once.

PR #444, opened 2018-09-07
Add waiter-init script for K8s pods

DaoWen:

Add `waiter-init` script into our Kitchen image for use in K8s pods.

We need a wrapper script for the user's command in order to cleanly support the combination of the following:

* SIGTERM capture to keep pods from exiting too early and causing a race in Waiter's scale-down code.
* Redirecting stdout and stderr to files (mimicking Mesos+Marathon behavior).
* Supporting `dumb-init` in pods to get correct zombie handling (todo).

### DIFF
--- a/kitchen/bin/waiter-init
+++ b/kitchen/bin/waiter-init
@@ -1,0 +1,29 @@
+#!/bin/bash
+#
+# A wrapper script for the Waiter-specific setup for a user command in a Waiter-K8s pod.
+# The script is usually invoked by prepending it to the user's Waiter command.
+#
+# A single argument is expected, the user's command string,
+# which is executed as its own bash shell process.
+
+# Ignore SIGTERM sent by Kubernetes on pod deletion,
+# waiting for SIGKILL (force delete) before exiting.
+trap : SIGTERM
+
+# Run the user's Waiter app command,
+# copying stdout and stderr to respectively named files.
+# We tee the output so that stdout and stderr are still accessible
+# via the Kubernetes `kubectl logs <pod-name>` command.
+/bin/bash -c "$1" 1> >(tee stdout) 2> >(tee stderr 1>&2)
+exit_code=$?
+
+# If the sub-command exited 143 (usually indicated it was killed via SIGTERM),
+# then sleep for a long time (15 minutes) awaiting a SIGKILL from Kubernetes.
+# This delay gives Waiter time to safely update the desired replica count
+# before the pod actually terminates, avoiding a race to replace this pod.
+if [ $exit_code -eq 143 ]; then
+    sleep 900
+fi
+
+# Propagate the user's command's exit code
+exit $exit_code
--- a/kitchen/Dockerfile
+++ b/kitchen/Dockerfile
@@ -2,3 +2,4 @@ FROM containersol/mesos-agent:1.0.0-0.1.0
 
 RUN mkdir -p /opt/kitchen
 COPY bin/kitchen /opt/kitchen
+COPY bin/waiter-init /usr/bin/waiter-init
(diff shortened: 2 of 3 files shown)


kitchen/bin/waiter-init
```diff
@@ -0,0 +1,29 @@
...
+
+# Ignore SIGTERM sent by Kubernetes on pod deletion,
+# waiting for SIGKILL (force delete) before exiting.
+trap : SIGTERM
+
+# Run the user's Waiter app command,
+# copying stdout and stderr to respectively named files.
+# We tee the output so that stdout and stderr are still accessible
+# via the Kubernetes `kubectl logs <pod-name>` command.
+/bin/bash -c "$1" 1> >(tee stdout) 2> >(tee stderr 1>&2)
```
shamsimam: On a related note, should we consider supporting `WAITER_SANDBOX` (while temporarily keeping support for `MESOS_SANDBOX`)?

kitchen/bin/waiter-init
```diff
@@ -0,0 +1,29 @@
...
+exit_code=$?
+
+# If the sub-command exited 143 (usually indicated it was killed via SIGTERM),
+# then sleep indefinitely, awaiting a SIGKILL from Kubernetes
```
shamsimam: Should we sleep for some definite period instead of indefinitely?

kitchen/bin/waiter-init
DaoWen: Setting `WAITER_SANDBOX` for Waiter on Kubernetes is easy (we'd just do it the same way we're currently setting `MESOS_SANDBOX`).

My impression was that injecting such environment variables into the Mesos containers in Waiter on Marathon was the problem (because we don't have the value until after the container is already created), and it wouldn't make sense to run this script in a Mesos container.

shamsimam: Digression: What are the reservations for running a similar script around Mesos tasks?

DaoWen: > Digression: What are the reservations for running a similar script around Mesos tasks?

There's no problem with auto-injecting a *similar* script into Marathon commands. Just not this script, as it's tailored for Kubernetes pods.

kitchen/bin/waiter-init
DaoWen: No, I don't see any advantage to picking an arbitrary time period instead of just letting it sleep forever. In all cases, Kubernetes should send a SIGKILL to the pod's containers long before our arbitrary timeout expires, meaning that whatever limit we choose would be essentially indefinite anyway. The downside is that if we chose an arbitrary timeout that's too short, we might accidentally change something in the Waiter-K8s implementation that causes a race. It seems better to me to let this sleep forever.

shamsimam: I'd feel more at ease if we slept for a long enough period, e.g. 1 hour, that avoids concerns about the data race you mention above. Can you be more specific as to what kind of change would introduce a data race for an exited command?

DaoWen: If we set the sleep period to 15 minutes, but later change our default pod termination grace period in our ReplicaSet spec to 20 minutes, then the pod might die (due to this sleep ending) before we're expecting it to. However, if the other cleanup is taking so long that the grace-period timeout is kicking in, we're in a bad situation anyway... So I guess that's not really a concern after all.

I can't think of a benefit to picking an arbitrary cutoff here. If the pod was sigtermed by Kuberentes, then it's going to get sigkilled after the grace period expires. If the sigterm was part of a Waiter-K8s scale-down operation, then Waiter will do a force-delete (sigkill) as soon as it updates the replica count (hopefully < 1 second delay). In all other cases, Kubernetes will detect that the pod is unhealthy (health checks failing) and restart it.

DaoWen: I changed the delay to 15 minutes. That seems sufficiently long, but it's still arbitrary.

PR #445, opened 2018-09-07
Log file access in K8s pods

### LINKED ISSUE #357: Full directory-contents support for Waiter-K8s
We should expose the files inside the working directory of the pods via Waiter as we do with the Marathon scheduler implementation.

We could do this by starting a little python file server in the home directory. Upon receiving a directory-contents request, we'd start the embedded file server with some reasonable timeout (e.g., 5 minutes), and if no requests come within that period, then we kill the file server process.

DaoWen:

Adds support for the Waiter service-instance file-browsing API, mimicking the functionality we have with Waiter-Marathon. This is achieved by adding a sidecar-app running nginx and serving the contents of the main working directory (shared between the two containers).

Closes #357.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -334,6 +334,17 @@
                                                            ;; The default factory function accepts an option for the docker container to use in the pod.
                                                            :default-container-image "twosigma/kitchen:latest"}
 
+                                 ;; Configuration for creating and querying a sidecar container in each Waiter Service Instance's Kuberentes pod
+                                 ;; running a server for directory listings (in JSON) and serving file contents.
+                                 ;; Waiter expects the directory listings in the format returned by nginx's autoindex module when configured for JSON.
+                                 ;; See the behavior of the default container image below for reference.
+                                 ;; The sidecar container is only added to Waiter-managed pods if the :port number is set in this map,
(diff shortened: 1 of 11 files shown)


waiter/config-full.edn
```diff
@@ -334,6 +334,17 @@
...
                                                            :default-container-image "twosigma/kitchen:latest"}
 
+                                 ;; Configuration for creating and querying a sidecar container in each Waiter Service Instance's Kuberentes pod
+                                 ;; running a server for directory listings (in JSON) and serving file contents.
+                                 ;; Waiter expects the directory listings in the format returned by nginx's autoindex module when configured for JSON.
+                                 ;; See the behavior of the default container image below for reference.
+                                 ;; The sidecar container is only added to Waiter-managed pods if the :port number is set in this map,
+                                 ;; otherwise, the file browsing API is disabled (always returning an empty array).
+                                 :fileserver {:image "twosigma/waiter-fileserver:latest"
+                                              :port 9090
```
shamsimam: We should add a comment specifying that this port number is should not intersect with the ports starting from `pod-base-port` for instances.

DaoWen: Clarified.

PR #447, opened 2018-09-09
Skip token-syncer tests when untouched

DaoWen:

Skip the token-syncer tests in Travis-CI if both the `token-syncer` directory and the `.travis.yml` file are untouched. Only applies to PRs (i.e., all tests always run on master builds).

We waste 5 or 6 minutes on almost all of our Travis builds running the tests for the independent token syncer project, which is almost never modified.

In the unlikely event that we manage to break the token syncer behavior by modifying something in Waiter, we should still catch it when the Travis master build breaks.

### DIFF
--- a/token-syncer/bin/ci/run-all-tests.sh
+++ b/token-syncer/bin/ci/run-all-tests.sh
@@ -6,6 +6,15 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 if [ "$TRAVIS" ]; then
+    # Don't run for PRs that didn't modify the Token-Syncer
+    if [ "${TRAVIS_PULL_REQUEST}" != false ]; then
+        PR_URL="https://api.github.com/repos/${TRAVIS_REPO_SLUG}/pulls/${TRAVIS_PULL_REQUEST}/files"
+        if !(curl -s $PR_URL | jq '.[].filename' | grep -qE '^"(\.travis|token-syncer)'); then
+            echo "Skipping Token-Syncer tests on this build (project unmodified)."
+            exit 0
+        fi
+    fi
+
     # Don't exceed the Travis container 4GB max
     export JVM_OPTS="-Xmx700m"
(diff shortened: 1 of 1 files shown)


DaoWen:
I tested this patch on my fork. Here's a junk PR I opened to check that it skips when expected:
https://travis-ci.org/DaoWen/waiter/builds/426232430

Note that the Token-Syncer job only took 50 seconds on the above Travis build, whereas it takes over 6 minutes on this PR's Travis build. This PR obviously doesn't skip the token syncer tests since the script I modified is in the token-syncer directory.

It would be nice if we could get it to abort even faster, but this is a nice speedup for now.

shamsimam:
The tokens syncer can be affected by changes to Waiter also. Maybe this is more appropriate for kitchen?

DaoWen:
> The tokens syncer can be affected by changes to Waiter also.

Isn't the token syncer just a Waiter client? I.e., unless we make a breaking change to the public interface, the token syncer should be fine. (If I'm wrong about that, please let me know.)

I would prefer adding an integration test to the Waiter project that detects when we break the public interface. We can put that test in its own file, and I can update the condition in this patch so that the token syncer tests are also run when we update the Waiter interface integration test file.

> Maybe this is more appropriate for kitchen?

The Kitchen tests take virtually no time to run, so this wouldn't really free up any time there. Sometime later I want to look into restructuring our Travis build to use multiple stages. I haven't looked into all of the features yet, but it might be possible to disable jobs in later-stage builds based off things done in an earlier stage. If we can do that, then maybe we can completely disable the Kitchen job when we don't touch the Kitchen codebase. That would be nice.

shamsimam:
Token syncer tests used to take ~4 minutes, would you be okay keeping the token syncer tests if they went back to running inside 4 minutes?

DaoWen:
> Token syncer tests used to take ~4 minutes, would you be okay keeping the token syncer tests if they went back to running inside 4 minutes?

No. Why would we want to waste time on those tests for unrelated changes? Unless the tests run in < 1 minute, I still think this patch is worthwhile.

PR #448, opened 2018-09-10
Fix metric-group setting for Kitchen-based integration tests

DaoWen:

- Add `metric-group` to `kitchen-request-headers` output.
- Remove `metric-group` from service-description in token-hostname test.

Not having the `metric-group` set was causing a mismatch in all of the content-encoding tests between the "canary" request and the "real" request (i.e., it was causing two different services to be created for each, but we of course only wanted one).

Adding the `metric-group` caused `test-token-hostname` to fail, so we simply `dissoc` it there.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -411,8 +411,9 @@
           (log/info "basic hostname as token test")
           (let [current-user (retrieve-username)
                 service-description (assoc (kitchen-request-headers)
-                                      :x-waiter-permitted-user "*"
-                                      :x-waiter-run-as-user current-user)]
+                                           :x-waiter-metric-group token-prefix
+                                           :x-waiter-permitted-user "*"
+                                           :x-waiter-run-as-user current-user)]
             (testing "hostname-token-creation"
               (log/info "creating configuration using token" token)
               (let [{:keys [body status]}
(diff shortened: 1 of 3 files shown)


waiter/integration/waiter/token_request_test.clj
```diff
@@ -410,9 +410,10 @@
         (try
           (log/info "basic hostname as token test")
           (let [current-user (retrieve-username)
-                service-description (assoc (kitchen-request-headers)
-                                      :x-waiter-permitted-user "*"
-                                      :x-waiter-run-as-user current-user)]
+                service-description (-> (kitchen-request-headers)
+                                        (dissoc :x-waiter-metric-group)
```
shamsimam: Why is this `dissoc` necessary?

waiter/src/waiter/util/client_tools.clj
```diff
@@ -315,6 +315,7 @@
    :health-check-url "/status"
    :idle-timeout-mins 10
    :mem 256
+   :metric-group "waiter_kitchen"
```
shamsimam: Should we instead specify `waiter_test` so we can isolated metrics from running our tests later?

waiter/integration/waiter/token_request_test.clj
DaoWen: Changed it to `assoc :x-waiter-metric-group token-prefix` as per our discussion.

waiter/src/waiter/util/client_tools.clj
DaoWen: Renamed to `waiter_kitchen_test`.

DaoWen:
@shamsimam - Any more feedback?

PR #449, opened 2018-09-10
Rebalance k8s tests

DaoWen:

- Add a small script to periodically print the current Kubernetes pod state in our K8s integration tests.
- Fix incorrect service names for two content-encoding tests. (These were duplicate names. Looks like a copy/paste error.)
- Shuffle some tests between K8s the heavy/lite test jobs. These changes are based on the observed pod state in the integration tests. (We moved tests that obviously created many pods to *heavy*, and moved a couple tests that only created one or two pods back to *lite*.)

- The periodic printing of the Kubernetes pod state helps to diagnose why some tests fail. Specifically, it's more obvious when a 503 error is due to our minikube cluster being oversubscribed.
- Correctly marking `:resource-heavy` tests makes the tests more reliable, and helps the corresponding Travis build to finish as quickly as possible.

### DIFF
--- a/waiter/bin/ci/monitor-pods.sh
+++ b/waiter/bin/ci/monitor-pods.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+: ${cycle_delay_secs:=${1:-30}}
+
+while true; do
+    sleep ${cycle_delay_secs}
+    printf '\n***** Kubernetes pod state at %s *****\n%s\n\n' \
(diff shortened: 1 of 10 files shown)


DaoWen:
@shamsimam - Not sure why GitHub didn't pick up on the Travis build, but it's green now:
https://travis-ci.org/twosigma/waiter/builds/426842202

PR #451, opened 2018-09-11
stores source-tokens as a separate entity in the kv-store

shamsimam:

- stores source-tokens as a separate entity in the `kv-store` instead of in the service description

This re-allows services to be shared by on-the-fly requests and multiple tokens.

This PR only stores the `service-id` to source tokens mapping. We do not store the token to `service-id` mapping as it is not used currently.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -812,17 +812,13 @@
                                            (partial sd/service-id->source-tokens-entries kv-store))
    :start-new-service-fn (pc/fnk [[:scheduler scheduler]
                                   [:state authenticator start-app-cache-atom task-threadpool]
-                                  store-service-description-fn store-source-tokens-fn]
-                           (fn start-new-service [{:keys [service-id source-tokens] :as descriptor}]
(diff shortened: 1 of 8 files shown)


waiter/src/waiter/core.clj
```diff
@@ -794,24 +798,31 @@
...
-                                  store-service-description-fn]
-                           (fn start-new-service [{:keys [service-id] :as descriptor}]
+                                  store-service-description-fn store-source-tokens-fn]
+                           (fn start-new-service [{:keys [service-id source-tokens] :as descriptor}]
                              (let [run-as-user (get-in descriptor [:service-description "run-as-user"])]
                                (auth/check-user authenticator run-as-user service-id))
-                             (service/start-new-service
-                               scheduler descriptor start-app-cache-atom task-threadpool
-                               :pre-start-fn #(store-service-description-fn descriptor))))
+                             (let [pre-start-fn (fn pre-start-new-service-fn []
```
sradack: Won't `request->descriptor` be called on this path?  Is a change in this fn necessary?

PR #456, opened 2018-09-14
handles NPE due to empty service description being returned

shamsimam:

- ensures a refreshed copy is retrieved before returning an empty service description

We would like to avoid NPEs in our code paths that assume that a service description is always available for a service-id. An empty map can be returned by `service-id->service-description` due to an existing cached value from a previous lookup when the service description did not exist. This can lead to an NPE in the autoscaler for a newly created service.

Most code paths that use `service-id->service-description` expect a service description to be available. The other uses perform the lookup before performing some scheduler operation on the service. Since these invocation expect a service description to be available, we perform the force refresh to avoid returning an empty map for a service description if it exists in the kv store.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -645,6 +645,8 @@
   (defn fetch-core
     "Loads the service description for the specified service-id from the key-value store."
     [kv-store ^String service-id & {:keys [refresh nil-on-missing?] :or {refresh false nil-on-missing? true}}]
+    (when refresh
+      (log/info "force fetching core service description for" service-id))
     (let [service-description (kv/fetch kv-store (service-id->key service-id) :refresh refresh)]
       (if (map? service-description)
         service-description
@@ -834,14 +836,9 @@
 (defn service-id->service-description
   "Loads the service description for the specified service-id including any overrides."
   [kv-store service-id service-description-defaults metric-group-mappings & {:keys [effective?] :or {effective? true}}]
-  (let [service-description (fetch-core kv-store service-id :refresh false)
-        service-description (if (and (empty? service-description))
-                              (do
-                                (log/info "force refreshing fetch of service description for" service-id)
-                                (fetch-core kv-store service-id :refresh true))
-                              service-description)]
-    (cond-> service-description
-      effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id))))
+  (cond-> (or (fetch-core kv-store service-id :refresh false)
+              (fetch-core kv-store service-id :refresh true))
+    effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
 
 (defn can-manage-service?
   "Returns whether the `username` is allowed to modify the specified service description."
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -835,8 +835,14 @@
   "Loads the service description for the specified service-id including any overrides."
   [kv-store service-id service-description-defaults metric-group-mappings &
    {:keys [effective? nil-on-missing? refresh] :or {effective? true nil-on-missing? true refresh false}}]
-  (cond-> (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
-          effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
+  (let [service-description (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
+        service-description (if (and (empty? service-description) (not refresh))
+                              (do
+                                (log/info "force refreshing fetch of service description for" service-id)
+                                (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh true))
+                              service-description)]
+    (cond-> service-description
+      effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id))))
 
 (defn can-manage-service?
   "Returns whether the `username` is allowed to modify the specified service description."
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -836,8 +836,7 @@
 (defn service-id->service-description
   "Loads the service description for the specified service-id including any overrides."
   [kv-store service-id service-description-defaults metric-group-mappings & {:keys [effective?] :or {effective? true}}]
-  (cond-> (or (fetch-core kv-store service-id :refresh false)
-              (fetch-core kv-store service-id :refresh true))
+  (cond-> (fetch-core kv-store service-id :refresh false)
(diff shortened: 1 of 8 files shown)


shamsimam:
@sradack 

> Why are the forced refreshes necessary?

Updated the description.

waiter/src/waiter/service_description.clj
```diff
@@ -837,8 +837,14 @@
   "Loads the service description for the specified service-id including any overrides."
   [kv-store service-id service-description-defaults metric-group-mappings &
    {:keys [effective? nil-on-missing? refresh] :or {effective? true nil-on-missing? true refresh false}}]
```
DaoWen: Nitpick: Why is this parameter called `refresh` and not `refresh?`? Both of the other optional flags have the `?` suffix.

waiter/src/waiter/service_description.clj
```diff
@@ -837,8 +837,14 @@
...
-  (cond-> (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
-          effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
+  (let [service-description (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
+        service-description (if (and (empty? service-description) (not refresh))
+                              (do
+                                (log/info "force refreshing fetch of service description for" service-id)
+                                (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh true))
```
DaoWen: Why do we  "force refreshing fetch of service description" when `refresh` is *not* true?

Maybe the logic is backwards. But if that's the case, then why aren't the unit tests failing?

Assuming I'm reading this correctly and the logic *is* inverted, we should add a unit test for this (to verify that it fails as-is), and then fix it.

waiter/src/waiter/service_description.clj
shamsimam: I'd prefer to make the rename refactor in a separate PR, there are other instances of `refresh` in the codebase that will also need to be fixed.

waiter/src/waiter/service_description.clj
shamsimam: Doesn't "force refresh" indicate we are refreshing even though `refresh` is not true. Maybe we should check `(nil? refresh)` and default to `nil` instead of false for `refresh`? That way a caller can indicate when he doesn't want the force refresh behavior. What do you think?

shamsimam: The parameters `nil-on-missing?` and  `refresh` were unspecified from any of the callers, so ended up removing them.

waiter/src/waiter/service_description.clj
shamsimam: `refresh` no longer is a parameter to this function.

waiter/src/waiter/service_description.clj
```diff
@@ -833,10 +833,15 @@
 
 (defn service-id->service-description
   "Loads the service description for the specified service-id including any overrides."
-  [kv-store service-id service-description-defaults metric-group-mappings &
-   {:keys [effective? nil-on-missing? refresh] :or {effective? true nil-on-missing? true refresh false}}]
-  (cond-> (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
-          effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
+  [kv-store service-id service-description-defaults metric-group-mappings & {:keys [effective?] :or {effective? true}}]
+  (let [service-description (fetch-core kv-store service-id :refresh false)
+        service-description (if (and (empty? service-description))
```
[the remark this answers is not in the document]
DaoWen: Is the `empty?` check necessary here, or can we assume any non-`nil` value is OK?

If the latter, then we can refactor to get rid of the `let` block altogether:
```clojure
(cond-> (or (fetch-core kv-store service-id :refresh false)
            (do
              (log/info "force refreshing fetch of service description for" service-id)
              (fetch-core kv-store service-id :refresh true))
  effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
```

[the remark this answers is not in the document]

waiter/src/waiter/service_description.clj
```diff
@@ -645,6 +645,8 @@
   (defn fetch-core
     "Loads the service description for the specified service-id from the key-value store."
     [kv-store ^String service-id & {:keys [refresh nil-on-missing?] :or {refresh false nil-on-missing? true}}]
+    (when refresh
+      (log/info "force fetching core service description for" service-id))
```
DaoWen: 👍 I like that this logging call was moved here from `service-id->service-description`.

shamsimam:
@sradack 

> What are the implications of not caching empty maps? Is that a better solution?

We run into performance issues with that approach in the past and explicitly changed our cache code in `utils/atom-cache-get-or-load` to preserve nil values.

waiter/src/waiter/service_description.clj
```diff
@@ -833,10 +835,10 @@
 
 (defn service-id->service-description
   "Loads the service description for the specified service-id including any overrides."
-  [kv-store service-id service-description-defaults metric-group-mappings &
-   {:keys [effective? nil-on-missing? refresh] :or {effective? true nil-on-missing? true refresh false}}]
-  (cond-> (fetch-core kv-store service-id :nil-on-missing? nil-on-missing? :refresh refresh)
-          effective? (default-and-override metric-group-mappings kv-store service-description-defaults service-id)))
+  [kv-store service-id service-description-defaults metric-group-mappings & {:keys [effective?] :or {effective? true}}]
+  (cond-> (or (fetch-core kv-store service-id :refresh false)
```
sradack: Is there ever a time when we would want to continue with a `nil` value from `fetch-core` where we wouldn't want to perform refresh?  If so, does the double-checking belong in `fetch-core`?  

As I understand it, `refresh` forces a ZK `sync()`, which we'd always want to perform if we got a `nil` value.  Adding it to `fetch-core` ensures that any user of `fetch-core` benefits.

This would also make it unnecessary to make the above changes to the `suspend-or-resume-service-handler` and `override-service-handler` handlers.

shamsimam: We discussed offline and decide to handle it by:
1. introducing the refresh behavior in the router state maintainer to prevent other subsystems from receiving the nil service descriptions
2. Maintaining the checks on the handlers which can receive the service-id externally.

waiter/src/waiter/state.clj
```diff
@@ -1208,7 +1209,7 @@
           go-chan
           (async/go
             (try
-              (loop [{:keys [iteration routers] :as current-state} @state-atom]
```
shamsimam: `iteration` was unused

shamsimam:
@DaoWen  @pschorf @sradack This is ready for review after the latest round of changes.

PR #460, opened 2018-09-18
Service gc pop push issue

shamsimam:

- moves cid/with-correlation-id outside the go-block creation in service-gc-go-routine
- moves cid/with-correlation-id outside the go-block creation in transient-metrics-data-producer
- reformats handle-blacklist-request
- avoid with-correlation-id call in make-request
- moves cid/with-correlation-id outside the go-block creation in start-router-state-maintainer
- refactors monitor-async-request
- avoids cid/with-correlation-id call in service-scaling-multiplexer

We want to avoid push/pop errors due to mixing of dynamic binding and async-go blocks.

### DIFF
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -348,32 +348,32 @@
   (let [exit-chan (async/chan 1)
         query-chan (async/chan 1)
         has-metrics-except-outstanding? #(-> % (dissoc "outstanding") (not-empty))]
-    ;; launch go-block to populate service-id->metrics-chan
-    (async/go-loop [iteration 0
-                    timeout-chan (async/timeout metrics-gc-interval-ms)]
-      (let [[args channel] (async/alts! [exit-chan timeout-chan query-chan] :priority true)]
-        (condp = channel
-          exit-chan
-          (when (not= :exit args)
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/metrics.clj
```diff
@@ -348,32 +348,32 @@
...
+            (when (not= :exit args)
+              (recur (inc iteration) timeout-chan))
+
+            timeout-chan
+            (do
               (try
                 (let [service-id->metrics (utils/filterm (fn [[_ metrics]] (has-metrics-except-outstanding? metrics))
                                                          (or (service-id->metrics-fn) {}))]
                   (log/info "writing transient metrics for" (count service-id->metrics) "services")
                   (async/>! service-id->metrics-chan service-id->metrics))
```
shamsimam: This `async` operation was inside the binding block previously.

shamsimam:
I found them by manually inspecting the uses to `cid/with-correlation-id`. Since this was a manual inspection, I wouldn't be surprised if we have more problematic dynamic bindings. We need a better detection mechnaism, I'm somewhat surprised/disappointed that our detection in `cid/with-correlation-id` missed the push/pop issue in `service-gc-go-routine`.

PR #462, opened 2018-09-20
lifts with-correlation-id blocks outside async/go blocks

shamsimam:

- lifts with-correlation-id blocks outside async/go blocks

We want to reduce the chances of push/pop binding errors due to interactions with go blocks.

### DIFF
--- a/waiter/src/waiter/metrics_sync.clj
+++ b/waiter/src/waiter/metrics_sync.clj
@@ -312,18 +312,18 @@
   [router-metrics-agent local-usage-agent metrics-sync-interval-ms encrypt]
   (let [exit-chan (async/chan 1)
         query-chan (async/chan 1)]
-    (async/go-loop [iteration 0
-                    timeout-chan (async/timeout metrics-sync-interval-ms)]
-      (let [[data channel] (async/alts! [exit-chan timeout-chan query-chan] :priority true)]
-        (condp = channel
-          exit-chan
-          (when (not= :exit data)
-            (recur (inc iteration) timeout-chan))
+    (cid/with-correlation-id
+      (str "setup-metrics-syncer")
+      (async/go-loop [iteration 0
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/metrics_sync.clj
```diff
@@ -312,18 +312,18 @@
...
         query-chan (async/chan 1)]
-    (async/go-loop [iteration 0
-                    timeout-chan (async/timeout metrics-sync-interval-ms)]
-      (let [[data channel] (async/alts! [exit-chan timeout-chan query-chan] :priority true)]
-        (condp = channel
-          exit-chan
-          (when (not= :exit data)
-            (recur (inc iteration) timeout-chan))
+    (cid/with-correlation-id
+      (str "setup-metrics-syncer")
```
DaoWen: Nitpick: `(str "setup-metrics-syncer")` → `"setup-metrics-syncer"`

waiter/src/waiter/scaling.clj
```diff
@@ -480,54 +480,51 @@
...
-                              service-id->router-state' (pc/map-from-keys
-                                                          (fn [service-id]
-                                                            {:healthy-instances (count (service-id->healthy-instances service-id))
-                                                             :expired-instances (count (service-id->expired-instances service-id))})
-                                                          service-ids-set)]
-                          (cid/with-correlation-id
-                            correlation-id
+    (let [correlation-id "SCALING"]
+      (cid/with-correlation-id
+        correlation-id
```
DaoWen: Nitpick: The let block here seems superfluous.

waiter/src/waiter/metrics_sync.clj
shamsimam: Done.

waiter/src/waiter/scaling.clj
shamsimam: Done.

PR #464, opened 2018-09-25
updates core.cache and core.memoize versions to 0.7.1

shamsimam:

- updates core.cache and core.memoize versions to 0.7.1

core.cache version 0.7.1 (transitive dependency in core.memoize) avoids the stack overflow error. Please see https://dev.clojure.org/jira/browse/CCACHE-40 for details on the error.


shamsimam:
@DaoWen I went ahead and did the clojure cache upgrade before the guava cache changes as we have a transitive dependency on Clojure cache from memoize.

PR #465, opened 2018-09-25
Use guava cache instead of clojure.core.cache

shamsimam:

- introduces cache-utils.clj
- ensures use of cache-utils over clojure.core.cache in the codebase
- uses guava cache instead of clojure.core cache

The guava cache is simpler to use (and more performant) than the Clojure cache library.

In addition, we had some StackOverflowExceptions due to a [known bug](https://dev.clojure.org/jira/browse/CCACHE-40) in clojure.core.cache's FIFO caches, where `concat` was used incorrectly and easily lead to blown stacks.

I settled at version 20.0 as using newer versions of guava results in the following error:
```
ERROR in (test-zk-keys) (ListenerContainer.java:41)                                                                                                                                                       
Uncaught exception, not in assertion.                                                                                                                                     
expected: nil                                                                                                                                                                                             
  actual:  #error {                                                                                                                                                                           
 :cause "com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor()Lcom/google/common/util/concurrent/ListeningExecutorService;"                                                                
 :via                                                                                                                                                                        
 [{:type java.lang.NoSuchMethodError                                                                                                                                                                      
   :message "com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor()Lcom/google/common/util/concurrent/ListeningExecutorService;"                                                         
   :at [org.apache.curator.framework.listen.ListenerContainer addListener "ListenerContainer.java" 41]}]                                                                                                  
 :trace                                                                                                                                                                                                 
 [[org.apache.curator.framework.listen.ListenerContainer addListener "ListenerContainer.java" 41]                                                                                                         
  [org.apache.curator.framework.imps.CuratorFrameworkImpl start "CuratorFrameworkImpl.java" 257]                                                                                                          
  [waiter.kv_test$fn__60335 invokeStatic "kv_test.clj" 205]                                                                                                                                               
  [waiter.kv_test$fn__60335 invoke "kv_test.clj" 193]                                                                                                                                          
  [clojure.test$test_var$fn__7983 invoke "test.clj" 716]                  
```

### DIFF
--- a/waiter/project.clj
+++ b/waiter/project.clj
@@ -64,7 +64,8 @@
                  [org.apache.curator/curator-recipes "2.11.0"
                   :exclusions [io.netty/netty org.slf4j/slf4j-api]]
                  [org.apache.curator/curator-test "2.11.0"
-                  :exclusions [io.netty/netty]]
+                  :exclusions [com.google.guava/guava
+                               io.netty/netty]]
                  [org.apache.curator/curator-x-discovery "2.11.0"
                   :exclusions [io.netty/netty org.slf4j/slf4j-api]]
                  [org.clojure/clojure "1.8.0"]
--- a/waiter/project.clj
+++ b/waiter/project.clj
@@ -36,8 +36,8 @@
                  [twosigma/clj-http "1.0.2-20180124_201819-gcdf23e5"
                   :exclusions [commons-io org.clojure/tools.reader potemkin slingshot]]
                  [clj-time "0.12.0"
-                  :exclusions
-                  [joda-time]]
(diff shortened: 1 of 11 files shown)


shamsimam:
@DaoWen curator ends up using the method `com.google.common.util.concurrent.MoreExecutors.sameThreadExecutor()` which got removed in later versions of guava.

DaoWen:
Update: Shams tried to exclude guava from the curator dependency, but they're using some now-removed methods. Apparently upgrading our curator version is also non-trivial, so we're sticking with Guava 20 for now.

PR #470, opened 2018-10-04
Split auth protocol

DaoWen:

Split the current Authentication protocol into two pieces:
- Authentication: The part that handles client request authentication with the Waiter routers
- Authorization: The part that checks if users have the privileges to create a service on the underlying scheduler platform.

The Authorization piece remains where it lives now. The Authorization piece moves into the scheduler configuration(s).

These are orthogonal concepts that should not have been conflated. We need to split them apart so that we can give different Authorization configurations for different schedulers. This is especially important for the composite scheduler, where we'd need different Authorization for, e.g., Marathon and Kubernetes.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -265,7 +265,7 @@
                            ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
                            ;; to create services on the underlying scheduler platform.
                            ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
-                           :authorizer nil
+                           :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                            ;; The interval (milliseconds) at which instance failed jobs will be checked:
                            :failed-tracker-interval-ms 10000
@@ -337,7 +337,7 @@
                                  ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
                                  ;; to create services on the underlying scheduler platform.
                                  ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
-                                 :authorizer nil
+                                 :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                                  ;; Configuration for creating and querying a sidecar container in each Waiter Service Instance's Kuberentes pod
                                  ;; running a server for directory listings (in JSON) and serving file contents.
@@ -395,7 +395,7 @@
                                ;; to create services on the underlying scheduler platform.
                                :authorizer {
                                             ;; kerberos-prestash determines user authorization via the presence of prestashed tickets
-                                            :factory-fn waiter.auth.kerberos/kerberos-prestash-authorizer
+                                            :factory-fn waiter.auth.kerberos/kerberos-authorizer
                                             ;; the minimum interval after which prestash kerberos tickets needs to refreshed
                                             :prestash-cache-min-refresh-ms 100
                                             ;; the maximum interval after which prestash kerberos tickets are forcibly refreshed
@@ -447,6 +447,11 @@
                     ;:kind :shell
                     :shell {
                             :factory-fn waiter.scheduler.shell/shell-scheduler
+
+                            ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
+                            ;; to create services on the underlying scheduler platform.
+                            ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
+                            :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                             ;; The shell scheduler can persist its state and use this state when it restarts.
                             ;; This is an optional config, when not provided no backup state will be persisted.
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -262,6 +262,11 @@
                            ;; that returns an instance of waiter.scheduler.ServiceScheduler:
                            :factory-fn waiter.scheduler.cook/cook-scheduler
 
+                           ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
+                           ;; to create services on the underlying scheduler platform.
+                           ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
+                           :authorizer nil
+
                            ;; The interval (milliseconds) at which instance failed jobs will be checked:
                            :failed-tracker-interval-ms 10000
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -264,7 +264,8 @@
 
                            ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
                            ;; to create services on the underlying scheduler platform.
-                           ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
+                           ;; We specify the default no-op authorizer here.
+                           ;; See the :marathon :authorizer settings for a full example.
                            :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                            ;; The interval (milliseconds) at which instance failed jobs will be checked:
@@ -336,7 +337,8 @@
 
                                  ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
                                  ;; to create services on the underlying scheduler platform.
-                                 ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
+                                 ;; We specify the default no-op authorizer here.
+                                 ;; See the :marathon :authorizer settings for a full example.
                                  :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                                  ;; Configuration for creating and querying a sidecar container in each Waiter Service Instance's Kuberentes pod
@@ -450,7 +452,8 @@
 
                             ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
                             ;; to create services on the underlying scheduler platform.
-                            ;; We specify no authorizer with "nil" here. See the :marathon :authorizer settings for a full example.
+                            ;; We specify the default no-op authorizer here.
+                            ;; See the :marathon :authorizer settings for a full example.
                             :authorizer {:factory-fn waiter.authorization/noop-authorizer}
 
                             ;; The shell scheduler can persist its state and use this state when it restarts.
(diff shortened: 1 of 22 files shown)


waiter/src/waiter/auth/authentication.clj
```diff
@@ -22,20 +22,6 @@
 
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
 
-(defprotocol Authenticator
```
shamsimam: I like holding on to this protocol because it clearly specifies the signature of the function that must be supported. Maybe I am biased by years of using Java interfaces. Same comment for the `authorizer` defined in the scheduler below. It is also good for consistency, since most of our `create-component` calls return instances of protocols.

waiter/src/waiter/core.clj
```diff
@@ -814,11 +814,11 @@
    :service-id->source-tokens-entries-fn (pc/fnk [[:curator kv-store]]
                                            (partial sd/service-id->source-tokens-entries kv-store))
    :start-new-service-fn (pc/fnk [[:scheduler scheduler]
-                                  [:state authenticator start-service-cache task-thread-pool]
+                                  [:state start-service-cache task-thread-pool]
                                   store-service-description-fn]
                            (fn start-new-service [{:keys [service-id] :as descriptor}]
                              (let [run-as-user (get-in descriptor [:service-description "run-as-user"])]
-                               (auth/check-user authenticator run-as-user service-id))
+                               (scheduler/validate-user scheduler run-as-user service-id))
```
shamsimam: I would like this to have a more generic name and signature `(validate this service-id)`. We can avoid passing the `run-as-user` as the scheduler can retrieve it from the service description.

waiter/src/waiter/scheduler.clj
```diff
@@ -108,7 +108,17 @@
...
+
+  (validate-user [this ^String user ^String service-id]
+    "Verify that a user has the proper permissions to create services on the underlying scheduler platform."))
+
+(defn make-authorizer-fn
+  "Create an authorizer-fn via create-component, providing a default if the passed-in authorizer is nil."
+  [authorizer]
+  (if authorizer
+    (utils/create-component authorizer)
+    (constantly :OK)))
```
shamsimam: Why return `:OK` and not true. What is the signature expected from the component created by `make-authorizer-fn`?

waiter/src/waiter/scheduler.clj
```diff
@@ -108,7 +108,17 @@
     "Retrieves the state the scheduler is maintaining for the given service-id.")
 
   (state [this]
-    "Returns the global (i.e. non-service-specific) state the scheduler is maintaining"))
+    "Returns the global (i.e. non-service-specific) state the scheduler is maintaining")
+
+  (validate-user [this ^String user ^String service-id]
+    "Verify that a user has the proper permissions to create services on the underlying scheduler platform."))
```
shamsimam: Docstring should clearly state that an exception will be thrown if validation fails.

waiter/src/waiter/util/utils.clj
```diff
@@ -441,11 +441,13 @@
   "Creates a component based on the specified :kind"
   [{:keys [kind] :as component-config} & {:keys [context]}]
   (log/info "component:" kind "with config" component-config (if context (str "and context " context) ""))
-  (let [kind-config (get component-config kind)
-        factory-fn (:factory-fn kind-config)]
+  (let [config (if kind
```
shamsimam: I think we should not change this and have consistency in how we use kind to create components.

waiter/config-full.edn
```diff
@@ -160,20 +160,15 @@
                         ;; :kind :one-user allows anonymous access to services (for testing purposes only):
                         :kind :one-user
                         :one-user {;; Custom implementations should specify a :factory-fn
-                                   ;; that returns an instance of waiter.auth.authentication.Authenticator:
+                                   ;; that returns an authentication middleware function.
+                                   ;; See the one-user-authenticator reference below for an example.
```
shamsimam: This comment should not change.

waiter/src/waiter/util/utils.clj
DaoWen: Adding *another* sub-map just for the sake of inserting `:kind` seems very cumbersome. I don't like it, but I don't like this either. I'll change it to use `:kind` for now.

waiter/config-full.edn
```diff
@@ -431,6 +451,13 @@
                     :shell {
                             :factory-fn waiter.scheduler.shell/shell-scheduler
 
+                            ;; The authorizer is used by the scheduler to verify that a user has the proper permissions
+                            ;; to create services on the underlying scheduler platform.
+                            ;; We specify the default no-op authorizer here.
+                            ;; See the :marathon :authorizer settings for a full example.
+                            :authorizer {:kind :default
```
shamsimam: Does shell scheduler need an authorizer?

waiter/src/waiter/scheduler.clj
```diff
@@ -108,7 +108,11 @@
     "Retrieves the state the scheduler is maintaining for the given service-id.")
 
   (state [this]
-    "Returns the global (i.e. non-service-specific) state the scheduler is maintaining"))
+    "Returns the global (i.e. non-service-specific) state the scheduler is maintaining")
+
+  (validate-service [this ^String service-id]
+    "Verify that a user has the proper permissions to create services on the underlying scheduler platform.
```
shamsimam: Maybe rephrase to:
> Verify creating a services on the underlying scheduler platform, e.g. by checking that the run-as-user has the proper permissions.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -459,7 +461,7 @@
       (catch [:status 404] _
         (comment "App does not exist."))))
 
-  (create-service-if-new [this {:keys [service-id] :as descriptor}]
+  (create-service-if-new [this {:keys [run-as-user service-id] :as descriptor}]
```
shamsimam: `run-as-user` should be removed.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -666,7 +672,7 @@
 (defn kubernetes-scheduler
   "Returns a new KubernetesScheduler with the provided configuration. Validates the
    configuration against kubernetes-scheduler-schema and throws if it's not valid."
-  [{:keys [authentication http-options max-patch-retries max-name-length orchestrator-name
+  [{:keys [authentication authorizer http-options max-patch-retries max-name-length orchestrator-name
```
shamsimam: Should we add schema validation for `authorizer` in the pre-condition?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -477,11 +479,15 @@
...
+     :syncer (retrieve-syncer-state-fn)})
+
+  (validate-service [_ service-id]
+    (let [{:strs [run-as-user]} (service-id->service-description-fn service-id)]
+      (authz/check-user authorizer run-as-user service-id))))
 
 (s/defn ^:always-validate create-cook-scheduler
   "Returns a new CookScheduler with the provided configuration."
-  [{:keys [allowed-users backend-port home-path-prefix instance-priorities search-interval-days
+  [{:keys [allowed-users authorizer backend-port home-path-prefix instance-priorities search-interval-days
```
shamsimam: Should we add schema validation for `authorizer` in the pre-condition?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -560,8 +565,8 @@
 (defn marathon-scheduler
   "Returns a new MarathonScheduler with the provided configuration.
    Validates the configuration against marathon-scheduler-schema and throws if it's not valid."
-  [{:keys [home-path-prefix http-options force-kill-after-ms framework-id-ttl mesos-slave-port
-           slave-directory sync-deployment url
+  [{:keys [authorizer home-path-prefix http-options force-kill-after-ms framework-id-ttl
```
shamsimam: Should we add schema validation for `authorizer` in the pre-condition?

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -678,18 +679,23 @@
...
+  (validate-service [_ service-id]
+    (let [{:strs [run-as-user]} (service-id->service-description-fn service-id)]
+      (authz/check-user authorizer run-as-user service-id))))
 
 (s/defn ^:always-validate create-shell-scheduler
   "Returns a new ShellScheduler with the provided configuration. Validates the
   configuration against shell-scheduler-schema and throws if it's not valid."
-  [{:keys [failed-instance-retry-interval-ms health-check-interval-ms health-check-timeout-ms
-           port-grace-period-ms port-range work-directory
+  [{:keys [authorizer failed-instance-retry-interval-ms health-check-interval-ms
```
shamsimam: Should we add schema validation for `authorizer` in the pre-condition?

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -826,8 +835,9 @@
 
 (defn shell-scheduler
   "Creates and starts shell scheduler with loops"
-  [{:keys [backup-file-name failed-instance-retry-interval-ms health-check-interval-ms health-check-timeout-ms port-grace-period-ms port-range
-           scheduler-name scheduler-state-chan scheduler-syncer-interval-secs start-scheduler-syncer-fn] :as config}]
+  [{:keys [backup-file-name failed-instance-retry-interval-ms health-check-interval-ms health-check-timeout-ms
```
shamsimam: Nitpick: Please undo as noting appears to have changed here.

waiter/src/waiter/scheduler.clj
DaoWen: Done.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Removed.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Done. (`contains-kind-sub-map?`)

waiter/src/waiter/scheduler/cook.clj
DaoWen: Done.

waiter/src/waiter/scheduler/marathon.clj
DaoWen: Done.

waiter/src/waiter/scheduler/shell.clj
DaoWen: Undone.

PR #477, opened 2018-10-18
Correct task-stats numbers for K8s scheduler

DaoWen:

- Negate the `staged` task count for K8s (currently comes out negative)
- Don't include `staged` tasks in the `running` count

Negative staged counts were obviously wrong, and staged tasks shouldn't be included in the running-task count.

The counts make more sense now:
```
running == healthy + unhealthy
replicas == running + staged
```


DaoWen:
@shamsimam - Unit test added for the case when we have a non-zero staged task count.

PR #489, opened 2018-10-29
adds class name check in addition to instanceof check

shamsimam:
to avoid classloading issues

- adds class name check in addition to instanceof check to avoid classloading issues

Waiter fails to startup when the `chan?` calls fail.


shamsimam:
Temporarily addresses #492

PR #494, opened 2018-11-01
Add support for content-length bigger than an int

nsinkov:

Add support for content-length bigger than an int. Use Long.parseLong when parsing content-length header string

We can get a request with content-length bigger than int can hold

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -222,7 +222,7 @@
         waiter-debug-enabled? (utils/request->debug-enabled? request)]
     (try
       (let [content-length-str (get passthrough-headers "content-length")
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -222,7 +222,7 @@
         waiter-debug-enabled? (utils/request->debug-enabled? request)]
     (try
       (let [content-length-str (get passthrough-headers "content-length")
-            content-length (if content-length-str (Integer/parseInt content-length-str) 0)]
+            content-length (if content-length-str (Long/parseLong content-length-str) 0)] ;xxx
```
shamsimam:
```suggestion
            content-length (if content-length-str (Long/parseLong content-length-str) 0)]
```

PR #495, opened 2018-11-01
remove secrun endpoint

nsinkov:

- remove /secrun endpoint and functionality

not used

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -168,9 +168,7 @@
 (defn request->endpoint
   "Retrieves the relative url Waiter should use to forward requests to service instances."
   [{:keys [uri]} waiter-headers]
-  (if (= "/secrun" uri)
-    (headers/get-waiter-header waiter-headers "endpoint-path" "/req")
-    uri))
+  uri)
 
 (defn- handle-response-error
   "Handles error responses from the backend."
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -168,9 +168,7 @@
 (defn request->endpoint
```
shamsimam: Is this function needed anymore. Can we inline the `uri` at its use?

waiter/test/waiter/process_request_test.clj
```diff
@@ -35,40 +35,25 @@
...
-        waiter-headers {(str headers/waiter-header-prefix "endpoint-path") custom-legacy-endpoint}
-        test-endpoints (clojure.set/union legacy-endpoints passthrough-endpoints)]
+  (let [passthrough-endpoints #{"/foo" "/baz/bar" "/load/balancer" "/auto/scale/1/2/3"}
+        waiter-headers {}
+        test-endpoints (clojure.set/union passthrough-endpoints)]
     (doseq [item test-endpoints]
       (testing (str "Test retrieve endpoint with headers and query string: " item)
         (let [dummy-request (assoc (request item :post {:a 1 :b 2}) :query-string "foo=bar&baz=1234")
-              expected-endpoint (if (contains? legacy-endpoints item) custom-legacy-endpoint item)]
+              expected-endpoint item]
```
shamsimam: Maybe rename `item` to `expected-endpoint` in the `doseq` above?

shamsimam: Actually if we inline, we can delete this test :)

waiter/src/waiter/process_request.clj
nsinkov: inlined

PR #499, opened 2018-11-06
adds support for last-update-time in the token index

shamsimam:

- adds support for last-update-time in the token index

Includes the token last update time in the token index metadata.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -84,11 +84,7 @@
                              (let [new-owner-key (new-owner-key)]
                                (log/info "storing" new-owner-key "for" owner "in the token-owners-key")
                                (kv/store kv-store token-owners-key (assoc owner->owner-key owner new-owner-key))
-                               new-owner-key)))
-      delete-token-from-index (fn delete-token-from-index [index-entries token-to-remove]
-                                (dissoc index-entries token-to-remove))
-      insert-token-into-index (fn insert-token-into-index [index-entries token-to-insert index-entry]
-                                (assoc index-entries token-to-insert index-entry))]
+                               new-owner-key)))]
 
   (defn store-service-description-for-token
     "Store the token mapping of the service description template in the key-value store."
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/token.clj
```diff
@@ -213,10 +227,9 @@
                                      (fn [tokens]
                                        (pc/map-from-keys
                                          (fn [token]
-                                           (let [{:strs [deleted] :as token-data} (kv/fetch kv-store token)
+                                           (let [{:strs [deleted last-update-time] :as token-data} (kv/fetch kv-store token)
                                                  token-hash (sd/token-data->token-hash token-data)]
-                                             {:deleted (true? deleted)
-                                              :etag token-hash}))
+                                             (make-index-entry token-hash (true? deleted) last-update-time)))
```
DaoWen: `true?` here seems redundant since it's also invoked inside `make-index-entry`. Can we remove it?

waiter/src/waiter/token.clj
```diff
@@ -499,7 +512,9 @@
                              (fn [[token entry]]
                                (cond-> (assoc entry :owner owner :token token)
                                  (not show-metadata)
-                                 (dissoc :deleted :etag)))))))
+                                 (dissoc :deleted :etag :last-update-time)
+                                 show-metadata
+                                 (update :last-update-time tc/from-long)))))))
```
DaoWen: I think `cond->` here is odd, since the conditions are `x` and `not x`.

```clojure
(->
  (if show-metadata
    (update entry :last-update-time tc/from-long)
    (dissoc entry :deleted :etag :last-update-time))
  (assoc :owner owner :token token))
```

waiter/src/waiter/token.clj
```diff
@@ -79,8 +87,8 @@
                                new-owner-key)))
       delete-token-from-index (fn delete-token-from-index [index-entries token-to-remove]
                                 (dissoc index-entries token-to-remove))
-      insert-token-into-index (fn insert-token-into-index [index-entries token-to-insert token-hash deleted]
-                                (assoc index-entries token-to-insert {:deleted (true? deleted) :etag token-hash}))]
+      insert-token-into-index (fn insert-token-into-index [index-entries token-to-insert index-entry]
+                                (assoc index-entries token-to-insert index-entry))]
```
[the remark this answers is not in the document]
shamsimam: I made the change and then I liked your original (offline) suggestion better and inlined the `assoc` and `dissoc` calls.

waiter/src/waiter/token.clj
shamsimam: Removed.

waiter/src/waiter/token.clj

PR #503, opened 2018-11-09
Throw on K8s streaming api responses with error status

DaoWen:

-


DaoWen:
@shamsimam - Build is green! :smile:

PR #505, opened 2018-11-13
Make scaling operation async friendly

shamsimam:

- renames task-thread-pool to start-service-thread-pool
- executes execute-scale-service-request on its own thread to avoid blocking async threads
- executes execute-scale-down-request during scaling on its own thread to avoid blocking async threads

We would like to avoid blocking async threads that can lead to a deadlock with our implementation of synchronous requests using the jet library.

### Outstanding question

Should all the scaling scheduler interactions (create, scale, kill and delete) move to a common thread pool? If yes, they can be part of a follow-up PR.

### Why didn;t we try making the http calls to the scheduler full async?

An example call stack:
```
scaling/service-scaling-executor [go-block]
-> scaling/execute-scale-service-request
  -> scheduler/scale-service 
   -> extract-service-deployment-info
    -> marathon/get-deployments
   -> marathon/delete-deployment
   -> marathon/get-app
   -> marathon/update-app
```
Introducing async means each of these call paths need to be converted to async and need to handle throwing/returning exceptions.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -515,13 +515,13 @@
                 (cond->> (utils/unique-identifier)
                   (not (str/blank? router-id-prefix))
                   (str (str/replace router-id-prefix #"[@.]" "-") "-")))
-   :scale-service-thread-pool (pc/fnk [] (Executors/newFixedThreadPool 20))
    :scaling-timeout-config (pc/fnk [[:settings
                                      [:blacklist-config blacklist-backoff-base-time-ms max-blacklist-time-ms]
                                      [:scaling inter-kill-request-wait-time-ms]]]
                              {:blacklist-backoff-base-time-ms blacklist-backoff-base-time-ms
                               :inter-kill-request-wait-time-ms inter-kill-request-wait-time-ms
                               :max-blacklist-time-ms max-blacklist-time-ms})
+   :scheduler-interactions-thread-pool (pc/fnk [] (Executors/newFixedThreadPool 20))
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/scaling.clj
```diff
@@ -261,19 +262,25 @@
...
                           (counters/inc! (metrics/service-counter service-id "scaling" "scale-down" "total"))
                           (if (or (nil? last-scale-down-time)
                                   (t/after? (t/now) (t/plus last-scale-down-time inter-kill-request-wait-time-in-millis)))
-                            (if (or (async/<!
-                                      (execute-scale-down-request
-                                        notify-instance-killed-fn peers-acknowledged-blacklist-requests-fn
-                                        scheduler instance-rpc-chan timeout-config
-                                        service-id iter-correlation-id num-instances-to-kill response-chan))
+                            (if (or (-> (fn []
+                                          (-> (execute-scale-down-request
```
pschorf: I think the `->` is overkill here

waiter/src/waiter/core.clj
```diff
@@ -515,6 +515,7 @@
                 (cond->> (utils/unique-identifier)
                   (not (str/blank? router-id-prefix))
                   (str (str/replace router-id-prefix #"[@.]" "-") "-")))
+   :scale-service-thread-pool (pc/fnk [] (Executors/newFixedThreadPool 20))
```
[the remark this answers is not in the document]

waiter/src/waiter/scaling.clj
shamsimam: Moved the function inside `au/execute`

PR #507, opened 2018-11-14
adds support for star in the run-as-user filter on /apps endpoint

shamsimam:

- adds support for `*` in the `run-as-user` filter on the `/apps` endpoint

We would like to be able to explicitly search and see all services running on Waiter.


shamsimam:
@dpo addressed feedback. Build is green too.

PR #509, opened 2018-11-14
Enforces limit on length of the command string

shamsimam:

- enforces validation limit on cmd - must be no larger than 1K characters
- issues friendly error message when command string validation fails

We would like to prevent users from sending huge command strings as inputs with a goal of avoiding stress on our backing store.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -52,7 +52,7 @@
 
 (def service-description-schema
   {;; Required
-   (s/required-key "cmd") (s/both s/Str (s/pred #(<= 1 (count %) 1024)'at-most-1K-chars))
+   (s/required-key "cmd") (s/both s/Str (s/pred #(<= 1 (count %) 1024) 'at-most-1K-chars))
    (s/required-key "cpus") schema/positive-num
    (s/required-key "mem") schema/positive-num
    (s/required-key "run-as-user") schema/non-empty-string
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -52,7 +52,7 @@
 
 (def service-description-schema
   {;; Required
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -52,7 +52,7 @@
 
 (def service-description-schema
   {;; Required
-   (s/required-key "cmd") schema/non-empty-string
+   (s/required-key "cmd") (s/both s/Str (s/pred #(<= 1 (count %) 1024)'at-most-1K-chars))
```
DaoWen: Missing a space before the quoted symbol `'at-most-1K-chars`.

DaoWen: Maybe maximum length should be configurable, but with the default of 1024.

waiter/test/waiter/token_test.clj
```diff
@@ -1179,6 +1179,24 @@
           (is (= 400 status))
           (is (str/includes? body "Cannot modify previous token metadata"))))
 
+      (testing "post:new-service-description:bad-token-command"
+        (let [kv-store (kv/->LocalKeyValueStore (atom {}))
+              service-description (walk/stringify-keys
+                                    {:cmd (str "tc " (str/join "" (repeat 1024 "x")))
```
DaoWen:
```suggestion
                                    {:cmd (apply str "tc " (repeat 1024 "x"))
```

waiter/src/waiter/service_description.clj
DaoWen: Add missing space:
```suggestion
   (s/required-key "cmd") (s/both s/Str (s/pred #(<= 1 (count %) 1024) 'at-most-1K-chars))
```

shamsimam: I will introduce it as a `constraint`.

PR #511, opened 2018-11-15
Include waiter in the user-agent header

shamsimam:

- consolidates the http-client factory method
- adds support for custom user-agent field
- configures user-agent string per http client

It is considered good practice to include a custom user-agent as it can help with debugging.

With the change:
```
$ curl -s -H 'User-Agent:' -H"x-waiter-token: kitchen.localtest.me" http://127.0.0.1:9091/request-info | jq .
{
  "headers": {
    "user-agent": "waiter/59a650a",
...
}

$ curl -s -H"x-waiter-token: kitchen.localtest.me" http://127.0.0.1:9091/request-info | jq .
{
  "headers": {
    "user-agent": "curl/7.49.1",
...
}
```


dposada:
Should we use something more meaningful than `1.0` for the version part?

shamsimam:
@dposada Will try and use the commit hash.

PR #519, opened 2018-11-28
adds support for service status

shamsimam:

- adds support for service status

We would like to explicit report the status of a service.

service-status


waiter/src/waiter/service.clj
```diff
@@ -293,3 +293,19 @@
...
                  ^Runnable (fn [] (cid/with-correlation-id correlation-id (start-fn)))))
       (log/info service-id "has been started on another thread" cache-value))))
+
+(defn resolve-service-status
+  "Determines the service status at any point in time.
+   A service can be one of the following states:
+   - Starting: the service has no healthy instances and is starting one up,
+   - Running: the service is successfully running at its desired scale,
+   - Failing: the service has instances failing to start,
+   - Waiting: the service is waiting to be started,
```
dposada: How about "Idle" instead of "Waiting"?

waiter/src/waiter/service.clj
```diff
@@ -293,3 +293,19 @@
...
+   - Scaling: the service is being scaled."
```
DaoWen: Having both "Running" and "Scaling" seems confusing. We probably shouldn't differentiate between these two states in the Status.

waiter/src/waiter/service.clj
shamsimam: Done.

waiter/src/waiter/service.clj
shamsimam: Used a single "Running" status.

PR #520, opened 2018-11-28
removes reference to jobsystem

shamsimam:

- removes reference to jobsystem in message

Avoids any confusion about jobsystem tickets.


shamsimam:
@dpo addressed merge conflict

PR #529, opened 2018-12-06
upgrades curator and guava versions

shamsimam:

- upgrades curator and guava versions

Running the latest versions gives us performance improvements and bug fixes (hopefully).

**Note**: Curator 4.0 has a hard dependency on ZooKeeper 3.5.x


dposada:
Can we delete the branch?

PR #532, opened 2018-12-07
reduces error logs for client error scenarios

shamsimam:

- reduces error logs for client error scenarios
- removes the `suppress-logging` flag

Avoids corrupting the log files with exception stack traces.

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -256,18 +256,19 @@
   (let [{:keys [status]} (ex-data ex)]
     (if status
       ex
-      (ex-info "Internal error" {:status 500} ex))))
+      (ex-info (str "Internal error: " (.getMessage ex)) {:status 500} ex))))
 
 (defn exception->response
   "Converts an exception into a ring response."
   [^Exception ex request]
   (let [wrapped-ex (wrap-unhandled-exception ex)
-        {:keys [friendly-error-message headers message status suppress-logging] :as data} (ex-data wrapped-ex)
+        {:keys [friendly-error-message headers logging message status] :as data} (ex-data wrapped-ex)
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -256,18 +256,19 @@
...
       ex
-      (ex-info "Internal error" {:status 500} ex))))
+      (ex-info (str "Internal error: " (.getMessage ex)) {:status 500} ex))))
 
 (defn exception->response
   "Converts an exception into a ring response."
   [^Exception ex request]
   (let [wrapped-ex (wrap-unhandled-exception ex)
-        {:keys [friendly-error-message headers message status suppress-logging] :as data} (ex-data wrapped-ex)
+        {:keys [friendly-error-message headers logging message status] :as data} (ex-data wrapped-ex)
```
dposada: Maybe `log-level` instead of `logging`?

waiter/src/waiter/util/utils.clj
```diff
@@ -256,18 +256,19 @@
...
         response-msg (if (or message friendly-error-message)
                        (str/trim (str message \newline friendly-error-message))
                        (.getMessage wrapped-ex))
         processed-headers (into {} (for [[k v] headers] [(name k) (str v)]))]
-    (when-not suppress-logging
+    (condp = logging
+      :info (log/info (.getMessage wrapped-ex))
```
dposada: Should we add `:warn` support?

waiter/src/waiter/util/utils.clj
shamsimam: Done.

waiter/src/waiter/util/utils.clj
shamsimam: Done.

PR #533, opened 2018-12-13
 adds changes suggested by kibit

shamsimam:

- adds changes suggested by kibit

Get us closer to using Clojure idioms.

### DIFF
--- a/waiter/src/waiter/cookie_support.clj
+++ b/waiter/src/waiter/cookie_support.clj
@@ -61,8 +61,7 @@
   "Inserts the provided name-value pair as a Set-Cookie header in the response"
   [response password name value age-in-days]
   (letfn [(add-cookie-into-response [response]
-            (let [encoded-cookie (-> (encode-cookie value password)
-                                     UrlEncoded/encodeString)
+            (let [encoded-cookie (UrlEncoded/encodeString (encode-cookie value password))
                   max-age (-> age-in-days t/days t/in-seconds)
                   path "/"
                   set-cookie-header (str name "=" encoded-cookie ";Max-Age=" max-age ";Path=" path ";HttpOnly=true")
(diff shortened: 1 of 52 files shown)


waiter/src/waiter/statsd.clj
```diff
@@ -234,10 +233,9 @@
       "Publishes a Statsd timer metric for each of the 50th, 75th, 95th,
       and 100th percentiles of the values that have been locally aggregated"
       [metric-group metric values]
-      (dorun
-        (map (fn [[k v]]
-               (gauge #(metric-path metric-group (str metric "_p" k)) v))
-             (percentiles values 50 75 95 99 100))))
+      (run! (fn [[k v]]
```
DaoWen: I would have preferred `doseq` here.

DaoWen: I think it's more clear to use `doseq` rather than `run!` + an anonymous `fn`, but I guess kibit's heuristics are just looking for `(dorun (map ....))`.

shamsimam: There is a performance implication here as `(do (map ...))` will return a sequence with the return values of invoking the nested function. `run!` just returns `nil`.

DaoWen: Right. This is a good change—I just think `doseq` would be even better. But I know these are automatic from kibit. I'm not asking you to change it.

waiter/src/waiter/cookie_support.clj
```diff
@@ -61,8 +61,7 @@
   "Inserts the provided name-value pair as a Set-Cookie header in the response"
   [response password name value age-in-days]
   (letfn [(add-cookie-into-response [response]
-            (let [encoded-cookie (-> (encode-cookie value password)
-                                     UrlEncoded/encodeString)
+            (let [encoded-cookie (UrlEncoded/encodeString (encode-cookie value password))
```
pschorf: I'm surprised you like this change :)

PR #536, opened 2018-12-14
throttles the number of simultaneous kerberos library invocations during authentication

shamsimam:

- throttles the number of simultaneous kerberos library invocations during authentication
- reports a 503 temporarily unavailable when there are too many pending kerberos authentications

This prevents a router from crashing due to too many concurrent calls to the GSS library. The concurrency level controls the number of threads present in the Thread pool and hence throttles the number of concurrent calls to the GSS library. Using the custom thread pool also offloads the relatively expensive native GSS calls to not block the `core.async` threads. In addition, the PR also responds with `503`s to requests that cause the pending kerberos authentication requests to grow beyond a pre-configured limit.

The [Semaphore](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html) based [solution](https://github.com/twosigma/waiter/tree/e3b803368623e85532c85dd96e08c6b6dc0ff892) looks much simpler but can potentially block async threads waiting on the `acquire` call. As this scenario can lead to poor performance in processing requests through Waiter, the [Semaphore](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html) based approach was avoided.

Screenshot of Kerberos metrics:
image

### DIFF
--- a/waiter/src/waiter/auth/kerberos.clj
+++ b/waiter/src/waiter/auth/kerberos.clj
@@ -117,10 +117,16 @@
          (integer? keep-alive-mins)
          (pos? keep-alive-mins)]}
   (let [thread-pool (ThreadPoolExecutor. 1 concurrency-level keep-alive-mins TimeUnit/MINUTES (LinkedBlockingQueue.))]
-    (metrics/waiter-gauge #(.getActiveCount thread-pool) "core" "kerberos" "throttle" "active-thread-count")
-    (metrics/waiter-gauge #(.getMaximumPoolSize thread-pool) "core" "kerberos" "throttle" "max-thread-count")
-    (metrics/waiter-gauge #(-> thread-pool .getQueue .size) "core" "kerberos" "throttle" "pending-task-count")
-    (metrics/waiter-gauge #(.getTaskCount thread-pool) "core" "kerberos" "throttle" "scheduled-task-count")
+    (metrics/waiter-gauge #(.getActiveCount thread-pool)
+                          "core" "kerberos" "throttle" "active-thread-count")
+    (metrics/waiter-gauge #(- concurrency-level (.getActiveCount thread-pool))
+                          "core" "kerberos" "throttle" "available-thread-count")
+    (metrics/waiter-gauge #(.getMaximumPoolSize thread-pool)
+                          "core" "kerberos" "throttle" "max-thread-count")
+    (metrics/waiter-gauge #(-> thread-pool .getQueue .size)
+                          "core" "kerberos" "throttle" "pending-task-count")
+    (metrics/waiter-gauge #(.getTaskCount thread-pool)
+                          "core" "kerberos" "throttle" "scheduled-task-count")
     (->KerberosAuthenticator thread-pool password)))
 
 (defrecord KerberosAuthorizer
--- a/waiter/src/waiter/auth/kerberos.clj
+++ b/waiter/src/waiter/auth/kerberos.clj
@@ -22,8 +22,9 @@
             [waiter.auth.authentication :as auth]
             [waiter.authorization :as authz]
             [waiter.auth.spnego :as spnego]
+            [waiter.metrics :as metrics]
             [waiter.util.utils :as utils])
-  (:import (java.util.concurrent ExecutorService Executors)))
+  (:import (java.util.concurrent LinkedBlockingQueue ThreadPoolExecutor TimeUnit)))
 
 (defn get-opt-in-accounts
   "Returns the list of users whose tickets are prestashed on host"
@@ -102,18 +103,24 @@
                          :status 403
                          :user run-as-user}))))))
 
-(defrecord KerberosAuthenticator [^ExecutorService thread-pool password]
+(defrecord KerberosAuthenticator [thread-pool password]
   auth/Authenticator
   (wrap-auth-handler [_ request-handler]
     (spnego/require-gss request-handler thread-pool password)))
 
 (defn kerberos-authenticator
   "Factory function for creating Kerberos authenticator middleware"
-  [{:keys [concurrency-level password]}]
+  [{:keys [concurrency-level keep-alive-mins password]}]
   {:pre [(not-empty password)
          (integer? concurrency-level)
-         (pos? concurrency-level)]}
-  (let [thread-pool (Executors/newFixedThreadPool concurrency-level)]
+         (pos? concurrency-level)
+         (integer? keep-alive-mins)
+         (pos? keep-alive-mins)]}
+  (let [thread-pool (ThreadPoolExecutor. 1 concurrency-level keep-alive-mins TimeUnit/MINUTES (LinkedBlockingQueue.))]
+    (metrics/waiter-gauge #(.getActiveCount thread-pool) "core" "kerberos" "throttle" "active-thread-count")
+    (metrics/waiter-gauge #(.getMaximumPoolSize thread-pool) "core" "kerberos" "throttle" "max-thread-count")
+    (metrics/waiter-gauge #(-> thread-pool .getQueue .size) "core" "kerberos" "throttle" "pending-task-count")
+    (metrics/waiter-gauge #(.getTaskCount thread-pool) "core" "kerberos" "throttle" "scheduled-task-count")
     (->KerberosAuthenticator thread-pool password)))
 
 (defrecord KerberosAuthorizer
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/auth/kerberos.clj
```diff
@@ -101,16 +102,19 @@
...
 (defn kerberos-authenticator
   "Factory function for creating Kerberos authenticator middleware"
-  [{:keys [password]}]
-  {:pre [(not-empty password)]}
-  (->KerberosAuthenticator password))
+  [{:keys [concurrency-level password]}]
+  {:pre [(not-empty password)
+         (integer? concurrency-level)
+         (pos? concurrency-level)]}
+  (let [thread-pool (Executors/newFixedThreadPool concurrency-level)]
```
scrosby: I don't like creating a possibly idle pool of 30+ extra threads. This should be self-scaling, up to the limit you set. I'd also rename the config value max-concurrency-level, not concurrency-level. 

What about 'new ThreadPoolExecutor' directly to construct it, that way you can make the pool auto-scale. Set a core pool size of, say, 1 thread, scale to max-concurrency-level maximum, with a 5 minute idle thread timeout. You can also put in a ThreadFactory and name the threads as they're created! :) That'll make stack traces and monitoring work nicer.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -83,32 +86,55 @@
...
+                (try
+                  (timers/stop timer-context)
+                  (let [^GSSContext gss-context (gss-context-init)
+                        token (do-gss-auth-check gss-context request)]
+                    (if (.isEstablished gss-context)
+                      (let [principal (gss-get-principal gss-context)
+                            user (first (str/split principal #"@" 2))
+                            response (auth/handle-request-auth request-handler request user principal password)]
+                        (log/debug "added cookies to response")
+                        (if token
```
scrosby: I don't know my core-async to review this. Please have someone else do this part.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -83,32 +86,55 @@
...
-                  (rr/header resp "WWW-Authenticate" token)
-                  (async/go
-                    (rr/header (async/<! resp) "WWW-Authenticate" token)))
-                resp))
-            (response-401-negotiate)))
+        (get-in request [:headers "authorization"])
+        (let [current-correlation-id (cid/get-correlation-id)
+              response-chan (async/promise-chan)
+              timer-context (timers/start (metrics/waiter-timer "core" "spnego" "throttle-delay"))]
+          ;; launch task that will populate the response in response-chan
```
scrosby: Could also metric the current size of the thread pool here. (Or rather, metric the number of free threads; getMaxPoolSize()-getActiveCount().... If it gets to zero, then we know we're bottlenecking here. Also metric'ing the blockingQueueSize is possible, but IMO, less important. I'd rather know we're backlogging and how often, not how long the backlog is.

shamsimam: Added metrics for active and max threads. Also added metrics for the number of pending tasks in the queue.

waiter/src/waiter/auth/kerberos.clj
shamsimam: Thanks for the suggestion. Agreed and implemented (except the custom thread factory).

shamsimam:
@scrosby ready for another round of review.

waiter/src/waiter/auth/spnego.clj
scrosby: I'd metric not the number of active threads, but the available threads. (max-active). 
The absolute number of active threads isn't an interesting number, its when the number gets close the the max, specifically when max-active gets low, that we care about.

The cognitive load of 'when this number gets close to zero there's a problem', is simpler than 'when the number gets close to another number, there's a problem'.  It'll also graph nicer, and we can put an alert in when, say, we get to available threads < 2 for more than X times per minute.

shamsimam: Done.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -83,32 +86,55 @@
...
+              (cid/with-correlation-id
+                current-correlation-id
+                (try
+                  (timers/stop timer-context)
+                  (let [^GSSContext gss-context (gss-context-init)
+                        token (do-gss-auth-check gss-context request)]
+                    (if (.isEstablished gss-context)
+                      (let [principal (gss-get-principal gss-context)
+                            user (first (str/split principal #"@" 2))
+                            response (auth/handle-request-auth request-handler request user principal password)]
```
sradack: Processing the request should happen outside of the threadpool.  The async action should only cover authentication, returning only the principal through the channel, after which processing request can proceed.

waiter/src/waiter/metrics.clj
```diff
@@ -93,6 +94,11 @@
   [classifier & nested-path]
   `(counters/counter ~(metric-name (concat ["waiter" classifier "counters"] nested-path))))
 
+(defmacro waiter-gauge
```
sradack: Using gauges is obviously new.  We avoided guages in the past, but unsure if we need to continue to avoid them.

sradack:
@shamsimam What are your thoughts on an approach that rejects requests that exceed the limit (with a request limit exceeded error) instead of allowing them to queue up inside Waiter?

shamsimam:
@sradack added support for failing requests when the kerberos queue length reaches a configured limit.

It is ready for review.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -60,9 +64,22 @@
...
       (rr/header "WWW-Authenticate" "Negotiate")
       (cookies/cookies-response)))
 
+(defn response-503-temporarily-unavailable
+  "Tell the client you're overloaded and would like them to try later"
+  []
+  (log/info "triggering 401 negotiate for spnego authentication")
+  (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
+  (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
+  (-> (rr/response "Service Temporarily Unavailable - Too Many Kerberos authentication requests")
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -73,42 +90,88 @@
...
-(defn gss-get-princ
+(defn gss-get-principal
   [^GSSContext gss]
   (str (.getSrcName gss)))
 
+(defn too-many-pending-auth-requests?
+  "Returns true if there are too many pending Kerberos auth requests."
+  [thread-pool max-queue-length]
+  (-> thread-pool
+      .getQueue
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -60,9 +64,22 @@
...
 
+(defn response-503-temporarily-unavailable
+  "Tell the client you're overloaded and would like them to try later"
+  []
+  (log/info "triggering 401 negotiate for spnego authentication")
+  (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
+  (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
+  (-> (rr/response "Service Temporarily Unavailable - Too Many Kerberos authentication requests")
+      (rr/status 503)
+      (rr/header "Content-Type" "text/html")
```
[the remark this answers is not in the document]
shamsimam: Changed to `text/plain`

waiter/src/waiter/auth/spnego.clj
```diff
@@ -60,9 +64,22 @@
...
+      (rr/header "Server" (utils/get-current-server-name))
       (rr/header "WWW-Authenticate" "Negotiate")
       (cookies/cookies-response)))
 
+(defn response-503-temporarily-unavailable
+  "Tell the client you're overloaded and would like them to try later"
+  []
+  (log/info "triggering 401 negotiate for spnego authentication")
+  (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
+  (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
```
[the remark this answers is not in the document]
shamsimam: Noted.

shamsimam:
@sradack any further feedback?

PR #541, opened 2018-12-19
remove explicit flag from fixed tests

shamsimam:

- remove explicit flag from fixed tests


dposada:
Can we delete the branch?

PR #542, opened 2018-12-27
includes server header in output generated from waiter

shamsimam:

- includes server name in output generated from waiter

Adding the `Server` header allows clients to differentiate between responses originating in Waiter and the backend.

**Before**:
```
$ curl -I -s -H"x-waiter-cpus: 1" -H"x-waiter-mem: 1024" -H"x-waiter-cmd: foo-bar" -H"x-waiter-version: fv" -H"x-waiter-timeout: 10000" http://localhost:9091/foo-bar 
HTTP/1.1 503 Service Unavailable
Content-Type: text/plain
Transfer-Encoding: chunked
...
```

**After** (note the extra `server` header):
```
$ curl -I -s -H"x-waiter-cpus: 1" -H"x-waiter-mem: 1024" -H"x-waiter-cmd: foo-bar" -H"x-waiter-version: fv" -H"x-waiter-timeout: 10000" http://localhost:9091/foo-bar                                      
HTTP/1.1 503 Service Unavailable
Content-Type: text/plain
Server: waiter/099b03d
Transfer-Encoding: chunked
...
```

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -93,12 +93,30 @@
   "Returns true if x is a non-negative integer"
   (and (integer? x) (non-neg? x)))
 
+(defn assoc-if-absent
+  "If the specified key, k, is not already associated with a value, v, in the map, m, associate k with v in m."
+  [m k v]
+  (cond-> m
+    (not (contains? m k)) (assoc k v)))
+
 (defn generate-secret-word
   [src-id dest-id processed-passwords]
   (let [password (second (first processed-passwords))
         secret-word (digest/md5 (str src-id ":" dest-id ":" password))]
     (log/debug "generate-secret-word" [src-id dest-id] "->" secret-word)
     secret-word))
+
+(let [server-name-atom (atom "waiter")]
+  (defn get-current-server-name
+    "Returns the name of the server used to handle the request."
+    []
+    (deref server-name-atom))
+
+  (defn reset-server-name-atom!
+    "Resets the server name."
+    [server-name]
+    (log/info "server name has been initialized to" server-name)
+    (reset! server-name-atom server-name)))
 
 (defn keyword->str
   "Converts keyword to string including the namespace."
@@ -140,14 +158,17 @@
   [data-map & {:keys [headers status] :or {headers {} status 200}}]
   {:body (clj->json data-map)
    :status status
-   :headers (assoc headers "content-type" "application/json")})
+   :headers (assoc headers
+              "content-type" "application/json"
+              "server" (get-current-server-name))})
 
 (defn clj->streaming-json-response
   "Converts the data into a json response which can be streamed back to the client."
   [data-map & {:keys [status] :or {status 200}}]
   (let [data-map (doall data-map)]
     {:status status
-     :headers {"content-type" "application/json"}
+     :headers {"content-type" "application/json"
+               "server" (get-current-server-name)}
      :body (fn [^ServletResponse resp]
              (let [writer (OutputStreamWriter. (.getOutputStream resp))]
                (try
@@ -246,7 +267,9 @@
                  render-error-text
                  (str/replace #"\n" "\n  ")
                  (str/replace #"\n  $" "\n")))
-     :headers (merge {"content-type" content-type} headers)
+     :headers (-> headers
+                  (assoc-if-absent "content-type" content-type)
+                  (assoc-if-absent "server" (get-current-server-name)))
      :status status}))
 
 (defn- wrap-unhandled-exception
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -93,13 +93,31 @@
...
+    (not (contains? m k)) (assoc k v)))
+
 (defn generate-secret-word
   [src-id dest-id processed-passwords]
   (let [password (second (first processed-passwords))
         secret-word (digest/md5 (str src-id ":" dest-id ":" password))]
     (log/debug "generate-secret-word" [src-id dest-id] "->" secret-word)
     secret-word))
 
+(let [server-name-atom (atom "waiter")]
```
scrosby: Why wrap this at all? Why not just have the bare atom? Less code, and I can do a 'find uses' of the atom and see all gets and sets, globally. Now, I'd need to search for the atom, then for these function uses.

waiter/test/waiter/core_test.clj
```diff
@@ -434,7 +434,9 @@
         (let [request {:request-method :delete, :uri (str "/apps/" service-id), :authorization/user user}
               {:keys [body headers status]} (async/<!! ((ring-handler-factory waiter-request?-fn handlers) request))]
           (is (= 200 status))
-          (is (= {"content-type" "application/json"} headers))
+          (is (= {"content-type" "application/json"
```
scrosby: These look like standard headers, repeated about 40 times.

Why not: 

(def expected-error-response-headers {....})
Then reference that dictionary throughout?

waiter/test/waiter/scaling_test.clj
```diff
@@ -166,7 +166,10 @@
                         {:basic-authentication {:src-router-id src-router-id} :route-params {:service-id test-service-id}}))
                     {:keys [body headers status]} (async/<!! response-chan)]
                 (is (= 200 status))
-                (is (= {"content-type" "application/json", "x-cid" correlation-id} headers))
+                (is (= {"content-type" "application/json"
```
scrosby: Could factor this out in a similar way.
(defn verify-response-headers 
  [headers expected-corrolation-id]
  (is ....))

waiter/test/waiter/util/utils_test.clj
```diff
@@ -118,7 +118,9 @@
     (testing "should convert empty map"
       (let [{:keys [body headers status]} (clj->json-response {})]
         (is (= 200 status))
-        (is (= {"content-type" "application/json"} headers))
+        (is (= {"content-type" "application/json"
```
scrosby: Ditto here.

waiter/test/waiter/core_test.clj
shamsimam: Introduced a new variable to hold the expected headers and left the assertions in the tests.

waiter/test/waiter/scaling_test.clj
shamsimam: Introduced a new variable to hold the expected headers and left the assertions in the tests.

waiter/test/waiter/util/utils_test.clj
shamsimam: Introduced a new variable to hold the expected headers and left the assertions in the tests.

shamsimam:
@scrosby ready for another round :)

waiter/src/waiter/util/utils.clj
scrosby: Almost everything in clojure is global, by design. What we lose by hiding it behind these accessors is 
#1: We have 9 lines of code; 10% of the lines in this patch and 33% of the new code added in this patch.
#2: We can no longer track accesses to the server name atom so easily, with, e.g., the IDE. We have to now trace indirectly via the finding the references to these two functions (versus searching for the one atom.)

The atom is still mutable, except instead of doing (reset! ...) its now (reset-server-name-atom!) and (get-current-server-name ...) instead of (deref).

shamsimam: The flexibility is, currently, a hypothetical concern. But I would not be surprised if we needed to change the implementation in the future. E.g. in the case of messages chose to dynamically reload the messages from a properties file or some other store.

"Handful of accesses" is the current state, but may not be as the code evolves. 

I feel like this is a scenario ideal for the use of `mount`. We want to initialize some state (we generate from the settings) and then let it be freely read without worry of it being mutated. When we can control the mutation to only the initialization, there is no harm in having everyone else access it directly to read. I am mimicking that by restricting access to the raw atom and instead requiring invocation of a helper function to read the state.

I'd very much appreciate if you accept the patch, we can have a more general discussion on how these should be handled and perform a more complete refactoring on the codebase for such uses of atom once we reach a consensus.

scrosby: I'm fine with approving it, but apparently I don't have commit permissions to waiter

PR #543, opened 2018-12-31
Kitchen: log to stdout by default

DaoWen:

- Change the default logging location from `./kitchen.log` to *stdout*.
- Add a new command line option to choose the log output destination. Options are `file` (which gives the current behavior), `stdout`, or `stderr`.

We usually expect logs in stdout/stderr in Waiter. The current proposal for backing up K8s pod logs in S3 doesn't capture arbitrary files like `kitchen.log` by default. This change will make Python Kitchen's logging behavior more compatible with Waiter-K8s.

### DIFF
--- a/kitchen/bin/kitchen
+++ b/kitchen/bin/kitchen
@@ -913,6 +913,7 @@ if __name__ == '__main__':
 
     parser = argparse.ArgumentParser(description='A toy HTTP Service for testing the Waiter platform')
     parser.add_argument('--hostname', metavar='HOSTNAME', default='', help='Server host name')
+    parser.add_argument('--log-output', metavar='LOG_OUTPUT', choices=['stdout', 'stderr', 'file'], default='stdout', help='Log output destination')
     parser.add_argument('-p', '--port', metavar='PORT_NUMBER', type=int, default=8080, help='Server port number')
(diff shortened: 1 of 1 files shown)


kitchen/bin/kitchen
```diff
@@ -913,6 +913,7 @@ if __name__ == '__main__':
 
     parser = argparse.ArgumentParser(description='A toy HTTP Service for testing the Waiter platform')
     parser.add_argument('--hostname', metavar='HOSTNAME', default='', help='Server host name')
+    parser.add_argument('--log-output', metavar='LOG_OUTPUT', choices=['stdout', 'stderr', 'file'], default='stdout', help='Log output destination')
```
shamsimam: Should we make `file` the default for backwards compatibility?

DaoWen: None of our tests are using the `kitchen.log` file for anything. Our Clojure version of Kitchen wrote to the console ([source](https://github.com/twosigma/waiter/blob/c3575bd/kitchen/resources/log4j.properties)), so we can consider this as fixing backward compatibility with the old Kitchen code. I believe I made the change originally because we weren't originally capturing *stdout* and *stderr* into files on Kubernetes. Now we emulate the Mesos behavior by creating those files with our `waiter-init` script. I think changing the output to `kitchen.log` was a bad change in hindsight.

PR #549, opened 2019-01-16
Enable Kerberos when accessing S3 buckets

DaoWen:

Enable Kerberos when accessing S3 buckets.

These requests fail (HTTP 401) if the buckets are kerberized. Enabling Kerberos/SPNEGO when they're not kerberized doesn't hurt anything.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -596,7 +596,10 @@
                   result (http-utils/http-request
                            http-client
                            log-bucket-url
-                           :query-string query-string)
+                           :query-string query-string
+                           ;; Enabling Kerberos/SPNEGO when the bucket is not kerberized does not
+                           ;; cause an error, and the extra flag is ignored on non-kerberized systems.
+                           :spnego-auth true)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -596,7 +596,8 @@
                   result (http-utils/http-request
                            http-client
                            log-bucket-url
-                           :query-string query-string)
+                           :query-string query-string
+                           :spnego-auth true)
```
shamsimam: Can we default this to false and configure it to be true?

DaoWen: We discussed this offline, and for now I've added a comment both here and on the curl command in the waiter-init script. Ideally we'd make this configurable, but for now we'll settle for the comment since the current code works in both kerberized and non-kerberized environments.

DaoWen: Diff: https://github.com/twosigma/waiter/compare/134f7c59a432bb516f1643cc934290ecf32c0da4..76966f795fc3626bb2744ed86bd193c2a6ec229e

DaoWen:
Green build here: https://travis-ci.org/DaoWen/waiter/builds/480593769

PR #553, opened 2019-01-17
adds support for file-based persistent kv store

shamsimam:

- adds support for file-based persistent kv store

This provides a persistent key-value store that can be used for local development. Having a persistent store allows storing tokens and service descriptions and being reused across server restarts.


waiter/config-shell.edn
```diff
@@ -26,6 +26,12 @@
  ;; Set the port:
  :port #config/env-int "WAITER_PORT"
 
+ ; ---------- Token Storage ----------
+
+ :kv-config {:kind :persistent
```
dposada: Maybe `:kind` could be `:file` or `:file-based`? `:persistent` seems too general

waiter/config-shell.edn
```diff
@@ -26,6 +26,12 @@
...
+             :persistent {:factory-fn waiter.kv/new-persistent-kv-store
+                          :target-file "kv-store.bin"}}
+
```
dposada: Does this make the shell scheduler in travis actually run with this? I'm curious if all of the integration tests pass.

shamsimam: I will make the composite scheduler run with this, that's the file that is used.

waiter/config-shell.edn
shamsimam: okay

PR #555, opened 2019-01-18
adds support for token query parameter on post

shamsimam:

- adds support for token query parameter on post

We would like to avoid having to include the token entry in the json POST-ed to the /token endpoint. This change allows us to pass the token as a query parameter instead thus clearing the way to reuse the same json payload to register multiple tokens.


dposada:
@shamsimam there are conflicts

PR #556, opened 2019-01-20
Add support for https-redirect parameter

shamsimam:

- add support for https-redirect parameter

We would like service owners to be able to opt-in to require clients to use https connections while making requests for their services.


sradack:
Originally we were discussing having this be a token parameter and not a service parameter.  What are your thoughts along these lines?

shamsimam:
I thought about that, it depends on how we present the feature. 

Currently, I have the documentation claiming that with the `https-redirect` parameter, a service (which is potentially worried about man in the middle attacks and leaking any sensitive data) is guaranteed to be receiving https traffic from the client.

waiter/src/waiter/service_description.clj
```diff
@@ -94,6 +94,7 @@
    (s/optional-key "scale-up-factor") schema/positive-fraction-less-than-1
    ; per-request related
    (s/optional-key "blacklist-on-503") s/Bool
+   (s/optional-key "https-redirect") s/Bool
```
sradack: Can this be removed now that `https-redirect` is a token-only parameter?

PR #560, opened 2019-01-23
add codahale metrics to kubernetes api calls

nsinkov:

add codahale metrics to kubernetes api calls

We will be able to monitor metrics on k8s api calls

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -242,13 +242,13 @@
   "Make an HTTP request to the Kubernetes API server using the configured authentication.
    If data is provided via :body, the application/json content type is added automatically.
    The response payload (if any) is automatically parsed to JSON."
-  [client url {:keys [scheduler-name]} metric-name & {:keys [body content-type request-method] :as options}]
+  [url {:keys [http-client scheduler-name]} metric-name & {:keys [body content-type request-method] :as options}]
   (scheduler/log "making request to K8s API server:" url request-method body)
   (ss/try+
     (let [auth-str @k8s-api-auth-str
           result (timers/start-stop-time!
                    (metrics/waiter-timer "scheduler" scheduler-name metric-name)
-                   (pc/mapply http-utils/http-request client url
+                   (pc/mapply http-utils/http-request http-client url
                               :accept "application/json"
                               (cond-> options
                                       auth-str (assoc-in [:headers "Authorization"] auth-str)
@@ -257,7 +257,7 @@
       result)
     (catch [:status 400] _
       (log/error "malformed K8s API request: " url options))
-    (catch [:client client] response
+    (catch [:client http-client] response
       (log/error "request to K8s API server failed: " url options body response)
       (ss/throw+ response))))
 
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -240,15 +242,17 @@
   "Make an HTTP request to the Kubernetes API server using the configured authentication.
    If data is provided via :body, the application/json content type is added automatically.
    The response payload (if any) is automatically parsed to JSON."
-  [client url & {:keys [body content-type request-method] :as options}]
+  [client url {:keys [scheduler-name]} & {:keys [body content-type request-method] :as options}]
```
DaoWen: In other places where I refactored to include the scheduler as a parameter, I replaced the `http-client` argument with the `scheduler` argument, since you can destructure the `scheduler` to get the `http-client`. I think that would be a good strategy to use here as well.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -240,15 +242,17 @@
...
   (scheduler/log "making request to K8s API server:" url request-method body)
   (ss/try+
     (let [auth-str @k8s-api-auth-str
-          result (pc/mapply http-utils/http-request client url
-                            :accept "application/json"
-                            (cond-> options
-                              auth-str (assoc-in [:headers "Authorization"] auth-str)
-                              (and (not content-type) body) (assoc :content-type "application/json")))]
+          result (timers/start-stop-time!
+                   (metrics/waiter-timer "scheduler" scheduler-name request-method)
```
DaoWen: `request-method` may be `nil` here. Maybe we should set the default here explicitly. Currently the default is set to `:get` in the call to `http-utils/http-request` below.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -813,11 +819,13 @@
                 (try
                   (loop [version (reset-watch-state! scheduler options)
                          iter 0]
-                    (let [watch-url (str resource-url "&watch=true&resourceVersion=" version)]
-                      ;; process updates forever (unless there's an exception)
-                      (doseq [json-object (streaming-api-request-fn watch-url)]
-                        (when json-object
-                          (update-fn json-object))))
+                    (timers/start-stop-time!
```
DaoWen: Looking at [the definition of `start-stop-time!`](https://github.com/metrics-clojure/metrics-clojure/blob/fe6b5f44e2a661cfbdd3dd09ee43567c030b5ff3/metrics-clojure-core/src/metrics/timers.clj#L139-L144), it doesn't appear to handle exceptions gracefully. I don't think that's an issue here, but we should keep that in mind.

nsinkov:
@DaoWen updated

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -240,20 +242,22 @@
...
   (scheduler/log "making request to K8s API server:" url request-method body)
   (ss/try+
     (let [auth-str @k8s-api-auth-str
-          result (pc/mapply http-utils/http-request client url
-                            :accept "application/json"
-                            (cond-> options
-                              auth-str (assoc-in [:headers "Authorization"] auth-str)
-                              (and (not content-type) body) (assoc :content-type "application/json")))]
+          result (timers/start-stop-time!
+                   (metrics/waiter-timer "scheduler" scheduler-name (str request-method))
```
DaoWen:
```suggestion
                   (metrics/waiter-timer "scheduler" scheduler-name (name request-method))
```

PR #563, opened 2019-01-25
introduces status inactive for accessing deleted services

shamsimam:

- introduces status inactive for accessing deleted services

screen shot 2019-01-25 at 2 55 39 am

### DIFF
--- a/waiter/src/waiter/service.clj
+++ b/waiter/src/waiter/service.clj
@@ -300,10 +300,10 @@
    - Starting: the service has no healthy instances and is starting one up,
    - Running: the service is running successfully with at least one healthy instance,
    - Failing: the service has instances failing to start,
-   - Idle: the service is waiting to be started."
+   - Inactive: the service has no scheduled and running tasks."
   [deployment-error {:keys [healthy requested scheduled] :or {healthy 0 requested 0 scheduled 0}}]
   (cond
     deployment-error :service-state-failing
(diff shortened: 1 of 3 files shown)


sradack:
I'm not clear on what we mean by "accessing deleted services."

shamsimam:
@sradack any further feedback on this?

waiter/src/waiter/service.clj
```diff
@@ -304,5 +304,6 @@
   [deployment-error {:keys [healthy requested scheduled] :or {healthy 0 requested 0 scheduled 0}}]
   (cond
     deployment-error :service-state-failing
+    (and (zero? requested) (zero? scheduled) (zero? healthy)) :service-state-inactive
```
sradack: Should this simply be `(zero? requested)`?  When would `scheduled` or `healthy` be positive when `requested` is zero?  We should also fix the doc for this fn.

sradack: Don't we 404 when asked about a service that has no requested instances?  And remove it from the service listing?

shamsimam: There can be a gap between requested and actually scheduled instances (delay in killing or scaling up), so we need to check for both values.

We 404 only when the service description is not known. For deleted services, we render the page with empty instances. https://github.com/twosigma/waiter/blob/master/waiter/src/waiter/handler.clj#L442

PR #564, opened 2019-01-25
Fix codahale reporter meter scale

nsinkov:

re-write codahale GraphiteReporter functionality in clojure

The GraphiteReporter we're using scales values from meters differently from what we need and truncates the values

### DIFF
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -134,30 +134,37 @@
        (finally
          (counters/dec! counter#)))))
 
-(defn get-metrics
-  "Return a nested map of metrics data."
-  ([] (get-metrics mc/default-registry MetricFilter/ALL))
-  ([^MetricRegistry registry ^MetricFilter metric-filter]
-   (utils/keys->nested-map
+(let [percentiles [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0]]
+  (defn metric-registry->map
+    "Unpack codahale metrics into a map of values"
+    ([] (metric-registry->map mc/default-registry MetricFilter/ALL))
+    ([^MetricRegistry registry ^MetricFilter metric-filter]
      (merge
        (pc/map-vals (fn [c] (counters/value c))
                     (.getCounters registry metric-filter))
        (pc/map-vals (fn [^Gauge m] {"value" (.getValue m)})
                     (.getGauges registry metric-filter))
        (pc/map-vals (fn [^Histogram h] {"count" (.getCount h)
-                                        "value" (->> (histograms/percentiles h [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0])
+                                        "value" (->> (histograms/percentiles h percentiles)
                                                      (pc/map-keys str))})
                     (.getHistograms registry metric-filter))
        (pc/map-vals (fn [^Meter m] {"count" (.getCount m)
-                                    "value" (meters/rate-one m)})
+                                    "value" {:1min-rate (meters/rate-one m)}})
                     (.getMeters registry metric-filter))
        (pc/map-vals (fn [^Timer t]
                       {"count" (.getCount t)
                        ; nanos -> seconds
-                       "value" (->> (timers/percentiles t [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0])
+                       "value" (->> (timers/percentiles t percentiles)
                                     (pc/map-vals #(/ % 1e9))
                                     (pc/map-keys str))})
-                    (.getTimers registry metric-filter)))
+                    (.getTimers registry metric-filter))))))
+
+(defn get-metrics
+  "Return a nested map of metrics data."
+  ([] (get-metrics mc/default-registry MetricFilter/ALL))
+  ([^MetricRegistry registry ^MetricFilter metric-filter]
+   (utils/keys->nested-map
+     (metric-registry->map registry metric-filter)
      #"\.")))
 
 (defn get-core-codahale-metrics
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -135,9 +135,9 @@
          (counters/dec! counter#)))))
 
 (let [percentiles [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0]]
-  (defn metric-registry->map
+  (defn metric-registry->metric-filter->metric-map
     "Unpack codahale metrics into a map of values"
-    ([] (metric-registry->map mc/default-registry MetricFilter/ALL))
+    ([] (metric-registry->metric-filter->metric-map mc/default-registry MetricFilter/ALL))
     ([^MetricRegistry registry ^MetricFilter metric-filter]
      (merge
        (pc/map-vals (fn [c] (counters/value c))
@@ -164,7 +164,7 @@
   ([] (get-metrics mc/default-registry MetricFilter/ALL))
   ([^MetricRegistry registry ^MetricFilter metric-filter]
    (utils/keys->nested-map
-     (metric-registry->map registry metric-filter)
+     (metric-registry->metric-filter->metric-map registry metric-filter)
      #"\.")))
 
 (defn get-core-codahale-metrics
--- a/waiter/src/waiter/metrics.clj
+++ b/waiter/src/waiter/metrics.clj
@@ -149,7 +149,7 @@
                                                      (pc/map-keys str))})
                     (.getHistograms registry metric-filter))
        (pc/map-vals (fn [^Meter m] {"count" (.getCount m)
-                                    "value" {:1min-rate (meters/rate-one m)}})
+                                    "value" (meters/rate-one m)})
                     (.getMeters registry metric-filter))
        (pc/map-vals (fn [^Timer t]
                       {"count" (.getCount t)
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,100 @@
...
+(let [df (DecimalFormat. "#.####################")]
+  (defn- format-value
+    "Convert double to string to send to graphite server"
+    [value]
+    (.format df value)))
+
+(defn- report-meter
+  "Report values from a codahale meter"
+  [prefix name ^Metered meter timestamp ^GraphiteSender graphite]
+  (let [send #(.send graphite (make-reporting-name prefix name %1) (format-value (%2 meter)) timestamp)]
```
shamsimam: Would it be better to create this function as a `defn`? It seems to be used in both `report-meter` and `report-snapshot`.

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,100 @@
...
+    (send "count" #(.getCount %))
+    (send "m1_rate" #(.getOneMinuteRate %))
+    (send "m5_rate" #(.getFiveMinuteRate %))
+    (send "m15_rate" #(.getFifteenMinuteRate %))
+    (send "mean_rate" #(.getMeanRate %))))
+
+(defn- report-snapshot
+  "Report values from a codahale snapshot"
+  [prefix name ^Snapshot snapshot timestamp ^GraphiteSender graphite]
+  (let [send #(.send graphite (make-reporting-name prefix name %1) (format-value (%2 snapshot)) timestamp)]
```
shamsimam: On the Waiter routers metrics, we report the following percentils: `[0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0]`. Should we be doing the same here?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,100 @@
...
+  (let [send #(.send graphite (make-reporting-name prefix name %1) (format-value (%2 meter)) timestamp)]
+    (send "count" #(.getCount %))
+    (send "m1_rate" #(.getOneMinuteRate %))
+    (send "m5_rate" #(.getFiveMinuteRate %))
+    (send "m15_rate" #(.getFifteenMinuteRate %))
+    (send "mean_rate" #(.getMeanRate %))))
+
+(defn- report-snapshot
+  "Report values from a codahale snapshot"
+  [prefix name ^Snapshot snapshot timestamp ^GraphiteSender graphite]
```
shamsimam: On the Waiter routers metrics, we report the one minute rate (`meters/rate-one`). Should we be doing the same here?

waiter/src/waiter/reporter.clj
```diff
@@ -31,21 +34,29 @@
   (close! [this]
     "Stops the reporter and performs any cleanup")
 
+  (report [this]
+    "Forces the reporter to report current metrics")
+
+  (start [this]
```
shamsimam: 👍

waiter/src/waiter/reporter.clj
shamsimam: I meant: `#(.send graphite (make-reporting-name prefix name %1) (format-value (%2 meter)) timestamp)`

waiter/src/waiter/metrics.clj
```diff
@@ -134,30 +134,37 @@
        (finally
          (counters/dec! counter#)))))
 
-(defn get-metrics
-  "Return a nested map of metrics data."
-  ([] (get-metrics mc/default-registry MetricFilter/ALL))
-  ([^MetricRegistry registry ^MetricFilter metric-filter]
-   (utils/keys->nested-map
+(let [percentiles [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0]]
+  (defn metric-registry->map
```
shamsimam: Nitpick: this is more like `metric-registry->metric-filter->metric-map`

waiter/src/waiter/metrics.clj
```diff
@@ -134,30 +134,37 @@
...
        (pc/map-vals (fn [^Gauge m] {"value" (.getValue m)})
                     (.getGauges registry metric-filter))
        (pc/map-vals (fn [^Histogram h] {"count" (.getCount h)
-                                        "value" (->> (histograms/percentiles h [0.0 0.25 0.5 0.75 0.95 0.99 0.999 1.0])
+                                        "value" (->> (histograms/percentiles h percentiles)
                                                      (pc/map-keys str))})
                     (.getHistograms registry metric-filter))
        (pc/map-vals (fn [^Meter m] {"count" (.getCount m)
-                                    "value" (meters/rate-one m)})
+                                    "value" {:1min-rate (meters/rate-one m)}})
```
shamsimam: Have we verified nothing breaks due to this change? I'm not sure what the effects of changing the schema for a Meter display are going to be.

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,53 @@
...
+    "Convert numbers to string to send to graphite server"
+    [value]
+    (if (number? value)
+      (.format df value)
+      (str value))))
+
+(defn- report-to-graphite-helper
+  "Recursively traverse metrics map and send to graphite"
+  [prefix map ^GraphiteSender graphite timestamp]
+  (if (map? map)
```
shamsimam: Can we flatten the map and avoid the recursion here?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,53 @@
...
+    (doseq [[k v] map]
+      (report-to-graphite-helper (make-reporting-name prefix (str k)) v graphite timestamp))
```
shamsimam: Nitpick: if we do stay with the recursive solution, we can use `recur` here to make it clearer that we are recursing here.
`(recur (make-reporting-name prefix (str k)) v graphite timestamp)`

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,53 @@
...
+    (.send graphite prefix (format-value map) timestamp)))
+
+(defn- report-to-graphite
+  "Report values from a codahale MetricRegistry"
+  [^MetricRegistry registry prefix ^MetricFilter filter ^GraphiteSender graphite]
+  (let [timestamp (.getTime (Clock/defaultClock))
+        map (metrics/metric-registry->map registry filter)]
+    (try
+      (when-not (.isConnected graphite)
+        (.connect graphite))
```
shamsimam: Can we add a log here that we are connecting to the graphite server?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,53 @@
...
+
+(let [df (DecimalFormat. "#.####################")]
+  (defn- format-value
+    "Convert numbers to string to send to graphite server"
+    [value]
+    (if (number? value)
+      (.format df value)
+      (str value))))
+
+(defn- report-to-graphite-helper
```
shamsimam: A log summarizing how many entries were sent to graphite in a log would be useful for debugging later.

waiter/src/waiter/reporter.clj
```diff
@@ -119,17 +170,18 @@
...
-                               (.filter (make-metrics-filter filter-regex))
-                               (.prefixedWith prefix)
-                               (.convertRatesTo TimeUnit/SECONDS)
-                               (.convertDurationsTo TimeUnit/MILLISECONDS)
-                               (.build ^GraphiteSender graphite-wrapper))]
-     [graphite-reporter state-atom])))
+         filter (filter-regex->metric-filter filter-regex)
+         period-ms-period (t/millis period-ms)]
+     (reify CodahaleReporter
+       (close! [_] (.close graphite-wrapper) (swap! state-atom assoc :run-state :closed))
```
shamsimam: Can we please avoid placing multiple expressions on a single line?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,53 @@
...
+        map (metrics/metric-registry->map registry filter)]
+    (try
+      (when-not (.isConnected graphite)
+        (.connect graphite))
+      (report-to-graphite-helper prefix map graphite timestamp)
+      (.flush graphite)
+      (catch IOException e
+        (try
+          (.close graphite)
+          (catch IOException e
```
shamsimam: Should this be catch throwable so that if this throws some other exception, we do not miss reporting the initial exception `e`?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,55 @@
                s/Any s/Any}
               (merge {:pickled? true} config)))
 
+(defn- make-reporting-name
```
shamsimam: We don't seem to be using the varargs version (found only one caller). Can we just replace this with `(str prefix "." k)` in the caller?

waiter/src/waiter/reporter.clj
```diff
@@ -87,15 +100,55 @@
...
+  (defn- format-value
+    "Convert numbers to string to send to graphite server"
+    [value]
+    (if (number? value)
+      (.format df value)
+      (str value))))
+
+(defn- report-to-graphite-helper
+  "Recursively traverse metrics map and send to graphite"
+  [prefix map ^GraphiteSender graphite timestamp]
```
shamsimam: Can we rename `map` as `data` as it will not be a map for the leaf value?

nsinkov:
@shamsimam  ready for merge

Issue #557, opened 2019-01-20
`test-request-queue-timeout-slow-start-app` is flaky with kubernetes scheduler


shamsimam:
```
lein parallel-test :only waiter.request-timeout-test/test-request-queue-timeout-slow-start-app
FAIL in (test-request-queue-timeout-slow-start-app) (request_timeout_test.clj:55)
test-request-queue-timeout-slow-start-app
expected: (clojure.string/includes? body "Check that your service is able to start properly!")
  actual: (not (clojure.string/includes? "\n  Waiter Error 503\n  ================\n  \n    After 10 seconds, no instance available to handle request.\n  \n  Request Info\n  ============\n  \n            Host: 127.0.0.1:9091\n            Path: /req\n    Query String: \n          Method: POST\n             CID: test-request-queue-timeout-slow-start-app-b6a8186bb9-113a633c93c46be1\n            Time: 2019-01-20T16:49:28.869Z\n       Principal: travis\n      Service Id: waiter-service-testrequestqueuetimeoutslowstartapptravis1291530-f0735b4b12fa87ece3afa084afa0a2a3\n  \n  \n  Additional Info\n  ===============\n  \n    {:outstanding-requests 1,\n     :service-id\n     \"waiter-service-testrequestqueuetimeoutslowstartapptravis1291530-f0735b4b12fa87ece3afa084afa0a2a3\",\n     :work-stealing-offers-sent 0,\n     :work-stealing-offers-received 0,\n     :slots-assigned 0,\n     :slots-in-use 0,\n     :waiting-for-available-instance 1,\n     :status 503,\n     :slots-available 0,\n     :requests-waiting-to-stream 0}\n    \n  Getting Help\n  ============\n  \n    Waiter on GitHub: http://github.com/twosigma/waiter \n  \n" "Check that your service is able to start properly!"))
```

PR #566, opened 2019-01-29
adds the response Server to the request log

shamsimam:

- adds the response Server to the request log

Helps quickly identify the source of a response from the request log.


waiter/src/waiter/request_log.clj
```diff
@@ -46,8 +46,10 @@
 (defn response->context
   "Convert a response into a context suitable for logging."
   [{:keys [authorization/principal backend-response-latency-ns descriptor latest-service-id get-instance-latency-ns
-           handle-request-latency-ns instance status] :as response}]
-  (let [{:keys [service-id service-description]} descriptor]
+           handle-request-latency-ns headers instance status] :as response}]
+  (let [{:keys [service-id service-description]} descriptor
+        response-server (or (get headers "server")
+                            (get headers "Server"))]
```
sradack: Aren't we always supposed to get lowercase headers?

shamsimam: Changed to all lower-case "server" headers in our code.

shamsimam:
Build failed due to a flaky test on k8s scheduler: https://github.com/twosigma/waiter/issues/557
All other tests passed.

PR #567, opened 2019-01-29
increases the fallback period limit to 1 day

shamsimam:

- increases the fallback period limit to 1 day

We would like to support larger fallback intervals.


waiter/src/waiter/service_description.clj
```diff
@@ -98,7 +98,7 @@
    s/Str s/Any})
 
 (def user-metadata-schema
-  {(s/optional-key "fallback-period-secs") (s/both s/Int (s/pred #(<= 0 % (t/in-seconds (t/hours 1))) 'at-most-1-hour))
+  {(s/optional-key "fallback-period-secs") (s/both s/Int (s/pred #(<= 0 % (t/in-seconds (t/days 1))) 'at-most-1-day))
```
dposada: Should we add a test for this?

dposada:

shamsimam:
Build failed due to a flaky test on k8s scheduler: https://github.com/twosigma/waiter/issues/557
All other tests passed.

PR #575, opened 2019-01-30
Graphite expects timestamp in seconds

nsinkov:

change timestamp sent to graphite from milliseconds to seconds

Graphite expects timestamp in seconds


shamsimam:
Awaiting green build.

PR #579, opened 2019-02-04
adds support for dynamically calculating token clusters

shamsimam:

- adds support for dynamically calculating token clusters

We would like to know the target cluster for a token. The `root` metadata does not suffice since while running in failure mode the request maybe capture by a different Waiter cluster than the one determined from the `host` header.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -801,7 +801,7 @@
                                              ;; A regex to determine the root from the incoming request host header.
                                              ;; The first group from the match is used to determine the root.
                                              ;; If no match is found, it defaults to the :cluster-config :name.
-                                             :root-regex #config/regex "(.*)(:.*)?"}}
+                                             :root-regex #config/regex "^waiter-([a-zA-Z][0-9a-zA-Z\\-]*)"}}
 
                 ;; Configures the amount of history we store for a token.
                 ;; It must be a positive integer.
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -790,6 +790,19 @@
                        :failed-check-threshold 5}
 
  :token-config {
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/settings.clj
```diff
@@ -388,7 +392,11 @@
    :support-info [{:label "Waiter on GitHub"
                    :link {:type :url
                           :value "http://github.com/twosigma/waiter"}}]
-   :token-config {:history-length 5
+   :token-config {:cluster-calculator {:kind :configured
+                                       :configured {:factory-fn 'waiter.token/new-configured-cluster-calculator}
+                                       :regex {:factory-fn 'waiter.token/new-regex-cluster-calculator
+                                               :root-regex #"(.*)(:.*)?"}}
```
DaoWen:
```suggestion
                                               :root-regex #"^([^:]*)"}}
```

waiter/config-full.edn
```diff
@@ -790,6 +790,19 @@
...
+                                     :kind :configured
+                                     :configured {:factory-fn 'waiter.token/new-configured-cluster-calculator}
+
+                                     ;; :kind :configured create a calculator that always assigns root as :cluster-config :name
+                                     ;; :kind :regex
+                                     :regex {:factory-fn 'waiter.token/new-regex-cluster-calculator
+                                             ;; A regex to determine the root from the incoming request host header.
+                                             ;; The first group from the match is used to determine the root.
+                                             ;; If no match is found, it defaults to the :cluster-config :name.
+                                             :root-regex #config/regex "(.*)(:.*)?"}}
```
DaoWen:
```suggestion
                                             :root-regex #config/regex #"^([^:]*)"}}
```

PR #585, opened 2019-02-08
Custom image support

nsinkov:

Add image field to service description

Allow users to specify a custom image for their app


waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -105,6 +105,17 @@
          (clojure.data/diff expected# actual#)))
      (is (= expected# actual#))))
 
+(deftest replicaset-spec-no-image
+  (let [scheduler (make-dummy-scheduler ["test-service-id"])
+        replicaset-spec ((:replicaset-spec-builder-fn scheduler) scheduler "test-service-id" dummy-service-description)]
+    (is (= "twosigma/kitchen:latest" (get-in replicaset-spec [:spec :template :spec :containers 0 :image])))))
```
DaoWen: This test will fail if a different configuration file is used. We should try reading this value from the settings instead.

We might need a new helper function to pull this value, since it's in a different place depending on whether or not we're running a composite scheduler.

Maybe use this function as a starting point:
https://github.com/twosigma/waiter/blob/d09cb7cbd3942da7b541b936873af9834eb2264c/waiter/src/waiter/util/client_tools.clj#L857-L863

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -40,6 +40,19 @@
...
             (is (< rs-snapshot-version' rs-watch-version'))))))))
 
+(deftest ^:parallel ^:integration-slow test-kubernetes-custom-image
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (let [{:keys [body]} (make-kitchen-request
+                                          waiter-url
+                                          {:x-waiter-name (rand-name)
+                                           :x-waiter-image "twosigma/integration"
+                                           :x-waiter-cmd "cd /tmp && python3 -m http.server $PORT0"
```
DaoWen: You put the canary file in `/tmp`? You should probably add a comment explaining how/why this test works.

In hindsight, I would have preferred that we add a custom script to the `$PATH` in the test image, and then the command could have simply been `run-custom-image-test-server` or something like that. (The script would have handled the logic for starting the server with the custom content.) But it's probably not worth refactoring the docker image at this point.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -40,6 +40,19 @@
             (is (<= initial-rs-snapshot-version rs-snapshot-version'))
             (is (< rs-snapshot-version' rs-watch-version'))))))))
 
+(deftest ^:parallel ^:integration-slow test-kubernetes-custom-image
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (let [{:keys [body]} (make-kitchen-request
+                                          waiter-url
```
DaoWen: Please fix the indentation here.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -40,6 +40,20 @@
...
             (is (< rs-snapshot-version' rs-watch-version'))))))))
 
+; test that we can provide a custom docker image that contains /tmp/index.html with "Integration Test Image" in it
+(deftest ^:parallel ^:integration-slow test-kubernetes-custom-image
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (let [{:keys [body]} (make-kitchen-request
+                             waiter-url
+                             {:x-waiter-name (rand-name)
+                              :x-waiter-image (System/getenv "INTEGRATION_TEST_CUSTOM_IMAGE")
```
DaoWen: What happens if this is not set? It seems like we should have an assertion for that.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -40,6 +40,20 @@
...
+                              :x-waiter-cmd "cd /tmp && python3 -m http.server $PORT0"
```
DaoWen: This seems like it will be very difficult to test with anything but the "twosigma/integration" image you built for this test. Something a little more flexible might be to use a custom environment variable set to some sentinel value (e.g., `WAITER_CUSTOM_IMAGE_SENTINEL='Integration Test Image'`), which is used in the custom command. The assertion at the end of this test would compare with the sentinel value.

shamsimam:
`waiter.scheduler/validate-service` for all the schedulers that do not support images (i.e. except the kubernetes scheduler) should throw an exception when the `image` field is provided in the service description.

PR #586, opened 2019-02-12
fixes the token syncer tests to handle for the new cluster metadata parameter

shamsimam:

- fixes the token syncer tests to handle for the new cluster metadata parameter

We wish to have green builds.

### DIFF
--- a/token-syncer/integration/token_syncer/basic_test.clj
+++ b/token-syncer/integration/token_syncer/basic_test.clj
@@ -38,6 +38,10 @@
         (str/split #",")
         sort)))
 
+(defn- waiter-url->cluster
+  [waiter-url]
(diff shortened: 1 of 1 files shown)


token-syncer/integration/token_syncer/basic_test.clj
```diff
@@ -38,6 +38,10 @@
         (str/split #",")
         sort)))
 
+(defn- waiter-url->cluster
```
nsinkov: documentation? :D

nsinkov: documentation? :D

shamsimam: added docstrings.

PR #587, opened 2019-02-13
upgrades clojure version to 1.10

shamsimam:

- upgrades clojure version to 1.10

Running the latest version of Clojure simplifies our move to JDK9/10 when required.


shamsimam:
Un-wip-ed the PR :)

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -759,12 +759,13 @@
   "Detects and marks lost processes."
   [running-pids id->instance]
   (pc/map-vals
-    (fn [{:strs [killed? shell-scheduler/pid] :as instance}]
-      (cond-> (-> (pc/map-keys keyword instance)
-                  (update :started-at str-to-date-safe))
-        (and (not killed?) (not (contains? running-pids pid)))
-        (assoc :killed? true
-               :message "Process lost after restart")))
+    (fn [{:strs [killed?] :as instance}]
```
DaoWen:
```suggestion
    (fn [{:strs [killed?] pid "shell-scheduler/pid" :as instance}]
```

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -759,12 +759,13 @@
...
+      (let [pid (get instance "shell-scheduler/pid")]
```
DaoWen: if you move this back to the destructuring above, then we don't need this `let`.

shamsimam: removed.

waiter/src/waiter/scheduler/shell.clj
shamsimam: Done.

PR #588, opened 2019-02-16
defaults the kerberos throttler core pool size to the concurrency level

shamsimam:

- defaults the kerberos throttler core pool size to the concurrency level

ThreadPoolExecutor spawns additional threads only when the queue becomes full. Instead, we want to have threads available to be processing tasks from the queue as soon as there are pending tasks.


sradack:
This was the change I expected.

shamsimam:
Build is now green.

PR #589, opened 2019-02-17
Persist custom K8s scheduler options

DaoWen:

Persist custom (namespaced) options to the Kubernetes scheduler.

This provides a way for custom "plugin" logic (e.g., a custom ReplicaSet Builder function) to read options from the Waiter configuration file.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -471,6 +471,7 @@
 (defrecord KubernetesScheduler [api-server-url
                                 authorizer
                                 cluster-name
+                                custom-options
                                 daemon-state
                                 fileserver
                                 http-client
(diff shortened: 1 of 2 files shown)


waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -776,7 +776,8 @@
                     :replicaset-api-version "extensions/v1beta1"
                     :replicaset-spec-builder {:factory-fn 'waiter.scheduler.kubernetes/default-replicaset-builder
                                               :default-container-image "twosigma/kitchen:latest"}
-                    :url "http://127.0.0.1:8001"}
+                    :url "http://127.0.0.1:8001"
+                    :custom-plugin/option "special"}
```
shamsimam: Nitpick: lexicographic order

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -977,6 +978,7 @@
          (fn? start-scheduler-syncer-fn)
          (or (nil? watch-retries) (integer? watch-retries))]}
   (let [authorizer (utils/create-component authorizer)
+        custom-options (into {} (filter (comp namespace key) context))
```
shamsimam: Why are any namespaced keys automatically promoted to custom options? I would prefer we reserve a special key for custom options that maps to a dictionary of values, e.g.
```:custom-optiiosn {}```

Comment (DaoWen) [reply waiter/src/waiter/scheduler/kubernetes.clj]: That does seem more straightforward. I'll refactor to do that instead.

**Comment by DaoWen** [comment]
@shamsimam - Refactored

**Comment by DaoWen** [comment]
> We should also add a unit tests to verify the values received by the k8s scheduler.

Do you want additional tests, or does the test I already wrote cover this?

File: waiter/test/waiter/scheduler/kubernetes_test.clj [hunk 3]
```diff
@@ -813,6 +814,10 @@
         (testing "should work with valid configuration"
           (is (instance? KubernetesScheduler (kubernetes-scheduler base-config))))
 
+        (testing "should retain custom plugin options"
+          (is (= (:custom-options base-config)
+                 (-> base-config kubernetes-scheduler :custom-options))))
```
DaoWen: > We should also add a unit tests to verify the values received by the k8s scheduler.

Did you want additional tests, or was this test sufficient?

shamsimam: This is good.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -761,6 +761,7 @@
                     :authorizer {:kind :default
                                  :default {:factory-fn 'waiter.authorization/noop-authorizer}}
                     :cluster-name "waiter"
+                    :custom-options '{a 1 b "two"}
```
shamsimam:
```suggestion
                    :custom-options {:a 1 :b "two"}
```

Why is there a leading tick in front of the map? Also, should we require the keys to be keywords?

DaoWen: I don't see the value in restricting the keys to be keywords. I don't think we enforce that anywhere else, and I don't like restricting the user when I don't have a strong reason for doing so. However, I did change the test map to use keywords, as you suggested, for consistency.

PR #613, opened 2019-03-09
loads only valid previous descriptors

shamsimam:

- loads only valid previous descriptors

Fallback support should look back in history to a valid service description and the lookup should not be cut short by an intermediate invalid description.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -129,7 +129,7 @@
 (defmacro assert-token-response
   "Asserts the token data in the response"
   ([response token-root token-cluster service-id-prefix deleted]
-   `(assert-token-response ~response ~token-root~token-cluster ~service-id-prefix ~deleted true))
+   `(assert-token-response ~response ~token-root ~token-cluster ~service-id-prefix ~deleted true))
   ([response token-root token-cluster service-id-prefix deleted include-metadata]
    `(let [body# (:body ~response)
           token-description# (parse-token-description body#)]
@@ -374,10 +374,10 @@
                 services (json/read-str body)
                 service-tokens (mapcat (fn [entry]
                                          (some->> entry
-                                                  walk/keywordize-keys
-                                                  :source-tokens
-                                                  flatten
-                                                  (map :token)))
+                                           walk/keywordize-keys
+                                           :source-tokens
+                                           flatten
+                                           (map :token)))
                                        services)]
             (assert-response-status response 200)
             (is (= (count @service-ids-atom) (count service-tokens))
@@ -394,10 +394,10 @@
                 services (json/read-str body)
                 service-tokens (mapcat (fn [entry]
                                          (some->> entry
-                                                  walk/keywordize-keys
-                                                  :source-tokens
-                                                  flatten
-                                                  (map :token)))
+                                           walk/keywordize-keys
+                                           :source-tokens
+                                           flatten
+                                           (map :token)))
                                        services)]
             (assert-response-status response 200)
             (is (= 3 (count service-tokens))
@@ -415,10 +415,10 @@
                   services (json/read-str body)
                   service-token-versions (mapcat (fn [entry]
                                                    (some->> entry
-                                                            walk/keywordize-keys
-                                                            :source-tokens
-                                                            flatten
-                                                            (map :version)))
+                                                     walk/keywordize-keys
+                                                     :source-tokens
+                                                     flatten
+                                                     (map :version)))
                                                  services)]
               (assert-response-status response 200)
               (is (= 1 (count service-token-versions))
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/descriptor.clj
```diff
@@ -232,30 +232,36 @@
     (-> (headers/split-headers (:headers request))
         (sd/merge-service-description-sources kv-store waiter-hostnames service-description-defaults token-defaults)
         (sd/merge-service-description-and-id kv-store service-id-prefix current-request-user metric-group-mappings
-                                             service-description-builder assoc-run-as-user-approved?)
+                                             service-description-builder assoc-run-as-user-approved? true)
         (sd/merge-suspended kv-store))))
 
 (defn descriptor->previous-descriptor
```
sradack: `descriptor->previous-descriptor` doesn't match what this function is doing anymore.   The goal of this change is to change our fallback descriptors.  We have a function for that called `retrieve-fallback-descriptor`.  Should we change it there instead?  Are we intentionally ignoring `fallback-period-secs` for invalid descriptors?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1397,3 +1397,49 @@
...
+      (-> token-description-1
+          (as-> token-description (post-token waiter-url token-description))
+          (assert-response-status 200))
+      (let [service-id-1 (retrieve-service-id waiter-url request-headers)]
+        (with-service-cleanup
+          service-id-1
+          (let [response-1 (make-request-with-debug-info request-headers #(make-request waiter-url "/hello" :headers %))]
+            (assert-response-status response-1 200)
+            (is (= service-id-1 (:service-id response-1))))
+          (-> token-description-2
```
sradack: Threading a small number of calls while also utilizing `as->` feels a bit awkward.  Would it be simpler to use `let` or write it normally?  e.g.  `(assert-response-status (post-token waiter-url token-description-2) 200))`

waiter/src/waiter/service_description.clj
```diff
@@ -854,7 +854,7 @@
      If after the merge a permitted-user is not available, then `username` becomes the permitted-user."
     [{:keys [defaults headers service-description-template source-tokens token-authentication-disabled token-preauthorized]}
      waiter-headers passthrough-headers kv-store service-id-prefix username metric-group-mappings
-     service-description-builder assoc-run-as-user-approved?]
+     service-description-builder assoc-run-as-user-approved? throw-if-invalid?]
```
sradack: Is a better approach to return `nil` when a service description would be invalid, and change the existing callers to throw on `nil`?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -374,10 +374,10 @@
                 services (json/read-str body)
                 service-tokens (mapcat (fn [entry]
                                          (some->> entry
-                                                  walk/keywordize-keys
```
sradack: Nitpick: Unrelated formatting changes throughout this file

waiter/src/waiter/service_description.clj
shamsimam: We still need metadata on why the service description is invalid to report in our error messages.

waiter/src/waiter/descriptor.clj
shamsimam: > Are we intentionally ignoring fallback-period-secs for invalid descriptors?

Only the `fallback-period-secs` of the latest version of the token is considered. We ignore values of `fallback-period-secs` from previous versions of the token.

waiter/integration/waiter/token_request_test.clj

waiter/src/waiter/descriptor.clj
shamsimam: Moved the logic to `retrieve-fallback-descriptor`

waiter/integration/waiter/token_request_test.clj
shamsimam: Undone.

waiter/src/waiter/service_description.clj
sradack: It could still return a map with partial results instead of nil.  I guess my larger point is that when it becomes desirable to have a function not throw in some cases, it seems like we should make it not throw in all cases, and rework the callers that expected an exception, as opposed to passing a parameter into the function causing it to throw in certain circumstances.  That seems cleaner to me.

shamsimam:
@sradack ready for review

waiter/src/waiter/service_description.clj
```diff
@@ -894,34 +894,43 @@
...
+         :service-preauthorized service-preauthorized
+         :source-tokens source-tokens})))
+
+  (defn validate-service-description
+    "Returns nil if the provided descriptor contains a valid service description.
+     Else it returns an instance of Throwable that reflects the validation error."
+    [kv-store service-description-builder
+     {:keys [core-service-description passthrough-headers service-description service-id waiter-headers]}]
+    (sling/try+
+      (let [stored-service-description? (fetch-core kv-store service-id)]
```
sradack: The `?` on the end to me implies it's a boolean, but it's not.

shamsimam: Removed the `?`

shamsimam:
@sradack ready for review

PR #637, opened 2019-03-21
allows new instance knowledge to be synced across clusters

shamsimam:

- allows new instance knowledge to be synced across clusters

Flaky test!

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1404,8 +1404,6 @@
           token (create-token-name waiter-url service-name)
           request-headers {:x-waiter-token token}
           fallback-period-secs 300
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/token_request_test.clj
```diff
@@ -1429,6 +1431,8 @@
           (let [response-1 (make-request-with-debug-info request-headers #(make-request waiter-url "/hello" :headers %))]
             (assert-response-status response-1 200)
             (is (= service-id-1 (:service-id response-1))))
+          ;; allow every router to learn about the new instance
+          (Thread/sleep scheduler-sync-allowance-sleep-duration-ms)
```
dposada: Why not use the wait-for pattern?

shamsimam: Done, ready for another round of review.

PR #640, opened 2019-03-22
avoids displaying key and trust store passwords

shamsimam:

- avoids displaying key and trust store passwords

We wish to avoid displaying passwords in our apis.

Screen Shot 2019-03-22 at 4 15 47 PM

### DIFF
--- a/waiter/src/waiter/settings.clj
+++ b/waiter/src/waiter/settings.clj
@@ -225,11 +225,13 @@
                              (map? nested-data)
                              (contains? nested-data :factory-fn))))
              (select-keys data))))
-    (cond-> (utils/dissoc-in settings [:zookeeper :connect-string])
+    (cond-> settings
       (get-in settings [:server-options :key-password])
-      (assoc-in [:server-options :key-password] "_hidden_")
+      (assoc-in [:server-options :key-password] "<hidden>")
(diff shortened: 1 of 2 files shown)


sradack:
Why aren't we using the `#config/env` reader tag?

shamsimam:
@sradack I also don't want them to show up in the settings endpoint. With this change, I do not have to worry about remembering to delete the password entries before display in the handler.

waiter/src/waiter/settings.clj
```diff
@@ -214,22 +214,29 @@
...
+      (if-not (and (map? data) (contains? data :kind))
+        data
+        (->> (keys data)
+             (remove #(let [nested-data (get data %)]
+                        (and (not= % :kind)
+                             (not= % (:kind data))
+                             (map? nested-data)
+                             (contains? nested-data :factory-fn))))
+             (select-keys data))))
+    (cond-> (utils/dissoc-in settings [:zookeeper :connect-string])
```
sradack: Why are we `dissoc`ing some and using `_hidden_` with others?

shamsimam: Done. Use `<hidden>` consistently.

shamsimam:
@sradack ready for review.

PR #644, opened 2019-03-26
converts the x-forwarded-proto to a keyword in request->scheme

shamsimam:

- converts the x-forwarded-proto to a keyword in request->scheme

Correct the mechanism for determining the request scheme.


dposada:

PR #648, opened 2019-03-26
adds support for accessing request and response trailers

shamsimam:

- adds support for accessing request and response trailers
- updates jet version that also [has changes for trailers support](https://github.com/twosigma/jet/pull/18).

HTTP requests/responses have first-class support for trailers.

### Note:

This PR acknowledges that trailer support in Jetty is inherently racy for HTTP/1.1 requests.


shamsimam:
Removing the hop-by-hop headers in the response means we need to rework this solution.

shamsimam:
@sradack 

> Shouldn't we have some code that is passing proxied trailers from the client to the backend

We do not need code in the Waiter router as it is handled by jet:

- The chunked response generated from the backend, has a [:trailer channel in the response](https://github.com/twosigma/jet/blob/21599451a0a4d69fb260307ab99d64e47368ee53/src/clj/qbits/jet/client/http.clj#L115).
- The :trailer gets returned as [part of the response](https://github.com/twosigma/waiter/blob/master/waiter/src/waiter/process_request.clj#L386) from the server (waiter router).
- This trailer then gets [picked up from the response map](https://github.com/twosigma/jet/blob/21599451a0a4d69fb260307ab99d64e47368ee53/src/clj/qbits/jet/servlet.clj#L273) and is used to populate the response from the router.

shamsimam:

sradack:
> @sradack
> 
> > Shouldn't we have some code that is passing proxied trailers from the client to the backend
> 
> We do not need code in the Waiter router as it is handled by jet:
> 
> * The chunked response generated from the backend, has a [:trailer channel in the response](https://github.com/twosigma/jet/blob/21599451a0a4d69fb260307ab99d64e47368ee53/src/clj/qbits/jet/client/http.clj#L115).
> * The :trailer gets returned as [part of the response](https://github.com/twosigma/waiter/blob/master/waiter/src/waiter/process_request.clj#L386) from the server (waiter router).
> * This trailer then gets [picked up from the response map](https://github.com/twosigma/jet/blob/21599451a0a4d69fb260307ab99d64e47368ee53/src/clj/qbits/jet/servlet.clj#L273) and is used to populate the response from the router.

We should add the following tests:

1. Waiter client speaking HTTP/1.1 whose request's trailer headers are verified to be passed to backend (not just Waiter itself)
2. Waiter client speaking HTTP/2 whose request's  trailer headers are verified to be passed to backend 
3. Waiter client speaking HTTP/1.1 whose request's  trailer headers are verified to be passed to backend, where the client introduces configurable delay between body and trailers.
4. Waiter client speaking HTTP/2 whose request's trailer headers are verified to be passed to backend, where the client introduces configurable delay between body and trailers.
5. Backend responding with HTTP/2 trailers in response where trailers are verified by client.
6. Backend responding with HTTP/1.1 trailers in response where backend introduces a configurable delay between completion of body and the sending of trailers, where trailers are verified by client.
7. Backend responding with HTTP/2 trailers in response where backend introduces a configurable delay between completion of body and the sending of trailers, where trailers are verified by client.
8. Backend responding with HTTP/1.1 trailers in response where backend responds with a failure status code (4XX/5XX), where trailers are verified by client.
9. Backend responding with HTTP/2 trailers in response where backend responds with a failure status code (4XX/5XX), where trailers are verified by client.

shamsimam:
I found that trailer support in Jetty is inherently racy for HTTP/1.1 requests, the trailers are sometimes (1 in 20 anecdotally) missing in responses. I haven't seen them missing in http/2 responses yet.

test-apps/README.md
```diff
@@ -10,3 +10,6 @@ The following test apps are included:
 - nginx:
```
sradack: TODO Do we still need nginx?

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+            while (true) {
+                final int bytesRead = inputStream.read(buffer);
+                if (bytesRead == -1) {
+                    LOGGER.info("total bytes read = " + totalBytesRead);
+                    break;
+                }
+                totalBytesRead += bytesRead;
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
sradack: Should we allow this exception to be thrown instead of catching & printing?

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+        LOGGER.info(correlationId + " received " + request.getProtocol() + " " + request.getMethod() +
+            " request at " + request.getRequestURI());
+        response.setHeader("x-cid", correlationId);
+        response.setHeader("server", "sediment:" + HttpConfiguration.SERVER_VERSION);
+
+        final Request jettyRequest = (Request) request;
+        final Response jettyResponse = (Response) response;
+
+        switch (request.getRequestURI()) {
+            case "/":
```
sradack: Let's remove "/" unless needed.  Do we need two endpoints that report status?

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+                    }
+                }
+                break;
+            }
+            respOutputStream.write(tempBytes, 0, bytesRead);
+            respOutputStream.flush();
+        }
+    }
+
+    private HttpFields retrieveTrailers(Request request) {
```
sradack: Could we name this `createResponseTrailers` instead of `retrieveTrailers`?

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.function.Function;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class Servlet extends HttpServlet {
+
+    private final static Logger LOGGER = Logger.getLogger(Servlet.class.getName());
+
+    static String retrieveCorrelationId(HttpServletRequest request) {
+        final Enumeration<String> headerNames = request.getHeaderNames();
```
sradack: Can we use `request.getHeader("x-cid")`?

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+
+    Map<String, Object> convertFieldsToMap(final Collection<String> names, final Function<String, Object> retrieveValue) {
+        final Map<String, Object> result = new HashMap<>();
+        for (String name : names) {
+            result.put(name, retrieveValue.apply(name));
+        }
+        return result;
+    }
+
+    Map<String, Object> populateRequestInfo(final Request request) throws IOException {
```
sradack: The goal of this function seems to limited to creating the response JSON.  It seems better to directly use a [JsonWriter](https://static.javadoc.io/com.google.code.gson/gson/2.6.2/com/google/gson/stream/JsonWriter.html) created on the response's output stream, avoiding the `ByteArrayInputStream`, etc.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+    private HttpFields retrieveTrailers(Request request) {
+        final HttpFields trailerFields = new HttpFields();
+        final String trailerPrefix = "x-sediment-response-trailer-";
+        for (String name : Collections.list(request.getHeaderNames())) {
+            if (name.toLowerCase().startsWith(trailerPrefix)) {
+                final String trailerName = name.substring(trailerPrefix.length());
+                trailerFields.add(trailerName, request.getHeader(name));
+            }
+        }
+        if (trailerFields.getFieldNamesCollection().isEmpty()) {
```
sradack: Nitpick: `trailerFields.size() == 0`

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+            return null;
+        } else {
```
sradack: Nitpick: remove `else`

test-apps/sediment/src/test/java/com/twosigma/waiter/sediment/ServletTest.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.ByteArrayInputStream;
+
+import static junit.framework.TestCase.assertEquals;
+
+/**
+ * Unit test for simple App.
+ */
+public class ServletTest {
+    /**
+     * Rigorous Test :-)
```
sradack: Much test

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,222 @@
...
+        final ServletOutputStream respOutputStream = response.getOutputStream();
+        while (true) {
+            final byte[] tempBytes = new byte[128];
+            final int bytesRead = respInputStream.read(tempBytes);
+            if (bytesRead == -1) {
+                final String sleepMsString = request.getHeader("x-sediment-sleep-before-response-trailer-ms");
+                if (sleepMsString != null) {
+                    try {
+                        LOGGER.info("sleeping " + sleepMsString + " ms before returning response trailers");
+                        Thread.sleep(Long.parseLong(sleepMsString));
```
sradack: This sleep is potentially preventing the emission of the terminating chunk.  The goal is to introduce the delay between the terminating chunk and terminating CRLF.

waiter/integration/waiter/basic_test.clj
```diff
@@ -379,6 +392,71 @@
...
+                                       (str "x-sediment-response-trailer-" k) (str v)))
+                                   (assoc request-headers
+                                     "x-sediment-response-status" response-status
+                                     "x-sediment-sleep-before-response-trailer-ms" response-trailer-delay-ms)
+                                   (seq response-trailers))
+                                 :body (ByteArrayInputStream. (.getBytes long-request))
+                                 :path "/trailers"
+                                 :protocol backend-proto
+                                 :trailers-fn (fn []
+                                                (Thread/sleep request-trailer-delay-ms)
```
sradack: I would love to confirm through other means that this sleep introduces the delay in the correct part of the stream.

test-apps/kitchen/bin/kitchen
```diff
@@ -554,7 +556,17 @@ class Kitchen(HTTPWebSocketsHandler):
                     chunk = bytes(itertools.islice(response_bytes, self.__chunk_size))
                     if len(chunk) == 0:
                         if self.__chunked:
-                            self.wfile.write(b'0\r\n\r\n')
+                            self.wfile.write(b'0\r\n')
+                            if self.__trailer_delay_secs > 0:
```
shamsimam: I verified with curl that this mechanism does indeed send the 0, sleeps and then sends the trailers.

I was unable to mimic the same behavior with a [Netty support for HTTP2 server](https://netty.io/4.1/api/io/netty/handler/codec/http2/package-summary.html). With the following code:
```
    private class EchoTrailersResponder extends RequestResponder {
        @Override
        void onHeadersRead(ChannelHandlerContext ctx, int streamId, boolean endOfStream, Http2Headers headers) {
            encoder().writeHeaders(ctx, streamId, createDefaultResponseHeaders(), 0, false, ctx.newPromise());
            ctx.flush();
            Thread.sleep(10000);

            encoder().writeData(ctx, streamId, RESPONSE_BYTES.duplicate(), 0, false, ctx.newPromise());
            ctx.flush();
            Thread.sleep(10000);

            Http2Headers responseTrailers = createResponseHeadersFromRequestHeaders(headers).add("trailer", "value1", "Value2");
            encoder().writeHeaders(ctx, streamId, responseTrailers, 0, true, ctx.newPromise());
            ctx.flush();
        }
    }
```
I saw in curl that we would receive the headers and then sleep 20 secs before receiving the body and trailers together. I verified that making requests with a jet client to this server received the body and headers though.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,225 @@
...
+
+        final String responseStatusString = request.getHeader("x-sediment-response-status");
+        if (responseStatusString == null) {
+            response.setStatus(200);
+        } else {
+            response.setStatus(Integer.parseInt(responseStatusString));
+        }
+
+        final ByteArrayInputStream respInputStream = new ByteArrayInputStream(outputBytes);
+        final ServletOutputStream respOutputStream = response.getOutputStream();
```
shamsimam: I use the output stream instead of the writer to mimic chunked encoding response below while working with byte arrays.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
```diff
@@ -0,0 +1,225 @@
...
+        final HttpFields trailerFields = new HttpFields();
+        final String trailerPrefix = "x-sediment-response-trailer-";
+        for (String name : Collections.list(request.getHeaderNames())) {
+            if (name.toLowerCase().startsWith(trailerPrefix)) {
+                final String trailerName = name.substring(trailerPrefix.length());
+                trailerFields.add(trailerName, request.getHeader(name));
+            }
+        }
+        if (trailerFields.size() == 0) {
+            return null;
```
shamsimam: `null` means we are not going to sending trailers, whereas a non-null return means we will send possibly empty trailers. Is not a big deal either way and would prefer not to change the code.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: As we spoke offline, jetty doesn't allow us to do that. I have added a similar behavior to kitchen (and added a test), but it works only for http 1.1 servers.

waiter/integration/waiter/basic_test.clj
shamsimam: I haven't confirmed this yet, will look into it and let you know what I find.

test-apps/sediment/src/main/java/com/twosigma/waiter/sediment/Servlet.java
shamsimam: Done.

PR #651, opened 2019-03-27
add image field validation per scheduler

nsinkov:

- add image field validation per scheduler

The new service image field is not supported by all schedulers

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -939,9 +939,7 @@
             (let [kitchen-image (System/getenv "INTEGRATION_TEST_KITCHEN_IMAGE")
                   _ (is (not (str/blank? kitchen-image)) "You must provide a kitchen image in the INTEGRATION_TEST_KITCHEN_IMAGE environment variable")]
               (make-kitchen-request-fn kitchen-image 200))
-            (using-cook? waiter-url)
-            (make-kitchen-request-fn "dummy/image" 500)
-            (using-marathon? waiter-url)
-            (make-kitchen-request-fn "dummy/image" 500)
-            (using-shell? waiter-url)
+            (or (using-cook? waiter-url)
+                (using-marathon? waiter-url)
+                (using-shell? waiter-url))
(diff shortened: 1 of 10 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -926,3 +926,22 @@
...
+                                                    waiter-url
+                                                    {:x-waiter-name (rand-name)
+                                                     :x-waiter-image %1}
+                                                    :path "/hello")
+                                         _ (assert-response-status response %2)])]
+      (cond (using-k8s? waiter-url)
+            (let [kitchen-image (System/getenv "INTEGRATION_TEST_KITCHEN_IMAGE")
+                  _ (is (not (str/blank? kitchen-image)) "You must provide a kitchen image in the INTEGRATION_TEST_KITCHEN_IMAGE environment variable")]
+              (make-kitchen-request-fn kitchen-image 200))
+            (using-cook? waiter-url)
```
shamsimam: Why not put the remaining clauses in an `or`?

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -488,8 +488,9 @@
      :syncer (retrieve-syncer-state-fn)})
 
   (validate-service [_ service-id]
-    (let [{:strs [run-as-user]} (service-id->service-description-fn service-id)]
-      (authz/check-user authorizer run-as-user service-id))))
+    (let [{:strs [run-as-user image]} (service-id->service-description-fn service-id)]
+      (authz/check-user authorizer run-as-user service-id)
+      (when image (throw (ex-info "Image field is set. Images are not supported with Marathon scheduler" {:image image}))))))
```
DaoWen: Copy-paste error: Marathon should be Cook here.

DaoWen: Would it be better to make a `reject-image-field` helper (maybe in scheduler.clj), and simply call it from each of the schedulers that doesn't support images? The scheduler name can be pulled from the `:name` on the scheduler object, and then added to the `ex-info` object along with the `:image` value.

PR #653, opened 2019-03-29
removes hop-by-hop headers in responses

shamsimam:

- removes connection-specific header fields when transmitting HTTP/1.x message via HTTP/2

> HTTP/2 does not use the Connection header field to indicate connection-specific header fields; in this protocol, connection-specific metadata is conveyed by other means. An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields; any message containing connection-specific header fields MUST be treated as malformed (Section 8.1.2.6).
> 
> The only exception to this is the TE header field, which MAY be present in an HTTP/2 request; when it is, it MUST NOT contain any value other than "trailers".
> 
> This means that an intermediary transforming an HTTP/1.x message to HTTP/2 will need to remove any header fields nominated by the Connection header field, along with the Connection header field itself. Such intermediaries SHOULD also remove other connection-specific header fields, such as Keep-Alive, Proxy-Connection, Transfer-Encoding, and Upgrade, even if they are not nominated by the Connection header field.

https://http2.github.io/http2-spec/

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -499,6 +499,7 @@
                           (catch Exception e
                             (async/close! request-state-chan)
                             (handle-process-exception e request)))
+                        (update :headers headers/dissoc-hop-by-hop-headers)
                         (assoc :get-instance-latency-ns instance-elapsed
                                :instance instance
                                :protocol proto-version)
--- a/test-apps/nginx/bin/run-nginx-server.sh
+++ b/test-apps/nginx/bin/run-nginx-server.sh
@@ -5,12 +5,28 @@
 # Launches kitchen fronted by a nginx instance to handle
 # incoming requests using the specified protocol.
 # Supported protocols are: http, https, h2c, and h2.
+# Expects the following environment variables to be set:
+# - KITCHEN_CMD the command to run kitchen, port will be specified with the -p flag
+# - MESOS_SANDBOX the sandbox directory location
+# - PORT0 the port nginx server will listen on for configured protocol requests
+# - PORT1 the port nginx server will listen on for http/1.1 requests
+# - PORT2 the port the kitchen server will listen
 
 set -ux
 
 # Log a message to stdout
-nginx_log() {
+function nginx_log() {
   printf '%s run-nginx-server.sh: %s\n' "$(date +'%Y-%m-%dT%H:%M:%S%z')" "$1"
+}
+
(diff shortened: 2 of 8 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -408,13 +409,23 @@
   (track-process-error-metrics descriptor)
   (utils/exception->response exception request))
 
+(defn cleanup-response-headers
+  "Removes connection-specific header fields when transmitting HTTP/1.x message via HTTP/2."
+  [{:keys [headers] :as response} internal-protocol backend-protocol]
+  (cond-> response
+    (and (= internal-protocol "HTTP/2.0") (not= backend-protocol "HTTP/2.0"))
```
sradack: Shouldn't we be removing these headers in all cases, not use HTTP/2?  See https://www.mnot.net/blog/2011/07/11/what_proxies_must_do "Remove Hop-by-hop headers"

sradack: We do this already for the request headers.

shamsimam: I had missed that hop-by-hop headers need to be removed for responses too. Thanks, done.

waiter/src/waiter/process_request.clj
```diff
@@ -23,6 +23,7 @@
             [metrics.histograms :as histograms]
             [metrics.meters :as meters]
             [metrics.timers :as timers]
+            [plumbing.core :as pc]
```
sradack: unused

waiter/src/waiter/process_request.clj
```diff
@@ -414,7 +415,7 @@
     [make-request-fn instance-rpc-chan start-new-service-fn
      instance-request-properties determine-priority-fn process-backend-response-fn
      request-abort-callback-factory local-usage-agent
-     {:keys [ctrl descriptor request-id request-time] :as request}]
+     {:keys [ctrl descriptor internal-protocol request-id request-time] :as request}]
```
sradack: unused

shamsimam: Removed.

waiter/src/waiter/process_request.clj
shamsimam: Removed.

shamsimam:
@sradack ready to merge (unrelated CLI tests failing).

PR #658, opened 2019-03-30
removes protocol field from instance

shamsimam:

- removes protocol field from instance

Refactoring. The protocol should not be a part of the service instance. This refactoring simplifies the [introduction of health check protcol](https://github.com/twosigma/waiter/pull/659).


DaoWen:
> The protocol should not be a part of the service instance.

Could you please clarify why?

shamsimam:
@DaoWen 

>> The protocol should not be a part of the service instance.

> Could you please clarify why?

1. The protocol is service-specific, so it doesn't make sense to repeat it on the instances.

2. We are working towards making the health check proto be separate from the backend proto (https://github.com/twosigma/waiter/pull/659). This feature will allow us to have http healthchecks for grpc services which will have backend-proto of grpc. In that world, we do not want the protocol to be tied to the instance while generating urls, instead the protcol should be passed in explicitly.

waiter/src/waiter/scheduler.clj
```diff
@@ -190,24 +189,23 @@
...
-  [{:keys [host port protocol]}]
+  [^String protocol ^String host port]
   (let [scheme (hu/backend-proto->scheme protocol)]
     (str scheme "://" host ":" port)))
 
 (defn end-point-url
   "Returns the endpoint url which can be queried on the url specified by the url-info map."
-  [url-info ^String end-point]
-  (str (base-url url-info)
+  [^String protocol ^String host port ^String end-point]
```
DaoWen: nitpick: why does everything have a hint except `port`?

shamsimam: I didn't add the type hint as I wasn't sure if it is an int or a string from the different paths where this is invoked. I can make the check and force it to `int` if you like.

shamsimam: We don't have the type hint on `ServiceInstance` either, let's skip it for now.

PR #665, opened 2019-04-02
avoids race in kitchen starting up before nginx proxy

shamsimam:

- avoids race in kitchen starting up before nginx proxy

The test is currently flaky since the nginx proxy starts up after the kitchen server. Since the health checks run against kitchen, it can trigger sending a request from Waiter to the nginx port and cause the connection to fail.


shamsimam:
@DaoWen build is green :)

PR #666, opened 2019-04-02
Fix race in test-create-service-if-new unit test

DaoWen:

- Fix race with agent in `test-create-service-if-new` unit test.

Deflake this unit test.


DaoWen:
@shamsimam - The patched test passed in a loop on another branch:
https://travis-ci.org/DaoWen/waiter/jobs/514841302#L1496

PR #684, opened 2019-04-10
Disable S3 log server tests on Travis

DaoWen:

Disable S3 log server tests on Travis.

Copying logs to S3 is causing pods to stay around too long in Travis,
which in turn is causing tests to flake due to unschedulable pods.


DaoWen:
This doesn't seem to actually help the Travis job times or stability. I'm closing it for now.

PR #688, opened 2019-04-11
adds support for populating WAITER_CLUSTER environment variable

shamsimam:

- adds support for waiter cluster name env variable

We would like to allow services launched via Waiter to know which Waiter cluster they are running on. This allows the service to make cluster-specific decisions inside its configuration.

### Note:

This PR also exposes the settings in a promise inside `config.clj` enabling mount-like behavior inside waiter for the settings which are effectively constants. This allows us to avoid threading through the cluster name across all the schedulers down to the function that builds the environment. The dependency on the need for the config is still explicit as the tests need to introduce the appropriate `with-redef`.

### DIFF
--- a/waiter/src/waiter/config.clj
+++ b/waiter/src/waiter/config.clj
@@ -1,0 +1,37 @@
+;;
+;; Copyright (c) Two Sigma Open Source, LLC
+;;
+;; Licensed under the Apache License, Version 2.0 (the "License");
+;; you may not use this file except in compliance with the License.
+;; You may obtain a copy of the License at
+;;
+;;  http://www.apache.org/licenses/LICENSE-2.0
+;;
+;; Unless required by applicable law or agreed to in writing, software
+;; distributed under the License is distributed on an "AS IS" BASIS,
+;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
+;;
+(ns waiter.config)
+
+(def ^:private config-promise (promise))
+
+(defn initialize-config
+  "Initializes the config."
+  [config-value]
+  (if-not (realized? config-promise)
+    (deliver config-promise config-value)
+    (throw (IllegalStateException. "The config has already been initialized"))))
+
+(defn- check-config-initialized
+  "Throws an exception if the config has not been initialized."
+  []
+  (when-not (realized? config-promise)
+    (throw (IllegalStateException. "The config has not been initialized"))))
+
+(defn retrieve-cluster-name
+  "Retrieves the configured cluster name."
+  []
+  (check-config-initialized)
+  (get-in @config-promise [:cluster-config :name]))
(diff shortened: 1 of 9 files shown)


waiter/src/waiter/config.clj
```diff
@@ -0,0 +1,37 @@
...
+;;
+(ns waiter.config)
+
+(def ^:private config-promise (promise))
+
+(defn initialize-config
+  "Initializes the config."
+  [config-value]
+  (if-not (realized? config-promise)
+    (deliver config-promise config-value)
```
DaoWen: This operation is racy. Should we have a lock to ensure the `realized?` and `deliver` operations happen atomically?
Alternatively, we could `deliver` the value, and then require `(identical? @config-promise config-value)`. (If the contents are identical, do we really care about the double-initialization?)

waiter/src/waiter/config.clj
```diff
@@ -0,0 +1,37 @@
...
+;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;; See the License for the specific language governing permissions and
+;; limitations under the License.
+;;
+(ns waiter.config)
+
+(def ^:private config-promise (promise))
+
+(defn initialize-config
+  "Initializes the config."
```
DaoWen: I think we need a more complete docstring here. Maybe explain that this is the global waiter config, as read at launch time, after post-processing. The configuration is constant for a given Waiter instantiation/process.

waiter/src/waiter/config.clj
```diff
@@ -0,0 +1,37 @@
...
+  [config-value]
+  (if-not (realized? config-promise)
+    (deliver config-promise config-value)
+    (throw (IllegalStateException. "The config has already been initialized"))))
+
+(defn- check-config-initialized
```
DaoWen: I like the word `require` better than `check` here. We could also use Clojure's precondition support.

waiter/src/waiter/config.clj
shamsimam: Updated the docstring. Let me know whether you like it.

waiter/src/waiter/config.clj
shamsimam: Done.

waiter/src/waiter/config.clj
shamsimam: Changed to use pre-conditions.

shamsimam:
@DaoWen I thought about this some and I agree with you regarding locality sensitive decisions. I still feel though having the cluster information available in the environment enables other features e.g. it allows services to track when they are running on a cluster other than where the token was registered.

@sradack also brought up an interesting idea of having cluster-specific environment entries for tokens where downstream dependencies could be configured. This would have to be supported as a new parameter. In such a world, having the cluster environment variable would still help with debugging as it informing which cluster name was used to load the cluster-specific entries.

shamsimam:
@sradack renamed to `WAITER_CLUSTER`

shamsimam:
Rebased to avoid merge conflicts.

PR #690, opened 2019-04-11
delete service after kitchen call

nsinkov:

- add call to delete service after kitchen call

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -1016,12 +1016,13 @@
 
 (deftest ^:parallel ^:integration-slow test-image-field-validation
   (testing-using-waiter-url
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -1016,12 +1016,13 @@
 
 (deftest ^:parallel ^:integration-slow test-image-field-validation
   (testing-using-waiter-url
-    (let [make-kitchen-request-fn #(let [response (make-kitchen-request
+    (let [make-kitchen-request-fn #(let [{:keys [service-id] :as response} (make-kitchen-request
```
shamsimam: Is `service-id` available in the response? I thought that only gets added with `make-request-with-debug-info`.

We also have a `with-service-cleanup` macro that can help cleanup the service.

PR #691, opened 2019-04-12
Token syncer additional success scenarios

shamsimam:

- handles skipping syncing tokens if the user-specified keys are the same
- handles hard-deleting tokens if they are already soft-deleted, irrespective of other parameters

Reduces the number of failure scenarios in the token syncer.

### DIFF
--- a/token-syncer/src/token_syncer/commands/syncer.clj
+++ b/token-syncer/src/token_syncer/commands/syncer.clj
@@ -74,11 +74,12 @@
 (defn sync-token-on-clusters
   "Syncs a given token description on all clusters.
    If the cluster-url->token-data says that a given token was not successfully loaded, it is skipped.
-   Token sync-ing is also skipped if the roots of the tokens are different."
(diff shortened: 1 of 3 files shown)


token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -97,6 +98,26 @@
...
 
+                  (and latest-root
+                       (= latest-token-description description))
+                  {:code :success/token-match}
+
+                  ;; deleted on both clusters, irrespective of remaining values
+                  (and (get latest-token-description "deleted")
+                       (get description "deleted"))
+                  {:code :error/tokens-deleted
+                   :details {:message "token deleted on both clusters should be handled separately"}}
```
dposada: What does "should be handled separately" mean?

shamsimam: Updated the messaging.

PR #694, opened 2019-04-15
Add WAITER_SANDBOX environment variable

DaoWen:

- Add `WAITER_SANDBOX` environment variable
- Rename existing waiter-init -> waiter-k8s-init
- Add waiter-mesos-init
- Update docker containers with both new scripts

We want a `WAITER_SANDBOX` environment variable that apps can use across both Mesos and Kubernetes instead of using `MESOS_SANDBOX`.

### DIFF
--- a/waiter/bin/ci/run-integration-tests-k8s-scheduler.sh
+++ b/waiter/bin/ci/run-integration-tests-k8s-scheduler.sh
@@ -32,14 +32,14 @@ fi
 ${CONTAINERS_DIR}/bin/build-docker-images.sh
 
 # Start waiter
-export WAITER_PORT=9091
+: ${WAITER_PORT:=9091}
 ${WAITER_DIR}/bin/run-using-k8s.sh ${WAITER_PORT} &
 
 # Start monitoring state of Kubernetes pods
 bash +x ${DIR}/monitor-pods.sh &
 
 # Run the integration tests
-export INTEGRATION_TEST_CUSTOM_IMAGE="twosigma/integration"
+export INTEGRATION_TEST_CUSTOM_IMAGE="twosigma/integration:latest"
 export INTEGRATION_TEST_CUSTOM_IMAGE_ALIAS="alias/integration"
 export INTEGRATION_TEST_KITCHEN_IMAGE="twosigma/waiter-test-apps"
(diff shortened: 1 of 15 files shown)


waiter/bin/ci/run-integration-tests-k8s-scheduler.sh
```diff
@@ -32,14 +32,14 @@ fi
...
-export WAITER_PORT=9091
+: ${WAITER_PORT:=9091}
 ${WAITER_DIR}/bin/run-using-k8s.sh ${WAITER_PORT} &
 
 # Start monitoring state of Kubernetes pods
 bash +x ${DIR}/monitor-pods.sh &
 
 # Run the integration tests
-export INTEGRATION_TEST_CUSTOM_IMAGE="twosigma/integration"
+export INTEGRATION_TEST_CUSTOM_IMAGE="twosigma/integration:latest"
```
shamsimam: Why do we need to provide the tag explicitly?

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -246,8 +246,9 @@
 
 (defn default-marathon-descriptor-builder
   "Returns the descriptor to be used by Marathon to create new apps."
-  [home-path-prefix service-id->password-fn {:keys [service-id service-description]} _]
-  (let [health-check-url (sd/service-description->health-check-url service-description)
+  [home-path-prefix service-id->password-fn {:keys [service-id service-description]} {:keys [container-init-commands]}]
+  (let [base-command-vector (vec (or container-init-commands ["waiter-mesos-init"]))
```
shamsimam: Should we allow empty/missing `container-init-commands` instead of providing a default here? Instead, we can require that `container-init-commands` ne non-empty instead of providing a default.

waiter/bin/ci/run-integration-tests-k8s-scheduler.sh
DaoWen: We don't. Removed.

waiter/src/waiter/scheduler/marathon.clj
DaoWen: Done!

PR #696, opened 2019-04-16
add marathon descriptor builder

nsinkov:

- add marathon descriptor builder. remove image aliases from marathon scheduler

we want to add default marathon constraints. instead of extending the constraint logic in the marathon scheduler, we'll allow a custom descriptor builder and remove the existing image alias logic


DaoWen:
This looks good to me. I like the symmetry with the RepicaSet builder function in our K8s scheduler.

PR #697, opened 2019-04-16
Don't retry k8s watch returning failure status

DaoWen:

- Generally improve handling of errors in the k8s watch threads
- Specifically handle "resource version too old" errors by immediately falling back to a global state query

Currently, we always exhaust our watch `iter` retry limit before falling back to a global query, even when our current resource version is too old (and thus we know all retries will fail until we do a global sync). This bad behavior was caused by refactoring #568. This patch cleans up the error handling code, improves logging, and specifically addresses the "resource version too old" error case.


DaoWen:
This patch is much easier to review if you suppress whitespace changes in the diff.

PR #703, opened 2019-04-22
Adds gGRPC example service and integration test

shamsimam:

- adds gGRPC example service named courier
- adds different flavors of grpc invocations as integration test
- reduces number of pending puts by avoiding offering empty data into the body channel
- avoids using chunked response folding for http/2 requests

We would like to confirm that gRPC services can work on top of Waiter.


containers/test-apps/courier/src/main/java/com/twosigma/waiter/courier/GrpcClient.java
```diff
@@ -0,0 +1,280 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class GrpcClient {
+
+    private static Function<String, Void> logFunction = new Function<String, Void>() {
+        @Override
+        public Void apply(final String message) {
+            System.out.println(message);
```
sradack: Should we use a `LOGGER` like we have in the server?

containers/test-apps/courier/src/main/java/com/twosigma/waiter/courier/GrpcClient.java
```diff
@@ -0,0 +1,280 @@
...
+                .newBuilder()
+                .setId(id)
+                .setFrom(from)
+                .setMessage(message)
+                .build();
+            final CourierReply response;
+            try {
+                response = futureStub.sendPackage(request).get();
+            } catch (final StatusRuntimeException e) {
+                logFunction.apply("RPC failed, status: " + e.getStatus());
```
sradack: Should we be making direct calls to a `LOGGER` in nearly every case that `logFunction` is currently being used?

containers/test-apps/courier/src/main/java/com/twosigma/waiter/courier/GrpcClient.java
```diff
@@ -0,0 +1,280 @@
...
+                            if (lockStepMode) {
+                                logFunction.apply("releasing semaphore after receiving response");
+                                lockStep.release();
+                            }
+                        }
+
+                        @Override
+                        public void onError(final Throwable throwable) {
+                            logFunction.apply("error in collecting summaries " + throwable);
+                            responsePromise.complete(null);
```
sradack: Should we be somehow interrupting the client message loop if there's an error here?  Could it become deadlocked?

containers/test-apps/courier/src/main/java/com/twosigma/waiter/courier/HealthCheckServer.java
```diff
@@ -0,0 +1,71 @@
...
+
+            final Headers responseHeaders = httpExchange.getResponseHeaders();
+            responseHeaders.add("content-type", "text/plain");
+            responseHeaders.add("server", "courier-health-check/jdk" + System.getProperty("java.version"));
+            if (correlationId != null) {
+                responseHeaders.add("x-cid", correlationId);
+            }
+
+            final String response = "OK";
+            httpExchange.sendResponseHeaders(200, response.length());
```
sradack: Should this be the length of `response.getBytes()`?

PR #705, opened 2019-04-23
Restructure apps and scripts for docker image builds

DaoWen:

- Reorganized the test-apps directory to make it more flat
- Moved the kubernetes fileserver docker image files into test-apps
- Refactored build-docker-images.sh builds all of our docker images
- Always use "latest" tag for docker images (which we expect local)
- Update integration scrips for smoother experience, both when run locally and when run in Travis.

We want to ensure we're always running the latest version of our code/configuration in our test containers when running in Travis, and we want to do that by always running a locally-built "latest" container image.


shamsimam:
We spoke offline, awaiting directory layout changes.

token-syncer/bin/ci/run-integration-tests.sh
```diff
@@ -16,7 +16,7 @@ TEST_SELECTOR=${2:-integration}
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 PROJECT_DIR=${DIR}/../../../
 WAITER_DIR=${PROJECT_DIR}/waiter
-KITCHEN_DIR=${PROJECT_DIR}/test-apps/kitchen
+KITCHEN_DIR=${PROJECT_DIR}/containers/test-apps/kitchen
```
DaoWen: I missed updating this path on my first pass after reorganizing the directory structure, but no tests failed. Do we even need this?

DaoWen:
Renamed the old top-level `test-apps` directory to `containers`, moved the test-apps projects into a new `test-apps` subdirectory, and added some READMEs. Ready for review.

containers/test-apps/Dockerfile
```diff
@@ -3,14 +3,11 @@ FROM twosigma/mesos-agent:1.0.0-1.0.1604
...
-  && apt-get install -y nginx \
-  && apt-get install -y python3
+RUN apt-get update && apt-get install -y gettext-base nginx python3
 
-COPY kitchen/bin/waiter-init /usr/bin/waiter-init
-COPY kitchen/bin/kitchen /opt/kitchen/kitchen
+COPY waiter-init/bin/waiter-init /usr/bin/waiter-init
 COPY nginx/bin/run-nginx-server.sh /opt/nginx/bin/run-nginx-server.sh
 COPY nginx/data/nginx-template.conf /opt/nginx/data/nginx-template.conf
+COPY kitchen/bin/kitchen /opt/kitchen/kitchen
```
shamsimam: Nitpick: why was this moved lower down than in the previous version?

containers/test-apps/waiter-init/bin/waiter-mesos-init
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
shamsimam: Can we delete this file?

token-syncer/bin/ci/run-integration-tests.sh
shamsimam: I'm surprised that it worked, the `test-ping-tokens` should use it.

DaoWen: I checked the logs and the tests were skipped. Mystery solved.

containers/test-apps/Dockerfile
DaoWen: I think this is an artifact of me splitting this off from my WAITER_SANDBOX PR. I'll sort these lexicographically instead.

containers/test-apps/waiter-init/bin/waiter-mesos-init
DaoWen: Yes, this was supposed to be added in the WAITER_SANDBOX PR. Deleted.

PR #708, opened 2019-04-24
avoids clearing TE headers while sending requests to the backend

shamsimam:

- avoids clearing TE headers while sending requests to the backend

The te header is not removed as it is needed by grpc to detect incompatible proxies: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md


sradack:
TE should only be removed for HTTP/2.  See discussion https://github.com/envoyproxy/envoy/issues/5541

shamsimam:
> TE should only be removed for HTTP/2. 

@sradack done, ready for review.

PR #710, opened 2019-04-24
adds match->cluster to further limit the values produced by a regex-cluster-calculator

shamsimam:

- adds match->cluster to further limit the values produced by a regex-cluster-calculator

We would like to restrict the cluster names produced by the regex-based calculator.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -843,20 +843,13 @@
 
  :token-config {
                 ;; Define the calculator that compute the root assigned to a token when they are created.
-                :cluster-calculator {;; :kind :configured create a calculator that always assigns root as :cluster-config :name
+                :cluster-calculator {;; :kind :configured create a calculator that looks up the cluster name using a configured map.
+                                     ;; When not matching entry is found in the map, the cluster is defined as :cluster-config :name
                                      :kind :configured
-                                     :configured {:factory-fn 'waiter.token/new-configured-cluster-calculator}
-
-                                     ;; :kind :regex create a calculator that uses a regex and a map to determine the cluster name
-                                     ;; :kind :regex
-                                     :regex {:factory-fn 'waiter.token/new-regex-cluster-calculator
-                                             ;; A regex to determine the root from the incoming request host header.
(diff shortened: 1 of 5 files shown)


waiter/config-full.edn
```diff
@@ -847,13 +847,16 @@
                                      :kind :configured
                                      :configured {:factory-fn 'waiter.token/new-configured-cluster-calculator}
 
-                                     ;; :kind :configured create a calculator that always assigns root as :cluster-config :name
+                                     ;; :kind :regex create a calculator that uses a regex and a map to determine the cluster name
```
dposada:
```suggestion
                                     ;; :kind :regex creates a calculator that uses a regex and a map to determine the cluster name
```

waiter/config-full.edn
```diff
@@ -847,13 +847,16 @@
...
                                      :regex {:factory-fn 'waiter.token/new-regex-cluster-calculator
                                              ;; A regex to determine the root from the incoming request host header.
-                                             ;; The first group from the match is used to determine the root.
+                                             ;; The first group from the match is used to determine the root using match->cluster.
                                              ;; If no match is found, it defaults to the :cluster-config :name.
-                                             :root-regex #config/regex "^waiter-([a-zA-Z][0-9a-zA-Z\\-]*)"}}
+                                             :host-regex #config/regex "^waiter-([a-zA-Z][0-9a-zA-Z\\-]*)"
+                                             ;; The entry against the first group of the regex match is used to determine the cluster
+                                             :match->cluster {"alias" "name"
+                                                              "name" "name"}}}
```
dposada: How about if we avoid the regex and just have the mapping use the host portion of the URL? For example:

`waiter-foo.example.com -> foo`
`waiter-bar.example.com -> bar`
`waiter-baz.example.com -> baz`

waiter/config-full.edn
shamsimam: Deleted.

waiter/config-full.edn
shamsimam: Done.

shamsimam:
@dposada ready for another round

PR #712, opened 2019-04-25
updates jet version for improved trailers support

shamsimam:

- updates jet version for [improved trailers support](https://github.com/twosigma/jet/pull/20)

We want to avoid sending empty trailer Header frames for HTTP/2 requests when we know that there are no trailers.


shamsimam:
@sradack this is good to merge.

PR #718, opened 2019-04-29
Decouple k8s namespace from run-as-user

DaoWen:

Decouple k8s objects' `namespace` from the Waiter `run-as-user` option.

We would like to make the `namespace` independently configurable from the `run-as-user` (e.g., via a new `run-in-namespace` Waiter service parameter). Separating the `namespace` logic from the `run-as-user` logic is the first step for supporting independent configuration. This also seems like an overall architectural improvement, since the concerns of `run-as-user` and k8s `namespace` are semantically distinct.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -155,9 +155,9 @@
       :pod-name pod-name
       :restart-number restart-number}))
 
-(defn- log-dir-path [namespace restart-count]
+(defn- log-dir-path [user restart-count]
   "Build log directory path string for a containter run."
-  (str "/home/" namespace "/r" restart-count))
+  (str "/home/" user "/r" restart-count))
 
 (defn- killed-by-k8s?
   "Determine whether a pod was killed (restarted) by its corresponding Kubernetes liveness checks."
@@ -185,7 +185,7 @@
                                                 :flags failure-flags
                                                 :healthy? false
                                                 :id newest-failure-id
-                                                :log-directory (log-dir-path (:k8s/namespace live-instance)
+                                                :log-directory (log-dir-path (:k8s/user live-instance)
                                                                              (dec restart-count))
                                                 :started-at newest-failure-start-time)
                                         ;; To match the behavior of the marathon scheduler,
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -280,9 +283,11 @@
       (log/error "request to K8s API server failed: " url options body response)
       (ss/throw+ response))))
 
-(defn- service-description->namespace
-  [{:strs [run-as-user]}]
-  run-as-user)
+(defn- get-service-user
```
shamsimam: Please add a docstring to explain what this is meant to return.

shamsimam: Maybe rename to `retrieve-run-as-user`?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -727,15 +729,20 @@
        :apiVersion replicaset-api-version
        :metadata {:annotations {:waiter/service-id service-id}
                   :labels {:app k8s-name
-                           :waiter-cluster cluster-name}
-                  :name k8s-name}
+                           :waiter-cluster cluster-name
+                           :waiter/user run-as-user}
+                  :name k8s-name
+                  ;; TODO - set namespace via run-in-namespace service parameter
```
shamsimam:
```suggestion
                  ;; TODO - set namespace via namespace service parameter
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -727,15 +729,20 @@
...
+                  :namespace run-as-user}
        :spec {:replicas min-instances
               :selector {:matchLabels {:app k8s-name
-                                       :waiter-cluster cluster-name}}
+                                       :waiter-cluster cluster-name
+                                       :waiter/user run-as-user}}
```
shamsimam: ANy reason we chose `waiter-cluster` over `waiter/cluster` earlier? Does that same restriction still apply for `waiter/user`?

DaoWen: Yes, I noticed that too. We should switch to `waiter/cluster`, but I was planning to open a separate PR for that change.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Added

DaoWen: Renamed

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Renamed

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Created issue #721 for this.

PR #729, opened 2019-05-02
avoids 500 error when invalid owner name is provided

shamsimam:

- avoids 500 error when invalid owner name is provided

An invalid owner in the query string should not result in a server error, instead should return an empty result.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -201,9 +201,9 @@
     "List all tokens for a given user."
     [kv-store owner]
     (let [owner->owner-key (kv/fetch kv-store token-owners-key)]
-      (if-let [owner-key (owner->owner-key owner)]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/token.clj
```diff
@@ -201,9 +201,9 @@
     "List all tokens for a given user."
     [kv-store owner]
     (let [owner->owner-key (kv/fetch kv-store token-owners-key)]
-      (if-let [owner-key (owner->owner-key owner)]
+      (if-let [owner-key (get owner->owner-key owner)]
```
dposada: Why was this change needed?

shamsimam: Because `owner->owner-key` is `nil` in some cases (e.g. waiter just started up with an no tokens)

PR #735, opened 2019-05-07
Disable h2c tests

DaoWen:

Disable two integration tests:
- `test-internal-protocol`
- `test-trailers-support-h2c-proto-sediment`

Waiting for a new haproxy version before re-enabling these tests.


DaoWen:
@shamsimam - We'll need to remember to turn these back on again later.

PR #736, opened 2019-05-07
Add JSON spec to create Kitchen service via waiter cli

DaoWen:

- Add JSON spec to create Kitchen service via waiter cli

This makes it easier to test Waiter in a local dev environment by creating the Kitchen test service.


DaoWen:
@dposada - Updated the README.

PR #723, opened 2019-04-29
X-Waiter-Namespace

DaoWen:
## Changes proposed in this PR

- Add `X-Waiter-Namespace` option
- Allow setting a default namespace in config.edn
- Use `run-as-user` as namespace if neither `namespace` nor `default-namespace` is provided.

## Why are we making these changes?

Allow users the flexibility to instruct Waiter to k8s objects in either their own namespace, or a default Waiter-managed namespace. This might be helpful if, e.g., users lack permissions to inspect objects in the default namespace via `kubectl`.

PR #745, opened 2019-05-10
Allow k8s namespace with run-as-requestor

DaoWen:

Run run-as-requestor apps in the user's namespace when `x-waiter-namespace=*`.

The strict requirement that the namespace matches the run-as-user field (or is omitted) made it impossible to create run-as-requestor apps in the requestor's namespace.

See the original patch in #723.


waiter/docs/parameters.md
```diff
@@ -42,7 +42,7 @@ Additional (optional) parameters that can be set:
...
 |`X-Waiter-Metric-Group`|"other"|[A-Za-z0-9-_]+|Metric groups allow services to be grouped together for the purpose of metric collection.|If you have `:statsd` enabled in your [config file](../config-full.edn), then specify a metric group. If you have multiple services, choose the same metric group if you'd like combined metrics, otherwise choose independent metric groups.|
 |`X-Waiter-Min-Instances`|1|1 or 2|The minimum number of instances that Waiter should keep running at all times.|Go with the default.|
-|`X-Waiter-Namespace`|empty|user-name|The namespace in which to create back-end scheduler objects. Must match the run-as-user when specified.|Leave this field blank unless you need direct access to resources Waiter creates on your behalf.|
+|`X-Waiter-Namespace`|empty|user-name|The namespace in which to create back-end scheduler objects. Must match the run-as-user when specified, or set to `*`.|Leave this field blank unless you need direct access to resources Waiter creates on your behalf.|
```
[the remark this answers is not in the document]
DaoWen: Reworded.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -194,9 +192,13 @@
       (let [current-user (retrieve-username)
             default-namespace (-> waiter-url default-scheduler-settings :replicaset-spec-builder :default-namespace)]
         (testing "Default namespace"
-          (check-pod-namespace waiter-url nil default-namespace))
+          (check-pod-namespace waiter-url {} default-namespace))
         (testing "Custom namespace"
-          (check-pod-namespace waiter-url current-user current-user))
+          (check-pod-namespace waiter-url {:x-waiter-namespace current-user} current-user))
+        (testing "Run-as-requestor namespace"
+          (check-pod-namespace waiter-url {:x-waiter-namespace "*" :x-waiter-run-as-user "*"} current-user))
```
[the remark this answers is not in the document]
DaoWen: > x-waiter-namespace is not specified and x-waiter-run-as-user: *

That's already covered by the default-namespace case in the unit tests.

Added integration and unit tests for the other cases.

DaoWen:
@shamsimam - All green!

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -194,12 +192,20 @@
...
+          (check-pod-namespace waiter-url {:x-waiter-namespace current-user} current-user))
+        (testing "Run-as-requestor and namespace *"
+          (check-pod-namespace waiter-url {:x-waiter-namespace "*" :x-waiter-run-as-user "*"} current-user))
+        (testing "Invalid namespace for implicit current user"
           (is (thrown? Exception #"Service namespace must either be omitted or match the run-as-user"
-                       (check-pod-namespace waiter-url "not-current-user" current-user))))))))
+                       (check-pod-namespace waiter-url {:x-waiter-namespace "not-current-user"} current-user))))
+        (testing "Invalid namespace for run-as-requestor"
+          (is (thrown? Exception #"Cannot use run-as-requestor with a specific namespace"
+                       (check-pod-namespace waiter-url {:x-waiter-namespace "not-current-user" :x-waiter-run-as-user "*"} current-user))))
```
shamsimam: Missing the case where the namespace is the same as the current running user.

waiter/src/waiter/service_description.clj
```diff
@@ -891,10 +891,22 @@
...
                                                        ;; * run-as-user is the same as a missing run-as-user
-                                                       (= "*" (get service-description-from-headers-and-token-sources "run-as-user"))
-                                                       (dissoc service-description-from-headers-and-token-sources "run-as-user"))
+                                                       (= "*" raw-run-as-user)
+                                                       (dissoc "run-as-user")
+                                                       ;; * namespace means match the current user (for use with run-as-requestor)
+                                                       (= "*" raw-namespace)
+                                                       (assoc "namespace" username))
+          sanitized-run-as-user (get sanitized-service-description-from-sources "run-as-user")
+          _ (when (and (= "*" raw-namespace) (some? sanitized-run-as-user))
```
shamsimam: Nitpick: `some?` is not necessary.

waiter/src/waiter/service_description.clj
```diff
@@ -891,10 +891,22 @@
...
+              (throw (ex-info "Cannot use namespace * with specific run-as-user"
+                              {:namespace raw-namespace :run-as-user raw-run-as-user :status 400})))
+          _ (when (and (nil? sanitized-run-as-user) (some? raw-namespace) (not= "*" raw-namespace))
+              (throw (ex-info "Cannot use run-as-requestor with a specific namespace"
```
shamsimam:
```suggestion
              (throw (ex-info "Cannot use run-as-requester with a specific namespace"
```

waiter/src/waiter/service_description.clj
```diff
@@ -891,10 +891,22 @@
...
+                                                       (= "*" raw-run-as-user)
+                                                       (dissoc "run-as-user")
+                                                       ;; * namespace means match the current user (for use with run-as-requestor)
+                                                       (= "*" raw-namespace)
+                                                       (assoc "namespace" username))
+          sanitized-run-as-user (get sanitized-service-description-from-sources "run-as-user")
+          _ (when (and (= "*" raw-namespace) (some? sanitized-run-as-user))
+              (throw (ex-info "Cannot use namespace * with specific run-as-user"
+                              {:namespace raw-namespace :run-as-user raw-run-as-user :status 400})))
+          _ (when (and (nil? sanitized-run-as-user) (some? raw-namespace) (not= "*" raw-namespace))
```
shamsimam: Nitpick: `some?` is not necessary.

waiter/test/waiter/service_description_test.clj
```diff
@@ -2080,6 +2080,27 @@
...
+                                    "namespace" "some-user")
+                             validation-settings))))
+
+    (testing "validate-service-description-namespace-invalid"
+      (is (thrown? Exception #"Cannot use namespace \* with specific run-as-user"
+                   (validate builder
+                             (assoc basic-service-description
+                                    "namespace" "*"
+                                    "run-as-user" "some-user")
+                             validation-settings))))
```
shamsimam: Can you verify we have unit tests for and if now, can we add the test for:
- run-as-user:* and namespace:*
- run-as-user:* and namespace:foo
- run-as-user:foo and namespace:foo
- run-as-user:bar and namespace:foo

waiter/src/waiter/service_description.clj
shamsimam: For an on-the-fly request, run-as-user will be missing, but it should be allowed to specify the namespace:requester. This scenario should not flag an error. I think we should be more explicit on our run-as-user checks with '*'

shamsimam:
I think the validation should have the following logic:

Assume requesting user is `foo` for the following table:

| Run-As-User | Namespace | Validation |
| --- | --- | --- |
| Missing | Missing | OK |
| Missing | * | OK |
| Missing | `foo` | OK |
| Missing | `bar` | FAIL |
| `foo` | Missing | OK |
| `foo` | * | OK |
| `foo` | `foo` | OK |
| `foo` | `bar` | FAIL |
| * | Missing | OK |
| * | * | OK |
| * | `foo` | FAIL |
| * | `bar` | FAIL |

waiter/src/waiter/service_description.clj
DaoWen: Removed.

waiter/src/waiter/service_description.clj
DaoWen: Renamed

waiter/src/waiter/service_description.clj
DaoWen: Removed

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -189,17 +187,46 @@
         (is (= expected-namespace pod-namespace))))))
 
 (deftest ^:parallel ^:integration-fast test-pod-namespace
+  "Expected behavior for services with namespaces:
+   Run-As-User    Namespace   Validation
+   Missing        Missing     OK
```
DaoWen: Added the matrix from your comment into the docstring here. All 12 cases are explicitly tested below.

waiter/test/waiter/service_description_test.clj
DaoWen: I'm now testing 12 cases from your table in the integration tests. Only 9 of the 12 cases are tested in this unit test because it's only testing the cases covered by the `validate` method.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
DaoWen: All 12 cases are covered now.

PR #748, opened 2019-05-10
Update copyright to include year

twosigmajab:
Testing gitpod!


shamsimam:
Do we want this bot to catch all such instances? If so, the clojure files have similar year-less copyrights.

shamsimam:
> Do we want this bot to catch all such instances? If so, the clojure files have similar year-less copyrights.

@twosigmajab ⬆️

PR #751, opened 2019-05-16
supports reading the input stream in custom thread pool

shamsimam:
This PR needs an associated jet change for [improved AsyncContentProvider support](https://github.com/twosigma/jet/pull/22).

- supports reading the input stream in a custom thread pool
- report request body length for all requests in the metrics (codahale and statsd) and logs

InputStreamContentProvider, used for InputStreams by jet, uses an iterator that blocks while determining if more content is available. This blocking is problematic for interactive clients who send some data and then expect a response from the server before sending the next chunk of data. In particular, in such scenarios, this leads to a deadlock in org.eclipse.jetty.client.HttpContent.advance(java.util.Iterator<java.nio.ByteBuffer>) which cannot return when it makes the second call to iterator.hasNext().

Using a DeferredContentProvider with asynchronous population of chunks avoids the aforementioned deadlock. In addition, we use a custom thread pool to avoid exhausting threads from Jetty's or core.async's thread pools while reading from the input stream.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -830,11 +830,13 @@
             :not-found "Not found"
             :prestashed-tickets-not-available "Prestashed jobsystem tickets not available"}
 
- ;; Waiter launches a separate thread pool xecutor to perform blocking reads on incoming request bodies
+ ;; Waiter launches a separate thread pool executor to perform blocking reads on incoming request bodies
  :stream-reader {;; the maximum number of concurrent input stream reads allowed in the executor (default 150)
                  :concurrency-level 150
                  ;; the idle time before cached threads from the executor are terminated (default 5)
-                 :keep-alive-mins 5}
+                 :keep-alive-mins 5
+                 ;; the number of concurrent requests that are supported by the executor (default 50000)
+                 :queue-limit 50000}
 
  ;; Support information used to provide user guidance in error messages
  :support-info [{:label "Email Support Team" :link {:type :email :value "support@example.com"}}
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -136,8 +136,8 @@
                   ;; When SSL is enabled, the password to the keystore
                   :key-password "keystore-password"
 
-                  ;; The maximum number of threads to use (default 250)
-                  :max-threads 250
+                  ;; The maximum number of threads to use (default 200)
+                  :max-threads 200
 
                   ;; The maximum size in bytes of the request header.
                   ;; Larger headers will allow for more and/or larger cookies plus larger form content encoded in a URL.
@@ -829,6 +829,12 @@
             :not-enough-memory "Not enough memory allocated"
             :not-found "Not found"
             :prestashed-tickets-not-available "Prestashed jobsystem tickets not available"}
+
+ ;; Waiter launches a separate thread pool xecutor to perform blocking reads on incoming request bodies
+ :stream-reader {;; the maximum number of concurrent input stream reads allowed in the executor (default 150)
+                 :concurrency-level 150
+                 ;; the idle time before cached threads from the executor are terminated (default 5)
+                 :keep-alive-mins 5}
 
  ;; Support information used to provide user guidance in error messages
  :support-info [{:label "Email Support Team" :link {:type :email :value "support@example.com"}}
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -180,17 +183,121 @@
     (deliver reservation-status-promise promise-value)
     (utils/exception->response (ex-info message (assoc metrics-map :status status) error) request)))
 
+(defn make-stream-reader-executor
+  "Creates and returns the ThreadPoolExecutor used for reading the input stream."
+  [concurrency-level keep-alive-mins]
+  (let [queue (LinkedBlockingQueue.)
```
pschorf: Should we fix the size of the `LinkedBlockingQueue`? Unbounded could be a problem :)

waiter/config-full.edn
```diff
@@ -136,8 +136,8 @@
                   ;; When SSL is enabled, the password to the keystore
                   :key-password "keystore-password"
 
-                  ;; The maximum number of threads to use (default 250)
-                  :max-threads 250
+                  ;; The maximum number of threads to use (default 200)
+                  :max-threads 200
```
pschorf: Why did this change?

waiter/src/waiter/process_request.clj
shamsimam: We will have at most one entry in the queue per queued/in-flight request. I feel like we do not need a bound here, the request queue bounds can handle the limiting.

waiter/config-full.edn
shamsimam: I reduced this since we no longer perform the InputStream reads on the Jetty threads. Having lower number of threads (though net we are ending up increasing it by a 100 due to the input stream reader thread pool) reduces the pressure on the JVM and the heap.

waiter/src/waiter/process_request.clj
shamsimam: Actually, let me add a reasonably high bound for safety.

waiter/config-full.edn
```diff
@@ -830,6 +830,14 @@
...
             :prestashed-tickets-not-available "Prestashed jobsystem tickets not available"}
 
+ ;; Waiter launches a separate thread pool executor to perform blocking reads on incoming request bodies
+ :stream-reader {;; the maximum number of concurrent input stream reads allowed in the executor (default 150)
+                 :concurrency-level 150
+                 ;; the idle time before cached threads from the executor are terminated (default 5)
+                 :keep-alive-mins 5
+                 ;; the number of concurrent requests that are supported by the executor (default 50000)
+                 :queue-limit 50000}
+
```
dposada: How did you come up with these numbers?

dposada: Should `:stream-reader` be a sub-map under `:server-options`?

waiter/src/waiter/process_request.clj
```diff
@@ -180,17 +183,140 @@
...
+                          "core" "stream-reader" "scheduled-task-count")
+    executor))
+
+(defn- submit-request-streaming-task
+  "Submits a task to the executor and invokes the error handler with custom error message if the submission fails."
+  [^ThreadPoolExecutor executor runnable-task]
+  (try
+    (.execute executor runnable-task)
+    (catch Throwable throwable
+      (log/error "unable to submit task to queue")
```
dposada: Is there any other info about the task that we can log here?

waiter/src/waiter/process_request.clj
```diff
@@ -180,17 +183,140 @@
...
+      (throw (ex-info "Too many concurrent requests on router"
```
dposada: Is it possible to write an integration test for this?

waiter/config-full.edn
shamsimam: The `server-options` are used to initialize the Waiter jetty server. The `stream-reader` is more the client part that is streaming data to the backend, than the server part which is receiving requests.

waiter/src/waiter/process_request.clj
shamsimam: It will be hard to simulate such large numbers of concurrent requests.

Regarding integration tests as a whole, this is mainly a refactoring task and existing integration tests, including streaming tests, all pass without modification.

waiter/config-full.edn
shamsimam: These are configurable and we can tweak them during launch if required. They have been set high enough to avoid issues under load we introduce in testing.

waiter/src/waiter/process_request.clj
shamsimam: I'll add log in the error handler about the number of bytes streamed. Any other info you are looking for? The CID should capture other info about the request.

PR #756, opened 2019-05-20
minimizes redundant puts into the body channel

shamsimam:

- minimizes redundant puts into the body channel
- increases the buffer size for data read in each iteration

We would like to minimize the number of pending puts on a body channel by not offering empty data. 
This prevents us from overwhelming the body channel and failing requests when `puts!` fail due to a backlog full of empty data buffers.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -212,56 +212,68 @@
                                  :queue-size (.size (.getQueue executor))}
                        :status 503}
                       throwable)))))
+(let [initial-buffer-size 1024
+      max-buffer-size 32768
+      max-buffer-increment-size 4096]
+  (defn stream-http-request
+    "Reads data from the input stream and queues it into the provided body channel as a ByteBuffer.
+     When no more data is available on the input stream, an asynchronous task is scheduled
+     on the provided executor to read data from the input stream later.
+     Reports an error to the error handler whenever:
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -212,56 +212,68 @@
...
+          stream-http-request-task (fn invoke-stream-http-request []
+                                     (cid/with-correlation-id
+                                       correlation-id
+                                       (stream-http-request executor service-id metric-group error-handler-fn
+                                                            input-stream body-ch @bytes-streamed-atom)))]
+      (try
+        (loop [unreported-bytes-to-statsd 0]
+          ;; TODO potential optimization to size the buffer based on available bytes
+          (let [available-bytes (.available input-stream)
+                buffer-size (loop [iter-size initial-buffer-size
```
pschorf: Why not do something like `(min max-buffer-size (* 1024 (/ available 1024)))`?

shamsimam: The intent is to have buffer size larger than the available bytes, else we'll have to send twice.
`(* 1024 (int (/ 5000 1024))) == 4096`

The other reason for this loop, is a micro-optimization to avoid division and multiplication in favor of additional and min/max comparisons.

shamsimam: Changed :)

PR #758, opened 2019-05-20
Includes request length in chunked encoding tests

shamsimam:

- includes request length in chunked encoding tests
- adds support for request body length in kitchen

We would like to verify that the correct number of bytes are being transferred to the backend regardless of the transfer encoding being used.

### DIFF
--- a/containers/test-apps/kitchen/bin/kitchen
+++ b/containers/test-apps/kitchen/bin/kitchen
@@ -406,7 +406,7 @@ class Kitchen(HTTPWebSocketsHandler):
         else:
             self.logger().debug('Got data on websocket: {}'.format(in_data))
 
-        self.__request_buffer = in_data
+        self.__request_body_length = len(in_data)
 
         if opcode == self._opcode_binary:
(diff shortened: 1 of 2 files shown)


containers/test-apps/kitchen/bin/kitchen
```diff
@@ -662,34 +690,9 @@ class Kitchen(HTTPWebSocketsHandler):
 
         # Check if client requested echo server mode
         echo_server = self.headers.get('x-kitchen-echo') is not None
-        echo_buffer = bytearray() if echo_server else None
```
DaoWen: Replacing this line with `self.__request_buffer = bytearray()` changes the behavior of the program. The `__slurp_bytes` helper could previously process a request with size _O(n)_ using _O(1)_ memory, but now it needs _O(n)_ memory because you always initialize this to a `bytearray`.

shamsimam: Thanks. Changed the code to avoid storing the entire request body in memory.

PR #765, opened 2019-05-23
adds support for filtering metrics by name

shamsimam:

- adds support for filtering metrics by name

We would like the /metrics endpoint less heavy to load by being more specific about the metrics we are interested in.


shamsimam:
can we merge this?

PR #767, opened 2019-05-28
uses bitwise operators while computing buffer size

shamsimam:

- uses bitwise operators while computing buffer size

Introduces a micro-optimization by avoiding integer division.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -218,7 +218,8 @@
       max-back-pressure-delay-ms 4000
       min-buffer-size 1024
       max-buffer-size 32768
-      buffer-increment-power-of-2 (int (/ (Math/log min-buffer-size) (Math/log 2)))]
+      min-buffer-increment-size 1024
+      buffer-increment-mask (bit-not (dec min-buffer-increment-size))]
 
   (defn stream-http-request
     "Reads data from the input stream and queues it into the provided body channel as a ByteBuffer.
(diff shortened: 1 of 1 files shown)


DaoWen:
> Introduces a micro-optimization.

What's the specific goal? Avoiding integer division? Creating allocator-friendly power-of-2 buffers?

waiter/src/waiter/process_request.clj
```diff
@@ -242,14 +242,11 @@
...
                 buffer-size (-> available-bytes
-                              (max 0)
-                              double
-                              (/ buffer-increment-size)
-                              int
-                              inc
-                              (* buffer-increment-size)
+                              (max min-buffer-size)
+                              (bit-shift-right buffer-increment-power-of-2)
+                              (bit-shift-left buffer-increment-power-of-2)
```
DaoWen: I think you could use [Integer/highestOneBit](https://docs.oracle.com/javase/10/docs/api/java/lang/Integer.html#highestOneBit(int)) here, replacing these bit shifts as well as the `Math/log` calls above.

shamsimam:
> > Introduces a micro-optimization.
> 
> What's the specific goal? Avoiding integer division? Creating allocator-friendly power-of-2 buffers?

Avoiding integer division.

waiter/src/waiter/process_request.clj
shamsimam: The goal isn't a power of 2.

`(Integer/highestOneBit 7000) => 4096` vs `(-> 7000 (bit-shift-right 10) (bit-shift-left 10)) => 6144`

DaoWen: Ah, I misread and thought the Math/log was computed on available-bytes. Masking off the last 9 bits would be easier to understand (and faster) than shift-right followed by shift-left.

waiter/src/waiter/process_request.clj
```diff
@@ -242,14 +243,10 @@
...
-                ;; get a buffer size between buffer-increment-size and max-buffer-size
+                ;; get a buffer size between min-buffer-size and max-buffer-size
                 buffer-size (-> available-bytes
-                              (max 0)
-                              double
-                              (/ buffer-increment-size)
-                              int
-                              inc
-                              (* buffer-increment-size)
+                              (bit-and buffer-increment-mask)
```
DaoWen: Nitpick: The indentation is off here, but it looks like it may have been off before your patch as well. The `(bit-and` line should be aligned with `available-bytes` above, rather than being indented 2 spaces like an `(if` block.

PR #771, opened 2019-05-28
Fix deep-merge-settings

nsinkov:

when merging settings maps we should not have special treatment for the value of :kind

If we are specifying defaults for components, we should set the defaults even if the current :kind is a different component. The defaults are still valid


shamsimam:
Please address the question posed above or close the PR.

nsinkov:
it looks good in production

PR #775, opened 2019-05-30
delete run-integration-tests-shell-scheduler

nsinkov:

- delete run-integration-tests-shell-scheduler.sh 

It's not used and it's confusing when making changes. We can re-create if we need it.


DaoWen:
I think we can delete this, but we should probably check with @shamsimam and @dposada to make sure they don't use this script locally.

shamsimam:
I do not use this script locally.

dposada:
I don't use it either.

Issue #721, opened 2019-04-29
Rename waiter-cluster K8s label to waiter/cluster for consistency


DaoWen:
For Kubernetes objects created by Waiter, we should be using the `namespace/property` naming convention for all of our Waiter-specific annotations and labels.

This change would be two stages:

1. Add the `waiter/cluster` label redundantly.
2. After some reasonable amount of time (i.e., when we're confident that all of the objects without `waiter/cluster` have been deleted or replaced), remove the `waiter-cluster` label and switch our queries (e.g., watches) to use the new `waiter/cluster` label.

PR #779, opened 2019-06-05
Update Kubernetes object labels

DaoWen:

- Move `service-id->k8s-app-name` docstring
- Add `waiter/cluster` Kubernetes label
- Add `waiter/service-hash` Kubernetes label

- Moving the docstring is just refactoring for correct placement before the argument list.
- Adding the `waiter/cluster` label helps address #721.
- Adding the `waiter/service-hash` label will make it easier to search for Kubernetes objects corresponding to a specific Waiter service. Since the existing `waiter/service-id` is an annotation (due to label length limitations), it can't be used in Kubernetes selectors when searching.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -67,7 +67,7 @@
   ;; If we have fewer than 48 characters, then we'll probably want to shorten the hash.
   (< k8s-max-name-length 48))
 
-(defn service-id->k8s-app-name [{:keys [max-name-length pod-suffix-length] :as scheduler} service-id]
+(defn service-id->k8s-app-name
   "Shorten a full Waiter service-id to a Kubernetes-compatible application name.
    May return the service-id unmodified if it doesn't violate the
    configured name-length restrictions for this Kubernetes cluster.
@@ -78,6 +78,7 @@
      {:max-name-length 32}
      \"waiter-myapp-e8b625cc83c411e8974c38d5474b213d\")
    ==> \"myapp-e8b625cc474b213d\""
+  [{:keys [max-name-length pod-suffix-length] :as scheduler} service-id]
   (let [[_ app-prefix x y z] (re-find #"([^-]+)-(\w{8})(\w+)(\w{8})$" service-id)
         k8s-max-name-length (- max-name-length pod-suffix-length 1)
         suffix (if (use-short-service-hash? k8s-max-name-length)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -728,13 +738,24 @@
...
+                  ;; https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+                  :annotations {:waiter/service-id service-id}
                   :labels {:app k8s-name
+                           ;; TODO - remove waiter-cluster
+                           ;; after waiter/cluster is exclusively in use
+                           ;; (see GitHub issue #721)
                            :waiter-cluster cluster-name
+                           :waiter/cluster cluster-name
+                           :waiter/service-hash service-hash
                            :waiter/user run-as-user}
```
shamsimam: In a separate PR, we should perform a length check on the `run-as-user` label and trim accordingly.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -728,13 +738,24 @@
...
+                  ;; we store just the 32-char hash portion of the service-id as a searchable label,
+                  ;; but store the full service-id as an annotation.
+                  ;; https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+                  ;; https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+                  :annotations {:waiter/service-id service-id}
                   :labels {:app k8s-name
+                           ;; TODO - remove waiter-cluster
+                           ;; after waiter/cluster is exclusively in use
+                           ;; (see GitHub issue #721)
                            :waiter-cluster cluster-name
```
shamsimam: Are we preserving the `waiter-cluster` label for backwards compatibility? Or is there a plan to remove it in a subsequent PR?

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: The OS ensures that isn't a problem: user names already have a length limit of 32 chars on posix systems.

https://serverfault.com/questions/294121/what-is-the-maximum-username-length-on-current-gnu-linux-systems

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: We're going to remove it in the near future, but we need to allow time for all the current services to be restarted. We probably won't remove it until after the ReplicaSet → Pod refactoring.

DaoWen:
@shamsimam - RTM

PR #795, opened 2019-06-12
Use lower-case headers consistently

shamsimam:

- Use lower-case headers consistently

Consistently use lower-case headers in waiter codebase to avoid confusion in performing gets from headers.


waiter/docs/how-to-waiter.md
```diff
@@ -4,7 +4,7 @@ Check out [Waiter Checklist](service-checklist.md) before you try running your s
 
 # Overview
 
-There are two ways to launch services on Waiter: the On-the-Fly API and the Token API.  The On-the-Fly API automatically starts HTTP services using custom HTTP headers coming from the client.  Alternatively, the Token API allows service maintainers to predefine a service.  Token-based services are reachable by HTTP clients using a DNS name or using a custom HTTP header, `X-Waiter-Token`.
```
nsinkov: what about editing waiter/docs/parameters.md and waiter/docs/rest-api.md

waiter/integration/waiter/token_request_test.clj
```diff
@@ -917,27 +917,27 @@
         (assert-response-status response 400)))
 
     (testing "can't use invalid token that's valid for zookeeper"
-      (let [response (make-request waiter-url "/pathabc" :headers {"X-Waiter-Token" "bad#token"})]
```
nsinkov: also edit waiter/integration/waiter/busy_instance_test.clj and waiter/integration/waiter/streaming_test.clj?

just run `waiter$ grep -r X-Waiter .` ?

nsinkov:
maybe for every change, do a top-level `grep -r <string> .` to see that another upper-case instance still exists

waiter/docs/how-to-waiter.md
shamsimam: Too many to edit, so left them.

waiter/integration/waiter/token_request_test.clj

waiter/docs/how-to-waiter.md
shamsimam: Done now.

shamsimam:
@nsinkov ready for another round of review.

PR #801, opened 2019-06-14
use the session-not-on-or-after field for auth expiry time

nsinkov:

- change to use the correct field for SAML auth expiry

Currently using the assertion's not-on-or-after, but that's only 5 minutes and intended to allow for the original redirects between the IDP and waiter. If there is a limit on the auth, then it's the optional session-not-on-or-after

### DIFF
--- a/waiter/integration/waiter/authentication_test.clj
+++ b/waiter/integration/waiter/authentication_test.clj
@@ -1,5 +1,6 @@
 (ns waiter.authentication-test
-  (:require [clojure.data.json :as json]
+  (:require [clj-time.core :as t]
+            [clojure.data.json :as json]
             [clojure.string :as string]
             [clojure.test :refer :all]
             [reaver :as reaver]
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/auth/saml.clj
```diff
@@ -130,17 +129,16 @@
...
                           attributes))
         conditions (.getConditions assertion)
-        audiences (mapcat #(let [audiences (.getAudiences %)]
-                             (map (fn [a] (.getAudienceURI a)) audiences))
-                          (.getAudienceRestrictions conditions))]
+        authn-statements (.getAuthnStatements assertion)
+        min-session-not-on-or-after (when (not-empty authn-statements)
+                                      (apply t/min-date '(map (.getSessionNotOnOrAfter %) authn-statements)))]
     {:attrs attrs
-     :audiences audiences
```
nsinkov: dropping what I don't use already

nsinkov:
Yeah, didn't have time to do yet

waiter/integration/waiter/authentication_test.clj
```diff
@@ -121,6 +121,11 @@
                                                                  :headers {"content-type" "application/x-www-form-urlencoded"}
                                                                  :method :post)
             _ (assert-response-status response 303)
+            cookie-fn (fn [cookies name] (some #(when (= name (:name %)) %) cookies))
```
shamsimam: Nitpick: This is a useful one for `client_tools`, maybe name it `extract-cookie`

waiter/integration/waiter/authentication_test.clj
```diff
@@ -121,6 +121,11 @@
...
+            auth-cookie (cookie-fn cookies "x-waiter-auth")
+            _ (is (not (nil? auth-cookie)))
+            one-hour (* 60 60)
```
shamsimam: Nitpick: `one-hour-in-secs (-> 1 t/hour t/in-seconds)`, can even inline the variable in the assertion below

PR #808, opened 2019-06-17
makes test-kubernetes-pod-expiry less flaky

shamsimam:

- makes test-kubernetes-pod-expiry less flaky when deployment error logic is triggered

We don't like flaky tests.

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -267,7 +267,7 @@
             (make-request-with-debug-info
               {:x-waiter-distribution-scheme "simple"
                :x-waiter-name (rand-name)}
-              #(make-kitchen-request waiter-url % :method :get :path "/"))]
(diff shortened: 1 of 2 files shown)


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -267,15 +267,15 @@
             (make-request-with-debug-info
               {:x-waiter-distribution-scheme "simple"
                :x-waiter-name (rand-name)}
-              #(make-kitchen-request waiter-url % :method :get :path "/"))]
+              #(make-kitchen-request waiter-url % :method :get :path "/hello"))]
```
nsinkov:
```suggestion
              #(make-kitchen-request waiter-url % :method :get :path "/" :query-params {"test-name" "test-kubernetes-pod-expiry"}))]
```

shamsimam: Undid the change.

PR #809, opened 2019-06-18
Eagerly upload completed container logs to S3

DaoWen:

* Move S3 log backup logic into the fileserver sidecar.
* Re-enable integration test for logs backed up to S3.

Rather than waiting until the pod is being deleted before uploading,
the sidecar uploads the logs for each run of the main container as soon
as it detects that container's exit (via a sentinel file).

This also addresses a bug where log files are not uploaded to S3 if the
main app container is in a CrashBackOffLoop when the pod is deleted.

### DIFF
--- a/containers/fileserver/Dockerfile
+++ b/containers/fileserver/Dockerfile
@@ -1,6 +1,9 @@
-FROM nginx:1.15-alpine
+FROM nginx:1.16-alpine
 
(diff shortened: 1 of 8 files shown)


nsinkov:
As per discussion, would this detect the main container getting a SIGKILL for some reason and react to it?

PR #816, opened 2019-06-19
Handle missing service on k8s delete

DaoWen:

Return 404 when service is nil or incomplete.

Currently this fails when the HTTP DELETE request comes back with an error. We should fail before making the HTTP request.

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -551,6 +551,14 @@
                       (= 200)))
                 :interval 5 :timeout 60)))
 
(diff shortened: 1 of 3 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -551,6 +551,14 @@
                       (= 200)))
                 :interval 5 :timeout 60)))
 
+        (testing "delete service again (should get 404)"
```
shamsimam: Should we wait on all routers to agree that the service has been deleted?

DaoWen: Done

PR #822, opened 2019-06-25
logging and import cleanup in process_request.clj

shamsimam:

- logs backend response status and headers when debug is enabled
- cleans up unused imports
- makes logs consistently start with lower-case characters 

Additional logging helps with debugging individual requests. Code cleanup is general hygiene.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -513,7 +513,7 @@
    {:keys [uri] :as request} reason-map reservation-status-promise confirm-live-connection-with-abort
    request-state-chan {:keys [status] :as response}]
   (when (utils/request->debug-enabled? request)
-    (log/info "backend response status:" (:status response)"and headers:" (:headers response)))
+    (log/info "backend response status:" (:status response) "and headers:" (:headers response)))
   (let [{:keys [service-description service-id]} descriptor
         {:strs [backend-proto blacklist-on-503 metric-group]} service-description
         waiter-debug-enabled? (utils/request->debug-enabled? request)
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -507,6 +512,8 @@
   [post-process-async-request-response-fn _ instance-request-properties descriptor instance
    {:keys [uri] :as request} reason-map reservation-status-promise confirm-live-connection-with-abort
    request-state-chan {:keys [status] :as response}]
+  (when (utils/request->debug-enabled? request)
+    (log/info "backend response status:" (:status response)"and headers:" (:headers response)))
```
nsinkov: space missing ```response)"and headers:"```

waiter/src/waiter/process_request.clj
```diff
@@ -364,7 +369,7 @@
           (histograms/update! (metrics/service-histogram service-id "request-content-length") content-length)
           (statsd/inc! metric-group "request_content_length" content-length)))
       (catch Exception e
-        (log/error e "Unable to track content-length on request")))
```
nsinkov: Just make all these capitalizations a separate PR. Search systematically. Then it will also cover the ```(log/error "there was a generic error``` from above

waiter/src/waiter/process_request.clj

waiter/src/waiter/process_request.clj
shamsimam: We are switching to lower-case. This PR changes the instances in this file.

nsinkov: oops

nsinkov: Let's open a task to replace ```log/(error|info|warn) .* ?"[A-Z]```

shamsimam: Created: https://github.com/twosigma/waiter/issues/825

PR #826, opened 2019-06-27
leftover changes from PR 746

nsinkov:

- some small fixes 

already committed to making these changes in PR 746


waiter/integration/waiter/authentication_test.clj
```diff
@@ -27,8 +29,9 @@
 (deftest ^:parallel ^:integration-fast test-token-authentication-parameter-error
   (testing-using-waiter-url
     (when (using-composite-authenticator? waiter-url)
-      (let [authentication-providers (map name (keys (get-in (waiter-settings waiter-url)
-                                                             [:authenticator-config :composite :authentication-providers])))
+      (let [authentication-providers (map name (keys (get-in
```
shamsimam: thrading macro `->>` would make this easier to read

waiter/src/waiter/util/client_tools.clj
```diff
@@ -987,6 +987,18 @@
   [waiter-url]
   (= "composite" (get-authenticator-kind waiter-url)))
 
+(defn using-saml-authenticator?
```
shamsimam: Should we make this private so someone doesn't inadvertently use this function?

nsinkov: if someone wants to use it, they can

waiter/integration/waiter/authentication_test.clj
nsinkov: updated

waiter/src/waiter/util/client_tools.clj
nsinkov: the function does what it says

nsinkov:
@shamsimam  all green, can you merge

PR #829, opened 2019-06-28
do not return 404 for token with only user-specified metadata

shamsimam:

- do not return 404 for token with only user-specified metadata

We should render tokens that do not have service parameters instead of reporting a 404.


dposada:
Will merge when Travis is green

shamsimam:
@dposada build is green :)

PR #830, opened 2019-06-28
disables sending trailers to non-http/2 backends

shamsimam:

- disables sending trailers to non-http/2 backends

Some of our http/1 backends do not support trailers, so disable sending request trailers for all of them.


pschorf:
Waiting on green build

shamsimam:
@pschorf  Build is now green.

PR #836, opened 2019-07-02
Make python idp server

nsinkov:

- create a custom python SAML IDP server

use this for integration testing as it is more flexible


dposada:
Will merge when Travis is green

PR #838, opened 2019-07-03
Add current-for-tokens

pschorf:
- Add `current-for-tokens` to service details
- Add `run-using-marathon.sh`

Allows users to determine if a service is the current version for a token.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1388,6 +1388,41 @@
         (finally
           (delete-token-and-assert waiter-url token))))))
 
+(deftest ^:parallel ^:integration-fast test-current-for-tokens-multiple-source-tokens
+  (testing-using-waiter-url
+    (let [service-name (rand-name)
+          token-name-a (create-token-name waiter-url (str service-name "-A"))
+          token-name-b (create-token-name waiter-url (str service-name "-B"))
+          first-service-description (assoc (kitchen-request-headers :prefix "")
+                                      :name service-name)
+          combined-token-header (str token-name-a "," token-name-b)]
+      (try
+        (let [response (post-token waiter-url (assoc first-service-description :token token-name-a))]
+            (assert-response-status response 200))
+          (let [response (post-token waiter-url (assoc first-service-description :token token-name-b))]
+            (assert-response-status response 200))
+
+        (let [service-id-a (retrieve-service-id waiter-url {:x-waiter-token combined-token-header})
+              new-service-description (update first-service-description :cpus #(+ % 0.1))]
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -419,7 +430,9 @@
                      (:time service-suspended-state)
                      (assoc :service-suspended-state service-suspended-state)
                      (seq source-tokens-entries)
-                     (assoc :source-tokens source-tokens-entries))]
+                     (assoc :source-tokens source-tokens-entries)
+                     (seq current-for-tokens)
```
dposada: Don't we want to add this key even if `current-for-tokens` is empty?

waiter/src/waiter/handler.clj
```diff
@@ -352,11 +352,21 @@
      :failed-instances (get service-id->failed-instances service-id)
      :killed-instances (get service-id->killed-instances service-id)}))
 
+(defn- get-current-for-tokens
+  [source-token-entries token->token-hash]
+  (->> source-token-entries
+       (reduce (fn [acc tokens] (into acc tokens)) [])
+       (filter (fn [{:strs [token version]}]
+                 (let [current-version (token->token-hash token)]
+                   (log/info "checking token" token ". Current version" current-version "source version" version)
```
dposada: `log/debug`?

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1358,6 +1358,36 @@
...
+
+          (testing "old service is no longer current for token"
+            (let [initial-service-details' (service-settings waiter-url initial-service-id)
+                  new-service-id (retrieve-service-id waiter-url {:x-waiter-token token})
+                  new-service-details (service-settings waiter-url new-service-id)]
+              (is (nil? (:current-for-tokens initial-service-details')))
+              (is (= [token] (:current-for-tokens new-service-details))))))
+        (finally
+          (delete-token-and-assert waiter-url token))))))
+
```
dposada: There's a good test for this in the CLI tests:

https://github.com/twosigma/waiter/blob/master/cli/integration/tests/waiter/test_cli.py#L1087-L1153

Should we replicate it here?

waiter/src/waiter/handler.clj
pschorf: I was just following the convention of all the other optional stuff in the file. Even service description isn't added if it's empty or missing.

waiter/src/waiter/handler.clj
pschorf: 👍 done

waiter/integration/waiter/token_request_test.clj
pschorf: Done

PR #843, opened 2019-07-09
adds support for asynchronous reads using the ReadListener from servlet api

shamsimam:

- adds support for asynchronous reads using the ReadListener from servlet api

We would like to avoid manually managing a custom thread pool to pump the request bytes to the backend.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -207,7 +207,7 @@
                         (min max-buffer-size))
           buffer-bytes (byte-array buffer-size)
           bytes-read (.read input-stream buffer-bytes)]
-      (log/info "processed bytes from the request input stream" {:available bytes-available :read bytes-read})
+      (log/debug "processed bytes from the request input stream" {:available bytes-available :read bytes-read})
       (when (pos? bytes-read)
         (let [byte-buffer (ByteBuffer/wrap buffer-bytes 0 bytes-read)]
           (when-not (au/timed-offer!! body-ch byte-buffer streaming-timeout-ms)
@@ -263,7 +263,7 @@
             (try
               (cid/with-correlation-id
                 correlation-id
-                (log/info "reading available bytes from request input stream")
+                (log/debug "reading available bytes from request input stream")
                 (loop []
                   (let [bytes-read (read-available-bytes!! streaming-timeout-ms complete-request-streaming input-stream body-ch)]
                     (when (pos? bytes-read)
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -190,143 +190,107 @@
...
+     Returns the bytes read from the input stream."
+    [streaming-timeout-ms error-handler-fn ^ServletInputStream input-stream body-ch]
+    (let [bytes-available (.available input-stream)
+          buffer-size (-> bytes-available
+                        (bit-and buffer-increment-mask)
+                        (max min-buffer-size)
+                        (min max-buffer-size))
+          buffer-bytes (byte-array buffer-size)
+          bytes-read (.read input-stream buffer-bytes)]
+      (log/info "processed bytes from the request input stream" {:available bytes-available :read bytes-read})
```
sradack: Should we switch this to `log/debug`?

waiter/src/waiter/process_request.clj
```diff
@@ -190,143 +190,107 @@
...
+      (.setReadListener
+        input-stream
+        (reify ReadListener
+          (onAllDataRead [_]
+            (complete-request-streaming nil))
+          (onDataAvailable [_]
+            (try
+              (cid/with-correlation-id
+                correlation-id
+                (log/info "reading available bytes from request input stream")
```
sradack: Should we switch this to `log/debug`?

waiter/src/waiter/process_request.clj
```diff
@@ -190,143 +190,107 @@
...
+                (loop []
+                  (let [bytes-read (read-available-bytes!! streaming-timeout-ms complete-request-streaming input-stream body-ch)]
+                    (when (pos? bytes-read)
+                      (report-request-size-metrics bytes-read false))
+                    (when (and (utils/non-neg? bytes-read) (.isReady input-stream))
```
sradack: All the examples I've seen have `isReady()` guarding `read()` calls.  Should we check `isReady()` before calling `read()`?

shamsimam: The example at https://docs.oracle.com/javaee/7/tutorial/servlets013.htm is using a `do ... while()` loops where the call to `read()` is before the call to `isReady()`

shamsimam:
> What wasn't working before?

1. Using the ReadListener inside jet at https://github.com/twosigma/jet/blob/waiter-patch/src/clj/qbits/jet/servlet.clj#L71 was not working.

2. My previous attempts (before the thread pool pump) did not use the call to `isReady()` and were not receiving the `onDataAvailable()` callbacks.

3. I also received [`IllegalStateException: ReadListener already set` errors](https://github.com/twosigma/waiter/pull/804) before, but did not dig deeper since my `onDataAvailable()` callbacks were not working.

waiter/src/waiter/process_request.clj

waiter/src/waiter/process_request.clj

PR #847, opened 2019-07-10
Clean up redundant error messages.

scrosby:
We log both the thrown error as well as the log/error.

-


waiter/src/waiter/service.clj
```diff
@@ -165,9 +164,8 @@
          (throw (ex-info "Unable to put instance on release-chan."
                          {:instance ~instance})))
        (do
-         (log/error "Unable to find release-chan for service" service-id#)
          (throw (ex-info "Unable to find release-chan."
-                         {:instance ~instance}))))))
+                         {:instance ~instance :service service-id#}))))))
```
sradack:
```suggestion
                         {:instance ~instance :service-id service-id#}))))))
```

PR #849, opened 2019-07-10
Change log level to warn for bad service names.

scrosby:
We should probably be rejecting these service names at injection, instead
of at the zookeeper level. Oh well.

-

### DIFF
--- a/waiter/src/waiter/kv.clj
+++ b/waiter/src/waiter/kv.clj
@@ -95,9 +95,9 @@
   given the underlying ZooKeeper implementation."
   [key]
   (when (re-matches #".*/.*" key)
-    (throw (ex-info "Key may not contain '/'" {:key key})))
+    (throw (ex-info "Key may not contain '/'" {:key key :log-level :warn})))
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/kv.clj
```diff
@@ -95,9 +95,9 @@
   given the underlying ZooKeeper implementation."
   [key]
   (when (re-matches #".*/.*" key)
-    (throw (ex-info "Key may not contain '/'" {:key key})))
+    (throw (ex-info "Key may not contain '/'" {:key key :log-level :warn})))
```
sradack: Service names shouldn't reach this level.  Can we validate service names at the appropriate level?

scrosby: Yup. See my comment on the pullrequest. I was keeping this deliberately scoped narrow. We can keep it open to make that change.

sradack: The trouble is that it should be considered an error to try to write an invalid key to ZK.  Let's make the change there or close this PR out.

PR #857, opened 2019-07-12
disable flaky tests

nsinkov:

- disable flaky backend-proto tests

too many PRs blocked by these right now


shamsimam:
Can you confirm that all of these are flaking? I thought only the `https` one is flaking internally.

PR #858, opened 2019-07-12
disable flaky tests only on kubernetes

nsinkov:

-disable flaky proto backend tests only on kubernetes


shamsimam:
Awaiting green build.

PR #859, opened 2019-07-12
Add service description builder state

nsinkov:

- expose service description builder state

stateful custom service description builder needs to expose state for health monitoring


nsinkov:
@shamsimam green

PR #861, opened 2019-07-15
Adds run-description to test metrics collection

nsinkov:
Adds run-description to test metrics collection

- adding the run-description field

To allow for a description of what the run is doing/testing.


dposada:
Will merge when checks are green

nsinkov:
@dposada green now

PR #864, opened 2019-07-17
make tests compatible with composite scheduler

nsinkov:

- update functions that read scheduler settings to use the default-scheduler-settings function

code needs to still work even if a composite scheduler is used

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -134,7 +134,7 @@
 
           ;; Test that the killed instances' logs were persisted to S3.
           ;; This portion of the test logic was modified from the active-instances tests above.
-          (let [log-bucket-url (k8s-log-bucket-url waiter-url)
+          (let [log-bucket-url (-> (default-scheduler-settings waiter-url) :log-bucket-url)
                 killed-instances (get-in (service-settings waiter-url service-id :cookies cookies)
                                          [:instances :killed-instances])
                 log-url (:log-url (first killed-instances))
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -134,7 +134,7 @@
 
           ;; Test that the killed instances' logs were persisted to S3.
           ;; This portion of the test logic was modified from the active-instances tests above.
-          (let [log-bucket-url (-> (default-scheduler-settings waiter-url) :log-bucket-url)
+          (let [log-bucket-url (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url)
                 killed-instances (get-in (service-settings waiter-url service-id :cookies cookies)
                                          [:instances :killed-instances])
                 log-url (:log-url (first killed-instances))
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/util/client_tools.clj
```diff
@@ -563,38 +563,38 @@
...
   "Returns the Marathon URL setting"
-  [waiter-url & {:keys [verbose] :or {verbose false}}]
-  (setting waiter-url [:scheduler-config :marathon :url] :verbose verbose))
-
-(defn k8s-log-bucket-url
-  "Returns the S3 bucket url for K8s service log backups"
-  [waiter-url & {:keys [verbose] :or {verbose false}}]
-  (setting waiter-url [:scheduler-config :kubernetes :log-bucket-url] :verbose verbose))
+  [waiter-url]
+  (-> (default-scheduler-settings waiter-url) :url))
```
shamsimam: This doesn't work, it is retrieving the url from a scheduler, not necessarily marathon. It should retrieve the marathon scheduler settings and then lookup `url`.

waiter/integration/waiter/request_timeout_test.clj
```diff
@@ -223,7 +223,7 @@
                                                  (quot 1000000))] ; truncated nanos->millis
               (is (<= startup-delay-ms instance-acquired-delay-ms)
                   (str "Healthy instance was found in just " instance-acquired-delay-ms " ms (too short)")))
-            (when (using-marathon? waiter-url)
+            (when (can-query-for-grace-period? waiter-url)
               (is (= (t/in-seconds grace-period) (service-id->grace-period waiter-url service-id)))))
```
shamsimam: `service-id->grace-period` should be fixed so it can retrieve it for both marathon and kubernetes. The `check-pod-namespace` has an example of retricing data from the pod state.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -134,7 +134,7 @@
 
           ;; Test that the killed instances' logs were persisted to S3.
           ;; This portion of the test logic was modified from the active-instances tests above.
-          (let [log-bucket-url (k8s-log-bucket-url waiter-url)
+          (let [log-bucket-url (-> (default-scheduler-settings waiter-url) :log-bucket-url)
```
shamsimam: This only works if the default scheduler is kubernetes. Instead the function should retrieve kubernetes scheduler settings and then lookup `log-bucket-url`

waiter/integration/waiter/request_timeout_test.clj
nsinkov: separate PR

waiter/src/waiter/util/client_tools.clj
nsinkov: this is only called when the scheduler is marathon

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
nsinkov: this is only called when the scheduler is k8s

nsinkov: updated

waiter/integration/waiter/request_timeout_test.clj
nsinkov: updated

waiter/src/waiter/util/client_tools.clj
nsinkov: updated

nsinkov:
@shamsimam added internal green

PR #865, opened 2019-07-17
adds support for logging request trailers

shamsimam:

- adds support for logging request trailers

Without this logging, there is no easy way for us to historically track the response trailers and/or whether they were emitted for a given request.

### Example logs:
```
2019-07-17 14:11:45,148 INFO  waiter.process-request [async-dispatch-10] - [CID=courier-request-447826860536844.1] response trailers: {grpc-status 0}
...
2019-07-17 14:11:50,304 INFO  waiter.process-request [async-dispatch-37] - [CID=courier-request-447826860536844.3] response trailers: {grpc-status 1, grpc-message Cancelled by server}
```

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -469,6 +469,20 @@
                     (log/warn "unable to abort as request not found inside response!"))]
       (log/info "aborted backend request:" aborted))))
 
+(defn- introspect-trailers
+  "Introspects and logs trailers received in the response"
+  [{:keys [trailers] :as response}]
+  (if trailers
+    (let [trailers-ch (async/promise-chan)
+          correlation-id (cid/get-correlation-id)]
+      (async/go
+        (when-let [trailers-data (async/<! trailers)]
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -469,6 +469,20 @@
...
+(defn- introspect-trailers
+  "Introspects and logs trailers received in the response"
+  [{:keys [trailers] :as response}]
+  (if trailers
+    (let [trailers-ch (async/promise-chan)
+          correlation-id (cid/get-correlation-id)]
+      (async/go
+        (when-let [trailers-data (async/<! trailers)]
+          (async/>! trailers-ch trailers-data)
+          (cid/cinfo correlation-id "response trailers:" trailers-data))
```
sradack: Debug?

sradack: This has me thinking that we could set up a middleware request log that logs on request completion.  These headers would be good candidates for that log.

Issue #842, opened 2019-07-09
Refactor usage of java.util.URLDecoder


nsinkov:
make a url-encode method, move these methods to utils, and refactor usages of java.util.URLDecoder java.util.URLEncoder

nsinkov:
org.eclipse.jetty.util.UrlEncoded is supposed to be faster, unify usage

nsinkov:
in waiter/src/waiter/cookie_support.clj on line 27:
TODO: make a url-encode method, move these methods to utils, and refactor usages of java.util.URLDecoder java.util.URLEncoder

nsinkov:
org.eclipse.jetty.util.UrlEncoded does not encode the same way as java.util.URLEncoder

PR #869, opened 2019-07-19
remove usage of java.util.URLDecoder and java.util.URLEncoder

nsinkov:

- remove usage of java.util.URLDecoder and java.util.URLEncoder
- refactor url de/encoding to utils

We have two ways to url de/encode: java.util.URLDecoder and org.eclipse.jetty.util.UrlEncoded

Just use the jetty one, since it should be faster

see: https://github.com/twosigma/waiter/issues/842

### DIFF
--- a/waiter/integration/waiter/authentication_test.clj
+++ b/waiter/integration/waiter/authentication_test.clj
@@ -4,8 +4,9 @@
             [clojure.string :as string]
             [clojure.test :refer :all]
             [reaver :as reaver]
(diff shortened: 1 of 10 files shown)


nsinkov:
This code doesn't actually work... https://github.com/eclipse/jetty.project/issues/3565

shamsimam:
Are you going to replace these uses with a utlity function that Jetty provides?

PR #872, opened 2019-07-22
adds support for grpc-status on waiter generated responses

shamsimam:

- adds support for grpc-status on waiter generated responses

Provide additional information to the GRPC clients when available on Waiter responses.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -295,10 +295,8 @@
                                 (= status 500) ["13" "Internal Server Error"]
                                 (= status 503) ["14" "Service Unavailable"]
                                 (= status 504) ["4" "Gateway Timeout"])]
-      (let [[grpc-status grpc-message] grpc-status-data
-            grpc-message (if (string? body)
-                           (utils/truncate body 128)
-                           grpc-message)
+      (let [[grpc-status standard-message] grpc-status-data
+            grpc-message (if (string? body) body standard-message)
             trailers-ch (async/promise-chan)
             new-headers (assoc headers
                           "content-type" "application/grpc"
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -285,6 +285,48 @@
       (catch Exception e
         (utils/exception->response e request)))))
 
+(defn add-grpc-headers-and-trailers
+  "Finds and attaches the equivalent grpc status codes for the provided http status code."
+  [server-name {:keys [body headers status trailers] :as response}]
+  (if (= server-name (get-in response [:headers "server"]))
+    (if-let [grpc-status-data (cond
+                                (= status 400) ["3" "Bad Request"]
+                                (= status 401) ["16" "Unauthorized"]
+                                (= status 500) ["13" "Internal Server Error"]
+                                (= status 503) ["14" "Service Unavailable"]
+                                (= status 504) ["4" "Gateway Timeout"])]
+      (let [[grpc-status grpc-message] grpc-status-data
+            grpc-message (if (string? body)
+                           (utils/truncate body 128)
+                           grpc-message)
+            trailers-ch (async/promise-chan)
+            new-headers (assoc headers
+                          "content-type" "application/grpc"
+                          "grpc-message" grpc-message
+                          "grpc-status" grpc-status)]
+        (async/go
+          (let [trailers-data (and trailers (async/<! trailers))
+                new-trailers-data (assoc trailers-data
+                                    "grpc-message" grpc-message
+                                    "grpc-status" grpc-status)]
+            (async/>! trailers-ch new-trailers-data)))
+        (-> response
+          (dissoc :body)
+          (assoc :headers new-headers
+                 :trailers trailers-ch)))
+      response)
+    response))
+
+(defn wrap-grpc-status
+  "Attaches grpc-status on Waiter generated responses based on http status codes."
+  [handler server-name]
+  (fn wrap-grpc-status-fn
+    [{:keys [client-protocol headers] :as request}]
+    (let [response (handler request)]
+      (cond-> response
+        (http-utils/grpc? headers client-protocol)
+        (ru/update-response #(add-grpc-headers-and-trailers server-name %))))))
+
 (defn- make-blacklist-request
   [make-inter-router-requests-fn blacklist-period-ms dest-router-id dest-endpoint {:keys [id] :as instance} reason]
   (log/info "peer communication requesting" dest-router-id "to blacklist" id "via endpoint" dest-endpoint)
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/core.clj
```diff
@@ -266,6 +266,48 @@
...
+  (if (= server-name (get-in response [:headers "server"]))
+    (if-let [grpc-status-data (cond
+                                (= status 400) ["3" "Bad Request"]
+                                (= status 401) ["16" "Unauthorized"]
+                                (= status 500) ["13" "Internal Server Error"]
+                                (= status 503) ["14" "Service Unavailable"]
+                                (= status 504) ["4" "Gateway Timeout"])]
+      (let [[grpc-status grpc-message] grpc-status-data
+            grpc-message (if (string? body)
+                           (utils/truncate body 128)
```
sradack: Why are we truncating to 128 bytes?

waiter/src/waiter/util/utils.clj
```diff
@@ -251,6 +252,10 @@
   (let [error-context (build-error-context data-map request)
         content-type (request->content-type request)]
     {:body (case content-type
+             "application/grpc"
```
sradack: What do gRPC clients generally do with the body content when `grpc-status` indicates an error?

shamsimam: I get a `gRPC frame header malformed: reserved bits not zero` message. I believe it has to do with the grpc encoding of the body.

https://github.com/grpc/grpc-java/blob/d7b9438d399e44d9dccd0f7830cb544bb1840f0d/core/src/main/java/io/grpc/internal/MessageDeframer.java#L372

Even if the header was set, I think this is problematic because grpc will try to parse the body as the declared response type from the proto file.

waiter/src/waiter/core.clj
shamsimam: Removed the truncation. I verified that I could send 1000s of characters in the status message.

waiter/src/waiter/main.clj
```diff
@@ -77,14 +77,15 @@
...
-                         [:state cors-validator router-id]
+                         [:state cors-validator router-id server-name]
                          handlers] ; Insist that all systems are running before we start server
                   (let [options (merge (cond-> server-options
                                          (:ssl-port server-options) (assoc :ssl? true))
                                        websocket-config
                                        {:ring-handler (-> (core/ring-handler-factory waiter-request?-fn handlers)
                                                           (cors/wrap-cors-preflight cors-validator (:max-age cors-config))
                                                           core/wrap-error-handling
+                                                          (core/wrap-grpc-status server-name)
```
sradack: Should we install the middleware around the `process-request` handler only?  I can't think of any other path where gRPC statuses need to be generated.

waiter/src/waiter/util/utils.clj
sradack: Should we avoid providing a body then?  Should we look at what other proxies do?

waiter/src/waiter/core.clj
```diff
@@ -266,6 +266,46 @@
       (catch Exception e
         (utils/exception->response e request)))))
 
+(defn add-grpc-headers-and-trailers
+  "Finds and attaches the equivalent grpc status codes for the provided http status code."
+  [server-name {:keys [body headers status trailers] :as response}]
+  (if (= server-name (get-in response [:headers "server"]))
```
sradack: We should definitely make a clear distinction between when Waiter is generating a response and when its a proxied response.  I'm not sure checking the server header is the best way to check for that distinction.  To illustrate my point, a backend should not be able to trigger this code by emitting a `server` header that is the same as Waiter's.

waiter/src/waiter/util/utils.clj
shamsimam: Will fix.

waiter/src/waiter/main.clj
shamsimam: The authentication responses are not generated by `process-request`

waiter/src/waiter/core.clj
shamsimam: I'll add a `:source` key to the responses to track Waiter generated responses.

shamsimam: done.

waiter/src/waiter/util/utils.clj
shamsimam: Avoided body.

Envoy produces responses with approrpiate grpc status headers, example:

https://github.com/envoyproxy/envoy/blob/fd7c70b3f3739554974769f1d0e72b706eb5c5a3/source/extensions/filters/http/grpc_http1_reverse_bridge/filter.cc#L154-L158

https://github.com/envoyproxy/envoy/blob/15706964a29e595c045a5d7ef0646d04f347dcc1/source/common/grpc/status.cc#L6-L26

waiter/src/waiter/util/utils.clj
```diff
@@ -270,6 +274,9 @@
                  (str/replace #"\n  $" "\n")))
        :headers (-> headers
                   (assoc-if-absent "content-type" content-type))
+       :waiter/message (-> (or (-> data-map :details :friendly-error-message)
```
sradack: Why isn't the appropriate message already passed in via the `message` argument?

waiter/src/waiter/core.clj
```diff
@@ -285,6 +285,48 @@
...
+                                (= status 503) ["14" "Service Unavailable"]
+                                (= status 504) ["4" "Gateway Timeout"])]
+      (let [[grpc-status standard-message] grpc-status-data
+            grpc-message (if (string? message) message standard-message)
+            trailers-ch (async/promise-chan)
+            new-headers (assoc headers
+                          "content-type" "application/grpc"
+                          "grpc-message" grpc-message
+                          "grpc-status" grpc-status)]
+        (async/go
```
sradack: Do we expect to ever have trailers?  Do gRPC responses always require trailers?

shamsimam: gRPC responses do not always require trailers, when both headers and trailers are present, it looks into the trailers for the status.

We need to send the trailers as a workaround to jetty limtation of closing the stream with an empty data frame when only headers are sent. The presence of a data frame confuses grpc and it reports a different error than the one sent by Waiter:

HTTP/2 traffic:
```
2019-07-26 01:46:04 DEBUG io.grpc.netty.NettyClientHandler - [id: 0x878392ea, L:/127.0.0.1:60283 - R:/127.0.0.1:9091] INBOUND HEADERS: streamId=3 headers=GrpcHttp2ResponseHeaders[:status: 504, x-cid: wgttguc150048199015780, grpc-status: 4, grpc-message: Request to service backend timed out, ...] padding=0 endStream=false
2019-07-26 01:46:04 DEBUG io.grpc.netty.NettyClientHandler - [id: 0x878392ea, L:/127.0.0.1:60283 - R:/127.0.0.1:9091] INBOUND DATA: streamId=3 padding=0 endStream=true length=0 bytes=
```

Status code:
```
{:code "INTERNAL", :description "Received unexpected EOS on DATA frame from server."}
```

waiter/src/waiter/util/utils.clj
shamsimam: I initially avoided it since `message` is more verbose. But since we have plenty of room with the header value, using `message` now.

The difference is between: (message) `Service description using waiter headers/token improperly configured; concurrency-level must be an integer in the range [1, 10000].` and (friendly-error-message) `concurrency-level must be an integer in the range [1, 10000].`

Issue #737, opened 2019-05-07
Kitchen should filter auth on /request-info


DaoWen:
We remove the `WAITER_PASSWORD` field from the `/environment` endpoint, but we don't remove the `authorization` header from the `/request-info` endpoint output.

PR #873, opened 2019-07-22
truncates authorization and cookie values in kitchen request-info response

shamsimam:

- truncates authorization and cookie values in kitchen request-info response

We should not display back sensitive information in the response.

Fixes https://github.com/twosigma/waiter/issues/737

### DIFF
--- a/containers/test-apps/kitchen/bin/kitchen
+++ b/containers/test-apps/kitchen/bin/kitchen
@@ -123,7 +123,10 @@ def gzip_compress(in_data):
 
 def truncate(in_string, truncate_length):
     "Collapse and truncate the given text to the given length."
-    return in_string[:truncate_length] + (in_string[truncate_length:] and '...')
+    if len(in_string) > truncate_length:
+        return in_string[:truncate_length] + '...'
+    else:
+        return in_string
 
 
 class LoggerMixin():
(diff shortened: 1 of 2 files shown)


containers/test-apps/kitchen/bin/kitchen
```diff
@@ -121,6 +121,11 @@ def gzip_compress(in_data):
     return out_data
 
 
+def truncate(in_string, truncate_length):
+    "Collapse and truncate the given text to the given length."
+    return in_string[:truncate_length] + (in_string[truncate_length:] and '...')
```
DaoWen: I find `(in_string[truncate_length:] and '...')` confusing... Why not `if len(in_string) > truncate_length`? I'd prefer not having the extra slicing operations when that condition fails.

containers/test-apps/kitchen/bin/kitchen
```diff
@@ -943,10 +948,22 @@ class Kitchen(HTTPWebSocketsHandler):
                  'request-length': self.__request_body_length,
                  'request-method': self.__method,
                  'uri': self.__path }
+
         if len(self.headers) > 0:
             info['headers'] = { k.lower(): v for k, v in self.headers.items() }
+
+            cookie_value = info['headers'].get('authorization', None)
```
DaoWen: The explicit `None` here is redundant:
https://docs.python.org/3/library/stdtypes.html#dict.get

containers/test-apps/kitchen/bin/kitchen
```diff
@@ -943,10 +948,22 @@ class Kitchen(HTTPWebSocketsHandler):
...
+            if cookie_value is not None:
+                info['headers']['authorization'] = truncate(cookie_value, 15)
+
+            cookie_value = info['headers'].get('cookie', None)
```
DaoWen: The explicit `None` here is redundant:
https://docs.python.org/3/library/stdtypes.html#dict.get

containers/test-apps/kitchen/bin/kitchen
```diff
@@ -943,10 +948,22 @@ class Kitchen(HTTPWebSocketsHandler):
...
+            if cookie_value is not None:
+                cookie_entries = [e for e in cookie_value.split(';') if e]
+                info['headers']['cookie'] = [truncate(v, 25) for v in cookie_entries]
```
DaoWen: Combine these two lines?
```
info['headers']['cookie'] = [truncate(v, 25) for v in cookie_value.split(';') if v]
```

containers/test-apps/kitchen/bin/kitchen
shamsimam: Done, uses `if-else` block

containers/test-apps/kitchen/bin/kitchen
shamsimam: Done, remove `None`.

containers/test-apps/kitchen/bin/kitchen
shamsimam: Done, remove `None`.

containers/test-apps/kitchen/bin/kitchen
shamsimam: Combined two lines.

PR #876, opened 2019-07-23
enables proto tests

shamsimam:

- corrects scheme for https requests from proto tests
- enables proto tests for all schedulers

We would like to test the full complement of our proto support while streaming data.


shamsimam:
@pschorf build is green

PR #879, opened 2019-07-24
Catch and log errors in k8s track-failed-instances fn

DaoWen:

Catch and log all thrown errors in the Kubernetes `track-failed-instances!` function.

We don't want higher-level operations (e.g., the scheduler update loop) to get killed because there was some error reporting a failed instance from a restarted pod.


shamsimam:
Awaiting green build

PR #882, opened 2019-07-25
aborts backend requests using the abort-ch

shamsimam:

- aborts backend requests using the abort-ch

Consistently abort backend requests using the `abort-ch` mechanism. Fallback to invoking `abort` directly only when the channel mechanism fails.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -264,7 +264,7 @@
                                                                   correlation-id
                                                                   (log/info "request byte stream error, result of aborted request:" aborted?)
                                                                   (async/close! body-ch)))]
-                                                 (log/error throwable "unable to stream request bytes, aborting request")
+                                                 (log/info throwable "unable to stream request bytes, aborting request")
                                                  (async/>!! abort-ch [throwable callback])))
                                              (report-request-size-metrics 0 true)
                                              (if throwable
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -250,15 +250,29 @@
                                        (cid/with-correlation-id
                                          correlation-id
                                          (let [complete-trigger-id (utils/unique-identifier)]
+                                           (log/debug throwable "received from ctrl-chan" message)
                                            (deliver complete-triggered-promise complete-trigger-id)
                                            (when (= complete-trigger-id @complete-triggered-promise)
                                              (log/debug "closing request body as" message)
                                              (counters/dec! request-body-streaming-counter)
+                                             ;; the callback is necessary as there is a data race between aborting the
+                                             ;; request and closing of the body channel triggering a normal complete before
+                                             ;; the abort request gets processed.
                                              (when throwable
-                                               (log/error throwable "unable to stream request bytes, aborting request")
-                                               (async/>!! abort-ch throwable))
+                                               (let [callback (fn complete-request-streaming-abort-result-callback [aborted?]
+                                                                (cid/with-correlation-id
+                                                                  correlation-id
+                                                                  (log/info "request byte stream error, result of aborted request:" aborted?)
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -250,15 +250,29 @@
                                        (cid/with-correlation-id
                                          correlation-id
                                          (let [complete-trigger-id (utils/unique-identifier)]
+                                           (log/debug throwable "received from ctrl-chan" message)
                                            (deliver complete-triggered-promise complete-trigger-id)
                                            (when (= complete-trigger-id @complete-triggered-promise)
                                              (log/debug "closing request body as" message)
                                              (counters/dec! request-body-streaming-counter)
+                                             ;; the callback is necessary as there is a data race between aborting the
```
sradack: Don't we control when `body-ch` gets closed later in the function?

waiter/src/waiter/process_request.clj
```diff
@@ -250,15 +250,29 @@
...
+                                                                (cid/with-correlation-id
+                                                                  correlation-id
+                                                                  (log/info "request byte stream error, result of aborted request:" aborted?)
+                                                                  (async/close! body-ch)))]
+                                                 (log/error throwable "unable to stream request bytes, aborting request")
+                                                 (async/>!! abort-ch [throwable callback])))
                                              (report-request-size-metrics 0 true)
-                                             (async/close! body-ch))))))]
+                                             (if throwable
+                                               ;; TODO avoid the need for the timeout while sending to abort channel
```
sradack: What would it take to avoid the timeout?

waiter/src/waiter/process_request.clj
```diff
@@ -485,13 +499,19 @@
...
-                    (log/warn "unable to abort as request not found inside response!"))]
-      (log/info "aborted backend request:" aborted))))
+  [{:keys [abort-ch] :as response}]
+  (fn abort-backend-request-callback [^Exception ex]
+    (let [correlation-id (cid/get-correlation-id)
+          callback (fn abort-http-request-result-callback [aborted?]
+                     (cid/with-correlation-id
+                       correlation-id
+                       (log/debug ex "aborted backend request:" aborted? "due to" (.getClass ex))))]
+      (when abort-ch
```
sradack: When do we not have an `abort-ch`?

waiter/src/waiter/process_request.clj
```diff
@@ -485,13 +499,19 @@
...
+        (log/debug "requesting backend to be aborted via abort-ch"))
+      (when-not (and abort-ch (async/>!! abort-ch [ex callback]))
+        (log/debug "aborting backend request directly")
+        (let [aborted? (some-> response :request (.abort ex))]
```
sradack: Should we log when we try to abort but are unable to?

waiter/src/waiter/process_request.clj
```diff
@@ -250,15 +250,29 @@
...
+                                             ;; the abort request gets processed.
                                              (when throwable
-                                               (log/error throwable "unable to stream request bytes, aborting request")
-                                               (async/>!! abort-ch throwable))
+                                               (let [callback (fn complete-request-streaming-abort-result-callback [aborted?]
+                                                                (cid/with-correlation-id
+                                                                  correlation-id
+                                                                  (log/info "request byte stream error, result of aborted request:" aborted?)
+                                                                  (async/close! body-ch)))]
+                                                 (log/error throwable "unable to stream request bytes, aborting request")
```
sradack:
```suggestion
                                                 (log/info throwable "unable to stream request bytes, aborting request")
```

waiter/src/waiter/process_request.clj
shamsimam: I created https://github.com/twosigma/waiter/pull/884 to show what we would need.

PR #883, opened 2019-07-25
explicitly classify error cause during request/response streaming

shamsimam:

- explicitly classify error cause during request/response streaming

We should correctly attribute streaming errors to the client or the backend. This assignment also has implications on instance blacklisting.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -165,7 +165,7 @@
                               (log/info "done processing request" status)
                               (when (= :success status)
                                 (counters/inc! (metrics/service-counter service-id "request-counts" "successful")))
-                              (when (str/ends-with? status "error")
+                              (when (contains? #{:client-error :generic-error :instance-error} status)
                                 (counters/inc! (metrics/service-counter service-id "request-counts" (name status))))
                               (when (= :generic-error status)
                                 (log/error "there was a generic error in processing the request;"
@@ -481,7 +481,7 @@
                     (when more-bytes-possibly-available?
                       (recur bytes-streamed' bytes-reported-to-statsd'))))))))
         (catch Exception e
-          (log/error e "exception occurred while streaming response for" service-id)
+          (log/info e "exception occurred while streaming response for" service-id)
           (meters/mark! stream-exception-meter)
           (let [[error-cause _ _] (classify-error e)]
             (deliver reservation-status-promise error-cause))
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -15,6 +15,7 @@
 ;;
 (ns waiter.process-request
   (:require [clojure.core.async :as async]
+            [clojure.core.async.impl.protocols :as protocols]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [full.async :as fa]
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -162,9 +165,11 @@
                               (log/info "done processing request" status)
                               (when (= :success status)
                                 (counters/inc! (metrics/service-counter service-id "request-counts" "successful")))
+                              (when (str/ends-with? status "error")
```
sradack: Should we be explicit instead of have convention on the name of the keyword?

waiter/src/waiter/process_request.clj
```diff
@@ -175,18 +180,34 @@
...
       instance)))
 
+(defn- classify-error
+  "Classifies the error responses from the backend into the following vector:
+   - error cause (:client-error, :instance-error or :generic-error),
+   - associated error message, and
+   - the http status code."
+  [error]
+  (let [classification (cond (instance? ExceptionInfo error)
+                             (let [[error-cause message status] (classify-error (ex-cause error))
```
sradack: I can't find any docs on an `ex-cause` function.

sradack: Are there currently instances where the recursive call is needed?

waiter/src/waiter/process_request.clj
```diff
@@ -175,18 +180,34 @@
...
+                                   error-cause (or (-> error ex-data :error-cause) error-cause)]
```
sradack: Should `classify-error` be responsible for determining `error-cause` instead of overriding it here?

waiter/src/waiter/process_request.clj
```diff
@@ -455,16 +481,17 @@
                     (when more-bytes-possibly-available?
                       (recur bytes-streamed' bytes-reported-to-statsd'))))))))
         (catch Exception e
+          (log/error e "exception occurred while streaming response for" service-id)
```
sradack:
```suggestion
          (log/info e "exception occurred while streaming response for" service-id)
```

waiter/src/waiter/process_request.clj
```diff
@@ -175,18 +180,34 @@
...
+                             (instance? IllegalStateException error)
+                             [:generic-error (.getMessage error) 400]
+                             (instance? EofException error)
+                             [:client-error "Connection unexpectedly closed while streaming request" 400]
+                             (instance? TimeoutException error)
+                             [:instance-error (utils/message :backend-request-timed-out) 504]
+                             :else
+                             [:instance-error (utils/message :backend-request-failed) 502])
+        [error-cause _ _] classification]
+    (log/info (-> error .getClass .getCanonicalName) (.getMessage error) "identified as" error-cause)
```
sradack: Are there overloads of the logging functions (e.g. `error`, `log`) that take a throwable?  Is the serialization/formatting of that argument insufficient?  Does it report the class name and message?

shamsimam: Placing the raw exception as an argument to the `log/info|error` methods causes the entire stack trace to be printed in various formats, java native stacktrace format or clojure stacktrace format, depending on the position of the argument.

waiter/src/waiter/process_request.clj

shamsimam:
Ready for another round of review.

waiter/src/waiter/process_request.clj
sradack: Why don't we want the stacktrace?

PR #884, opened 2019-07-25
aborts backend requests using the abort-ch - v2

shamsimam:
## Jet PR:

[handles multiple abort attempts to a request](https://github.com/twosigma/jet/pull/35/files)

- aborts backend requests using the abort-ch
- handles multiple abort attempts to a request

Consistently abort backend requests using the `abort-ch` mechanism. Fallback to invoking `abort` directly only when the channel mechanism fails.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -271,28 +271,27 @@
                                        (cid/with-correlation-id
                                          correlation-id
                                          (let [complete-trigger-id (utils/unique-identifier)]
-                                           (log/debug throwable "received from ctrl-chan" message)
                                            (deliver complete-triggered-promise complete-trigger-id)
                                            (when (= complete-trigger-id @complete-triggered-promise)
-                                             (log/debug "closing request body as" message)
+                                             ;; avoid decrementing the counter multiple times
                                              (counters/dec! request-body-streaming-counter)
-                                             ;; the callback is necessary as there is a data race between aborting the
-                                             ;; request and closing of the body channel triggering a normal complete before
-                                             ;; the abort request gets processed.
-                                             (when throwable
-                                               (let [callback (fn complete-request-streaming-abort-result-callback [aborted?]
-                                                                (cid/with-correlation-id
-                                                                  correlation-id
-                                                                  (log/info "request byte stream error, result of aborted request:" aborted?)
-                                                                  (async/close! body-ch)))]
-                                                 (log/info throwable "unable to stream request bytes, aborting request")
-                                                 (async/>!! abort-ch [throwable callback])))
-                                             (report-request-size-metrics 0 true)
-                                             (if throwable
-                                               ;; TODO avoid the need for the timeout while sending to abort channel
-                                               (async/go
-                                                 (async/<! (async/timeout 1000))
-                                                 (async/close! body-ch))
+                                             (report-request-size-metrics 0 true))
+                                           (if throwable
+                                             (let [identifier-map {:identifier complete-trigger-id}
+                                                   ;; the callback is necessary as there is a data race between aborting the
+                                                   ;; request and closing of the body channel triggering a normal complete before
+                                                   ;; the abort request gets processed.
+                                                   callback (fn complete-request-streaming-abort-result-callback [aborted?]
+                                                              (cid/with-correlation-id
+                                                                correlation-id
+                                                                (log/info "result of aborted request:" aborted? identifier-map)
+                                                                (async/close! body-ch)))]
+                                               (log/info throwable "aborting request" message identifier-map)
+                                               (when-not (async/put! abort-ch [throwable callback])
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -252,13 +252,26 @@
...
+                                                   ;; the callback is necessary as there is a data race between aborting the
+                                                   ;; request and closing of the body channel triggering a normal complete before
+                                                   ;; the abort request gets processed.
+                                                   callback (fn complete-request-streaming-abort-result-callback [aborted?]
+                                                              (cid/with-correlation-id
+                                                                correlation-id
+                                                                (log/info "result of aborted request:" aborted? identifier-map)
+                                                                (async/close! body-ch)))]
+                                               (log/info throwable "aborting request" message identifier-map)
+                                               (when-not (async/put! abort-ch [throwable callback])
```
sradack: Why do we need to close the `body-ch` here?  The callback includes closing `body-ch`.  Doesn't the callback get called always?

waiter/src/waiter/process_request.clj
```diff
@@ -485,13 +498,22 @@
...
+    (let [correlation-id (cid/get-correlation-id)
+          callback (fn abort-http-request-result-callback [aborted?]
+                     (cid/with-correlation-id
+                       correlation-id
+                       (log/info ex "aborted backend request:" aborted?)))]
+      (when-not abort-ch
+        (log/warn "abort-ch not available in response"))
+      (log/debug "aborting backend request")
+      (when-not (and abort-ch (async/put! abort-ch [ex callback]))
+        (log/debug "aborting backend request directly")
```
sradack: If the `abort-ch` is closed, doesn't that mean the backend request is already aborted?

waiter/src/waiter/process_request.clj
```diff
@@ -485,13 +498,22 @@
...
+        (if-let [request (:request response)]
```
sradack: Does the response always have a `:request` key?

waiter/src/waiter/process_request.clj
```diff
@@ -252,13 +252,26 @@
...
+                                                   ;; the abort request gets processed.
+                                                   callback (fn complete-request-streaming-abort-result-callback [aborted?]
+                                                              (cid/with-correlation-id
+                                                                correlation-id
+                                                                (log/info "result of aborted request:" aborted? identifier-map)
+                                                                (async/close! body-ch)))]
+                                               (log/info throwable "aborting request" message identifier-map)
+                                               (when-not (async/put! abort-ch [throwable callback])
+                                                 ;; abort channel already closed, cleanup by closing the body channel
+                                                 (async/close! body-ch)))
```
sradack: Is it still possible that we're closing the `body-ch` before aborting the request?

sradack: Can you point me to the code that handles aborting the response to the client such that on the close of `body-ch` we don't complete normally?

waiter/src/waiter/process_request.clj
shamsimam: If the put fails, the callback will not be invoked, we want to go ahead and close the channel regardless as we are done streaming the request body.

waiter/src/waiter/process_request.clj
shamsimam: I'm not sure I understand the question completely. Closing `body-ch` means we are done streaming the request body. It does not directly trigger the abort of the backend request.

Invoking abort on the backend request causes Jetty to treat the request as a failure and stream appropriate frames to the client in the response body. The abort is called at:
https://github.com/twosigma/jet/blob/e1f0a7b279c564c4a457ee1d020797c818bc7f39/src/clj/qbits/jet/client/http.clj#L434

and the backend request cleanup is performed at:
https://github.com/twosigma/jet/blob/e1f0a7b279c564c4a457ee1d020797c818bc7f39/src/clj/qbits/jet/client/http.clj#L547

The above causes the exception to be propagated on the `error-chan` which we process at:
https://github.com/twosigma/waiter/blob/44685bb6d22fc242f9a741f25f5e7c12c00e330e/waiter/src/waiter/process_request.clj#L446

waiter/src/waiter/process_request.clj
shamsimam: I expect the response to always have the request key. I added the warn so we are notified when this is not true.

waiter/src/waiter/process_request.clj
shamsimam: The branch is also covering the case when the `abort-ch` is not available.

`abort-ch` closed means we do not have a way to asynchronously triggering the abort logic. Currently, `abort-ch` is only closed when we are done processing the request:
https://github.com/twosigma/jet/blob/e1f0a7b279c564c4a457ee1d020797c818bc7f39/src/clj/qbits/jet/client/http.clj#L411

PR #886, opened 2019-07-25
wrap test metrics in try catch

nsinkov:

- wrap test metrics in try catch

If we can't post test metrics, we should not have the test process error out.


dposada:

PR #889, opened 2019-07-26
Run test-s3-logs only when log-bucket-url is configured

DaoWen:

Change predicate on `test-s3-logs` to check if the `:log-bucket-url` setting was configured.

This is an optional setting, but our integration tests currently assume it's always set if you're using Kubernetes scheduler. The new predicate is more consistent.

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -77,12 +77,12 @@
   (testing-using-waiter-url
     (when (using-k8s? waiter-url)
       (when-let [log-bucket-url (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url)]
-        (let [headers {:x-waiter-name (rand-name)
-                       :x-waiter-concurrency-level 1
+        (let [headers {:x-waiter-concurrency-level 1
                        :x-waiter-distribution-scheme "simple"
                        :x-waiter-max-instances 2
-                       :x-waiter-scale-up-factor 0.99
-                       :x-waiter-scale-down-factor 0.99}
+                       :x-waiter-name (rand-name)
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -75,9 +75,9 @@
...
   (testing-using-waiter-url
-    (when (using-k8s? waiter-url)
-      (let [log-bucket-url (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url)
-            headers {:x-waiter-name (rand-name)
+    (when-let [log-bucket-url (when (using-k8s? waiter-url)
+                                (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url))]
+      (let [headers {:x-waiter-name (rand-name)
                      :x-waiter-concurrency-level 1
                      :x-waiter-distribution-scheme "simple"
                      :x-waiter-max-instances 2
```
shamsimam: Nitpick: lexicographic order.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -75,9 +75,9 @@
 ;; Fails on (is (> (count instance-ids) 1) (str instance-ids)) as there is only one instance
 (deftest ^:parallel ^:integration-slow ^:resource-heavy test-s3-logs
   (testing-using-waiter-url
-    (when (using-k8s? waiter-url)
-      (let [log-bucket-url (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url)
```
shamsimam: I think changing this `let` to a `when-let` makes the test simpler to read.

DaoWen: Refactored

PR #894, opened 2019-07-30
adds support for rate limiting the number of spnego auth responses by requesting host

shamsimam:

- adds support for rate-limiting the number of spnego auth responses by requesting host

Avoids a misbehaving client (cookies disabled) from affecting other services by overloading the Kerberos auth mechanism.

### DIFF
--- a/waiter/src/waiter/auth/spnego.clj
+++ b/waiter/src/waiter/auth/spnego.clj
@@ -69,16 +69,36 @@
       (cookies/cookies-response)))
 
 (defn response-503-temporarily-unavailable
-  "Tell the client you're overloaded and would like them to try later"
+  "Tell the client you're overloaded and would like them to try later."
   []
-  (log/info "triggering 401 negotiate for spnego authentication")
+  (log/info "triggering 503 temporarily unavailable")
   (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
   (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
-  (-> (rr/response "Too many Kerberos authentication requests")
+  (-> (rr/response "Too many Kerberos authentication requests; consider enabling cookies in your client")
       (rr/status 503)
       (rr/header "content-type" "text/plain")
       (utils/attach-waiter-source)
       (cookies/cookies-response)))
+
+(defn response-401-negotiate-or-503-temporarily-unavailable
+  "Depending upon the rate of requests requiring kerberos auth,
+   either tell the client you're overloaded and would like them to try later,
+   or tell the client you'd like them to use kerberos to authenticate."
+  [max-one-minute-rate-per-host request]
+  (let [host-name (-> request
+                    :headers
+                    (get "host")
+                    (or "_unknown_")
+                    utils/authority->host
+                    (str/replace "." "_")
+                    str/lower-case)
+        kerberos-meter (metrics/waiter-meter "core" "gss-auth-request" host-name)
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/auth/spnego.clj
```diff
@@ -69,17 +69,37 @@
...
 (defn response-503-temporarily-unavailable
-  "Tell the client you're overloaded and would like them to try later"
+  "Tell the client you're overloaded and would like them to try later."
   []
-  (log/info "triggering 401 negotiate for spnego authentication")
+  (log/info "triggering 503 temporarily unavailable")
   (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
   (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
-  (-> (rr/response "Too many Kerberos authentication requests")
+  (-> (rr/response "Too many Kerberos authentication requests; consider enabling cookies in your client")
```
sradack: It's possible to get this message if you have cookies enabled.

sradack: We should be able to add a custom message here that links to documentation.

waiter/src/waiter/auth/spnego.clj
```diff
@@ -69,17 +69,37 @@
...
 
+(defn response-401-negotiate-or-503-temporarily-unavailable
+  "Depending upon the rate of requests requiring kerberos auth,
+   either tell the client you're overloaded and would like them to try later,
+   or tell the client you'd like them to use kerberos to authenticate."
+  [max-one-minute-rate-per-host request]
+  (let [host-name (-> request
+                    :headers
+                    (get "host")
+                    (or "_unknown_")
```
sradack: `_no_host_`?  Could we better indicate that the client didn't specify the host rather then that we don't know the host?

waiter/src/waiter/auth/spnego.clj
```diff
@@ -69,17 +69,37 @@
...
+                    utils/authority->host
+                    (str/replace "." "_")
+                    str/lower-case)
+        kerberos-meter (metrics/waiter-meter "core" "gss-auth-request" host-name)
+        one-minute-rate (meters/rate-one kerberos-meter)]
+    (if (>= one-minute-rate max-one-minute-rate-per-host)
+      (response-503-temporarily-unavailable)
+      (do
+        (meters/mark! kerberos-meter)
```
sradack: Are we worried about cleaning up old metrics?

waiter/src/waiter/auth/spnego.clj
```diff
@@ -69,17 +69,37 @@
...
+  [max-one-minute-rate-per-host request]
+  (let [host-name (-> request
+                    :headers
+                    (get "host")
+                    (or "_unknown_")
+                    utils/authority->host
+                    (str/replace "." "_")
+                    str/lower-case)
+        kerberos-meter (metrics/waiter-meter "core" "gss-auth-request" host-name)
+        one-minute-rate (meters/rate-one kerberos-meter)]
```
sradack: Are we confident in the way this meter measures the rate?

waiter/src/waiter/auth/spnego.clj
sradack: Do we intend to limit all on-the-fly requests to Waiter's hostname to the same rate as an individual service?

shamsimam:
Closing as we are looking at other rate-limiting options.

PR #907, opened 2019-08-06
ensures the server header gets added to the request log for waiter generated responses

shamsimam:

- ensures the server header gets added to the request log for waiter generated responses

We would like to continue to report the server header from Waiter responses in the request log.


shamsimam:
Build is now green.

PR #909, opened 2019-08-07
adds state for kubernetes instances

shamsimam:

- adds state for kubernetes instances

Adds more visibility into the state of the Kubernetes pods.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -215,14 +215,16 @@
   [restart-expiry-threshold pod]
   (try
     (let [port0 (get-in pod [:spec :containers 0 :ports 0 :containerPort])
+          ;; waiter-app is the first container we register
           restart-count (get-in pod [:status :containerStatuses 0 :restartCount] 0)
           run-as-user (or (get-in pod [:metadata :labels :waiter/user])
                           ;; falling back to namespace for legacy pods missing the waiter/user label
                           (k8s-object->namespace pod))
+          ;; pod phase documentation: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
           {:keys [phase] :as pod-status} (:status pod)
-          ;; waiter-app is the first container we register
-          primary-container-state (pc/map-vals #(or (:reason %) "reason_unknown")
-                                               (get-in pod-status [:containerStatuses 0 :state]))]
+          primary-container-state (get-in pod-status [:containerStatuses 0 :state])]
+      (when (> (count primary-container-state) 1)
+        (log/warn "received multiple states for waiter-app:" primary-container-state))
       (scheduler/make-ServiceInstance
         (cond-> {:extra-ports (->> (get-in pod [:metadata :annotations :waiter/port-count])
                                    Integer/parseInt range next (mapv #(+ port0 %)))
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -214,28 +215,34 @@
   [restart-expiry-threshold pod]
   (try
     (let [port0 (get-in pod [:spec :containers 0 :ports 0 :containerPort])
-          restart-count (get-in pod [:status :containerStatuses 0 :restartCount])
+          restart-count (get-in pod [:status :containerStatuses 0 :restartCount] 0)
           run-as-user (or (get-in pod [:metadata :labels :waiter/user])
                           ;; falling back to namespace for legacy pods missing the waiter/user label
-                          (k8s-object->namespace pod))]
+                          (k8s-object->namespace pod))
+          {:keys [phase] :as pod-status} (:status pod)
```
DaoWen: It would be great to have a comment with a link to the Pod Phase documentation here.
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -214,28 +215,34 @@
...
+          ;; waiter-app is the first container we register
```
DaoWen: We already index to container `0` several places above, so this comment seems out of place. Either move it up higher or delete it.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -214,28 +215,34 @@
...
+          primary-container-state (pc/map-vals #(or (:reason %) "reason_unknown")
```
DaoWen: This would give us `{:running "reason_unknown"}`. Is that really what we want?

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -394,11 +394,30 @@
                              :annotations {:waiter/port-count "1"
                                            :waiter/service-id "test-app-1234"}}
                   :spec {:containers [{:ports [{:containerPort 8080 :protocol "TCP"}]}]}
-                  :status {:podIP "10.141.141.11"
+                  :status {:phase "Pending"
+                           :podIP "0.0.0.0"
```
DaoWen: Why add the `:podIP` field here? I think Kubernetes leaves this key out if an IP hasn't been assigned yet.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -468,9 +487,21 @@
...
+                        :log-directory "/home/myself/r0"
+                        :port 8080
+                        :service-id "test-app-1234"
+                        :started-at (du/str-to-date "2014-09-13T00:24:46Z" k8s-timestamp-format)})
+                     (scheduler/make-ServiceInstance
                        {:healthy? true
                         :host "10.141.141.11"
                         :id "test-app-1234.test-app-1234-abcd1-0"
+                        :k8s/pod-phase "Running"
+                        :k8s/primary-container-state {:running "reason_unknown"}
```
DaoWen: Again, is this really a helpful state? I'd rather have something like this:
```
:k8s/primary-container-state {:state :running}
```
and
```
:k8s/primary-container-state {:state :terminated
                              :reason "OOMKilled"}
 ```

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Added.

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Removed.

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Moved comment near top of function

waiter/test/waiter/scheduler/kubernetes_test.clj
shamsimam: Changed and added a warning for when multiple states are provided by k8s.

waiter/test/waiter/scheduler/kubernetes_test.clj
shamsimam: This was added for the tests as `live-pod?` filters out entries that do not have IPs. That will be fixed in the follow-up PR addressing container creating.

PR #921, opened 2019-08-14
Support annotating a Waiter pod as expired

DaoWen:

Consider a Waiter-K8s pod expired if it has the annotation `waiter/pod-expired` set to `"true"`.

This is a cleaner way to manually replace Waiter pods, rather than deleting the pods directly. The expiry logic will create a replacement and then delete the target after the replacement is healthy.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -212,9 +212,10 @@
   "Returns true when the pod instance can be classified as expired.
    An instance can be expired for the following reasons:
    - it has restarted too many times (reached the restart-expiry-threshold threshold)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -232,6 +233,12 @@
                 {:instance-id instance-id
                  :primary-container-status primary-container-status
                  :started-at pod-started-at})
+      true)
+    (= "true" pod-expired)
```
shamsimam: Nitpick: this is cheaper to compute, we can move it higher in the `cond` list.

DaoWen: I expect this to be orders of magnitude less frequent than the other two cases, which is why I put it last.

PR #922, opened 2019-08-14
remove parallel-test

nsinkov:

- remove parallel-test

parallel-test hangs, replace with eftest

### DIFF
--- a/waiter/project.clj
+++ b/waiter/project.clj
@@ -114,8 +114,7 @@
   :eftest {:report clojure.test/report}
   :resource-paths ["resources"]
   :main waiter.main
-  :plugins [[com.holychao/parallel-test "0.3.2"]
-            [lein-exec "0.3.7"]
+  :plugins [[lein-exec "0.3.7"]
             [test2junit "1.2.2"]
             [lein-eftest "0.5.8"]]
   ; In case of kerberos problems, export KRB5_KTNAME=/var/spool/keytabs/$(id -un)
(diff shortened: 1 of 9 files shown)


nsinkov:
@shamsimam updated

waiter/project.clj
```diff
@@ -111,11 +111,13 @@
                  [slingshot "0.12.2"]
                  [try-let "1.3.1"
                   :exclusions [org.clojure/clojure]]]
-  :eftest {:report clojure.test/report}
+  :eftest {:report clojure.test/report
+           :thread-count (fn []
+                           (or (some-> (System/getenv "LEIN_TEST_THREADS") Long/valueOf)
```
shamsimam: `thread-count` eventually calls `Executors/newFixedThreadPool` which expect an int argument.

nsinkov:
updated

waiter/project.clj
```diff
@@ -111,11 +111,12 @@
                  [slingshot "0.12.2"]
                  [try-let "1.3.1"
                   :exclusions [org.clojure/clojure]]]
-  :eftest {:report clojure.test/report}
+  :eftest {:report clojure.test/report
+           :thread-count (or (some-> (System/getenv "LEIN_TEST_THREADS") Integer/valueOf)
```
shamsimam: Should we do `parseInt` instead of `valueOf`?
`availableProcessors` also returns an `int`.

waiter/project.clj
```diff
@@ -111,11 +111,12 @@
...
+                           (.availableProcessors (Runtime/getRuntime)))}
   :resource-paths ["resources"]
   :main waiter.main
-  :plugins [[com.holychao/parallel-test "0.3.2"]
-            [lein-exec "0.3.7"]
+  :plugins [[lein-exec "0.3.7"]
             [test2junit "1.2.2"]
             [lein-eftest "0.5.8"]]
```
shamsimam: Nitpick: lexicographic order

PR #923, opened 2019-08-14
adds support for JWT access tokens

shamsimam:

- adds implementation of jwt authenticator
- adds jwt authenticator settings
- adds integration test

Adds support for authentication using JWT access tokens to Waiter services.

### DIFF
--- a/containers/test-apps/jwks-server/resources/jwks.json
+++ b/containers/test-apps/jwks-server/resources/jwks.json
@@ -1,0 +1,90 @@
+{
+  "keys": [
+    {
+      "alg": "RS256",
+      "e": "AQAB",
+      "kid": "7c368fc914ce6cb181fa0d670f63bd5df6db7b25",
+      "kty": "RSA",
+      "n": "vWmir2ZdXeMZkfsg0GTPfQw7CKmDNu50Sc76pndZPNyLf5JeR39JueHIPVXJ",
+      "use": "enc"
+    },
+    {
+      "alg": "RS256",
+      "d": "Eq5xpGnNCivDflJsRQBXHx1hdR1k6Ulwe2JZD50LpXyWPEAeP88vLNO97IjlA7_GQ5sLKMgvfTeXZx9SE-7YwVol2NXOoAJe46sui395IW_GO-pWJ1O0BkTGoVEn2bKVRUCgu-GjBVaYLU6f3l9kJfFNS3E0QbVdxzubSu3Mkqzjkn439X0M_V51gfpRLI9JYanrC4D4qAdGcopV_0ZHHzQlBjudU2QvXt4ehNYTCBr6XCLQUShb1juUO1ZdiYoFaFQT5Tw8bGUl_x_jTj3ccPDVZFD9pIuhLhBOneufuBiB4cS98l2SR_RQyGWSeWjnczT0QU91p1DhOVRuOopznQ",
+      "e": "AQAB",
+      "kid": "wwjcyqculjybrlzo0tzwjjniusfb4p4fakdotbf6",
+      "kty": "RSA",
+      "n": "ofgWCuLjybRlzo0tZWJjNiuSfb4p4fAkd_wWJcyQoTbji9k0l8W26mPddxHmfHQp-Vaw-4qPCJrcS2mJPMEzP1Pt0Bm4d4QlL-yRT-SFd2lZS-pCgNMsD1W_YpRPEwOWvG6b32690r2jZ47soMZo9wGzjb_7OMg0LOL-bSf63kpaSHSXndS5z5rexMdbBYUsLA9e-KXBdQOS-UTo7WTBEMa2R2CapHg665xsmtdVMTBQY4uDZlxvb3qCo5ZwKh9kG4LT6_I5IhlJH7aGhyxXFvUK-DWNmoudF8NAco9_h9iaGNj8q2ethFkMLs91kzk2PAcDTW9gb54h4FRWyuXpoQ",
+      "use": "sig"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "8OSgWqVALOqZyLTLBcVJSgRZ6yxo0Wk_FkNG7uRCV28",
+      "kid": "f8f78b86-af1b-46c5-8849-006020f34d83",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "H5UiUvUNr0OcxjEXpGCfFkYig-63Cs2SV_q6Y_5uDck"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "NIX7-o9XrTLAYNanapB9_fIeOnZUHug2jhwbZ6zs6dE",
+      "kid": "b66fd551-5e49-46e3-8c3b-06b953a16f6b",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "gLYQSUzOu4T0xq12F_RCc51YgSc8w9GZK3WIyYRlH44"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "SXbPPwVHzPYYTwwRG_27RuoTN15Xn8zctwND727B2no",
+      "kid": "3106d31a-87a2-482f-a8dd-b727a043de5a",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "qE4PwqfWP9pQ8iVpPhvQY433hiAn_rzjH-CPTnY_8yQ"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "dx_kwm1Kg47VwP-4DZHBO9bLdGgSu-w20VYwKyqHOw0",
+      "kid": "73711a87-eeb6-42be-b537-3fdeaa4872bb",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "SycUT4WVy3AKTR_FxBjkAa468LocX8QrM480EK2EOoc"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "kdE8RVHsC87AdRSOA9jrJct75r-oZtggCgPQmpZHQzg",
+      "kid": "bfc58a16-8e07-42af-a632-2d3165dbf859",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "nkKQj8rkxfvGFRcXNCeVz236ePxBZMtuxAp_F2PeIvk"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "oyi29vqepPO-ZCQLTnthB6-IWGME_88E3beFPwXK4Xs",
+      "kid": "e787c763-ffe1-405d-bdee-bc50ea0e3aa5",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "nv4W45wfSuWBAKeY9hEJ_SLp1o3d0kdViCs9TpbsEWk"
+    },
+    {
+      "crv": "Ed25519",
+      "d": "sH8QTNekLsfdXA29vipsLHiC8xQHqwteWcFtjM5gexs",
+      "kid": "9bf169b7-d8cc-4f60-82bd-2d060b11653b",
+      "kty": "OKP",
+      "use": "sig",
+      "x": "Tw32CCAlFNn2jz6wwVAQOgABuuA4pyLfaSnwzTEVhoY"
+    },
+    {
+      "crv": "P-256",
+      "kid": "e787c763-ffe1-d8cc-b537-bc50ea0e3aa5",
+      "kty": "EC",
+      "x": "f83OJ3D2xF1Bg8vub9tLe1gHMzV76e8Tus9uPHvRVEU",
+      "y": "x_FEzRu9m36HLN_tue659LNpXW6pCyStikYjKIWI5a0"
+    },
+    {
+      "e": "AQAB",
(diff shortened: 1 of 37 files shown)


shamsimam:
@sradack this is ready for review. No hurry merging it though.

waiter/src/waiter/auth/jwt.clj
```diff
@@ -0,0 +1,275 @@
...
+        (timers/stop timer-context)
+        (counters/inc! (metrics/waiter-counter "core" "jwt" "validation" "failed"))
+        (utils/exception->response th request)))))
+
+(defrecord JwtAuthenticator [token-type issuer keys-cache password subject-key]
+  auth/Authenticator
+  (wrap-auth-handler [_ request-handler]
+    (fn jwt-auth-handler [request]
+      (if (and (not (auth/request-authenticated? request))
+               (-> request :headers (get "authorization") retrieve-bearer-entry))
```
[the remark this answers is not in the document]
shamsimam: Made the JWT handler its own handler and no longer implements the Authenticator interface.

waiter/src/waiter/core.clj
```diff
@@ -573,6 +578,11 @@
    :instance-rpc-chan (pc/fnk [] (async/chan 1024)) ; TODO move to service-chan-maintainer
    :interstitial-state-atom (pc/fnk [] (atom {:initialized? false
                                               :service-id->interstitial-promise {}}))
+   :jwt-authenticator (pc/fnk [[:settings authenticator-config]
+                               passwords]
+                        (let [jwt-config (assoc (:jwt authenticator-config)
+                                           :password (first passwords))]
+                          (jwt/jwt-authenticator jwt-config)))
```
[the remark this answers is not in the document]
shamsimam: Added, we can now set `:jwt :disabled` in the map.

waiter/src/waiter/core.clj
```diff
@@ -767,19 +777,21 @@
...
                                            (auth/wrap-auth-cookie-handler
                                              password
-                                             (fn authenticate-request [request]
-                                               (cond
-                                                 (:skip-authentication request) (do
-                                                                                  (log/info "skipping authentication for request")
-                                                                                  (request-handler request))
-                                                 (auth/request-authenticated? request) (request-handler request)
-                                                 :else (auth-handler request)))))))
+                                             (auth/wrap-auth-handler
```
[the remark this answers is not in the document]
shamsimam: > Could we handle another scheme that also uses Bearer that isn't JWT?

I changed the code to trigger JWT authentication only if the bearer token has three fragments separated by dots.

> Could we handle both?

I'm checking the refactoring needed to support multiple authentication schemes.
If we do that, then it also helps with the above question of supporting multiple authentication schemes.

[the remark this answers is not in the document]
shamsimam: added support for trying other auths if access token validation fails.

shamsimam:
ready for another round of review.

waiter/src/waiter/auth/jwt.clj
```diff
@@ -0,0 +1,332 @@
...
+                (t/millis update-interval-ms)
+                (fn refresh-keys-cache-task []
+                  (refresh-keys-cache http-client http-options jwks-url supported-algorithms keys-cache)))
+   :query-state-fn (fn query-jwt-cache-state []
+                     @keys-cache)})
+
+(def ^:const bearer-prefix "Bearer ")
+
+;; access the private function for validating claims
+(def ^:const validate-claims #'jwt/validate-claims)
```
sradack: Can we avoid using a private function so that we can feel safe about upgrading?

containers/test-apps/jwks-server/resources/jwks.json
```diff
@@ -0,0 +1,90 @@
+{
```
sradack: We should generate these keys dynamically as part of the startup script.

containers/test-apps/jwks-server/src/jwks_server/handler.clj
```diff
@@ -0,0 +1,112 @@
...
+        options {:alg alg :header header}]
+    (jwt/sign payload private-key options)))
+
+(defn process-get-token-request
+  "Retrieves an JWT access token generated using a random EdDSA key."
+  [{:keys [query-string]}]
+  (log/info "query string:" query-string)
+  (let [{:strs [host]} (try
+                         (when-not (str/blank? query-string)
+                           (->> (str/split query-string #"&")
```
sradack: We should use ring's built in param parsing: https://github.com/ring-clojure/ring/blob/master/ring-core/src/ring/middleware/params.clj

waiter/integration/waiter/authentication_test.clj
```diff
@@ -126,3 +128,152 @@
...
+                                    :set-cookie set-cookie
+                                    :target-url target-url})]
+        (assert-response-status response 200)
+        (is (= (retrieve-username) (str body)))
+        (is (= "jwt" (get headers "x-waiter-auth-method")) assertion-message)
+        (is (= (retrieve-username) (get headers "x-waiter-auth-user")) assertion-message)
+        (assert-auth-cookie set-cookie assertion-message))
+      (log/info "JWT authentication is disabled"))))
+
+(deftest ^:parallel ^:integration-fast test-bypass-jwt-authentication-waiter-realm
```
sradack: Should this be named something like `test-fallback-to-alternate-auth-on-invalid-jwt-token`?  Are we testing that authentication falls back to another auth method if we provide an invalid JWT?  Is that what we want?

waiter/integration/waiter/authentication_test.clj
```diff
@@ -126,3 +128,152 @@
...
+            service-id
+            (assert-response-status response 200)
+            (is (= "jwt" (get headers "x-waiter-auth-method")) assertion-message)
+            (is (= (retrieve-username) (get headers "x-waiter-auth-user")) assertion-message)
+            (assert-auth-cookie set-cookie assertion-message))
+          (finally
+            (delete-token-and-assert waiter-url host))))
+      (log/info "JWT authentication is disabled"))))
+
+(deftest ^:parallel ^:integration-fast test-bypass-jwt-authentication-token-realm
```
sradack: It would be great if we could have some integration tests that result in failures, both unauthorized and forbidden.

waiter/src/waiter/auth/jwt.clj
sradack: I'd rather write our own claims validation function than use this private function.

waiter/src/waiter/auth/jwt.clj
```diff
@@ -0,0 +1,332 @@
...
+  [request-handler token-type issuer subject-key supported-algorithms key-id->jwk password request]
+  (let [claims-or-throwable (extract-claims token-type issuer subject-key supported-algorithms key-id->jwk request)]
+    (if (instance? Throwable claims-or-throwable)
+      (if (-> claims-or-throwable ex-data :status (= status-unauthorized))
+        ;; allow downstream processing before deciding on authentication challenge in response
+        (ru/update-response
+          (request-handler request)
+          (fn [{:keys [status] :as response}]
+            (if (and (= status status-unauthorized) (utils/waiter-generated-response? response))
+              ;; add to challenge initiated by Waiter
```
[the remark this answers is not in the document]
sradack: We're only sending challenges when the provided token is unauthorized, but we're falling through to another authentication method?  That feels odd.

I'm not sure what the desired behavior we're going for is.  As far as I see it:

1. JWT users
They want to speak JWT.  They indicate it by always providing the JWT.  If they provide an invalid JWT (say it's expired), they get challenged for a new, valid JWT.  They shouldn't get a challenged for another auth method (which seems to be what we're doing here.  Do you feel/think otherwise?)

2. Non-JWT users
They never provide JWTs, and they never get challenged for JWT. (This seems to be how things currently work in this PR, and this behavior is controlled elsewhere.)

waiter/src/waiter/auth/jwt.clj
```diff
@@ -0,0 +1,332 @@
...
+
+(defn wrap-auth-handler
+  "Wraps the request handler with a handler to trigger JWT access token authentication."
+  [{:keys [issuer keys-cache password subject-key supported-algorithms token-type]} request-handler]
+  (fn jwt-auth-handler [request]
+    (if (and (not (auth/request-authenticated? request))
+             (auth/select-auth-header request access-token?))
+      (authenticate-request request-handler token-type issuer subject-key supported-algorithms
+                            (:key-id->jwk @keys-cache) password request)
+      (request-handler request))))
```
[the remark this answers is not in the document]
sradack: Seems fine.  See my above response.

waiter/src/waiter/util/client_tools.clj
```diff
@@ -1136,3 +1143,15 @@
             (recur))))
       (async/close! body-ch))
     body-ch))
+
```
sradack: Can this be deleted?

waiter/test-files/jwt/jwks.json
```diff
@@ -0,0 +1,90 @@
+{
```
sradack: Can this be deleted?  Can we generate keys dynamically for our unit tests?  How are you generating these keys?

containers/test-apps/jwks-server/resources/jwks.json
sradack: 1. We don't have to answer the question of how these keys are generated.
2. We ensure there's nothing special about the values in this file.

containers/test-apps/jwks-server/src/jwks_server/handler.clj
sradack: URL parameters can be URL encoded, which wasn't being handled.

waiter/integration/waiter/authentication_test.clj
sradack: The integration tests show that our JWT behavior is correct all the way to the client, which is valuable.  My first thought was that we didn't test this because it was "hard", even though we probably both agree it would be useful.

waiter/integration/waiter/authentication_test.clj
sradack: JWT *could* stand on its own, meaning it can work without any other authentication method enabled.  You haven't implemented it to work that way.  That's fine.  I'm confused the way we've set up the challenges however.  See my other comment regarding challenges.

waiter/src/waiter/auth/jwt.clj
sradack: We shouldn't implement hacks to depend on private methods of outside projects.  Please implement the function ourselves or find another way to get that method called via methods that are available to us publicly.  Let's not leave this tech debt for ourselves.

shamsimam: Implemented our own `validate-claims`

containers/test-apps/jwks-server/resources/jwks.json
shamsimam: Since it is not particularly easy to generate these values with our current library, I'd vote to keep it as it is. This is a simple test server (or file used by unit tests in another context) implementation, I'd like to keep it simple and avoid additional features like generating these keys dynamically.

waiter/src/waiter/auth/jwt.clj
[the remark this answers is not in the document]
shamsimam: sradack wrote: 

> I'm confused the way we've set up the challenges however.
> 
> We're only sending challenges when the provided token is unauthorized, but we're falling through to another authentication method? That feels odd.
> I'm not sure what the desired behavior we're going for is. 
> As far as I see it:
> 
> JWT users
> They want to speak JWT. They indicate it by always providing the JWT. If they provide an invalid JWT (say it's expired), they get challenged for a new, valid JWT. They shouldn't get a challenged for another auth method (which seems to be what we're doing here. Do you feel/think otherwise?)
> 
> Non-JWT users
> They never provide JWTs, and they never get challenged for JWT. (This seems to be how things currently work in this PR, and this behavior is controlled elsewhere.)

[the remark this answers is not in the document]
shamsimam: JWT authentication is treated as an optimization.
When a client provides multiple authorization headers, she is asking us to try one of many authentication schemes.
If JWT authentication fails, we can still try the configured authentication of the service (e.g. saml, kerberos, etc.).
If that authentication succeeds, then we can go ahead and process the response.

However, if that authentication fails too, none of the provided authentication schemes have worked.
We have a `401` unauthorized response and a challenge from the configured auth method.
We add in the challenge for JWT auth scheme too into the response.
The client can now decide how to tackle the challenge knowing that none of the provided authorization headers worked.

containers/test-apps/jwks-server/resources/jwks.json
sradack: OK.  It would be helpful for you to document how you generated the keys.  This server must remain Travis-only for as long as private keys are in the repository.

waiter/integration/waiter/authentication_test.clj
shamsimam: Added integration tests for both unauthorized and forbidden responses.

containers/test-apps/jwks-server/resources/jwks.json
shamsimam: Added sources of the stored keys in the README.

shamsimam:
Ready for another round of review.

PR #935, opened 2019-08-23
Self service CORS

nsinkov:

- Add allowed CORS list on tokens

Token owners will have self-service CORS capability

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -281,7 +281,8 @@
                :max-age 3600
 
                ;; :kind :token-parameter uses allowed list that's set on the token
-               :token-parameter {:factory-fn waiter.cors/token-parameter-based-validator}}
+               :token-parameter {:factory-fn waiter.cors/token-parameter-based-validator
+                                 :supports-token-parameter? true}}
 
  :entitlement-config {
                       ;; :kind :simple requires that the requesting user be the same as the run-as-user:
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -278,7 +278,10 @@
                :exposed-headers ["etag"]
 
                ;; The value to use for the access-control-max-age header:
-               :max-age 3600}
+               :max-age 3600
+
+               ;; :kind :token-parameter uses allowed list that's set on the application token
+               :token-parameter {:factory-fn waiter.cors/token-parameter-based-validator}}
 
  :entitlement-config {
                       ;; :kind :simple requires that the requesting user be the same as the run-as-user:
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -280,7 +280,7 @@
                ;; The value to use for the access-control-max-age header:
                :max-age 3600
 
-               ;; :kind :token-parameter uses allowed list that's set on the application token
+               ;; :kind :token-parameter uses allowed list that's set on the token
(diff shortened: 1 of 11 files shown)


waiter/config-full.edn
```diff
@@ -278,7 +278,10 @@
                :exposed-headers ["etag"]
 
                ;; The value to use for the access-control-max-age header:
-               :max-age 3600}
+               :max-age 3600
+
+               ;; :kind :token-parameter uses allowed list that's set on the application token
```
shamsimam:
```suggestion
               ;; :kind :token-parameter uses allowed list that's set on the token
```

waiter/src/waiter/cors.clj
```diff
@@ -18,7 +18,8 @@
             [metrics.counters :as counters]
             [waiter.metrics :as metrics]
             [waiter.util.ring-utils :as ru]
-            [waiter.util.utils :as utils])
+            [waiter.util.utils :as utils]
+            [waiter.service-description :as sd])
```
shamsimam: lexicographic order

waiter/src/waiter/cors.clj
```diff
@@ -44,16 +45,16 @@
   "Preflight request handling middleware.
   This middleware needs to precede any authentication middleware since CORS preflight
   requests do not support authentication."
-  [handler cors-validator max-age]
-  (fn wrap-cors-preflight-fn [request]
+  [handler cors-validator max-age kv-store token-defaults waiter-hostnames]
+  (fn wrap-cors-preflight-fn [{:keys [headers] :as request}]
     (if (preflight-request? request)
```
shamsimam: We should populate `access-control-allow-methods` from the configired `methods` on the token `allowed-cors`?

waiter/src/waiter/cors.clj
```diff
@@ -147,3 +148,49 @@
...
+          target-scheme (name (utils/request->scheme request))]
+      (first (filter #(allowed-cors-rule-matches? origin-scheme origin-no-host target-scheme uri request-method %)
+                     (map-indexed vector allowed-cors))))))
+
+(defn token-parameter-based-validator
+  "Factory function for TokenParametersCorsValidator. Validates using token allowed-cors parameter.
+   Same logic for preflight and regular requests"
+  [config]
+  (let [allow-cors?
+        (fn [{:keys [headers] :as request}]
```
shamsimam: We should name this anonymous function

waiter/src/waiter/cors.clj
```diff
@@ -147,3 +148,49 @@
...
+  CorsValidator
+  (preflight-allowed? [_ request] (allow-cors? request))
+  (request-allowed? [_ request] (allow-cors? request)))
+
+(defn- allowed-cors-rule-matches?
+  "Checks if an allowed CORS rule matches"
+  [origin-scheme origin-no-host target-scheme path method
+   [_ {:strs [origin-regex origin-schemes target-path-regex target-schemes methods] :as xxx}]]
+  (cond
+    (and methods (not (.contains methods (str/upper-case (name method))))) false
```
shamsimam: `(not (.contains methods (str/upper-case (name method))))` =>
`(not-any? #(= % (str/upper-case (name method))) methods)`

shamsimam: Same for the expressions below

PR #941, opened 2019-09-01
incorporates feedback into work-stealing using moving averages of successful offers

shamsimam:

- introduces moving average implementations
- incorporates feedback into work-stealing using moving averages of successful offers

Work-stealing is an optimization but can trigger many failed offers when requests are short-lived. Incorporating feedback of successful offers and toning down offers accordingly allows us to reduce the number of failed offers at the cost of increased latency for successful offers.


waiter/src/waiter/util/moving_average.clj
```diff
@@ -0,0 +1,66 @@
...
+;;
+(ns waiter.util.moving-average)
+
+(defprotocol MovingAverage
+  "Data structure that computes moving averages."
+  (insert-value [_ value] "Update the moving average with `value` and return a new MovingAverage instance.")
+  (get-average [_] "Returns the moving average represented by this instance."))
+
+;; Computes the simple moving average over a defined window size.
+(defrecord SlidingWindowMovingAverage [window-size data data-sum]
```
shamsimam: Did not use [`com.codahale.metrics.SlidingWindowReservoir`](https://github.com/dropwizard/metrics/blob/3680c67b2b0db0346a2273fcbcaa80a7d29aacd8/metrics-core/src/main/java/com/codahale/metrics/SlidingWindowReservoir.java#L9) as computing the average requires [copying and sorting the data array](https://github.com/dropwizard/metrics/blob/3680c67b2b0db0346a2273fcbcaa80a7d29aacd8/metrics-core/src/main/java/com/codahale/metrics/UniformSnapshot.java#L24).

waiter/config-full.edn
```diff
@@ -752,6 +752,16 @@
  ; ---------- Load Balancing ----------
 
  :work-stealing {
+                 ;; Work-stealing maintains a simple moving average of successful offers which is
+                 ;; used as feedback to determine the probability of making subsequent offers.
+                 :moving-average {;; the initial value of the moving average
```
sradack: If a router accepts help, even after a period of not accepting help, why wouldn't we always (p=1) offer help the next time we think the router needs help? Does an exponential backoff model desired behavior better?  An exponential backoff model might improve how quickly react to situations where help is actually needed while reducing overall offers when help is usually rejected.

waiter/config-full.edn
```diff
@@ -752,6 +752,16 @@
...
+                                  :initial-value 1.00
+                                  ;; the maximum value of the moving average
+                                  :maximum-value 1.00
```
sradack: Can we use `max` and `min`, in line with how we define all other maximums and minimums in our config?

waiter/config-full.edn
sradack: If another router only accepts 5% of workstealing offers, and we end up only making offers 5% of the time, are we work stealing only 0.25% of the time (5%*5%)?  Does that outcome achieve our goal?

sradack: Can we name `moving-average`  something indicative of its purpose (e.g. `accept-rate`)?

shamsimam: > An exponential backoff model might improve how quickly react to situations where help is actually needed while reducing overall offers when help is usually rejected.

I'll add support for exponential and then we can configure the model we want accordingly.

> If another router only accepts 5% of work-stealing offers, and we end up only making offers 5% of the time, are we work-stealing only 0.25% of the time (5%*5%)?

The idea is to try and reach a state where offers are not rejected. If a router accepts only 5% offers, we will send it an offer with a 5% chance. The offer that gets sent will get accepted with higher probability (because we didn't make the other 19 offers that are expected to get rejected). That will bump up the average. When we start allocating (not necessarily sent due to the local probability filter) offers at a rate higher than 5% the average will get pulled down to the 5% acceptance rate of the remote router.

waiter/config-full.edn
shamsimam: Will rename to `max` and `min`.

waiter/config-full.edn
sradack: We're not realistically going to chose a model other than exponential.  Adding another model is going to leave code around that we're not using.  

If we like an exponential backoff, then the following configuration proposed parameters intuitive:

base-backoff-ms
backoff-factor 
max-backoff-ms

(We currently have blacklisting defined in similar terms.)

The following configuration parameters are not:
:max-value (This is never not 1.)
:weight-value (There's no reasonable value for this that results in a guaranteed followup help offer following a successful offer.)
:min-value (Setting this to >0 means we still have a chance to send a help offer right after we were rejected, which we don't want.)
:initial-value (This is never not 1.)
:window-size (Also set to something fairly arbitrary.)

It's confusing to try to convey the behavior we want indirectly through these parameters.

shamsimam: I created a [simple simulation](https://gist.github.com/shamsimam/481073999c6cbb435bc4f1344a9a7a92) with two modes:
- stable: where if a request needs help if stays until it is helped by work-stealing
- whimsical: where requests can get processed by another instance before work-stealing help arrives

Feedback is welcome on the simulation.

From the [output](https://gist.github.com/shamsimam/481073999c6cbb435bc4f1344a9a7a92#file-work-stealing-simulation-L168) of the current simulation:
- latency (`request-delay-average`) is lowest for current implementation, but at the cost of many failed offers
- simple moving average with minimum of `0.10` strikes a good balance of latency and failed requests for both cases.
- exponential moving average reports fewer failed requests at cost of higher latency. In the  `whimsical` case, almost no work-stealing help is provided.

shamsimam: Added [exponential backoff](https://gist.github.com/shamsimam/481073999c6cbb435bc4f1344a9a7a92#gistcomment-3017425). We can obtain similar results (offer success ratio and reduced total offers) as simple average but with additional implementation difficulty and configuration to support exponential backoff.

shamsimam: @sradack ⬆️

PR #942, opened 2019-09-03
wip: limits the maximum number of work-stealing offers in-flight

shamsimam:

- limits the maximum number of work-stealing offers in-flight per service per router

The number of work-stealing offers can grow to become quite large and affect the number of ports available on the hosts. By restricting the number of outstanding work-stealing offers per service per router, we can bound the number of ports assigned to these offers as a function of the number of live services.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -753,7 +753,7 @@
 
  :work-stealing {
                  ;; The maximum number of outstanding work-stealing offers per service per router:
-                 :max-work-stealing-offers-per-service 50
+                 :max-work-stealing-offers-per-service 20
 
                  ;; The interval (milliseconds) on which Waiter makes work-stealing offers:
                  :offer-help-interval-ms 100
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -752,6 +752,9 @@
  ; ---------- Load Balancing ----------
 
  :work-stealing {
+                 ;; The maximum number of outstanding work-stealing offers per service per router:
+                 :max-work-stealing-offers-per-service 50
+
                  ;; The interval (milliseconds) on which Waiter makes work-stealing offers:
                  :offer-help-interval-ms 100
 
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -941,14 +941,16 @@
                              (scheduler/validate-service scheduler service-id)
                              (service/start-new-service
                                scheduler descriptor start-service-cache scheduler-interactions-thread-pool)))
-   :start-work-stealing-balancer-fn (pc/fnk [[:settings [:work-stealing offer-help-interval-ms reserve-timeout-ms]]
+   :start-work-stealing-balancer-fn (pc/fnk [[:settings [:work-stealing max-work-stealing-offers-per-service
+                                                         offer-help-interval-ms reserve-timeout-ms]]
                                              [:state instance-rpc-chan router-id]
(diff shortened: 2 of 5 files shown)


waiter/config-full.edn
```diff
@@ -752,6 +752,9 @@
  ; ---------- Load Balancing ----------
 
  :work-stealing {
+                 ;; The maximum number of outstanding work-stealing offers per service per router:
+                 :max-work-stealing-offers-per-service 50
```
DaoWen:
```suggestion
                 :max-work-stealing-offers-per-service 20
```

DaoWen: I think it's good to keep the `config-full.edn` numbers the same as the defaults in `settings.clj`.

waiter/config-full.edn
```diff
@@ -752,6 +752,9 @@
  ; ---------- Load Balancing ----------
 
  :work-stealing {
+                 ;; The maximum number of outstanding work-stealing offers per service per router:
+                 :max-work-stealing-offers-per-service 20
```
sradack: It seems undesirable to not offer all available capacity to a lone router handling all the traffic, especially given that we have other ways of increasing the number of ports available to us.

sradack: Does this setting belong in the service description, given the wide variety of `concurrency-level`s among services and therefore the wide variety of optimal values for this setting?

shamsimam: We implemented work-stealing as an optimization. This configuration allows us to disable work-stealing (setting to `0`). We can also choose to make this value arbitrarily high if we have enough ports to play with.

sradack: Agreed that it can be seen as useful setting as a value for 0 or infinity.  Unclear when it's set to anything in between.

shamsimam: If we know we have 200 services running, and no more than 50 (for example) in-flight work-stealing offers per service, no more than 10K ports are in use the work-stealing mechanism. This is valuable in restricting the network resources used by the work-stealing mechanism.

sradack: The problem is we don't know we have 200 services running.

shamsimam: We can from other components. I can also change this to a global value where the limits are imposed on all work-stealing offers in flight.

sradack: Yes we could.  We'd have to think about fair allocation of work steals across services, etc.  It's an easy step from here.

Realistically,
1. We're not going to turn work stealing off.  I'd rather have a :disabled flag for this anyway.
2. We can increase the number of ports available to us: https://making.pusher.com/ephemeral-port-exhaustion-and-how-to-avoid-it/  We need to do this anyway to create headroom for more services.
3. The work going on in #942 will hopefully dramatically reduce the number of work steals.

I don't see a need for this PR as it currently stands.  

Should we do the work now to make this a global limit?  It would be a nice to have, but we have many nice to haves that we probably want more.

sradack: Maybe I'm missing some context on a specific instance where a service was sending out hundreds of work steals (that wasn't fixed by #942) and we simply don't want to provide that high of a level of service.  Let me know.

shamsimam: If we can get an increase in the number of ports then this PR is unnecessary.
#942 will help, but not sure it alone will help unless we can increase port numbers.
As you said global limit is nice to have, but we can add it later.

shamsimam: Created https://github.com/twosigma/waiter/pull/945 to enforce global limits on work-stealing offers.

PR #943, opened 2019-09-04
Scale down unhealthy expired instances more eagerly

DaoWen:

Change the `expired-instances` count passed to the `scale-service` function to exclude one expired-unhealthy instance for each unexpired-healthy instance.

If healthy instances are available to accept requests, we should not
alter target instance count results to keep those instances around.

### DIFF
--- a/waiter/src/waiter/scaling.clj
+++ b/waiter/src/waiter/scaling.clj
@@ -424,7 +424,12 @@
       (fn [service-id]
         (let [outstanding-requests (or (service-id->outstanding-requests service-id) 0)
               {:keys [healthy-instances expired-healthy-instances expired-unhealthy-instances]} (service-id->router-state service-id)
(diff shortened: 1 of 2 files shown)


DaoWen:
@shamsimam - Added all of the unit tests you suggested.

DaoWen:
Those new tests don't pass... Could you clarify whether or not you expected them to pass?

shamsimam:
> Those new tests don't pass... Could you clarify whether or not you expected them to pass?

I expect them to pass. 
I was seeking your opinion on whether you agree with the assertions.

DaoWen:
> I expect them to pass.

I think you mean that you know they don't currently pass, but you think the implementation should be such that they pass. Is that right?

shamsimam:
> > I expect them to pass.
> 
> I think you mean that you know they don't currently pass, but you think the implementation should be such that they pass. Is that right?

Yes.

DaoWen:
Closing this for now. We still haven't quite figured out the right semantics for this.

PR #949, opened 2019-09-10
adds support for references in the descriptor

shamsimam:

- adds support for references in the descriptor
- references are now used to determine if a service is stale

We need to allow custom builders to add references to services and also contribute to marking services as stale rather than relying only on tokens going stale.

### DIFF
--- a/waiter/docs/service-references.md
+++ b/waiter/docs/service-references.md
@@ -15,14 +15,17 @@ By default, Waiter relies on the service idle timeout to GC services after perio
 However, service reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
 
 When a service description is constructed from a request, the service references are also updated.
-These references are available as the `:references` key in the descriptor and are annotated by a `:type` parameter.
-The service GC process checks these references by type and marks a service as a candidate for eager GC
+These references are available as the `:reference` key in the descriptor.
+The :refrence enrty is a map where the keys represent a `:type` parameter.
+The service GC process checks individual references by type and marks a service as a candidate for eager GC
   if _all_ references used to access the service are stale.
-This staleness check is performed using the functions returned from `retrieve-reference-type->stale-fn` of the builder.
+An individual reference is stale if any of its value entries is stale.
+This staleness check on the value is performed using the functions returned from
+  `retrieve-reference-type->stale-fn` of the builder and invoking the corresponding 'type' function on the value.
 
 The default implementation of the `ServiceDescriptionBuilder` returns
   the following functions for the different reference types:
-  1. all services that can be directly accessed never go stale;
+  1. any services that can be directly accessed never goes stale;
   1. services accessed via tokens go stale if all tokens used to access the service have been updated.
 
 Custom builder implementations can add additional references to services and
--- a/waiter/docs/service-references.md
+++ b/waiter/docs/service-references.md
@@ -12,21 +12,35 @@ It is possible for the service to be reachable from various combinations of head
 ## Service GC
 
 By default, Waiter relies on the service idle timeout to GC services after periods of inactivity (not receiving requests).
-However, service reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
+However, services reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
 
 When a service description is constructed from a request, the service references are also updated.
-These references are available as the `:reference` key in the descriptor.
-The :refrence enrty is a map where the keys represent a `:type` parameter.
+These references are available as the `:reference-type->entry` key in the descriptor.
+The `reference-type->entry` is a map where the keys represent a `:type` parameter, e.g.
+a service created with `x-waiter-token: foo` header on a request will have the value:
+```
+  {:token {:sources [{"token" "foo" "version" "v1"}]}}
+```
+
+If the same service can be accessed by another token, we end up building multiple
+  `reference-type->entry` maps as references that refer to the service.
 The service GC process checks individual references by type and marks a service as a candidate for eager GC
   if _all_ references used to access the service are stale.
-An individual reference is stale if any of its value entries is stale.
+An individual reference, i.e. the `reference-type->entry` map, is stale if any of its value entries is stale.
 This staleness check on the value is performed using the functions returned from
   `retrieve-reference-type->stale-fn` of the builder and invoking the corresponding 'type' function on the value.
 
-The default implementation of the `ServiceDescriptionBuilder` returns
-  the following functions for the different reference types:
-  1. any services that can be directly accessed never goes stale;
-  1. services accessed via tokens go stale if all tokens used to access the service have been updated.
+The default implementation of the `ServiceDescriptionBuilder` returns a map with a single entry for `:token`.
+The provided `:token` staleness function deems services accessed via tokens to be stale if all tokens
+  used to access the service have been updated.
 
-Custom builder implementations can add additional references to services and
+Custom builder implementations can add additional reference types to services and
   need to provide appropriate staleness check functions for each reference type.
+E.g. a hypothetical implementation which treats the image parameter as docker images can
+  have the following entry for a service:
+```
+  {:image {:name "twosigma/kitchen" :build "20191001"}
+   :token {:sources [{"token" "foo" "version" "v1"}]}}
+```
+The `retrieve-reference-type->stale-fn` must then provide an implementation for a function that
+  can check staleness of `:image` reference types.
--- a/waiter/docs/service-references.md
+++ b/waiter/docs/service-references.md
@@ -1,10 +1,10 @@
 # Service References
 
-To process a http request, Waiter needs to resolve a request to a service description.
+To process an HTTP request, Waiter needs to resolve a request to a service description.
 This service description tells Waiter how to locate the service instance to use to process the current request.
 
 Waiter uses request headers to construct the service description using the `ServiceDescriptionBuilder`.
-It is possible for the service to be reachable from various combinations of headers:
+It is possible for the _same_ service to be reachable from various combinations of headers:
   1. token - single or multiple
   1. on-the-fly services without tokens
   1. on-the-fly services that extend tokens, e.g. run-as-requester
@@ -14,14 +14,21 @@ It is possible for the service to be reachable from various combinations of head
 By default, Waiter relies on the service idle timeout to GC services after periods of inactivity (not receiving requests).
 However, services reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
 
-When a service description is constructed from a request, the service references are also updated.
-These references are available as the `:reference-type->entry` key in the descriptor.
-The `reference-type->entry` is a map where the keys represent a `:type` parameter.
+**Note**: A service known to be referenced only via on-the-fly headers never goes stale.
+
+When a service description is constructed from a request, the set of references for a service are also updated.
+Individual references are available in the `:reference-type->entry` key in the descriptor,
+  the descriptor being built per request.
+It is possible to reference the same service using different combinations of on-the-fly headers and tokens.
+As such, there is a set of references (the union of all `:reference-type->entry` from the descriptor) that
+  can be used to reference a service.
 
 ### Reference (`:reference-type->entry` in the descriptor) examples
 
-- A service created with on-the-fly header without tokens will have an empty map
-  as the following value for `:reference-type->entry` in the descriptor:
+The `reference-type->entry` is a map where the keys represent a `:type` parameter used during staleness checks.
+
+- A service created with on-the-fly header without tokens never goes stale.
+  It has an empty map as the value for `:reference-type->entry` in the descriptor:
 ```
   {}
 ```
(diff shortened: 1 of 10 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -562,12 +579,21 @@
                                          user-service-description)))
           service-id (service-description->service-id service-id-prefix core-service-description)
           service-description (default-and-override core-service-description metric-group-mappings
-                                                    kv-store defaults service-id)]
+                                                    kv-store defaults service-id)
+          references (cond-> references
+                       (seq source-tokens) (conj {:sources source-tokens :type :token})
+                       ;; if no references were used to create the service, it is directly accessible
+                       (and (empty? references) (empty? source-tokens)) (conj {:type :direct-access}))]
```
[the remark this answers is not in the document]
shamsimam: Changed.

waiter/src/waiter/service_description.clj
```diff
@@ -1068,37 +1093,82 @@
...
+
+  (defn store-references!
+    "Stores the reference entries in the key-value store against a service."
+    [synchronize-fn kv-store service-id references]
+    (when (seq references)
+      (let [reference-lock-prefix "REFERENCES-LOCK-"
+            bucket (-> service-id hash int (Math/abs) (mod 16))
+            reference-lock (str reference-lock-prefix bucket)]
+        (meters/mark! (metrics/waiter-meter "core" "reference" "store" "all"))
+        (meters/mark! (metrics/waiter-meter "core" "reference" "store" (str "bucket-" bucket)))
```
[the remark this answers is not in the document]
shamsimam: Added the timer to `synchronize-fn`.

waiter/src/waiter/service_description.clj
```diff
@@ -1068,37 +1093,82 @@
...
+        (synchronize-fn
```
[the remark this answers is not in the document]
shamsimam: Good catch! Added the guard and comments.

shamsimam:
@sradack ready for another round of review

waiter/docs/service-references.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Service GC
+
+By default, Waiter relies on the service idle timeout to GC services after periods of inactivity (not receiving requests).
+However, service reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
+
+When a service description is constructed from a request, the service references are also updated.
+These references are available as the `:references` key in the descriptor and are annotated by a `:type` parameter.
+The service GC process checks these references by type and marks a service as a candidate for eager GC
+  if _all_ references used to access the service are stale.
```
sradack: Running service two references: token and image.  Image does stale.  New service goes up for new image.  Are we not GC'ing the old service, since it's token is not stale?

shamsimam: In the proposed example: the token reference entry will have additional metadata that it introduced an image and the token staleness check function will have to account for it.

shamsimam:
Rebasing changes to make it easier to merge.

shamsimam:
@sradack this is ready for another round of review.

waiter/src/waiter/service_description.clj
```diff
@@ -1068,37 +1087,85 @@
...
+    (let [keys (service-id->key service-id)]
+      (kv/fetch kv-store keys :refresh refresh)))
+
+  (defn store-reference!
+    "Stores the reference entries in the key-value store against a service."
+    [synchronize-fn kv-store service-id reference]
+    (when (-> (service-id->references kv-store service-id)
+            (or #{})
+            (contains? reference)
+            (not))
```
sradack: Can change this to a `when-not`?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,32 @@
...
+Waiter uses request headers to construct the service description using the `ServiceDescriptionBuilder`.
+It is possible for the service to be reachable from various combinations of headers:
+  1. token - single or multiple
+  1. on-the-fly services without tokens
+  1. on-the-fly services that extend tokens, e.g. run-as-requester
+
+## Service GC
+
+By default, Waiter relies on the service idle timeout to GC services after periods of inactivity (not receiving requests).
+However, service reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
```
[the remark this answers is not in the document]
shamsimam: Done.

waiter/docs/service-references.md
```diff
@@ -0,0 +1,32 @@
...
+
+When a service description is constructed from a request, the service references are also updated.
+These references are available as the `:reference` key in the descriptor.
```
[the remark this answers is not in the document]
shamsimam: Changed to `:reference-type->entry`

waiter/docs/service-references.md
```diff
@@ -0,0 +1,32 @@
...
+The :refrence enrty is a map where the keys represent a `:type` parameter.
```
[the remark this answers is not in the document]
shamsimam: Fixed typos, added example.

waiter/docs/service-references.md
```diff
@@ -0,0 +1,32 @@
...
+The service GC process checks individual references by type and marks a service as a candidate for eager GC
+  if _all_ references used to access the service are stale.
+An individual reference is stale if any of its value entries is stale.
+This staleness check on the value is performed using the functions returned from
+  `retrieve-reference-type->stale-fn` of the builder and invoking the corresponding 'type' function on the value.
+
+The default implementation of the `ServiceDescriptionBuilder` returns
+  the following functions for the different reference types:
```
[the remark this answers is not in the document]
shamsimam: Changed description to clarify it returns a map.

waiter/src/waiter/core.clj
```diff
@@ -747,8 +747,13 @@
    :synchronize-fn (pc/fnk [[:settings [:zookeeper base-path mutex-timeout-ms]]
                             curator]
                      (fn synchronize-fn [path f]
-                       (let [lock-path (str base-path "/" path)]
-                         (curator/synchronize curator lock-path mutex-timeout-ms f))))})
+                       (let [lock-path (str base-path "/" path)
+                             lock-name (apply str (filter #(Character/isLetterOrDigit ^char %) (str path)))]
```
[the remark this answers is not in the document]
shamsimam: Done, caller provides correct path.

waiter/src/waiter/service_description.clj
shamsimam: Changed.

waiter/docs/service-references.md
```diff
@@ -0,0 +1,46 @@
...
+
+By default, Waiter relies on the service idle timeout to GC services after periods of inactivity (not receiving requests).
+However, services reachable only via references, e.g. tokens, can be GC-ed eagerly if the reference has been updated.
+
+When a service description is constructed from a request, the service references are also updated.
+These references are available as the `:reference-type->entry` key in the descriptor.
+The `reference-type->entry` is a map where the keys represent a `:type` parameter, e.g.
+a service created with `x-waiter-token: foo` header on a request will have the value:
+```
+  {:token {:sources [{"token" "foo" "version" "v1"}]}}
```
sradack: Why are we using strings for the "`token"` and `"version"` keys?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+If the same service can be accessed by another token, we end up building multiple
+  `reference-type->entry` maps as references that refer to the service.
+The service GC process checks individual references by type and marks a service as a candidate for eager GC
+  if _all_ references used to access the service are stale.
+An individual reference, i.e. the `reference-type->entry` map, is stale if any of its value entries is stale.
```
sradack: I don't follow this sentence.  The `reference-type->entry` map may contain multiple references, right?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,46 @@
...
+This staleness check on the value is performed using the functions returned from
+  `retrieve-reference-type->stale-fn` of the builder and invoking the corresponding 'type' function on the value.
+
+The default implementation of the `ServiceDescriptionBuilder` returns a map with a single entry for `:token`.
```
sradack: from the `retrieve-reference-type->stale-fn`?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,46 @@
...
+The provided `:token` staleness function deems services accessed via tokens to be stale if all tokens
+  used to access the service have been updated.
+
+Custom builder implementations can add additional reference types to services and
+  need to provide appropriate staleness check functions for each reference type.
+E.g. a hypothetical implementation which treats the image parameter as docker images can
+  have the following entry for a service:
+```
+  {:image {:name "twosigma/kitchen" :build "20191001"}
```
sradack: Couldn't we have multiple image labels resolving to the same service?

waiter/docs/service-references.md
sradack: How are multiple tokens handled?  Is that modeled differently from various tokens resolving to the same service?

sradack: How does this map change if the same service was to be resolved directly without a token?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,46 @@
...
+   :token {:sources [{"token" "foo" "version" "v1"}]}}
```
sradack: Could the value for `:token` be modeled as an array without being wrapped inside a map?

waiter/docs/service-references.md
shamsimam: Added.

waiter/docs/service-references.md
```diff
@@ -0,0 +1,78 @@
...
+- A service created with on-the-fly header without tokens will have an empty map
+  as the following value for `:reference-type->entry` in the descriptor:
+```
+  {}
+```
+
+- A service created with `x-waiter-token: foo` header on a request will have the
+  following value for `:reference-type->entry` in the descriptor:
+```
+  {:token {:sources [{:token "foo" :version "v0"}]}}
```
sradack: If this service is later referenced entirely on the fly, how would this map change?

waiter/docs/service-references.md
```diff
@@ -0,0 +1,78 @@
...
+  {:token {:sources [{:token "bar" :version "v1"}
+                     {:token "baz" :version "v2"}]}}
+```
+
+If the same service can be accessed by another token, we end up building multiple
+  `reference-type->entry` maps as references that refer to the service.
+E.g. if all the above requests mapped to the same service, that service would have
+  the following references:
+```
+  #{{}
```
sradack: Is `:reference-type->entry` a set or a map?  In previous examples it was a map, but now it's a set.

waiter/docs/service-references.md
shamsimam: `:reference-type->entry` is computed per request. In the on-the-fly scenario, `:reference-type->entry` would be an empty map `{}`.

waiter/docs/service-references.md
shamsimam: `:reference-type->entry` is a map computed per request. Over the lifetime of the services individual `:reference-type->entry` are aggregated into a set to maintain all references that have been used to access the service.

waiter/docs/service-references.md
shamsimam: The references for the service would be a set with two entries: `{}` and `{:token {:sources [{:token "foo" :version "v0"}]}}`.

waiter/docs/service-references.md
sradack: Let's use the opportunity of my confusion to improve the docs and make this more explicit.

shamsimam: Documentation has been updated.

shamsimam:

waiter/docs/service-references.md
sradack: This is much better now.  Thank you for the updates.

waiter/src/waiter/core.clj
```diff
@@ -949,14 +955,19 @@
...
+                               (let [context {:token->token-hash token->token-hash}
+                                     reference-type->stale-fn (sd/retrieve-reference-type->stale-fn service-description-builder context)]
+                                 (fn service-id->idle-timeout [service-id]
+                                   (sd/service-id->idle-timeout
+                                     service-id->service-description-fn service-id->references-fn token->token-metadata
+                                     reference-type->stale-fn token-defaults service-id))))
    :service-id->password-fn (pc/fnk [[:scheduler service-id->password-fn*]]
                               service-id->password-fn*)
+   :service-id->references-fn (pc/fnk [[:state kv-store]]
+                                (partial sd/service-id->references kv-store))
```
sradack: Can we avoid `partial`?  It's quite slow and this is in the critical path.

sradack: I was curious about this and dug a bit further.

https://stuartsierra.com/2015/06/01/clojure-donts-optional-arguments-with-varargs

Were you aware that calling varargs functions always allocates a new sequence?  

https://gist.github.com/halgari/5d7780e6f068ae217bb2

We use these a lot and I wonder if we can improve performance and reduce GC pressure by avoiding these.

sradack: Separate PR though

PR #950, opened 2019-09-10
moves sanitize-history to utils

shamsimam:

- moves sanitize-history to utils

Refactoring of utility method to `utils`.

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -610,5 +610,5 @@
 (defn sanitize-history
   "Limits the history length stored in the nested map."
   [data history-length history-key]
-  (dissoc-in token-data (repeat history-length history-key)))
+  (dissoc-in data (repeat history-length history-key)))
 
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -606,3 +606,9 @@
     (pos? scale-amount) :scale-up
     (neg? scale-amount) :scale-down
     :else :stable))
+
+(defn sanitize-history
+  "Limits the history length stored in the nested map."
+  [token-data history-length history-key]
```
nsinkov:
```suggestion
  [data history-length history-key]
```

waiter/src/waiter/util/utils.clj
```diff
@@ -606,3 +606,9 @@
...
+  (dissoc-in token-data (repeat history-length history-key)))
```
nsinkov:
```suggestion
  (dissoc-in data (repeat history-length history-key)))
```

PR #955, opened 2019-09-11
Update kv.clj

nsinkov:

- fix metric name


waiter/src/waiter/kv.clj
```diff
@@ -144,7 +144,7 @@
     (validate-zk-key key)
     (meters/mark! (metrics/waiter-meter "core" "kv-zk" "store"))
     (timers/start-stop-time!
-      (metrics/waiter-timer "core" "kv-zk" "retrieve")
+      (metrics/waiter-timer "core" "kv-zk" "write")
```
shamsimam:
```suggestion
      (metrics/waiter-timer "core" "kv-zk" "store")
```

PR #956, opened 2019-09-11
validates health-check-port-index even if ports is not explicitly provided

shamsimam:

- validates health-check-port-index even if ports is not explicitly provided

Disallow invalid configurations from being accepted and run as services.


shamsimam:
Closing.

This is working as designed to allow on-the-fly to specify the a valid `ports` value.

PR #960, opened 2019-09-13
adds length validation for parameters

shamsimam:

- adds length validation for cmd, cmd-type, health-check-url, image, namespace, owner, permitted-user, run-as-user, scheduler, version

Prevents users from entering large data into service descriptions.


shamsimam:
We can achieve the same thing with constraints which is more dynamic.

PR #968, opened 2019-09-24
excludes command parameter from token

shamsimam:

- excludes command parameter from token

Avoid command in token tests unless needed.


dposada:
Will merge when green

PR #969, opened 2019-09-25
makes JWT auth per request opt-in by tokens

shamsimam:

- introduces function to guard JWT auth per request

We wish to decide on a per request basis whether or not to perform JWT authentication.

### DIFF
--- a/waiter/integration/waiter/authentication_test.clj
+++ b/waiter/integration/waiter/authentication_test.clj
@@ -170,12 +170,13 @@
                                     :target-url target-url})]
         (assert-response-status response 200)
         (is (= (retrieve-username) (str body)))
-        (is (= "jwt" (get headers "x-waiter-auth-method")) assertion-message)
+        (is (not= "jwt" (get headers "x-waiter-auth-method")) assertion-message)
         (is (= (retrieve-username) (get headers "x-waiter-auth-user")) assertion-message)
         (assert-auth-cookie set-cookie assertion-message))
       (log/info "JWT authentication is disabled"))))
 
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/auth/jwt.clj
```diff
@@ -325,6 +325,7 @@
   [{:keys [issuer keys-cache password subject-key supported-algorithms token-type]} request-handler]
   (fn jwt-auth-handler [request]
     (if (and (not (auth/request-authenticated? request))
+             (= "true" (get-in request [:waiter-discovery :service-parameter-template "env" "USE_BEARER_AUTH"]))
```
shamsimam: I didn't make this configurable as it is a small change this way. In any case, the tests end up using this environment variable.

waiter/integration/waiter/authentication_test.clj
```diff
@@ -311,7 +315,9 @@
     (if (jwt-auth-enabled? waiter-url)
       (let [waiter-host (-> waiter-url sanitize-waiter-url utils/authority->host)
             host (create-token-name waiter-url (rand-name))
-            service-parameters (assoc (kitchen-params) :name (rand-name))
+            service-parameters (assoc (kitchen-params)
+                                 :env {"USE_BEARER_AUTH" "true"}
```
sradack: Can we use WAITER_BEARER_AUTH?

shamsimam: `WAITER_*` environment variable names are reserved :(

https://github.com/twosigma/waiter/blob/0e6b8f406ae41076691e47fc2c06060d7d4383e0/waiter/src/waiter/service_description.clj#L41-L46

PR #971, opened 2019-09-27
adds waiter-api flag to request log

shamsimam:

- adds waiter-api flag to request log

We would like to easily distinguish api calls from backend calls.

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -153,19 +153,22 @@
 (defn ring-handler-factory
   "Creates the handler for processing http requests."
   [waiter-request?-fn {:keys [process-request-fn] :as handlers}]
-  (fn http-handler [{:keys [uri] :as request}]
-    (if-not (waiter-request?-fn request)
-      (do
-        (counters/inc! (metrics/waiter-counter "requests" "service-request"))
-        (process-request-fn request))
-      (let [{:keys [handler route-params]} (routes-mapper request)
-            request (assoc request :route-params (or route-params {}))
-            handler-fn (get handlers handler process-request-fn)]
-        (when (and (not= handler :process-request-fn) (= handler-fn process-request-fn))
-          (log/warn "using default handler as no mapping found for" handler "at uri" uri))
-        (when handler
-          (counters/inc! (metrics/waiter-counter "requests" (name handler))))
-        (handler-fn request)))))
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/request_log.clj
```diff
@@ -75,7 +76,8 @@
       principal (assoc :principal principal)
       protocol (assoc :backend-protocol protocol)
       server (assoc :server server)
-      handle-request-latency-ns (assoc :handle-request-latency-ns handle-request-latency-ns))))
+      handle-request-latency-ns (assoc :handle-request-latency-ns handle-request-latency-ns)
+      (some? waiter-api-call?) (assoc :waiter-api waiter-api-call?))))
```
sradack: Why is `some?` necessary?

waiter/src/waiter/core.clj
```diff
@@ -187,6 +187,15 @@
                                 (rr/header "server" server-name)))]
       (ru/update-response response add-server-header))))
 
+(defn attach-waiter-api-middleware
+  "Attaches a boolean value for :waiter-api-call? to the response."
+  [handler waiter-request?-fn]
+  (fn attach-waiter-api-middleware-fn [request]
+    (ru/update-response
+      (handler request)
```
sradack: Can we add it the request map and use that instead of calling `waiter-request?-fn` again downstream?

waiter/src/waiter/core.clj
```diff
@@ -187,6 +187,15 @@
...
+      (fn add-waiter-api-call-fn [response]
+        (assoc response :waiter-api-call? (waiter-request?-fn request))))))
```
sradack: How does ELK handle fields with question marks at the end?

waiter/src/waiter/request_log.clj
shamsimam: `some?` checks for non-nil value, else a `false` value would have been skipped here.

waiter/src/waiter/core.clj
shamsimam: Done. Preserved the `waiter-request?-fn` argument in `ring-handler-factory` to avoid wholesale changes in the unit tests.

waiter/src/waiter/core.clj
shamsimam: ELK gets to see `waiter-api`

PR #983, opened 2019-10-24
adds response location header, run-as-user, server-port and token to the request log

shamsimam:

- adds response location header, run-as-user, server-port and token to the request log

We would like to query the request log for additional data.

Example request log:
```
{"response-location":"/waiter-async/status/362ec0216e86-51aab0b0935230b7-http/r9091-35ebecf101e2-3bea06de4f36a8f2/w9091-warittmar59568416841740-2684ef42e803e41912f0a5ac1c32f3d1/127.0.0.4/10503/async/status?request-id=c3924624-cc48-4975-bc17-d9b1a3b71f6b",
 "run-as-user":"shamsimam",
 "server-port":9091,
 "token":"www.wtrttslf59657441429902.t1.127.0.0.1",
 ...}
```

### DIFF
--- a/waiter/src/waiter/request_log.clj
+++ b/waiter/src/waiter/request_log.clj
@@ -29,7 +29,8 @@
 
 (defn request->context
   "Convert a request into a context suitable for logging."
-  [{:keys [client-protocol headers internal-protocol query-string remote-addr request-id request-method request-time uri] :as request}]
+  [{:keys [client-protocol headers internal-protocol query-string remote-addr request-id
+           request-method request-time server-port uri] :as request}]
   (let [{:strs [content-length content-type host origin user-agent x-cid x-forwarded-for]} headers]
     (cond-> {:cid x-cid
              :host host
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/request_log.clj
```diff
@@ -45,25 +46,29 @@
...
 
 (defn response->context
   "Convert a response into a context suitable for logging."
   [{:keys [authorization/method authorization/principal backend-response-latency-ns descriptor latest-service-id
            get-instance-latency-ns handle-request-latency-ns headers instance instance-proto protocol status
            waiter-api-call?] :as response}]
-  (let [{:keys [service-id service-description]} descriptor
-        {:strs [content-length content-type grpc-status server]} headers
+  (let [{:keys [service-id service-description source-tokens]} descriptor
+        token (some->> source-tokens (map #(get % "token")) seq (str/join ","))
```
DaoWen: Is this sorted in some deterministic way? Is it worth doing a sort here to make sure partial tokens always show up in the same order? That seems important for filtering logs based on this value.

shamsimam: `source-tokens` are sorted in the order the user specified them as that also affects the service description computed in the end. So, it would produce them is the same order as they are specified in the `x-waiter-token` header.

DaoWen: Ah, that makes sense. I was thinking of disjoint tokens, but they can overlap too so the merge order matters. Thanks.

PR #984, opened 2019-10-26
Limit number of async request status checks

shamsimam:

- adds timer for async request status check

Making frequent status checks can end up generating too many requests from Waiter. The new scheme limits the number of status checks performed by Waiter.

### DIFF
--- a/waiter/test/waiter/async_request_test.clj
+++ b/waiter/test/waiter/async_request_test.clj
@@ -326,6 +326,50 @@
           (complete-async-request-fn :success)
           (is (nil? (get @async-request-store-atom request-id))))))))
 
+(deftest test-post-process-async-request-response-sanitized-check-interval
+  (let [{:keys [host port] :as instance} {:host "www.example.com", :port 1234}
+        router-id "my-router-id"
+        service-id "test-service-id"
+        metric-group "test-metric-group"
+        backend-proto "http"
+        async-request-store-atom (atom {})
+        request-id "request-2394613984619"
+        reason-map {:request-id request-id}
+        request-properties {:async-check-interval-ms 100, :async-request-timeout-ms 6000}
+        location (str "/location/" request-id)
+        query-string "a=b&c=d|e"
+        make-http-request-fn (fn [in-instance in-request end-route metric-group backend-proto]
+                               (is (= instance in-instance))
+                               (is (= {:body nil :headers {} :query-string "a=b&c=d|e" :request-method :get} in-request))
+                               (is (= "/location/request-2394613984619" end-route))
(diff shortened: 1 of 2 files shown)


waiter/test/waiter/async_request_test.clj
```diff
@@ -326,6 +326,50 @@
...
+        async-request-store-atom (atom {})
+        request-id "request-2394613984619"
+        reason-map {:request-id request-id}
+        request-properties {:async-check-interval-ms 100, :async-request-timeout-ms 6000}
+        location (str "/location/" request-id)
+        query-string "a=b&c=d|e"
+        make-http-request-fn (fn [in-instance in-request end-route metric-group backend-proto]
+                               (is (= instance in-instance))
+                               (is (= {:body nil :headers {} :query-string "a=b&c=d|e" :request-method :get} in-request))
+                               (is (= "/location/request-2394613984619" end-route))
```
DaoWen:
```suggestion
                               (is (= location end-route))
```

waiter/test/waiter/async_request_test.clj
```diff
@@ -326,6 +326,50 @@
...
           (is (nil? (get @async-request-store-atom request-id))))))))
 
+(deftest test-post-process-async-request-response-sanitized-check-interval
+  (let [{:keys [host port] :as instance} {:host "www.example.com", :port 1234}
+        router-id "my-router-id"
+        service-id "test-service-id"
+        metric-group "test-metric-group"
+        backend-proto "http"
+        async-request-store-atom (atom {})
+        request-id "request-2394613984619"
```
DaoWen: Where did this long number come from? If it's not actually important, maybe just `123` or something would be better?

waiter/test/waiter/async_request_test.clj
```diff
@@ -326,6 +326,50 @@
...
+                               (is (= "http" backend-proto)))
+        instance-rpc-chan (async/chan 1)
+        complete-async-request-atom (atom nil)
+        response {}]
+    (with-redefs [service/release-instance-go (constantly nil)
+                  monitor-async-request
+                  (fn [make-get-request-fn complete-async-request-fn request-still-active? _
+                       async-check-interval-ms async-request-timeout-ms correlation-id exit-chan]
+                    (is (request-still-active?))
+                    (is (= 120 async-check-interval-ms))
```
DaoWen: It's not obvious where this number comes from. Maybe we should just assert that it's greater than the interval we passed in, but less than some percentage of `async-request-timeout-ms`?

DaoWen:
@shamsimam - I'm going to merge this feature now, and we can refactor the unit test as a follow-on.

PR #989, opened 2019-11-02
enables synchronization when roots mismatch but updated by same user

shamsimam:

- enables synchronization when roots mismatch but updated by same user

Enables syncing tokens across clusters when they have been edited by the same user.

### DIFF
--- a/token-syncer/src/token_syncer/commands/syncer.clj
+++ b/token-syncer/src/token_syncer/commands/syncer.clj
@@ -83,8 +83,8 @@
             cluster-result
             (try
               (let [{:keys [description error status] :as token-data} (get cluster-url->token-data cluster-url)
-                    latest-root (get latest-token-description "root")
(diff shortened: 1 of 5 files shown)


token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -117,9 +117,10 @@
                           (apply dissoc description system-metadata-keys)))
                   {:code :skip/token-sync}
 
-                  ;; token user-specified content different, and roots also different
+                  ;; token user-specified content, last update user, or root different
```
dposada: Shouldn't this be "and"?

```suggestion
                  ;; token user-specified content, last update user, and root different
```

shamsimam:
@dpo accepted suggestion, good to merge

dposada:
Will merge when green

PR #992, opened 2019-11-21
adds the error class into the request log

shamsimam:

- adds the error class into the request log

Enables classification of error responses generated by Waiter.

### Example request log:
```
$ grep -o "waiter-error-class.*" log/request.log | cut -c-100
waiter-error-class":"java.util.concurrent.TimeoutException"}
waiter-error-class":"java.io.EOFException"}
waiter-error-class":"java.net.ConnectException"}
```

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -66,9 +66,9 @@
    - the http status code."
   [error]
   (let [classification (cond (instance? ExceptionInfo error)
-                             (let [[error-cause message status] (classify-error (ex-cause error))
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -66,9 +66,9 @@
    - the http status code."
```
sradack: Update docstring

shamsimam: Updated.

shamsimam:
Build is now green, this can be merged.

PR #993, opened 2019-11-23
fails integration tests early

shamsimam:

- fails integration tests early

Saves unnecessary use of resources and build time while running tests.


waiter/src/waiter/util/client_tools.clj
```diff
@@ -1078,10 +1078,13 @@
 (defmacro with-service-cleanup
   "Ensures a service is cleaned up."
   [service-id & body]
-  `(try
-     ~@body
-     (finally
-       (delete-service ~'waiter-url ~service-id))))
+  `(let [service-id# ~service-id]
+     (if service-id#
```
DaoWen: Nitpick: `if-let`?

PR #1000, opened 2019-12-07
uses waiter principal in async status check requests

shamsimam:

- adds the required waiter-principal settings
- uses waiter principal in async status check requests

Ensures requests to the backend from Waiter have a valid principal in the request.


shamsimam:
> We should also update Kitchen to reject empty x-waiter-auth-user headers.

We need to handle `authentication=disabled` case, e.g. with a new start-up flag to bypass auth.

PR #1001, opened 2019-12-09
Kitchen add auth check for X-Waiter-Auth-Principal

DaoWen:

Kitchen's waiter-auth implementation should check for a non-empty `X-Waiter-Auth-Principal` header.

Many real-world Waiter apps use a waiter-auth implementation that fails when the `X-Waiter-Auth-Principal` header is not set or is empty. We'd like Kitchen to model that same behavior.


DaoWen:
This is stale, and apparently not that important. Closing for now.

PR #1012, opened 2019-12-27
requires explicit inclusion of service metrics in /apps/<service-id> endpoint

shamsimam:

- requires (**backwards incompatible**) explicit inclusion of service metrics in /apps/<service-id> endpoint

Avoids relatively expensive inter-router communication when metrics is not needed in the `/apps/<service-id>` endpoint.


sradack:
Please confirm no usages of this endpoint outside of the UI.

shamsimam:
@sradack usages outside the UI are either sending the new flag or do not care about the metrics.

PR #1013, opened 2019-12-27
avoids throwing exceptions during error processing work-stealing channel

shamsimam:

- avoids throwing exceptions during error processing work-stealing channel
- improves upon status codes returned by work-stealing requests

Avoids throwing exceptions processing failed interactions with work-stealing channels. Failures in work-stealing (or resource-offering) attempts are common and should not trigger the exception path (and logging) to handle them. E.g. it is common for the work-stealing channel to be missing as a service/router starts up.

### DIFF
--- a/waiter/src/waiter/service.clj
+++ b/waiter/src/waiter/service.clj
@@ -87,8 +87,8 @@
        (do
          (log/info "received offer channel, making offer request.")
          (when-not (au/offer! work-stealing-chan# offer-params#)
-           (log/info "unable to put instance on work-stealing-chan"
-                     {:offer-params offer-params# :service-id ~service-id})
+           (log/error "unable to put instance on work-stealing-chan"
+                      {:offer-params offer-params# :service-id ~service-id})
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/service.clj
```diff
@@ -84,13 +85,17 @@
           (async/put! ~instance-rpc-chan))
      (if-let [work-stealing-chan# (async/<! response-chan#)]
        (do
-         (log/info "Received offer channel, making offer request.")
-         (when-not (au/offer! work-stealing-chan# ~offer-params)
-           (throw (ex-info "Unable to put instance on work-stealing-chan."
-                           {:offer-params ~offer-params, :service-id ~service-id}))))
+         (log/info "received offer channel, making offer request.")
+         (when-not (au/offer! work-stealing-chan# offer-params#)
+           (log/info "unable to put instance on work-stealing-chan"
```
sradack: Should this be `log/error`?

PR #1014, opened 2019-12-28
enables listing apps by individual service description parameters

shamsimam:

- enables listing apps by individual service description parameters

Simplifies searching for services by their service description parameters.

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -245,6 +245,12 @@
                     re-pattern)]
     #(re-matches pattern %)))
 
+(defn- strs->filter-fn
+  "Returns a name-filtering function that matches on any of the given sequence of user-provided names as filter string."
+  [names]
+  (let [filter-fns (map str->filter-fn names)]
+    (fn [value] (some #(%1 value) filter-fns))))
+
 (defn- retrieve-scaling-state
   "Retrieves the scaling state for the service from the autoscaler state."
   [query-autoscaler-state-fn service-id]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -254,34 +254,57 @@
     :scale-amount
     utils/scale-amount->scaling-state))
 
+(defn param-value->filter-fn
+  "Creates the filter function that does substring match for the input string or any string in the sequence."
```
scrosby: This description is unclear. Maybe something like:

"Takes a single string or a sequence of strings. Returns a filter function ...."

But then, that description makes this smell like it should be split into two functions  strs->filter-fn .... which handles the multiple case (the second branch of the if) and then this functon. Or maybe after splitting, inline this function into its caller?

waiter/src/waiter/handler.clj
```diff
@@ -254,34 +254,57 @@
...
-                                  (and (if (str/blank? run-as-user)
-                                         (authz/manage-service? entitlement-manager auth-user service-id service-description)
-                                         ((str->filter-fn run-as-user) (get service-description "run-as-user")))
-                                       (or (str/blank? token)
-                                           (let [filter-fn (str->filter-fn token)]
-                                             (->> source-tokens
+          viewable-service-ids (filter
+                                 (fn [service-id]
+                                   (let [service-description (service-id->service-description-fn service-id :effective? true)
+                                         source-tokens (-> (service-id->source-tokens-entries-fn service-id) vec flatten)]
```
scrosby: Why merge this into the same line?

waiter/src/waiter/handler.clj
```diff
@@ -254,34 +254,57 @@
...
+                                     (and (service-description-filter-predicate service-description)
+                                          (if (or (nil? run-as-user)
```
scrosby: I assume this nil? guard was to protect against a bug? Why only here, not token-version? or token?

waiter/src/waiter/handler.clj
shamsimam: Created `strs->filter-fn`

waiter/src/waiter/handler.clj
shamsimam: Minor change as I find it easier to read having to parse fewer lines.

waiter/src/waiter/handler.clj
shamsimam: The `nil` behavior for `run-as-user` is different for legacy reasons. When no `run-as-user` is provided, we display only services a user can manage.

PR #1016, opened 2019-12-30
adds request and response streaming telemetry

shamsimam:

- adds request and response streaming telemetry

Helps us diagnose whether services are sending small/large packets of data.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -297,8 +297,8 @@
                                           (when (pos? bytes-read)
                                             (meters/mark! throughput-meter bytes-read)
                                             (meters/mark! throughput-meter-global bytes-read)
-                                            (meters/mark! throughput-packets-meter 1)
-                                            (meters/mark! throughput-packets-meter-global 1)
+                                            (meters/mark! throughput-packets-meter)
+                                            (meters/mark! throughput-packets-meter-global)
                                             (swap! bytes-streamed-atom + bytes-read)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -287,10 +287,18 @@
...
+                                          throughput-meter (metrics/service-meter service-id "streaming" "request-bytes")
+                                          throughput-meter-global (metrics/waiter-meter "streaming" "request-bytes")
+                                          throughput-packets-meter (metrics/service-meter service-id "streaming" "request-packets")
+                                          throughput-packets-meter-global (metrics/waiter-meter "streaming" "request-packets")]
                                       (fn report-request-size-metrics [bytes-read complete?]
                                         (try
                                           (when (pos? bytes-read)
+                                            (meters/mark! throughput-meter bytes-read)
+                                            (meters/mark! throughput-meter-global bytes-read)
+                                            (meters/mark! throughput-packets-meter 1)
```
sradack: Why do we expect that each iteration here is 1 packet?  Are we talking about 1 IP packet?

sradack: 1 is optional. Let's remove.

shamsimam: > Why do we expect that each iteration here is 1 packet? Are we talking about 1 IP packet?

Not an IP packet, one buffered packet in Waiter code. It can help reflect inefficiency if there are many packets with small sizes.

> 1 is optional. Let's remove.

Removed.

PR #1023, opened 2020-01-22
Introduces a fast path that avoids async/go blocks to retrieve maintainer channels

shamsimam:

Introduces a fast path that avoids async/go blocks to retrieve maintainer channels. The maintainer state is maintained in an atom and the fast path looks up the atom and avoids the async block when there is a hit.

Avoiding go-blocks allows avoiding context switching to retrieve the channel data thus improving the time required to reserve an instance.

### Example metrics:

Screen Shot 2020-01-22 at 9 04 45 PM

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -120,6 +120,7 @@
                               ["/router-metrics" :state-router-metrics-handler-fn]
                               ["/scheduler" :state-scheduler-handler-fn]
                               ["/service-description-builder" :state-service-description-builder-handler-fn]
+                              ["/service-maintainer" :state-service-maintainer-handler-fn]
                               ["/statsd" :state-statsd-handler-fn]
                               ["/work-stealing" :state-work-stealing-handler-fn]
                               [["/" :service-id] :state-service-handler-fn]]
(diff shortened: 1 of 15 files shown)


waiter/src/waiter/core.clj
```diff
@@ -639,7 +640,7 @@
                               :socket-timeout client-connection-idle-timeout-ms})
    :http-clients (pc/fnk [http-client-properties]
                    (hu/prepare-http-clients http-client-properties))
-   :instance-rpc-chan (pc/fnk [] (async/chan 1024)) ; TODO move to service-chan-maintainer
+   :instance-rpc-chan (pc/fnk [] (async/chan 1024)) ; breaks cycle between daemons and routines
```
sradack: Can you explain this comment a bit more?

waiter/src/waiter/core.clj
```diff
@@ -946,17 +947,21 @@
                                                      instance short-circuit? router-ids "blacklist"
                                                      (partial make-blacklist-request make-inter-router-requests-sync-fn blacklist-period-ms)
                                                      reason))))
+   :populate-maintainer-chan-routine! (pc/fnk [[:state instance-rpc-chan]]
+                                        ;; avoid cycle in dependency resolution for functions defined in daemons
```
sradack: Can you explain this comment a bit more?

waiter/src/waiter/core.clj
```diff
@@ -1191,6 +1198,12 @@
    :messages (pc/fnk [[:settings {messages nil}]]
                (when messages
                  (utils/load-messages messages)))
+   :populate-maintainer-chan! (pc/fnk [[:state instance-rpc-chan]
```
sradack: Why is `populate-maintainer-chan!` in the `:daemon` section?

waiter/src/waiter/core.clj
sradack: Why shouldn't everything you've changed to use `:populate-maintainer-chan-routine!` use `:populate-maintainer-chan!` instead?

PR #1026, opened 2020-01-24
considers the x-waiter-http-redirect header for https redirects

shamsimam:

- considers the `x-waiter-http-redirect` header for https redirect

Consistent with Waiter philosophy where headers override token configuration. In addition, allows on-the-fly requests to also configure https redirects.


sradack:
I can't imagine a situation where someone would want to set this.  If someone is trying to use HTTPS, they would change the URL to https

shamsimam:
Okay.

PR #1028, opened 2020-01-31
refactor use of run-as-user parameter in service-list-handler

shamsimam:

- fixes typo in service-list-handler

Fixes handling of `run-as-user` parameter.


shamsimam:
@dposada this is now a refactoring PR :)

PR #1033, opened 2020-02-11
enables explicit grpc tests

shamsimam:

- enables grpc cancellation tests when custom grpc port environment variable is provided
- fixes ConcurrentModificationException while iterating on data returned from trackState()
- deflakes previously explicit flaky tests

We wish to run the full suite of gRPC integration tests and assertions including client and server cancellation tests.

### DIFF
--- a/waiter/src/waiter/util/client_tools.clj
+++ b/waiter/src/waiter/util/client_tools.clj
@@ -150,8 +150,13 @@
 
 (defn grpc-cancellations-supported?
   "Returns true if gRPC cancellations are supported.
-   Certain proxies, e.g. haproxy https://github.com/haproxy/haproxy/issues/172, do not support cancellations.
-   Other proxies, e.g. Envoy, do support gRPC cancellations."
+   Waiter routers support gRPC client and server cancellation.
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/util/client_tools.clj
```diff
@@ -148,6 +148,14 @@
   (not= (retrieve-waiter-port waiter-url)
         (retrieve-h2c-port waiter-url)))
 
+(defn grpc-cancellations-supported?
+  "Returns true if gRPC cancellations are supported.
+   Certain proxies, e.g. haproxy https://github.com/haproxy/haproxy/issues/172, do not support cancellations.
+   Other proxies, e.g. Envoy, do support gRPC cancellations."
+  [waiter-url]
+  (or (not (behind-proxy? waiter-url))
+      (not (str/blank? (System/getenv "WAITER_GRPC_CLEARTEXT_PORT")))))
```
sradack: Can you make the connection between the implementation and the docs for this fn?

shamsimam: Improved the doctring

PR #1035, opened 2020-02-21
ensures computed protocol is HTTP/2.0 instead of HTTP/2

shamsimam:

- ensures computed protocol is HTTP/2.0 instead of HTTP/2

The Waiter code expects the client protocol to be `HTTP/2.0` while processing http/2 requests. This change ensures that we compute the protocol correctly as the rest of the Waiter codes expects `HTTP/2.0` as the value for the client protocol.


nsinkov:
merge when green

PR #1037, opened 2020-02-23
Enable admin user to mark token as no-sync

DaoWen:

Tokens may be manually excluded from synchronization via an admin-mode update
by setting the last-update-user field to match a value set via the environment
in the token syncer: `WAITER_TOKEN_ADMIN_NO_SYNC_USER`

An admin may need to make some temporary modifications to a token in one cluster that should not be synced to other Waiter clusters. Setting the `last-update-user` to a well-known value is a simple way to signal this on the token itself.


DaoWen:
We decided we don't need this for now.

PR #1038, opened 2020-02-24
add ability to redact fields from request log

nsinkov:

- add field to token that list fields to redact from the request log

### DIFF
--- a/waiter/src/waiter/request_log.clj
+++ b/waiter/src/waiter/request_log.clj
@@ -93,7 +93,9 @@
 (defn log-request!
   "Log a request"
   [request response]
-  (log (merge (request->context request) (response->context response))))
+  (log (apply dissoc
+              (merge (request->context request) (response->context response))
+              (map keyword (get-in response [:descriptor :service-description "redacted-request-fields"])))))
 
 (defn wrap-log
   "Wraps a handler logging data from requests and responses."
--- a/waiter/src/waiter/request_log.clj
+++ b/waiter/src/waiter/request_log.clj
@@ -93,9 +93,11 @@
 (defn log-request!
   "Log a request"
   [request response]
-  (log (apply dissoc
-              (merge (request->context request) (response->context response))
-              (map keyword (get-in response [:descriptor :service-description "redacted-request-fields"])))))
+  (let [redacted-request-fields-string (get-in response [:descriptor :service-description "env" "WAITER_REDACTED_REQUEST_FIELDS"])]
+    (log (apply dissoc
+                (merge (request->context request) (response->context response))
+                (when-not (str/blank? redacted-request-fields-string)
+                  (map keyword (str/split redacted-request-fields-string #",")))))))
 
 (defn wrap-log
   "Wraps a handler logging data from requests and responses."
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/request_log.clj
```diff
@@ -93,7 +93,11 @@
 (defn log-request!
   "Log a request"
   [request response]
-  (log (merge (request->context request) (response->context response))))
+  (let [redacted-request-fields-string (get-in response [:descriptor :service-description "env" "REDACTED_REQUEST_FIELDS"])]
+    (log (apply dissoc
```
DaoWen: It's confusing having this much logic in the `log` statement. Let's lift the redacted request into a new `let` binding.

waiter/src/waiter/request_log.clj
```diff
@@ -93,7 +93,11 @@
...
+                (merge (request->context request) (response->context response))
+                (when-not (str/blank? redacted-request-fields-string)
```
DaoWen: Rather than `(apply dissoc ... (when-not ...))`, I think `(cond->> ... (not ...) (apply dissoc ...))` is more idiomatic.

waiter/src/waiter/request_log.clj
```diff
@@ -93,7 +93,11 @@
...
+                  (map keyword (str/split redacted-request-fields-string #",")))))))
```
DaoWen: Let's lift this into a new `let` binding too: `redacted-request-fields`

waiter/src/waiter/request_log.clj
```diff
@@ -93,7 +93,11 @@
 (defn log-request!
   "Log a request"
   [request response]
-  (log (merge (request->context request) (response->context response))))
+  (let [redacted-request-fields-string (get-in response [:descriptor :service-description "env" "REDACTED_REQUEST_FIELDS"])]
```
DaoWen: I feel like we should be prefixing our "reserved" environment variable names:
`WAITER_REDACTED_REQUEST_FIELDS`?

DaoWen: By "reserved" I mean "we reserve the right to take some action if you set this."

`WAITER_CONFIG_`? We'd have to exclude this from the regex. @shamsimam ?

shamsimam: Sounds good to me. 
We should also add checks in our code that disallows waiter code from generating environment variables starting with `WAITER_CONFIG_`.

waiter/src/waiter/request_log.clj
```diff
@@ -93,7 +93,10 @@
 (defn log-request!
   "Log a request"
   [request response]
-  (log (merge (request->context request) (response->context response))))
+  (let [redacted-request-fields-string (get-in response [:descriptor :service-description "env" "WAITER_CONFIG_REDACTED_REQUEST_FIELDS"])
+        redacted-request-fields (when-not (str/blank? redacted-request-fields-string)
+                                  (map keyword (str/split redacted-request-fields-string #",")))]
+    (log (apply dissoc (merge (request->context request) (response->context response)) redacted-request-fields))))
```
shamsimam: Was this already discussed: removing the keys vs replacing the values with a special string like `_REDACTED_`?

nsinkov: discussed offline

PR #1042, opened 2020-02-28
sorts nested map parameters to ensure deterministic service-id

shamsimam:

- sorts nested map parameters to ensure deterministic service-id

Service descriptions that contain nested maps that are equal (as they contain the same key-value mappings) should map to the same service-id. However, the current code does not sort the maps while computing the service-id and as a result, equal configurations produce different service-ids.

**Note**: This solution can end up changing the service-ids of existing services which uses map parameters like `env`, `metadata`, etc with multiple entries.

### DIFF
--- a/waiter/test/waiter/service_description_test.clj
+++ b/waiter/test/waiter/service_description_test.clj
@@ -206,7 +206,16 @@
                           :expected (str service-id-prefix "fum1234a-a1030ca63357baad681c25935eb4e494")}
                          {:name "service-description->service-id:invalid-chars-present-in--name"
                           :input-data {"name" "fum-!@#$%.,:()"}
-                          :expected (str service-id-prefix "fum-df72716b57632adfc64b74165eb7d7f2")})]
+                          :expected (str service-id-prefix "fum-df72716b57632adfc64b74165eb7d7f2")}
+                         {:name "service-description->service-id:map-data"
+                          :input-data {"env" {"bar" "baz"}}
+                          :expected (str service-id-prefix "23dcf01c78e18f56be504c40c236438a")}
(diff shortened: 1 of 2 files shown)


waiter/test/waiter/service_description_test.clj
```diff
@@ -206,7 +206,13 @@
...
                          {:name "service-description->service-id:invalid-chars-present-in--name"
                           :input-data {"name" "fum-!@#$%.,:()"}
-                          :expected (str service-id-prefix "fum-df72716b57632adfc64b74165eb7d7f2")})]
+                          :expected (str service-id-prefix "fum-df72716b57632adfc64b74165eb7d7f2")}
+                         {:name "service-description->service-id:map-data"
+                          :input-data {"env" {"bar" "baz", "fee" "fie", "foe" "fum"}}
+                          :expected (str service-id-prefix "54b332ed6fa91bf23ffc57ef82d3f1ff")}
+                         {:name "service-description->service-id:map-data-reordered"
+                          :input-data {"env" {"foe" "fum", "bar" "baz", "fee" "fie"}}
+                          :expected (str service-id-prefix "54b332ed6fa91bf23ffc57ef82d3f1ff")})]
```
shamsimam: These test-cases produce different service-ids without the fix: `test-waiter-prefix-49c060d06a02e18e102568b60abc48a9` and `test-waiter-prefix-cef7f9643e37ba5a5df0469202ddf7e3`, respectively.

shamsimam:
Creating follow-up PRs for fallback and GC.

shamsimam:
See: https://github.com/twosigma/waiter/pull/1043

PR #1043, opened 2020-03-02
adds fallback to and GC support for outdated service-IDs

shamsimam:

- handles sets in service parameters while producing deterministic service id
- GCs services whose service-IDs have changed after the deterministic ID fix
- adds support to fallback to older (v0) service-id

We need to cleanup services running with an outdated service-ID computation scheme as they will no longer receive main traffic and it is wasteful to keep them running.
Allow fallback to the older service.


shamsimam:
Spoke with @DaoWen, we can handle this situation better with script that identifies the services, starts the new services, and cleans up older services.

shamsimam:
Please see: https://github.com/twosigma/waiter/pull/1043

PR #1044, opened 2020-03-02
skips test-kubernetes-pod-expiry-grace-period when container-running-grace-secs is 0

shamsimam:

- skips test-kubernetes-pod-expiry-grace-period when container-running-grace-secs is 0

The test fails due to incorrect assertions when the `container-running-grace-secs` config is disabled.

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -297,10 +297,17 @@
                                :x-waiter-timeout 30000)
               service-id (retrieve-service-id waiter-url waiter-headers)
               timeout-secs 150]
-          (if (> container-running-grace-secs timeout-secs)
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -297,10 +297,17 @@
                                :x-waiter-timeout 30000)
               service-id (retrieve-service-id waiter-url waiter-headers)
               timeout-secs 150]
-          (if (> container-running-grace-secs timeout-secs)
+          (cond
+            (zero? container-running-grace-secs)
+            (log/info "skipping test as container-running-grace-secs is disabled"
+                      {:container-running-grace-secs container-running-grace-secs
+                       :waiter-url waiter-url})
+
```
DaoWen:
```suggestion
```

PR #1048, opened 2020-03-06
adds guard to protect against NPE when min-instances default is unavailable

shamsimam:

- adds guard to protect against NPE when min-instances default is unavailable

On the delete service code path, the default passed in as arguments do not include the min instances and this leads to an NPE preventing users from deleting the service.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -1075,8 +1075,9 @@
 (defn can-manage-service?
   "Returns whether the `username` is allowed to modify the specified service description."
   [kv-store entitlement-manager service-id username]
-  ; the stored service description should already have a run-as-user
-  (let [core-service-description (service-id->service-description kv-store service-id {} [])]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -1075,8 +1076,8 @@
   "Returns whether the `username` is allowed to modify the specified service description."
   [kv-store entitlement-manager service-id username]
   ; the stored service description should already have a run-as-user
-  (let [service-description (service-id->service-description kv-store service-id {} [])]
-    (authz/manage-service? entitlement-manager username service-id service-description)))
+  (let [core-service-description (service-id->service-description kv-store service-id {} [])]
```
nsinkov: This code actually just wants the run-as-user. It should get it directly via fetch-core kv-store service-id

shamsimam: Done, thanks for noticing this.

PR #1052, opened 2020-03-10
writes new entries before deleting old ones to avoid intervening index queries from reading empty results

shamsimam:

- writes new entries before deleting old ones to avoid intervening index queries from reading empty results

We want to avoid intervening index queries from reading empty results while the reindex operation is in progress.

### DIFF
--- a/token-syncer/project.clj
+++ b/token-syncer/project.clj
@@ -14,7 +14,7 @@
 ;; limitations under the License.
 ;;
 (defproject token-syncer "0.1.0-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


token-syncer/project.clj
```diff
@@ -14,7 +14,7 @@
 ;; limitations under the License.
 ;;
 (defproject token-syncer "0.1.0-SNAPSHOT"
-  :dependencies [[twosigma/jet "0.7.10-20190426_181314-g8d0a48b"]
+  :dependencies [[twosigma/jet "0.7.10-20191230_163338-g0dabe28"]
```
shamsimam: This edit triggers the token-syncer tests in travis.

PR #1054, opened 2020-03-19
allows http/2 requests to configure idle timeouts

shamsimam:

- allows http/2 requests to configure idle timeouts

Requests configuring idle timeouts should have their timeout values respected.

### DIFF
--- a/waiter/integration/waiter/grpc_test.clj
+++ b/waiter/integration/waiter/grpc_test.clj
@@ -257,7 +257,7 @@
           (is (= (count-items states "HALF_CLOSE") 1) assertion-message)
           (is (= ["CLOSE" "COMPLETE"] (take-last 2 states)) assertion-message))))))
 
-(deftest ^:parallel ^:integration-fast test-grpc-unary-call
+(deftest ^:parallel ^:integration-slow test-grpc-unary-call
   (testing-using-waiter-url
     (let [{:keys [h2c-port host request-headers service-id]} (start-courier-instance waiter-url)
           id (rand-name "m")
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/util/http_utils.clj
```diff
@@ -126,7 +126,7 @@
     {:http1-client (http-client-factory (cond-> config
                                           client-name (update :client-name str "-http1")
                                           user-agent (update :user-agent str ".http1")))
-     :http2-client (-> (cond-> config
+     :http2-client (-> (cond-> (dissoc config :conn-timeout :socket-timeout)
```
sradack: It feels wrong that we `dissoc`ing only in the h2 case.

sradack: Can you explain this change a bit more?

shamsimam: There are two http clients for handling http/2 requests: `HTTP2Client` and `HttpClient`. Setting the idle timeouts on both client instances interacts weirdly. The presence of the timeout values on the `HttpClient` instance means that the properties configured on the request (processed by the `HTTP2Client`) via the `x-waiter-*` headers are not respected. By having the timeout only on the `HTTP2Client`, we allow the custom request timeouts to be triggered

shamsimam: `(prepare-http2-transport conn-timeout socket-timeout)` already creates a client with the requested timeouts.

waiter/integration/waiter/grpc_test.clj
```diff
@@ -281,7 +281,33 @@
...
+                assertion-message (grpc-result->assertion-message correlation-id rpc-result)]
+            (assert-grpc-ok-status status assertion-message)
+            (is reply assertion-message)
+            (when reply
+              (is (= id (.getId reply)) assertion-message)
+              (is (= content (.getMessage reply)) assertion-message)
+              (is (= "received" (.getResponse reply)) assertion-message))
+            (assert-request-state grpc-client request-headers service-id correlation-id ::success)))
+
+        (testing "long request and reply"
```
shamsimam: This test fails without the fix with a `TimeoutException` in Waiter request to the backend.

waiter/src/waiter/util/http_utils.clj
sradack: Ahh, a comment that `prepare-http2-transport` handles the connect and socket timeouts would have helped me, but I see it now.  So we still have default connect and socket timeouts for H2 after this change?

shamsimam: Added comment.

I ended up not adding the test since the default timeout is 15 minutes.
I verified manually that the default timeout works:
```
2020-03-19 00:13:11,392 INFO  waiter.core [qtp1290884462-188] - [CID=wgttguc174478353882484] request received: {:internal-protocol HTTP/2.0, :request-id 9eafe8c087f8-6356d8278a9a1531-http, :remote-addr 127.0.0.1, :client-protocol HTTP/2.0, :headers {x-waiter-health-check-proto http, x-waiter-max-instances 1, x-waiter-backend-proto h2c, host 127.0.0.1:9091, x-waiter-cmd-type shell, x-waiter-name wgttguc174445802557383, x-waiter-version version-does-not-matter, grpc-timeout 914995m, grpc-accept-encoding gzip, user-agent grpc-java-netty/1.20.0, x-waiter-grace-period-secs 120, content-type application/grpc, cookie x-waiter-auth=TlBZAkew8ops0N00kM%2FUACUC6QdVyLkARR%2F0FFNg4RdfCmXmaQVc%2FEfnL..., x-waiter-health-check-port-index 1, x-waiter-concurrency-level 32, x-waiter-ports 2, x-waiter-idle-timeout-mins 10, x-cid wgttguc174478353882484, x-waiter-min-instances 1, x-waiter-debug true, x-waiter-cpus 0.2, te trailers, x-waiter-mem 512, x-waiter-cmd /Users/shamsimam/projects/github-projects/twosigma-waiter/containers/test-app...}, :content-length nil, :content-type application/grpc, :character-encoding nil, :uri /courier.Courier/SendPackage, :query-string nil, :router-id r9091-9e3e1860f755-639339e0b025f60, :scheme :http, :request-method :post}
2020-03-19 00:13:11,394 INFO  waiter.process-request [async-dispatch-48] - [CID=wgttguc174478353882484] process request to 127.0.0.1:9091 at path /courier.Courier/SendPackage
2020-03-19 00:13:11,395 INFO  waiter.process-request [async-dispatch-19] - [CID=wgttguc174478353882484] suggested instance: w9091-wgttguc174445802557383-8c5a07b600fe207ccb541916859596a6.9ea8dfde3dda-27119d460babf7c4 127.0.0.9 10500
2020-03-19 00:13:11,397 INFO  waiter.process-request [async-dispatch-19] - [CID=wgttguc174478353882484] connecting to http://127.0.0.9:10500/courier.Courier/SendPackage using HTTP/2.0
2020-03-19 00:13:11,397 INFO  waiter.process-request [async-dispatch-19] - [CID=wgttguc174478353882484] w9091-wgttguc174445802557383-8c5a07b600fe207ccb541916859596a6 {:idle-timeout 900000, :streaming-timeout 20000}
2020-03-19 00:13:11,397 INFO  waiter.process-request [qtp1290884462-116] - [CID=wgttguc174478353882484] streamed 1048 bytes from request input stream
...
2020-03-19 00:28:11,401 INFO  waiter.process-request [async-dispatch-8] - [CID=wgttguc174478353882484] java.util.concurrent.TimeoutException Idle timeout expired: 900004/900000 ms identified as :instance-error
2020-03-19 00:28:11,427 ERROR waiter.util.utils [async-dispatch-8] - [CID=wgttguc174478353882484] #error {
 :cause Idle timeout expired: 900004/900000 ms
```

PR #1059, opened 2020-04-03
includes last auth update time in k8s scheduler state

shamsimam:

- includes last auth token update time in k8s scheduler state
- changes the :action-fn signature to return a map

Allows visibility into the state of the scheduler, in this case, by letting us know the last authentication token update time.

### Example state output:

```
      "kubernetes": {
        "auth-token-renewer": {
          "auth-token": "Bearer foobarfief...",
          "expiry-time": "2020-04-05T04:21:43.1234Z",
          "k8s/retrieve-time": "2020-04-04T05:14:21.514Z"
        }
...
```

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -39,12 +39,14 @@
   (:import (java.io InputStreamReader)
            (org.joda.time.format DateTimeFormat)))
 
-(defn authorization-from-environment []
+(defn authorization-from-environment
   "Sample implementation of the authentication string refresh function.
-   Returns a string to be used as the value for the Authorization HTTP header,
-   reading the string from the WAITER_K8S_AUTH_STRING environment variable."
+   Returns a map with :auth-token mapped to a string that can be used as
+   the value for the Authorization HTTP header, reading the string from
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -913,16 +915,21 @@
...
-        auth-update-task
-        :delay-ms (* 60000 refresh-delay-mins)))))
+    {:cancel-fn (if refresh-delay-mins
+                  (du/start-timer-task
+                    (t/minutes refresh-delay-mins)
+                    auth-update-task
+                    :delay-ms (* 60000 refresh-delay-mins))
+                  (constantly nil))
+     :query-state-fn (fn query-auth-token-state []
+                       {:last-token-retrieve-time @auth-update-time-atom})}))
```
DaoWen, line 932: Maybe this function should just return the atom's contents (and the atom can have a more generic name). That way the implementation could decide all of the state it wants to expose... But I'm not sure how we'd achieve this since `action-fn` currently just returns a string? Maybe we could have it return a string or a 2-count sequence?

shamsimam: We can make a backwards incompatible change and have the `action-fn` return a map instead.

PR #1060, opened 2020-04-03
JWT auth defaults

shamsimam:

- adds option to enable JWT auth by default for tokens
- adds JWT auth support for waiter api requests

We would like to allow Waiter API requests and service requests to get JWT authentication by default.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -248,7 +248,11 @@
                               ;; The token type expected in the access token header
                               :token-type "JWT"
                               ;; The interval at which to refresh the keys from the authorization server:
-                              :update-interval-ms 60000}
+                              :update-interval-ms 60000
+                              ;; Whether JWT authentication is the default for services and Waiter API requests.
+                              ;; For services, the value of USE_BEARER_AUTH environment variable overrides this value.
+                              ;; The default is false.
+                              :use-bearer-auth-default? false}
 
                         ;; :kind :one-user allows anonymous access to services (for testing purposes only):
                         :one-user {;; Custom implementations should specify a :factory-fn
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -225,7 +225,14 @@
                         ;; Waiter supports JWT access token-based authentication before trying the configured authenticator.
                         ;; JWT authentication can be disabled by specifying a value of :disabled instead of configuring the map below:
                         ;; :jwt :disabled
-                        :jwt {;; JWT relies on periodically refreshing the list of available keys from a JWKS
+                        :jwt {;; Whether JWT authentication is allowed for Waiter API requests.
+                              ;; When not configured, the value is assumed to be true.
+                              :allow-bearer-auth-api? true
+                              ;; Whether JWT authentication is the allowed for service (proxy) requests.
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/auth/jwt.clj
```diff
@@ -365,12 +365,19 @@
...
-             (auth/select-auth-header request access-token?))
+             (auth/select-auth-header request access-token?)
+             (or
+               ;; service requests will enable JWT auth based on env variable or when absent, the use-bearer-auth-default?
+               (and (not waiter-api-call?)
+                    (= "true" (get-in request [:waiter-discovery :service-parameter-template "env" "USE_BEARER_AUTH"]
+                                      (str use-bearer-auth-default?))))
+               ;; waiter api requests will enable JWT auth based on use-bearer-auth-default?
+               (and waiter-api-call?
+                    use-bearer-auth-default?)))
```
shamsimam, line 380: Do you want a separate flag to configure the defaults for waiter api requests?

waiter/config-full.edn
```diff
@@ -248,7 +248,15 @@
                               ;; The token type expected in the access token header
                               :token-type "JWT"
                               ;; The interval at which to refresh the keys from the authorization server:
-                              :update-interval-ms 60000}
+                              :update-interval-ms 60000
+                              ;; Whether JWT authentication is the default for Waiter API requests.
+                              ;; For services, the value of USE_BEARER_AUTH environment variable overrides this value.
+                              ;; The default is true.
+                              :use-bearer-auth-default-for-api? true
```
sradack, line 255: Would this be clearer as `:allow-bearer-auth-default-for-api?`.  `use` implies other methods aren't allowed.

sradack: What does `default` mean here?  Should this just be `:allow-bearer-auth-api?`

waiter/config-full.edn
```diff
@@ -248,7 +248,15 @@
...
+                              ;; Whether JWT authentication is the default for services proxy requests.
+                              ;; The value of USE_BEARER_AUTH environment variable of the service overrides this value.
+                              ;; The default is false.
+                              :use-bearer-auth-default-for-service? false}
```
sradack, line 259: Would this be clearer as :allow-bearer-auth-default-for-service?. `use` implies other methods aren't allowed.

sradack: What does `default` mean here?  Should this just be `:allow-bearer-auth-services?`

waiter/config-full.edn
shamsimam: Renamed to `:allow-bearer-auth-api?`. Rephrased the docstring, this is an optional field, the default refers to the value used when the field is not configured.

waiter/config-full.edn
shamsimam: Renamed to `:allow-bearer-auth-services?`. Rephrased the docstring, this is an optional field, the default refers to the value used when the field is not configured.

waiter/src/waiter/auth/jwt.clj
shamsimam: Added separate flag.

PR #1064, opened 2020-04-10
adds support for the profile parameter

shamsimam:

- adds support for the `profile` parameter

Adding support for the `profile` parameter to allow different defaults for different applications.

### Notes
- profiles can override the default service parameters
- user-specified profiles must be non-empty strings and supported in the Waiter config
- services which do not specify profiles are treated internally in code to have specified a profile with an empty string

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1644,3 +1644,30 @@
                   (is (= service-id-1 (:service-id response-2))))))))
         (finally
           (delete-token-and-assert waiter-url token))))))
+
+(deftest ^:parallel ^:integration-fast test-profile-inside-token
+  (testing-using-waiter-url
+    (let [{:keys [profile-config]} (waiter-settings waiter-url)
+          token (rand-name)
+          base-description {:cmd "will-not-work"
+                            :cmd-type "shell"
+                            :cpus 1
+                            :health-check-url "/ping"
+                            :mem 100
+                            :name token
+                            :permitted-user "*"
+                            :run-as-user (retrieve-username)
+                            :version "1"}]
+      (doseq [profile (keys profile-config)]
+        (let [token-description (assoc base-description
+                                  :profile (name profile)
+                                  :token token)
+              register-response (post-token waiter-url token-description)
+              _ (assert-response-status register-response 200)
+              {:keys [body]} (get-token waiter-url token)]
+          (is (= (:profile token-description) (-> body str json/read-str (get "profile"))))
+          (let [service-id (retrieve-service-id waiter-url {"x-waiter-token" token})
+                _ (is service-id "No service created by using token!")
+                service-description (service-id->service-description waiter-url service-id)]
+            (is (= service-description (dissoc token-description :token))))
+          (delete-token-and-assert waiter-url token))))))
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1648,26 +1648,29 @@
 (deftest ^:parallel ^:integration-fast test-profile-inside-token
   (testing-using-waiter-url
     (let [{:keys [profile-config]} (waiter-settings waiter-url)
-          token (rand-name)
           base-description {:cmd "will-not-work"
                             :cmd-type "shell"
                             :cpus 1
                             :health-check-url "/ping"
                             :mem 100
-                            :name token
+                            :metric-group "waiter_test"
+                            :name "test-profile-inside-token"
                             :permitted-user "*"
                             :run-as-user (retrieve-username)
                             :version "1"}]
       (doseq [profile (keys profile-config)]
-        (let [token-description (assoc base-description
+        (let [token (rand-name)
+              token-description (assoc base-description
                                   :profile (name profile)
                                   :token token)
-              register-response (post-token waiter-url token-description)
-              _ (assert-response-status register-response 200)
-              {:keys [body]} (get-token waiter-url token)]
-          (is (= (:profile token-description) (-> body str json/read-str (get "profile"))))
-          (let [service-id (retrieve-service-id waiter-url {"x-waiter-token" token})
-                _ (is service-id "No service created by using token!")
-                service-description (service-id->service-description waiter-url service-id)]
-            (is (= service-description (dissoc token-description :token))))
-          (delete-token-and-assert waiter-url token))))))
+              register-response (post-token waiter-url token-description)]
+          (try
+            (assert-response-status register-response 200)
+            (let [{:keys [body]} (get-token waiter-url token)]
+              (is (= (:profile token-description) (-> body str json/read-str (get "profile"))))
+              (let [service-id (retrieve-service-id waiter-url {"x-waiter-token" token})
+                    _ (is service-id "No service created by using token!")
+                    service-description (service-id->service-description waiter-url service-id)]
+                (is (= service-description (dissoc token-description :token)))))
+            (finally
+              (delete-token-and-assert waiter-url token))))))))
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -701,6 +701,11 @@
                       _ (password-store/check-empty-passwords passwords)
                       processed-passwords (mapv #(vector :cached %) passwords)]
                   processed-passwords))
+   :profile->defaults (pc/fnk [[:settings profile-config service-description-defaults]]
+                        (let [custom-profiles (pc/for-map [[profile {:keys [service-parameters]}] profile-config]
+                                                  (name profile)
+                                                  (merge service-description-defaults service-parameters))]
+                          (assoc custom-profiles "" service-description-defaults)))
    :query-service-maintainer-chan (pc/fnk [] (au/latest-chan)) ; TODO move to service-chan-maintainer
    :router-metrics-agent (pc/fnk [router-id] (metrics-sync/new-router-metrics-agent router-id {}))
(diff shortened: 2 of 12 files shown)


shamsimam:
@sradack this is ready for an initial review as I polish the code and tests.

waiter/src/waiter/core.clj
```diff
@@ -701,6 +701,11 @@
                       _ (password-store/check-empty-passwords passwords)
                       processed-passwords (mapv #(vector :cached %) passwords)]
                   processed-passwords))
+   :profile->defaults (pc/fnk [[:settings profile-config service-description-defaults]]
+                        (let [custom-profiles (pc/for-map [[profile {:keys [service-parameters]}] profile-config]
+                                                  (name profile)
+                                                  (merge service-description-defaults service-parameters))]
+                          (assoc custom-profiles "" service-description-defaults)))
```
sradack, line 708: I'm not in love with empty string as a key.

waiter/src/waiter/core.clj
```diff
@@ -837,12 +842,12 @@
                                  (digest/md5 (str service-id (first passwords)))))
    ; This function is only included here for initializing the scheduler above.
    ; Prefer accessing the non-starred version of this function through the routines map.
-   :service-id->service-description-fn* (pc/fnk [[:settings service-description-defaults metric-group-mappings]
-                                                 [:state kv-store]]
+   :service-id->service-description-fn* (pc/fnk [[:settings metric-group-mappings]
```
sradack, line 845: Keeping service-description-defaults as independent from profiles feels doesn't feel awkward, where as sticking service-description-defaults in the profiles map as an empty string key does.

sradack: Keeping separate also helps reinforce the notion that profiles are not "layered".  We have defaults, and a single profile that can overlay a new set of defaults.

shamsimam: Profile providing a different set of layered defaults is how it is represented in the configuration with the different `service-description-defaults` and `profile-config` sections.

The implementation 'optimizes' by pre-computing the profiles to default parameters mapping instead of having to overlay the parameters on every request. It also saves us from having to pass a couple of parameters around every time. I can change the mapping to `nil` or a `:default` keyword to represent the no-explicit profile specified case.

What do you think?

waiter/src/waiter/core.clj
shamsimam: How do you feel about using a keyword, e.g. `:default`?

waiter/src/waiter/core.clj
sradack: They're not equivalent keys in the map.  A profile can be a subset, where as all defaults cannot be a subset.  The set of all default parameters are not a "profile."  The optimization doesn't sound like it saves that much and could be done higher up.  Having separate parameters increases clarity IMO and not worth the "savings" of having separate parameters.

waiter/src/waiter/core.clj
shamsimam: Removed the `:default` entry and renamed to `profile->overrides`.

waiter/src/waiter/core.clj
shamsimam: Changed the code to pass in two separate arguments: `service-description-defaults` as before and the new `profile->overrides`.

waiter/src/waiter/service_description.clj
```diff
@@ -478,6 +510,10 @@
                                              parameter->issues :name "name must be a non-empty string.")
                                            (attach-error-message-for-parameter
                                              parameter->issues :ports "ports must be an integer in the range [1, 10].")
+                                           (attach-error-message-for-parameter
+                                             parameter->issues :profile
+                                             (str "profile must be a non-empty string"
```
sradack, line 515: Must it?

waiter/src/waiter/service_description.clj
```diff
@@ -78,6 +78,7 @@
    (s/optional-key "namespace") schema/non-empty-string
    (s/optional-key "permitted-user") schema/non-empty-string
    (s/optional-key "ports") schema/valid-number-of-ports
+   (s/optional-key "profile") schema/non-empty-string
```
sradack, line 81: Should this be a all lowercase no hyphen etc etc string?

waiter/src/waiter/core.clj
```diff
@@ -1076,7 +1087,8 @@
                                               (throw (ex-info (str "authentication must be one of: '"
                                                                    (str/join "', '" (sort authentication-providers)) "'")
                                                               {:authentication authentication :status 400}))))
-                                          (sd/validate service-description-builder service-description {}))))
+                                          (sd/validate service-description-builder service-description
+                                                       {:profile->overrides profile->overrides}))))
```
sradack, line 1091: Why can't `service-description-builder` have access to `profile-overrides` in all cases?  (initialize in constructor)

waiter/src/waiter/service_description.clj
```diff
@@ -1012,14 +1064,19 @@
...
       (let [stored-service-description (fetch-core kv-store service-id)]
         ; Validating is expensive, so avoid validating if we've validated before, relying on the fact
         ; that we'll only store validated service descriptions
         (when-not (seq stored-service-description)
-          (validate service-description-builder core-service-description {:allow-missing-required-fields? false})
-          (validate service-description-builder service-description {:allow-missing-required-fields? false}))
+          (let [{:keys [profile->overrides]} sources]
+            (validate service-description-builder core-service-description
+                      {:allow-missing-required-fields? false
+                       :profile->overrides profile->overrides})
```
sradack, line 1076: Feels like `service-description-builder` should already have this.

waiter/src/waiter/service_description.clj
shamsimam: It is just a name, there is a separate check that validates whether the name is supported, i.e. matches something from the config.

waiter/src/waiter/service_description.clj
shamsimam: If a user provides a profile, it must be non-empty. If they want it empty, then don't specify it.

waiter/src/waiter/core.clj
shamsimam: let's do it in a separate PR, `service-description-defaults` can also move.

waiter/src/waiter/service_description.clj
shamsimam: let's do it in a separate PR, `service-description-defaults` can also move.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1644,3 +1644,30 @@
...
+                                  :token token)
+              register-response (post-token waiter-url token-description)
+              _ (assert-response-status register-response 200)
+              {:keys [body]} (get-token waiter-url token)]
+          (is (= (:profile token-description) (-> body str json/read-str (get "profile"))))
+          (let [service-id (retrieve-service-id waiter-url {"x-waiter-token" token})
+                _ (is service-id "No service created by using token!")
+                service-description (service-id->service-description waiter-url service-id)]
+            (is (= service-description (dissoc token-description :token))))
+          (delete-token-and-assert waiter-url token))))))
```
DaoWen, line 1673: Shouldn't `delete-token-and-assert` call should be in a `finally` block to help avoid leaving tokens around?

waiter/src/waiter/core.clj
```diff
@@ -701,6 +701,15 @@
                       _ (password-store/check-empty-passwords passwords)
                       processed-passwords (mapv #(vector :cached %) passwords)]
                   processed-passwords))
+   :profile->overrides (pc/fnk [[:settings profile-config service-description-constraints]]
+                         (pc/for-map [[profile {:keys [service-parameters]}] profile-config]
```
DaoWen, line 705: Kudos for using `pc/for-map`!

waiter/src/waiter/service_description.clj
DaoWen: Unless there's a compelling reason to _not_ do it now, it seems like it makes more sense to inject it in the right place now rather than merge in new code that we already know we want to immediately go back and restructure.

waiter/src/waiter/settings.clj
```diff
@@ -92,6 +92,7 @@
    ;; TODO port belongs in server-options?
    (s/required-key :port) (s/either schema/positive-int
                                     [schema/positive-int])
+   (s/required-key :profile-config) {schema/non-empty-string schema/profile-definition}
```
DaoWen, line 95: Why are we making this required? It seems like we should always make new keys optional if possible to avoid breaking our existing config files. I see no harm in the profile config coming back `nil` if it's not set (assuming we used nil-safe accessors—which I think we did here since it only seems to be used as a sequence in this patch).

waiter/src/waiter/service_description.clj
shamsimam: Lots of refactoring and adds noise to the PR.
I will do it in a follow-up PR.

waiter/src/waiter/settings.clj
shamsimam: `settings` will already provide a default value (empty map), just like we do for all config we add.

The required is necessary to simplify restructuring in `pc/fnk`

waiter/integration/waiter/token_request_test.clj
shamsimam: Moved to `finally` block

waiter/src/waiter/settings.clj
DaoWen: Ah, I thought this schema was applied directly to the config file. This makes sense now.

PR #1065, opened 2020-04-11
adds support for disabling liveness checks

shamsimam:

- adds support for disabling liveness checks

Liveness and readiness checks are different. The change allows disabling liveness checks on k8s pods.


waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -275,7 +275,8 @@
      :cpus cpus
      :healthChecks [{:protocol (-> (or health-check-proto backend-proto) hu/backend-proto->scheme str/upper-case)
                      :path health-check-url
-                     :gracePeriodSeconds grace-period-secs
+                     ;; configure grace period even when it is 0 as the value defaults to 300 when unspecified
+                     :gracePeriodSeconds (max grace-period-secs 1)
```
sradack, line 279: I was thinking this would be a really large value if 0 to achieve the same effect.  Maybe something like `instance-expiry-mins*60`.  What do you think?

shamsimam: Hmm, sounds good. Will update.

shamsimam: Done.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -275,7 +275,9 @@
      :cpus cpus
      :healthChecks [{:protocol (-> (or health-check-proto backend-proto) hu/backend-proto->scheme str/upper-case)
                      :path health-check-url
-                     :gracePeriodSeconds grace-period-secs
+                     :gracePeriodSeconds (if (pos? grace-period-secs)
+                                           grace-period-secs
+                                           (-> instance-expiry-mins t/minutes t/in-seconds))
```
DaoWen, line 280: Won't this cause the instance to immediately die right when we expect to mark it as expired? That seems to defeat the purpose of expiry. Maybe we should do 1.5 times the expiry period, or just add some fixed amount of time as a buffer (e.g., 5 minutes).

This is the scenario I'm thinking of:

The instance just expired, so we're going to seamlessly create a replacement. But now the database goes down for 30 seconds, so the app is now unhealthy. Since the grace period has passed, Marathon immediately kills the expired instance without waiting for the replacement. The behavior we want is for the expired instance to stay around until the replacement is ready in order to avoid downtime. In this case, maybe the app takes 2 minutes to start a new instance, so now we have a 2 minute downtime instead of just 30 seconds caused by the database blip.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -258,6 +258,22 @@
           (let [service-account (get-pod-service-account waiter-url current-user)]
             (is (= current-user service-account))))))))
 
+(deftest ^:parallel ^:integration-fast test-kubernetes-grace-period-disabled
```
DaoWen, line 261: Why is this a Kubernetes test? Shouldn't this work on all platforms?

DaoWen: In addition to this check, should we also have a check that uses the `--enable-status-change` that you added a while back? We could switch the app to unhealthy and assert that the connection times out (with an override for the timeout so it doesn't actually wait a whole 5 minutes) rather than getting routed to a new instance.

waiter/src/waiter/scheduler/marathon.clj
DaoWen: I just noticed that Steven's original recommendation was `instance-expiry-mins*60`. I think 2x the expiry would be fine (i.e., it "double-expired" so we're ok to kill it now). I'd be worried that if we pick some fixed interval like I suggested as an alternative (e.g., 5 minutes) that it would be too short and we'd end up in the same scenario that I described above.

shamsimam: Increased to twice the `instance-expiry-mins`

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
shamsimam: Moved the test and confirmed that requests get queue timeout due to lack of a healthy instance.

waiter/src/waiter/scheduler/marathon.clj
```diff
@@ -275,7 +275,9 @@
      :cpus cpus
      :healthChecks [{:protocol (-> (or health-check-proto backend-proto) hu/backend-proto->scheme str/upper-case)
                      :path health-check-url
-                     :gracePeriodSeconds grace-period-secs
+                     :gracePeriodSeconds (if (pos? grace-period-secs)
+                                           grace-period-secs
+                                           (-> instance-expiry-mins t/minutes t/in-seconds (* 2)))
```
DaoWen, line 280: Nitpick: Should we add a comment explaining why we're setting it like this? E.g., "Marathon's gracePeriodSeconds has a default value of 300. Using 2x the instance expiry limit is effectively equivalent to an infinite grace period."

shamsimam: added comment

PR #1069, opened 2020-04-15
allows specifying components to include in scheduler state

shamsimam:

- allows specifying components to include in scheduler state

Reduces the amount of scheduler state that is transmitted and rendered when only certain components are of interest.

### DIFF
--- a/waiter/src/waiter/scheduler/composite.clj
+++ b/waiter/src/waiter/scheduler/composite.clj
@@ -101,7 +101,7 @@
 
   (state [_ include-flags]
     (cond-> {:supported-include-params ["aggregator" "components"]
-             :type "CompositeScheduler"}
+             :type "Composite"}
       (contains? include-flags "aggregator")
       (assoc :aggregator (query-aggregator-state-fn))
       (contains? include-flags "components")
--- a/waiter/src/waiter/scheduler/composite.clj
+++ b/waiter/src/waiter/scheduler/composite.clj
@@ -99,9 +99,13 @@
         service-id->scheduler
         (scheduler/service-id->state service-id)))
 
-  (state [_]
-    {:aggregator (query-aggregator-state-fn)
-     :components (pc/map-vals scheduler/state scheduler-id->scheduler)})
+  (state [_ include-flags]
+    (cond-> {:supported-include-params ["aggregator" "components"]
+             :type "CompositeScheduler"}
+      (contains? include-flags "aggregator")
+      (assoc :aggregator (query-aggregator-state-fn))
+      (contains? include-flags "components")
+      (assoc :components (pc/map-vals #(scheduler/state % include-flags) scheduler-id->scheduler))))
 
   (validate-service [_ service-id]
     (-> service-id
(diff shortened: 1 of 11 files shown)


waiter/src/waiter/scheduler/composite.clj
```diff
@@ -99,9 +99,13 @@
         service-id->scheduler
         (scheduler/service-id->state service-id)))
 
-  (state [_]
-    {:aggregator (query-aggregator-state-fn)
-     :components (pc/map-vals scheduler/state scheduler-id->scheduler)})
+  (state [_ include-flags]
+    (cond-> {:supported-include-params ["aggregator" "components"]
+             :type "CompositeScheduler"}
```
DaoWen, line 104: Isn't it redundant to include the word "Scheduler" in each of these scheduler types?

shamsimam: Dropped the `Scheduler` portion from the types

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -686,10 +686,15 @@
...
-    {:id->service @id->service-agent
-     :port->reservation @port->reservation-atom
-     :syncer (retrieve-syncer-state-fn)})
+  (state [_ include-flags]
+    (cond-> {:supported-include-params ["port-reservation" "services" "syncer"]
+             :type "ShellScheduler"}
+      (contains? include-flags "port-reservation")
+      (assoc :port->reservation @port->reservation-atom)
+      (contains? include-flags "services")
+      (assoc :id->service @id->service-agent)
```
DaoWen, line 695: I think it's very confusing that the "include" flag doesn't match the key here. (And this isn't the only one.)

Is there a reason you're not just using the key name as the flag name?

shamsimam: The key name is very wordy, so used a shorter variant. The include flags are passed in the response so users don't have to guess what flags are supported.

shamsimam: Changed the names to match the keys

sradack: Nick looks like he covered this.

waiter/src/waiter/scheduler/composite.clj
sradack: Nick looks like he covered this.

sradack: I liked it how it was before.  The type name is supposed to match the actual type name, yes?

shamsimam: I'm happy to revert it :)

shamsimam: Reverted to include `Scheduler` suffix.

DaoWen: I still disagree that it's useful if _every possible value_ of this string is _always_ going to end in "Scheduler", but this is just an aesthetic thing so I'm not going to push it. I don't anticipate ever using this field for anything other than visually confirming that I'm looking at the state that I expected.

PR #1072, opened 2020-04-16
allows token parameters in profile defaults

shamsimam:

- renames service-parameters to defaults
- allows token parameters in profile defaults

Allow configuring token parameters in profiles.

### Example logs:

Error message during startup when profile has unsupported keys:
```
2020-04-16 01:56:49,065 ERROR waiter.main [main] -  encountered exception starting waiter
clojure.lang.ExceptionInfo: Profile has unsupported entries in defaults {:name "webapp", :profile-defaults {"concurrency-level" 120, "cpus" 1, "load-balancing" "random", "foo" "bar"}, :unsupported-keys ("foo")}
        at waiter.core$fn__47420$fnk47418_positional__47421.invoke(core.clj:716)
```

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -917,6 +917,10 @@
                                  (fn async-trigger-terminate-fn [target-router-id service-id request-id]
                                    (async-req/async-trigger-terminate
                                      async-request-terminate-fn make-inter-router-requests-sync-fn router-id target-router-id service-id request-id)))
+   :attach-token-defaults-fn (pc/fnk [[:settings [:token-config token-defaults]]
+                                      [:state profile->defaults]]
+                               (fn attach-token-defaults-fn [token-parameters]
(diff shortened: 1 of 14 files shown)


waiter/src/waiter/core.clj
```diff
@@ -703,14 +703,26 @@
...
+                              (throw (ex-info "Profile has unsupported entries in defaults"
+                                              {:name profile
+                                               :profile-defaults defaults
+                                               :unsupported-keys unsupported-keys})))
                             ;; validate the profile's service parameters
-                            (sd/validate-schema service-parameters max-constraints-schema initial-profile->defaults
+                            (sd/validate-schema defaults max-constraints-schema initial-profile->defaults
                                                 {:allow-missing-required-fields? true})
-                            service-parameters)))
+                            ;; validate the profile's service parameters
```
sradack, line 723: comment is wrong

shamsimam: fixed

PR #1074, opened 2020-04-16
excludes token parameters while comparing service description from profile test

shamsimam:

- excludes token parameters while comparing service description from profile test


shamsimam:
self-merging

PR #1076, opened 2020-04-20
avoids passing user-agent field to backend if client didn't provide one

shamsimam:

- avoids passing user-agent field to backend if client didn't provide one

Waiter shouldn't additional non-hop-by-hop headers to proxied requests.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.5.1

### DIFF
--- a/waiter/src/waiter/util/http_utils.clj
+++ b/waiter/src/waiter/util/http_utils.clj
@@ -102,9 +102,9 @@
       (.clear (.getContentDecoderFactories client)))
     (.setCookieStore client (HttpCookieStore$Empty.))
     (.setDefaultRequestContentType client nil)
-    (when user-agent
-      (let [new-user-agent-field (HttpField. HttpHeader/USER_AGENT (str user-agent))]
-        (.setUserAgentField client new-user-agent-field)))
+    (.setUserAgentField client
+                        (when-not (str/blank? user-agent)
+                          (HttpField. HttpHeader/USER_AGENT (str user-agent))))
     client))
 
 (defn- prepare-http2-transport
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/util/http_utils.clj
```diff
@@ -125,11 +125,11 @@
   (let [http2-transport (prepare-http2-transport conn-timeout socket-timeout)]
     {:http1-client (http-client-factory (cond-> config
                                           client-name (update :client-name str "-http1")
-                                          user-agent (update :user-agent str ".http1")))
+                                          (not (str/blank? user-agent)) (update :user-agent str ".http1")))
      ;; prepare-http2-transport already handles the connect and socket timeouts
      :http2-client (-> (cond-> (dissoc config :conn-timeout :socket-timeout)
                          client-name (update :client-name str "-http2")
-                         user-agent (update :user-agent str ".http2"))
+                         (not (str/blank? user-agent)) (update :user-agent str ".http2"))
```
DaoWen, line 132: It looks like this is always blank:
https://github.com/twosigma/waiter/blob/e96bcdde1622aa153233a5d6b6410b0995490464/waiter/src/waiter/core.clj#L660-L666

Should we set it to something more useful?

shamsimam: There are other code paths where the user-agent isn't blank, .e.g in the tests.

waiter/src/waiter/util/http_utils.clj
```diff
@@ -102,9 +102,9 @@
       (.clear (.getContentDecoderFactories client)))
     (.setCookieStore client (HttpCookieStore$Empty.))
     (.setDefaultRequestContentType client nil)
-    (when user-agent
-      (let [new-user-agent-field (HttpField. HttpHeader/USER_AGENT (str user-agent))]
-        (.setUserAgentField client new-user-agent-field)))
+    (.setUserAgentField client
+                        (when-not (str/blank? user-agent)
+                          (HttpField. HttpHeader/USER_AGENT (str user-agent))))
```
shamsimam, line 107: This is the main fix where we are explicitly setting the user-agent field to null instead of letting it be the default.
https://github.com/eclipse/jetty.project/blob/fa54c749466a1a2b032ae284c73ab559493c07d6/jetty-client/src/main/java/org/eclipse/jetty/client/HttpClient.java#L138

PR #1078, opened 2020-04-21
includes service resource usage in /apps endpoints

shamsimam:

- includes service resource usage in /apps endpoints

Avoid the need for clients to compute this value.

Screen Shot 2020-04-21 at 1 08 31 PM
Screen Shot 2020-04-21 at 1 08 20 PM

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -287,7 +287,7 @@
       (or (empty? param-predicates)
           (every? #(%1 service-description) param-predicates)))))
 
-(defn compute-service-usage
+(defn compute-resource-usage
   "Computes the resources used to a service."
   [{:strs [cpus mem]} instance-count]
   {:cpus (* instance-count cpus)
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -330,15 +336,19 @@
...
                                  :service-id service-id
-                                 :service-description service-description
+                                 :service-description core-service-description
                                  :status (service/retrieve-service-status-label service-id global-state)
-                                 :url (prepend-waiter-url (str "/apps/" service-id))}
+                                 :url (prepend-waiter-url (str "/apps/" service-id))
+                                 :usage usage-map}
                                 include-effective-parameters?
                                 (assoc :effective-parameters
                                        (service-id->service-description-fn service-id :effective? true))
```
sradack, line 354: We're doing this twice.

waiter/src/waiter/handler.clj
```diff
@@ -330,15 +336,19 @@
...
-                                {:instance-counts (retrieve-instance-counts service-id)
+                                {:instance-counts instance-counts
                                  :last-request-time (get-in service-id->metrics [service-id "last-request-time"])
                                  :service-id service-id
-                                 :service-description service-description
+                                 :service-description core-service-description
                                  :status (service/retrieve-service-status-label service-id global-state)
-                                 :url (prepend-waiter-url (str "/apps/" service-id))}
+                                 :url (prepend-waiter-url (str "/apps/" service-id))
+                                 :usage usage-map}
```
sradack, line 351: Nitpick: I think you're right.  `:resource-usage` feels better to me.

sradack: I considered naming this `:resources-requested` (or something similar) and and reserve `:resource-usage` to mean actual usage.

sradack: Ultimately decided `:resources-requested` feels awfully k8s-specific.

waiter/src/waiter/handler.clj
shamsimam: Removed duplicate computation.

waiter/src/waiter/handler.clj
shamsimam: Renamed to `resource-usage`

PR #1090, opened 2020-05-08
adds support for last-update-time string in admin token update

shamsimam:

- adds support for last-update-time string in admin token update

Maintain consistency between token output json and input accepted during token update.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -24,6 +24,7 @@
             [waiter.kv :as kv]
             [waiter.service-description :as sd]
             [waiter.status-codes :refer :all]
+            [waiter.util.date-utils :as du]
             [waiter.util.ring-utils :as ru]
             [waiter.util.utils :as utils])
   (:import (org.joda.time DateTime)))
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/token.clj
```diff
@@ -495,12 +508,16 @@
                         {:previous previous :status http-400-bad-request :token token :log-level :warn}))))
 
     ; Store the token
-    (let [new-token-metadata (merge {"cluster" (calculate-cluster cluster-calculator request)
+    (let [{:strs [last-update-time] :as new-token-metadata}
+          (merge {"cluster" (calculate-cluster cluster-calculator request)
                                      "last-update-time" (.getMillis ^DateTime (clock))
```
DaoWen, line 513: Indentation needs to be fixed here.

waiter/src/waiter/token.clj
```diff
@@ -495,12 +508,16 @@
...
                                      "last-update-user" authenticated-user
                                      "owner" owner
                                      "root" (or (get existing-token-metadata "root") token-root)}
                                     new-token-metadata)
+          new-token-metadata (cond-> new-token-metadata
+                               (string? last-update-time)
+                               (assoc "last-update-time" (parse-last-update-time last-update-time)))
```
DaoWen, line 520: I think using `update` rather than `assoc` here would help make it more clear that an existing value is being replaced, rather than a new entry is being added to the map.

```suggestion
                               (update"last-update-time" parse-last-update-time))
```

waiter/src/waiter/token.clj
shamsimam: Reformatted.

waiter/src/waiter/token.clj

PR #1091, opened 2020-05-08
CLI: avoid using user-provided service-description to load cpu and mem

shamsimam:
uses effective service description to retrieve `cpus` and `mem`

- uses resource usage for `cpus` and `mem`
- uses effective service description to retrieve `cpus` and `mem`

With profiles, `cpus` and `mem` may come from the profile and will not necessarily be available on the user-provided service description.

### Sample output:
```
$ waiter show wtrttsfs275748767724030.127.0.0.1

=== WAITER1 / wtrttsfs275748767724030.127.0.0.1 ===
...
# Services    3
# Failing     0
# Instances   3
Total Memory  768 MiB
Total CPUs    0.3

Service Id                                                          Instances  Run as user      CPUs  Memory    Version    Status    Last request    Current?
w9091-wtrttsfs275748767724030v3-1031cfbf922ed536561eab73681b8fb6            1  shamsimam         0.1  256 MiB   version-3  Running   6 minutes ago   Current
w9091-wtrttsfs275748767724030v2-ac72cc50c1e79ec1b9cd79f6e3fced6a            1  shamsimam         0.1  256 MiB   version-2  Running   7 minutes ago   Not Current
w9091-wtrttsfs275748767724030v1-ad290c15643910b39dd9524bf8799089            1  shamsimam         0.1  256 MiB   version-1  Running   8 minutes ago   Not Current
```


cli/waiter/subcommands/show.py
```diff
@@ -20,25 +21,34 @@ def format_using_current_token(service, token_etag, token_name):
...
+
+
 def tabulate_token_services(services, token_etag, token_name):
     """Returns a table displaying the service info"""
     num_services = len(services)
     if num_services > 0:
         num_failing_services = len([s for s in services if s['status'] == 'Failing'])
-        total_mem_usage = format_memory_amount(sum(s['service-description']['mem'] for s in services))
-        total_cpu_usage = round(sum(s['service-description']['cpus'] for s in services), 2)
+        num_instances = reduce(lambda acc, s: acc + retrieve_num_instances(s), services, 0)
```
dposada, line 35: Maybe this?

```suggestion
        num_instances = sum(retrieve_num_instances(s) for s in services)
```

waiter/src/waiter/handler.clj
```diff
@@ -331,7 +331,7 @@
                                       :unhealthy-instances (-> service-id->unhealthy-instances (get service-id) count)})
           service-id->metrics (service-id->metrics-fn)
           include-effective-parameters? (or (utils/request-flag request-params "effective-parameters")
-                                            (utils/param-contains? request-params "include" "request-info"))
```
dposada, line 334: Why are we removing `request-info` here?

shamsimam: This is a copy-paste bug, the flag should be called `effective-parameters`.
I couldn't use the `include=effective-parameters` query parameter as a result in the CLI.

cli/waiter/subcommands/show.py
shamsimam: Done.

PR #1097, opened 2020-05-15
adds x-waiter-jwt header containing JWT access token to backend requests

shamsimam:

- adds x-waiter-jwt header containing JWT access token to backend requests

The JWT access token should be available to the backend in case it wants to perform its own authorization.

### DIFF
--- a/waiter/src/waiter/auth/authentication.clj
+++ b/waiter/src/waiter/auth/authentication.clj
@@ -52,23 +52,27 @@
                     {:status http-400-bad-request}))))
 
 (defn- add-cached-auth
-  [response password principal age-in-seconds]
-  (cookie-support/add-encoded-cookie response password AUTH-COOKIE-NAME [principal (tc/to-long (t/now))]
-                                     (or age-in-seconds (-> 1 t/days t/in-seconds))))
+  [response password principal age-in-seconds auth-metadata]
+  (let [cookie-value [principal (tc/to-long (t/now)) auth-metadata]
+        cookie-expiry (or age-in-seconds (-> 1 t/days t/in-seconds))]
+    (cookie-support/add-encoded-cookie
+      response password AUTH-COOKIE-NAME cookie-value cookie-expiry)))
 
 (defn select-auth-params
   "Returns a map that contains only the auth params from the input map"
   [m]
   (select-keys m [:authorization/method :authorization/principal :authorization/user]))
 
-(defn auth-params-map
+(defn build-auth-params-map
   "Creates a map intended to be merged into requests/responses."
   ([method principal]
-   (auth-params-map method principal (first (str/split principal #"@" 2))))
-  ([method principal user]
-   {:authorization/method method
-    :authorization/principal principal
-    :authorization/user user}))
+   (build-auth-params-map method principal nil))
+  ([method principal metadata]
+   (let [user (utils/principal->username principal)]
+     (cond-> {:authorization/method method
+              :authorization/principal principal
+              :authorization/user user}
+       metadata (assoc :authorization/metadata metadata)))))
 
 (defn request-authenticated?
   "Returns true if the authorization info is already available in the input map."
@@ -78,13 +82,14 @@
 (defn handle-request-auth
   "Invokes the given request-handler on the given request, adding the necessary
   auth headers on the way in, and the x-waiter-auth cookie on the way out."
-  ([handler request method principal password]
-   (handle-request-auth handler request principal (auth-params-map method principal) password nil))
-  ([handler request principal auth-params-map password age-in-seconds]
-   (let [handler' (middleware/wrap-merge handler auth-params-map)]
+  ([handler request auth-params-map password]
+   (handle-request-auth handler request auth-params-map password nil))
+  ([handler request auth-params-map password age-in-seconds]
+   (let [{:keys [authorization/metadata authorization/principal]} auth-params-map
+         handler' (middleware/wrap-merge handler auth-params-map)]
      (-> request
        handler'
-       (add-cached-auth password principal age-in-seconds)))))
+       (add-cached-auth password principal age-in-seconds metadata)))))
 
 (defn decode-auth-cookie
   "Decodes the provided cookie using the provided password.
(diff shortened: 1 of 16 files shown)


waiter/src/waiter/auth/authentication.clj
```diff
@@ -52,23 +52,27 @@
...
+    (cookie-support/add-encoded-cookie
+      response password AUTH-COOKIE-NAME cookie-value cookie-expiry)))
 
 (defn select-auth-params
   "Returns a map that contains only the auth params from the input map"
   [m]
   (select-keys m [:authorization/method :authorization/principal :authorization/user]))
 
-(defn auth-params-map
+(defn build-auth-params-map
```
shamsimam, line 66: Rename function to avoid confusion with `auth-params-map` variables often used to store the result in callers.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -78,13 +82,14 @@
 (defn handle-request-auth
   "Invokes the given request-handler on the given request, adding the necessary
   auth headers on the way in, and the x-waiter-auth cookie on the way out."
-  ([handler request method principal password]
-   (handle-request-auth handler request principal (auth-params-map method principal) password nil))
-  ([handler request principal auth-params-map password age-in-seconds]
```
shamsimam, line 83: The principal was duplicated in `auth-params-map`.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -104,12 +109,17 @@
   "Verifies whether the decoded authenticated cookie is valid as per the following rules:
    The decoded value must be a sequence in the format: [auth-principal auth-time].
    In addition, the auth-principal must be a string and the auth-time must be less than a day old."
-  [[auth-principal auth-time :as decoded-auth-cookie]]
-  (log/debug "well-formed?" decoded-auth-cookie (integer? auth-time) (string? auth-principal) (= 2 (count decoded-auth-cookie)))
+  [[auth-principal auth-time auth-metadata :as decoded-auth-cookie]]
```
shamsimam, line 112: The cookie can now contain an optional metadata entry.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -182,9 +193,10 @@
   [password handler]
   (fn auth-cookie-handler
     [{:keys [headers] :as request}]
-    (let [[auth-principal _ :as decoded-auth-cookie] (get-and-decode-auth-cookie-value headers password)]
+    (let [decoded-auth-cookie (get-and-decode-auth-cookie-value headers password)
+          [auth-principal _ auth-metadata] decoded-auth-cookie]
       (if (decoded-auth-valid? decoded-auth-cookie)
-        (let [auth-params-map (auth-params-map :cookie auth-principal)
```
shamsimam, line 187: Example of where the function name and local variable name is the same.

waiter/src/waiter/auth/kerberos.clj
```diff
@@ -33,7 +33,7 @@
   (try
     (let [{:keys [exit out err]} (shell/sh "krb5_prestash" "query" "host" host)]
       (if (zero? exit)
-        (set (map #(first (str/split % #"@" 2)) (str/split-lines out)))
+        (set (map #(utils/principal->username %) (str/split-lines out)))
```
shamsimam, line 36: Helper function to avoid duplicate code to extract the username.

waiter/src/waiter/auth/saml.clj
```diff
@@ -203,8 +203,8 @@
         ; https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/technical-reference/the-role-of-claims
         upn (first (get attrs "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/upn"))
         saml-principal (or upn name-id-value)
-        {:keys [authorization/principal authorization/user]} (auth/auth-params-map :saml saml-principal)
```
shamsimam, line 206: Avoid unnecessarily building the `auth-params-map` just for the principal and user.

waiter/src/waiter/handler.clj
```diff
@@ -46,12 +46,6 @@
             [waiter.util.utils :as utils])
   (:import (java.io InputStream)))
 
-(defn make-auth-user-map
```
shamsimam, line 49: Moved to `process_request`

waiter/src/waiter/process_request.clj
```diff
@@ -52,6 +52,13 @@
            (org.eclipse.jetty.io EofException)
            (org.eclipse.jetty.server HttpChannel HttpOutput Response)))
 
+(defn make-auth-user-map
```
shamsimam, line 55: Moved from handler.clj and added TODO in the comment below

waiter/src/waiter/auth/authentication.clj
```diff
@@ -52,23 +52,27 @@
                     {:status http-400-bad-request}))))
 
 (defn- add-cached-auth
-  [response password principal age-in-seconds]
-  (cookie-support/add-encoded-cookie response password AUTH-COOKIE-NAME [principal (tc/to-long (t/now))]
-                                     (or age-in-seconds (-> 1 t/days t/in-seconds))))
+  [response password principal age-in-seconds auth-metadata]
+  (let [cookie-value [principal (tc/to-long (t/now)) auth-metadata]
```
sradack, line 56: Should we avoid adding `auth-metadata` to `cookie-value` if `auth-metadata` is nil?

waiter/src/waiter/process_request.clj
```diff
@@ -52,6 +52,13 @@
...
+  "Creates a map containing the username and principal from a request"
+  [{:keys [authorization/metadata authorization/principal]}]
+  ;; TODO creating this map is unnecessary, we can destructure from the request directly
```
sradack, line 58: This seems like a small TODO.  Can we either do it or remove the comment if we feel like we should leave this function in place?

waiter/src/waiter/headers.clj
```diff
@@ -113,13 +113,12 @@
...
-  ([headers username principal]
-   (cond-> headers
-     username (assoc "x-waiter-auth-principal" username)
-     principal (assoc "x-waiter-authenticated-principal" principal))))
+  ([headers principal {:keys [access-token]}]
+   (let [username (utils/principal->username principal)]
+     (cond-> headers
+       username (assoc "x-waiter-auth-principal" username)
+       principal (assoc "x-waiter-authenticated-principal" principal)
+       access-token (assoc "x-waiter-jwt" access-token)))))
```
sradack, line 121: We should only add `access-token` if the method is `:jwt`.  In theory other auth methods might add `access-token` and it wouldn't be correct to add `x-waiter-jwt`.

waiter/src/waiter/auth/authentication.clj
shamsimam: Avoided adding the metadata entry into the cookie.

waiter/src/waiter/headers.clj
shamsimam: Changed the key to `jwt-access-token`. When we support `OIDC`, the method will be `oidc` and will have a token to send to the backend with the same key.

waiter/src/waiter/process_request.clj
shamsimam: Removed the TODO, do not want to introduce those changes in this PR.

PR #1102, opened 2020-05-18
adds support for filtering tokens list by specific parameter values

shamsimam:

- adds support for filtering tokens list by specific parameter values

We want to query for tokens configured with specific values on the parameters.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -378,6 +378,13 @@
                        :token token
                        :log-level :warn})))))
 
+(defn- token-description->editable-token-parameters
+  "Retrieves the user editable token parameters from the token description."
+  [token-description]
+  (-> (merge (:service-parameter-template token-description)
+             (:token-metadata token-description))
+    (select-keys sd/token-user-editable-keys)))
+
 (defn- handle-post-token-request
   "Validates that the user is the creator of the token if it already exists.
    Then, updates the configuration for the token in the database using the newest password."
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/token.clj
```diff
@@ -584,25 +589,49 @@
...
+                             (fn list-tokens-parameters-predicate [[token _]]
+                               (let [token-parameters (token->token-parameters token)]
+                                 (every? #(% token-parameters) parameter-filter-predicates))))
                            (map
                              (fn [[token entry]]
                                (-> (if show-metadata
                                      (update entry :last-update-time tc/from-long)
                                      (dissoc entry :deleted :etag :last-update-time))
-                                   (assoc :owner owner :token token)))))))
+                                 (assoc :owner owner :token token)))))))
```
DaoWen, line 634: I disagree with the change in indentation here.

https://guide.clojure.style/#threading-macros
https://github.com/clojure/clojure/blob/ee3553362de9bc3bfd18d4b0b3381e3483c2a34c/src/clj/clojure/core.clj#L4095-L4096

shamsimam: The body indentation mechanism makes editing uses of the arrow macros easier.
E.g. changing `->` to `some->` will not introduce additional whitespace changes.

waiter/src/waiter/token.clj
```diff
@@ -584,25 +589,49 @@
...
+                                                        (contains? allowed-parameter-values
+                                                                   (str (get token-parameters parameter-name))))))
+                                               (pc/map-vals
+                                                 (fn [params]
+                                                   (cond
+                                                     (string? params) #{params}
+                                                     :else (set params)))
+                                                 (-> request-params
+                                                   (select-keys sd/token-user-editable-keys)
+                                                   (dissoc "owner"))))
```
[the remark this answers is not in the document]
shamsimam: Done (didn't commit this since the `set` in `filterable-param?` was unnecessary.

waiter/src/waiter/token.clj
DaoWen: I wouldn't have commented if this were new code, but this changed was applied to a line that was otherwise untouched in the patch. Our team just needs to decide on a standard to avoid auto format noise like this in the future.

PR #1113, opened 2020-06-02
Instance tracking logs and InstanceTracker Logger

pandashadopan:
- Added part of instance tracking logging 

- To have logs for instance tracking


waiter/bin/run-using-minimesos.sh
```diff
@@ -17,6 +17,8 @@ if [ ${EXIT_CODE} -eq 0 ]
 then
     export WAITER_MARATHON=${MINIMESOS_MARATHON%;}
     export WAITER_ZOOKEEPER_CONNECT_STRING=$(echo ${MINIMESOS_ZOOKEEPER} | awk -F/ '{print $3}')
+    export GRAPHITE_SERVER_PORT=5555
+    export JWKS_SERVER_URL=8040
```
DaoWen, line 21: Should this just be the port number? Or should there be a `localhost` here?

waiter/src/waiter/scheduler.clj
```diff
@@ -76,7 +76,8 @@
    ^String log-directory
    ^String message])
 
-(defn make-ServiceInstance [value-map]
+(defn make-ServiceInstance [{id :id started-at :started-at :as value-map}]
+  (log/log "InstanceTracker" :debug nil (str "Created an instance with id: " id " that started at " started-at))
```
DaoWen, line 80: We usually prefer creating a map with the fields we're interested in, and then logging the map. Same for extra data included in exception error messages.
```clojure
(log/log "InstanceTracker" :debug (str "Created an instance " {:id id :started-at started-at}))
```

waiter/src/waiter/handler.clj
```diff
@@ -176,6 +176,10 @@
                                                            (when started-at
                                                              (du/str-to-date started-at)))))]
                     (notify-instance-killed-fn instance)))
+                (log/log "InstanceTracker" :debug nil (str "Instance " instance-id " BLACKLIST " (utils/clj->json
```
DaoWen, line 179: I would prefer if you declared a macro similar to `scheduler/log` that delegates to `(log/log "InstanceTracker" :debug nil ...)`.

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(defmacro logI
```
DaoWen, line 50: Let's rename this to `log-service-instance`.

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
...
+  "Log InstanceTracker-specific messages."
+  [instance]
+  `(log/log "InstanceTracker" :debug nil (utils/clj->json instance)))
```
DaoWen, line 53: We want to `assoc` a couple of things here:
* `leader?`
* event-type (keepalive, running, killed, healthy, etc.)

waiter/src/waiter/scheduler/cook.clj
```diff
@@ -308,6 +308,8 @@
                             allowed-users)
         failed-instances (map job->service-instance failed-jobs)]
     (when (seq failed-instances)
+      (doseq [failed-instance failed-instances]
+        (scheduler/logI (assoc failed-instance :event-type "FAIL")))
```
DaoWen, line 312: `(scheduler/log-service-instance failed-instance "FAIL")`

waiter/src/waiter/scheduler.clj
DaoWen: We're not going to do the `leader?` field right now. Too much refactoring involved.

waiter/src/waiter/scheduler.clj
```diff
@@ -889,10 +905,10 @@
                                   (concat healthy-instances))
             known-instance-ids' (->> known-instances' (map :id) set)
             previously-known-instance? (comp known-instance-ids :id)
-            new-instance-ids (->> known-instances'
-                                  (remove previously-known-instance?)
-                                  (sort instance-comparator)
-                                  (mapv :id))
+            new-instances (->> known-instances'
```
DaoWen, line 908: Let's just log all of the instances from this collection with `:launch` as the event-type.

waiter/src/waiter/state/maintainer.clj
```diff
@@ -642,6 +642,13 @@
                                                        (if (seq new-instance-ids) (str "New healthy instances: " new-instance-ids ".") "")
```
DaoWen, line 642: We should be logging new healthy (and unhealthy) instances in our instance log.

waiter/src/waiter/state/maintainer.clj
```diff
@@ -628,11 +628,20 @@
...
-                                                 prev-instance-ids (set (map :id (get service-id->healthy-instances service-id)))
-                                                 new-instance-ids (filterv (complement prev-instance-ids) curr-instance-ids)
-                                                 rem-instance-ids (filterv (complement curr-instance-ids) prev-instance-ids)
+                                           (let [curr-instances (set healthy-instances)
+                                                 prev-instances (set (get service-id->healthy-instances service-id))
+                                                 new-instances (filterv (complement prev-instances) curr-instances)
+                                                 rem-instances (filterv (complement curr-instances) prev-instances)
+                                                 new-instance-ids (mapv :id new-instances)
+                                                 rem-instance-ids (mapv :id rem-instances)
+
```
DaoWen, line 637: Let's remove this blank line.

```suggestion
```

waiter/src/waiter/state/maintainer.clj
```diff
@@ -642,6 +651,13 @@
                                                        (if (seq new-instance-ids) (str "New healthy instances: " new-instance-ids ".") "")
                                                        (if (seq rem-instance-ids) (str "Removed healthy instances: " rem-instance-ids ".") "")
                                                        (if (seq unhealthy-instance-ids) (str "Unhealthy instances: " unhealthy-instance-ids ".") ""))))
+
```
DaoWen, line 654: Let's remove this blank line.

```suggestion
```

waiter/src/waiter/state/responder.clj
```diff
@@ -470,10 +471,10 @@
...
       {:current-state' (if (= :blacklisted response-code)
                          (-> current-state
-                           ; mark instance as blacklisted and set the expiry time
-                           (update-in [:instance-id->state instance-id] sanitize-instance-state)
-                           (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :blacklisted))
-                           (update-state-by-blacklisting-instance-fn cid instance-id blacklist-period-ms))
+                             ; mark instance as blacklisted and set the expiry time
+                             (update-in [:instance-id->state instance-id] sanitize-instance-state)
+                             (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :blacklisted))
+                             (update-state-by-blacklisting-instance-fn cid instance-id blacklist-period-ms))
```
DaoWen, line 477: Shams and I have been debating what our standard indention style should be for threading macros. For now, I think's its best to not change it if you're not actually editing. Please un-indent these to revert of the whitespace change.

waiter/src/waiter/state/responder.clj
```diff
@@ -482,9 +483,10 @@
   [{:keys [instance-id->state] :as current-state} update-instance-id->blacklist-expiry-time-fn update-status-tag-fn
    instance-id expiry-time]
   (log/info "unblacklisting instance" instance-id "as blacklist expired at" expiry-time)
-  (cond-> (update-instance-id->blacklist-expiry-time-fn current-state #(dissoc % instance-id))
```
DaoWen, line 485: `cond->` returns the first argument unmodified if all the conditions are falsy. The `when` version will return `nil` instead.

waiter/src/waiter/state/responder.clj
```diff
@@ -470,10 +471,10 @@
...
       {:current-state' (if (= :blacklisted response-code)
                          (-> current-state
-                           ; mark instance as blacklisted and set the expiry time
-                           (update-in [:instance-id->state instance-id] sanitize-instance-state)
-                           (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :blacklisted))
-                           (update-state-by-blacklisting-instance-fn cid instance-id blacklist-period-ms))
+                               ; mark instance as blacklisted and set the expiry time
+                               (update-in [:instance-id->state instance-id] sanitize-instance-state)
+                               (update-in [:instance-id->state instance-id] update-status-tag-fn #(conj % :blacklisted))
+                               (update-state-by-blacklisting-instance-fn cid instance-id blacklist-period-ms))
```
DaoWen, line 477: Whitespace here is off.

pandashadopan: Resolved

waiter/bin/run-using-k8s.sh
```diff
@@ -8,6 +8,10 @@
 # Runs Waiter, configured to use the Kubernetes scheduler.
 
 export WAITER_PORT=${1:-9091}
+export GRAPHITE_SERVER_PORT=5555
+export JWKS_PORT=8040
+export JWKS_SERVER_URL="http://127.0.0.1:${JWKS_PORT}/keys"
+export WAITER_TEST_JWT_ACCESS_TOKEN_URL="http://127.0.0.1:${JWKS_PORT}/get-token?host={HOST}"
```
shamsimam, line 14: Please move the test script (for k8s and minimesos) changes to separate PRs.

waiter/bin/run-using-minimesos.sh
```diff
@@ -17,6 +17,10 @@ if [ ${EXIT_CODE} -eq 0 ]
 then
     export WAITER_MARATHON=${MINIMESOS_MARATHON%;}
     export WAITER_ZOOKEEPER_CONNECT_STRING=$(echo ${MINIMESOS_ZOOKEEPER} | awk -F/ '{print $3}')
+    export GRAPHITE_SERVER_PORT=5555
+    export JWKS_PORT=8040
+    export JWKS_SERVER_URL="http://127.0.0.1:${JWKS_PORT}/keys"
+    export WAITER_TEST_JWT_ACCESS_TOKEN_URL="http://127.0.0.1:${JWKS_PORT}/get-token?host={HOST}"
```
shamsimam, line 23: Please move the test script (for k8s and minimesos) changes to separate PRs.

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(defmacro log-service-instance
+  "Log InstanceTracker-specific messages."
+  [instance event-type]
+  `(log/log "InstanceTracker" :debug nil (utils/clj->json (assoc ~instance :timestamp (t/now) :event-type ~event-type))))
```
shamsimam, line 53: Why is this set at `debug` by default?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -205,6 +205,7 @@
                                               ;; we don't include the exit code in failed instances that were killed by k8s.
                                               (not (killed-by-k8s? newest-failure))
                                               (assoc :exit-code (:exitCode newest-failure)))]
+                (scheduler/log-service-instance newest-failure-instance :fail)
                 (swap! service-id->failed-instances-transient-store
                        update-in [service-id] assoc newest-failure-id newest-failure-instance)))))))
```
shamsimam, line 210: We should wrap the `service-id->failed-instances-transient-store` update in a function and that function can log the service instance. E.g. that function can be reused in the marathon scheduler below.

waiter/src/waiter/scheduler/shell.clj
```diff
@@ -363,12 +363,15 @@
...
       (release-port! port->reservation-atom port port-grace-period-ms)
-      (assoc instance
-        :exit-code exit-value
-        :failed? (if (zero? exit-value) false true)
-        :healthy? false
-        :killed? true ; does not actually mean killed -- using this to mark inactive
-        :message message))
+      (let [newly-constructed-instance (assoc instance
+                                         :exit-code exit-value
+                                         :failed? (if (zero? exit-value) false true)
```
shamsimam, line 368: Extract `failed?` into a local variable?

waiter/src/waiter/state/maintainer.clj
```diff
@@ -628,11 +628,19 @@
                                                                               (dissoc service-id->instability-issue service-id))]
                                          (when (or (not= (get service-id->healthy-instances service-id) healthy-instances)
                                                    (not= (get service-id->unhealthy-instances service-id) unhealthy-instances))
-                                           (let [curr-instance-ids (set (map :id healthy-instances))
-                                                 prev-instance-ids (set (map :id (get service-id->healthy-instances service-id)))
-                                                 new-instance-ids (filterv (complement prev-instance-ids) curr-instance-ids)
-                                                 rem-instance-ids (filterv (complement curr-instance-ids) prev-instance-ids)
+                                           (let [curr-instances (set healthy-instances)
+                                                 prev-instances (set (get service-id->healthy-instances service-id))
+                                                 new-instances (filterv (complement prev-instances) curr-instances)
```
shamsimam, line 633: These can remain as `filter` instead of `filterv`

waiter/src/waiter/state/maintainer.clj
```diff
@@ -628,11 +628,19 @@
...
+                                                 rem-instances (filterv (complement curr-instances) prev-instances)
```
shamsimam, line 634: These can remain as `filter` instead of `filterv`

waiter/src/waiter/state/maintainer.clj
```diff
@@ -628,11 +628,19 @@
...
+                                                 new-instance-ids (mapv :id new-instances)
+                                                 rem-instance-ids (mapv :id rem-instances)
+                                                 prev-unhealthy-instances (set (get service-id->unhealthy-instances service-id))
+                                                 new-unhealthy-instances (filterv (complement prev-unhealthy-instances) unhealthy-instances)
```
shamsimam, line 638: These can remain as `filter` instead of `filterv`

waiter/src/waiter/state/maintainer.clj
```diff
@@ -711,7 +725,9 @@
       {:go-chan go-chan
        :notify-instance-killed-fn (fn notify-router-state-maintainer-of-instance-killed [instance]
                                     (log/info "received notification of killed instance" (:id instance))
-                                    (async/go (async/>! kill-notification-chan {:instance instance})))
+                                    (scheduler/log-service-instance instance :kill)
+                                    (async/go
+                                      (async/>! kill-notification-chan {:instance instance})))
```
shamsimam, line 730: Nitpick: unnecessary formatting

waiter/src/waiter/state/maintainer.clj
```diff
@@ -711,7 +725,9 @@
       {:go-chan go-chan
        :notify-instance-killed-fn (fn notify-router-state-maintainer-of-instance-killed [instance]
                                     (log/info "received notification of killed instance" (:id instance))
-                                    (async/go (async/>! kill-notification-chan {:instance instance})))
+                                    (scheduler/log-service-instance instance :kill)
```
shamsimam, line 728: The event is `kill-notification`. The `kill` has been triggered elsewhere, it should be added in the scheduler implementation.

waiter/src/waiter/state/responder.clj
```diff
@@ -482,9 +483,14 @@
   [{:keys [instance-id->state] :as current-state} update-instance-id->blacklist-expiry-time-fn update-status-tag-fn
    instance-id expiry-time]
   (log/info "unblacklisting instance" instance-id "as blacklist expired at" expiry-time)
-  (cond-> (update-instance-id->blacklist-expiry-time-fn current-state #(dissoc % instance-id))
-    (contains? instance-id->state instance-id)
-    (update-in [:instance-id->state instance-id] update-status-tag-fn #(disj % :blacklisted))))
```
shamsimam, line 487: I'd prefer this same structure. We can wrap the `update-in` into a helper function that adds the necessary instance tracking logging. The `if-else` feels clunky here.

waiter/src/waiter/state/responder.clj
```diff
@@ -482,9 +483,14 @@
...
+  (let [update-instance-id->blacklist-expiry-time-fn-ret
+        (update-instance-id->blacklist-expiry-time-fn current-state #(dissoc % instance-id))]
+    (if (contains? instance-id->state instance-id)
+      (do
+        (scheduler/log-service-instance {:id instance-id} :readmit)
```
shamsimam, line 490: Did we settle on the `readmit` and `eject` terms?

waiter/src/waiter/state/maintainer.clj
```diff
@@ -628,11 +628,19 @@
...
-                                           (let [curr-instance-ids (set (map :id healthy-instances))
-                                                 prev-instance-ids (set (map :id (get service-id->healthy-instances service-id)))
-                                                 new-instance-ids (filterv (complement prev-instance-ids) curr-instance-ids)
-                                                 rem-instance-ids (filterv (complement curr-instance-ids) prev-instance-ids)
+                                           (let [curr-instances (set healthy-instances)
+                                                 prev-instances (set (get service-id->healthy-instances service-id))
+                                                 new-instances (filterv (complement prev-instances) curr-instances)
+                                                 rem-instances (filterv (complement curr-instances) prev-instances)
+                                                 new-instance-ids (mapv :id new-instances)
+                                                 rem-instance-ids (mapv :id rem-instances)
```
shamsimam, line 636: We should track instances that got removed.

waiter/src/waiter/scheduler.clj
pandashadopan: To preserve the convention that Scheduler specific messages use debug by default. 

```
(defmacro log
  "Log Scheduler-specific messages."
  [& args]
  `(log/log "Scheduler" :debug nil (print-str ~@args)))
```

waiter/src/waiter/state/responder.clj
pandashadopan: No. We can change them later

DaoWen: We definitely had consensus on "eject", but we didn't really discuss the symmetric term.

waiter/src/waiter/state/maintainer.clj
```diff
@@ -642,6 +650,12 @@
                                                        (if (seq new-instance-ids) (str "New healthy instances: " new-instance-ids ".") "")
                                                        (if (seq rem-instance-ids) (str "Removed healthy instances: " rem-instance-ids ".") "")
                                                        (if (seq unhealthy-instance-ids) (str "Unhealthy instances: " unhealthy-instance-ids ".") ""))))
+                                         (when (not= (get service-id->expired-instances service-id) expired-instances)
```
DaoWen, line 653: Since this logic is only used for InstanceTracker logging, let's predicate this extra work on the logger actually being enabled.

https://clojure.github.io/tools.logging/#clojure.tools.logging/enabled?

waiter/src/waiter/state/maintainer.clj
pandashadopan: Done

waiter/src/waiter/state/maintainer.clj
pandashadopan: Changed

waiter/src/waiter/state/maintainer.clj
pandashadopan: Done

waiter/src/waiter/state/maintainer.clj
pandashadopan: Done

waiter/src/waiter/state/maintainer.clj
pandashadopan: Done

waiter/src/waiter/scheduler/shell.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/bin/run-using-minimesos.sh
pandashadopan: Done

waiter/bin/run-using-k8s.sh
pandashadopan: Done

waiter/src/waiter/state/responder.clj
pandashadopan: Done

waiter/src/waiter/state/maintainer.clj
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,16 @@
...
+
+(defn swap-atom-with-instance-logging
```
DaoWen, line 55: This function name should end with `!` (per Clojure conventions) since it mutates an atom.

shamsimam: Also, we should be specific about the fields we want to render rather than rendering the entire instance map. E.g. the request log does this where it extracts specific fields from the request/response.

DaoWen: We should delegate to `add-instance-to-buffered-collection!` here. I think us *not* using that function in kubernetes.clj was a bug that I introduced.

DaoWen: I would prefer our helper function have a simpler name, like `track-failed-instance!`

waiter/src/waiter/scheduler.clj
```diff
@@ -754,7 +754,9 @@
 (defn add-instance-to-transient-store!
   [transient-store service-id instance event-type & args]
   (log-service-instance instance event-type)
-  (add-instance-to-buffered-collection! transient-store service-id instance args))
+  (if (= (count args) 3)
+    (add-instance-to-buffered-collection! transient-store service-id instance (nth args 0) (nth args 1) (nth args 2))
+    (add-instance-to-buffered-collection! transient-store service-id instance (nth args 0))))
```
pandashadopan, line 759: Should I use (first ) or (nth ... 0) in this case? I used nth to preserve the fact that nth is used on the line above.

DaoWen:
```suggestion
    (apply add-instance-to-buffered-collection! transient-store service-id instance args)))
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -205,8 +205,8 @@
                                               ;; we don't include the exit code in failed instances that were killed by k8s.
                                               (not (killed-by-k8s? newest-failure))
                                               (assoc :exit-code (:exitCode newest-failure)))]
-                (swap! service-id->failed-instances-transient-store
-                       update-in [service-id] assoc newest-failure-id newest-failure-instance)))))))
+                (scheduler/add-instance-to-transient-store! service-id->failed-instances-transient-store
+                                             service-id newest-failure-instance :fail newest-failure-id)))))))
```
shamsimam, line 209: Nitpick: Indentation is off here.

pandashadopan: Done

waiter/src/waiter/scheduler.clj
DaoWen: OK, I like the idea of setting all of our current messages to `:info` level, and reserving the ability to set some more noisy events later to `:debug` level. @pandashadopan , let's met the log level an explicit argument to this macro.

I also with Shams that we should `select-keys` on the instance to choose the fields we want to log.

waiter/src/waiter/scheduler.clj
```diff
@@ -730,13 +735,25 @@
 
 (defn add-instance-to-buffered-collection!
   "Helper function to add/remove entries into the transient store"
-  [transient-store max-instances-to-keep service-id instance-entry initial-value-fn remove-fn]
```
shamsimam, line 733: Why did we reorder the argument? It has triggered a bunch of changes in the callers.

waiter/src/waiter/scheduler.clj
```diff
@@ -730,13 +735,25 @@
...
+  ([transient-store service-id instance-entry max-instances-to-keep initial-value-fn remove-fn]
+   (swap! transient-store
+          (fn [service-id->failed-instances]
+            (update-in service-id->failed-instances [service-id]
+                       #(cond-> (or % (initial-value-fn))
+                                (= max-instances-to-keep (count %)) (remove-fn)
+                                true (conj instance-entry))))))
+  ;; Since we have a map (failure-id to failure-instance) in k8s we just update
+  ;; the map instead of adding element to a set like in marathon's case
+  ([transient-store service-id instance-entry failure-id]
```
shamsimam, line 747: I think we should not have this override and force the caller to provide a limit.
Also, the caller can assoc the `failure-id` into the instance.

waiter/src/waiter/scheduler.clj
pandashadopan: I reordered the argument because I wanted to use & args in the `add-instance-to-transient-store!` without argument reordering it would be less idiomatic:

```
(defn add-instance-to-transient-store!
  [transient-store service-id instance event-type & args]
  (log-service-instance instance event-type)
  (apply add-instance-to-buffered-collection! transient-store service-id instance args))
```

shamsimam: If we don't have the override we don't have a need to use `args`.

shamsimam: Please do not change the signature of this function, `max-instances-to-keep` should be passed by the caller.

waiter/src/waiter/scheduler.clj
```diff
@@ -730,13 +735,25 @@
...
+  ([transient-store service-id instance-entry]
+   (swap! transient-store
+          (fn [service-id->failed-instances]
+            (update-in service-id->failed-instances [service-id]
+                       #(assoc % (:failure-id instance-entry) instance-entry))))))
+
+(defn add-instance-to-transient-store!
+  [transient-store service-id instance event-type & args]
+  (log-service-instance instance event-type)
+  (apply add-instance-to-buffered-collection! transient-store service-id instance args))
```
shamsimam, line 756: We should not need to use `& args` and `apply` in this function.

waiter/src/waiter/scheduler.clj
pandashadopan: Resolved

shamsimam: I still see a diff. There is no need for the extra set of prenthesis since we have only one signature for the function now.

waiter/src/waiter/state/responder.clj
```diff
@@ -30,7 +30,8 @@
             [waiter.util.date-utils :as du]
             [waiter.util.utils :as utils]
             [waiter.work-stealing :as work-stealing])
-  (:import clojure.lang.PersistentQueue))
+  (:import clojure.lang.PersistentQueue
+           (org.joda.time DateTime)))
```
shamsimam, line 34: Is this import unused? If so, please remove.

waiter/src/waiter/state/responder.clj
```diff
@@ -478,13 +479,21 @@
        :response-chan response-chan
        :response response-code})))
 
+(defn- update-in-with-log-instance
+  [mp log-body event-type log-level key-sequence func & args]
```
shamsimam, line 483: Based on the single invocation below isn't `log-body` always the `instance-id`?
We should avoid the use of `apply` whenever we can.

pandashadopan: I added the abstract layer so that we can reuse this function later. Should I make it only consider the current case and we would change it later if we would need to?

shamsimam: Yes, please change it to consider the current case.

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(defmacro log-service-instance
+  "Log InstanceTracker-specific messages."
+  [instance event-type log-level]
+  `(log/log "InstanceTracker" ~log-level nil (utils/clj->json (assoc ~instance :timestamp (t/now) :event-type ~event-type))))
```
shamsimam, line 53: Reminder: https://github.com/twosigma/waiter/pull/1113#discussion_r442259899

waiter/src/waiter/state/responder.clj
pandashadopan: Removed

waiter/src/waiter/state/responder.clj
pandashadopan: Done

waiter/src/waiter/scheduler.clj
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(defmacro log-service-instance
+  "Log InstanceTracker-specific messages."
+  [instance event-type log-level key-seq]
+  `(log/log "InstanceTracker" ~log-level nil (utils/clj->json (assoc (select-keys ~instance ~key-seq) :timestamp (t/now) :event-type ~event-type))))
```
shamsimam, line 53: The `key-seq` should be determined in this function. The caller should not be the one passing this list in.

waiter/src/waiter/state/responder.clj
```diff
@@ -478,13 +478,18 @@
        :response-chan response-chan
        :response response-code})))
 
+(defn- update-in-with-log-instance
+  [mp log-body event-type log-level key-sequence update-fn func]
```
shamsimam, line 482: `mp` is better referred to as the `current-state`.
`key-sequence` is actually  better referred to as the `path`.

Importantly, if this function is named `clear-rejection-state` (as that is the only use currently), then none of the arguments except current state (`mp`) and `instance-id` need to be passed to this function.

waiter/src/waiter/scheduler.clj
pandashadopan:
```
(let key-seq [:id :service-id :started-at :healthy? :health-check-status :flags :exit-code :host :port :extra-ports :log-directory :message]
  (log/log ...))
```

Would something like this suffice?

DaoWen: We should just `def` the key-seq outside the macro and reference it inside.

Also, `:flags` needs to be flattened, otherwise it won't show correctly up in ELK.

DaoWen: `:extra-ports` is also an array that needs to be flattened.

shamsimam: > How would we request certain keys from the caller side in this case?

The caller shouldn't have to care. This function will extract whatever keys it feels are necessary to log the instance and event.

DaoWen: I was thinking about `:extra-ports` again, and I think we should just log out the count. Since they are always sequential in both the Marathon and Kubernetes implementations, having port0 and the extra-ports count is sufficient.

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,11 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(defmacro log-service-instance
```
shamsimam, line 50: Can we have this as a function instead of a macro?

I do not expect instance logging events to be that frequent. In any case, they are not on the reserve instance hot paths.

pandashadopan: Done

waiter/src/waiter/state/responder.clj
pandashadopan: Done

waiter/src/waiter/scheduler.clj
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -918,6 +955,9 @@
             (when leader?
               (statsd/histo! metric-group "startup_time"
                              (du/interval->nanoseconds duration)))))
+        (doseq [new-instance new-instances]
+          (when (healthy-instances? new-instance)
```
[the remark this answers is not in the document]
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -737,6 +768,11 @@
                          (= max-instances-to-keep (count %)) (remove-fn)
                          true (conj instance-entry))))))
 
+(defn add-to-store-and-track-instance!
+  [transient-store max-instances-to-keep service-id instance event-type log-level initial-value-fn remove-fn]
```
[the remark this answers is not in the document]
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,37 @@
...
+                       :flags
+                       (map (fn [flag] [flag true]))
+                       (into {}))]
+    (log/log "InstanceTracker" log-level nil (utils/clj->json
+                                               (assoc (merge
+                                                        (dissoc instance-to-log :flags)
+                                                        flags-map)
+                                                 :extra-ports extra-ports
+                                                 :timestamp (t/now)
+                                                 :event-type event-type)))))
```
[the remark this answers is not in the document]
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,37 @@
...
+              :blackist-period-ms])
+
+(defn log-service-instance
+  "Log InstanceTracker-specific messages."
+  [instance event-type log-level]
+  (let [extra-ports (count (:extra-ports instance))
+        instance-to-log (select-keys instance key-seq)
+        flags-map (->> instance-to-log
+                       :flags
+                       (map (fn [flag] [flag true]))
```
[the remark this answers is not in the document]
pandashadopan: Done

waiter/src/waiter/scheduler.clj
```diff
@@ -47,6 +47,37 @@
   [& args]
   `(log/log "Scheduler" :debug nil (print-str ~@args)))
 
+(def key-seq [:id
```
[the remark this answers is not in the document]
pandashadopan: Done

PR #1114, opened 2020-06-04
account for token edits while GC-ing services

shamsimam:

- account for token edits while GC-ing services

We should avoid GC-ing services when there has been a recent token update to allow fallback to continue to function and avoid service downtime.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -1256,21 +1256,25 @@
 (defn service->gc-time
   "Computes the time when a given service should be GC-ed.
    If the service has GC disabled (by configuring idle-timeout-mins=0), nil is returned.
-   If the service is active or was created by on-the-fly, the GC time is retrieved from the service description.
+   If the service is active or was created by on-the-fly, the GC time is calculated using
+   the service's idle-timeout-mins and the time of the service's most recent request.
    Else, the GC time is the sum of the fallback period seconds and the stale service timeout."
   [service-id->service-description-fn service-id->references-fn token->token-parameters reference-type->stale-fn
    attach-token-defaults-fn service-id last-modified-time]
   (let [{:strs [idle-timeout-mins]} (service-id->service-description-fn service-id)
         references (service-id->references-fn service-id)]
-    (if (service-references-stale? reference-type->stale-fn references)
+    (cond
+      ;; when stale, use token info to compute GC time
+      (service-references-stale? reference-type->stale-fn references)
       (let [{:strs [stale-timeout-mins]} (attach-token-defaults-fn {})]
         (log/info service-id "that uses references is stale")
         (if-let [source-tokens (->> references (map :token) (remove nil?) (map :sources) seq)]
           (source-tokens->gc-time token->token-parameters attach-token-defaults-fn source-tokens)
           ;; use the default token stale timeout for a stale service built without tokens
           (t/plus last-modified-time (t/minutes stale-timeout-mins))))
-      (when (pos? idle-timeout-mins)
-        (t/plus last-modified-time (t/minutes idle-timeout-mins))))))
+      ;; when GC is enabled, use idle-timeout
+      (pos? idle-timeout-mins)
+      (t/plus last-modified-time (t/minutes idle-timeout-mins)))))
 
 (let [service-id->key #(str "^REFERENCES#" %)]
 
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -1224,8 +1224,11 @@
          (.getMillis ^DateTime (clock))))))
 
 (defn- source-tokens->gc-time
-  "Computes the time when a service created using provided source tokens should be GC-ed.
-   It ignores whether GC has been disabled by configuring idle-timeout-mins=0."
+  "Computes the time when a service created using provided source tokens should be GC-ed
+   using token last-update-time, fallback-period-secs and stale-timeout-mins.
+   The function also assumes that source-tokens-seq is non-empty.
+   It ignores whether GC has been disabled by configuring idle-timeout-mins=0.
+   The caller must take care to handle the scenario where GC has been disabled."
   [token->token-parameters attach-token-defaults-fn source-tokens-seq]
   (->> source-tokens-seq
     (map (fn source-tokens->gc-time-helper [source-tokens]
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -1252,22 +1253,24 @@
...
-        (log/info service-id "that uses tokens is stale")
+        (log/info service-id "that uses references is stale")
         (if-let [source-tokens (->> references (map :token) (remove nil?) (map :sources) seq)]
-          (source-tokens->idle-timeout token->token-parameters attach-token-defaults-fn source-tokens)
+          (source-tokens->gc-time token->token-parameters attach-token-defaults-fn source-tokens)
           ;; use the default token stale timeout for a stale service built without tokens
-          stale-timeout-mins))
-      idle-timeout-mins)))
+          (t/plus last-modified-time (t/minutes stale-timeout-mins))))
+      (when (pos? idle-timeout-mins)
```
DaoWen, line 1272: Let's use `cond` here instead of `if`+`when`.

`(if a b (when c d))` → `(cond a b c d)`

Note that `cond` returns `nil` when none of the conditions match:
https://clojuredocs.org/clojure.core/cond

waiter/src/waiter/service_description.clj
```diff
@@ -1252,22 +1253,24 @@
...
   (every? (fn [reference] (stale-reference? reference-type->stale-fn reference)) references))
 
-(defn service-id->idle-timeout
-  "Computes the idle timeout, in minutes, for a given service.
-   If the service is active or was created by on-the-fly, the idle timeout is retrieved from the service description.
-   Else, the idle timeout is the sum of the fallback period seconds and the stale service timeout."
+(defn service->gc-time
+  "Computes the time when a given service should be GC-ed.
+   If the service has GC disabled (by configuring idle-timeout-mins=0), nil is returned.
+   If the service is active or was created by on-the-fly, the GC time is retrieved from the service description.
```
DaoWen, line 1259:
```suggestion
   If the service is active or was created by on-the-fly, the GC time is calculated using
   the service's idle-timeout-mins and the time of the service's most recent request.
```

waiter/src/waiter/service_description.clj
```diff
@@ -1252,22 +1253,24 @@
...
+        (log/info service-id "that uses references is stale")
         (if-let [source-tokens (->> references (map :token) (remove nil?) (map :sources) seq)]
-          (source-tokens->idle-timeout token->token-parameters attach-token-defaults-fn source-tokens)
+          (source-tokens->gc-time token->token-parameters attach-token-defaults-fn source-tokens)
           ;; use the default token stale timeout for a stale service built without tokens
-          stale-timeout-mins))
-      idle-timeout-mins)))
+          (t/plus last-modified-time (t/minutes stale-timeout-mins))))
+      (when (pos? idle-timeout-mins)
+        (t/plus last-modified-time (t/minutes idle-timeout-mins))))))
```
DaoWen, line 1273: Shouldn't we still be using `last-request-time` here?

Maybe it makes more sense to have separate functions for computing the normal idle gc time and the eager stale gc time. Then we can take the min of the two.

Or does it make sense for us to extend an instance's life past its idle timeout for token fallback logic? I think I prefer that a token can idle-out as scheduled (since it's getting no traffic!) even if fallback is active. You can imagine a scenario where a service was about to idle-out, but then the token is modified, and the fallback period extends the service's life by several hours. However, I could be convinced that it's more user-friendly to keep the old one around during the fallback period.

shamsimam: > Or does it make sense for us to extend an instance's life past its idle timeout for token fallback logic?
> a scenario where a service was about to idle-out, but then the token is modified, and the fallback period extends the service's life by several hours.

This is exactly the behavior we want that triggered this bug fix. A user updating a token expects fallback to help out while the new instance starts up.

> Shouldn't we still be using last-request-time here?

`last-modified-time` is proxy for `last-request-time`, it may be off by a few seconds at worst.

waiter/src/waiter/service_description.clj
DaoWen: I still think this comment needs rewording, but based on the clarification I just got from you about why we're using last-modified-time, my suggestion is obviously not accurate.

waiter/src/waiter/service_description.clj
DaoWen: > This is the requested behavior that triggered this bug fix. A user updating a token expects fallback to help out while the new instance starts up.

OK, true. It's probably better to extend a service that would have normally idled-out so that we can avoid users being confused about why the old instance shut down too quickly. In practice this is unlikely to happen (a service that has no requests being updated right as it's about to idle-out), and even if it does I think you're right that the possible user confusion is not with the minor resource savings that we'd get by shutting the old service off a tiny bit earlier.

waiter/src/waiter/service_description.clj
shamsimam: The updated docstring uses: `the most recent completion time of the service's requests`

waiter/src/waiter/service_description.clj
```diff
@@ -1222,21 +1223,25 @@
...
-             (+ (-> (+ fallback-period-secs (dec (-> 1 t/minutes t/in-seconds))) ;; ceiling
-                  t/seconds
-                  t/in-minutes)
-                stale-timeout-mins))))
-    (reduce max)))
+                   (attach-token-defaults-fn))
+                 most-recent-token-update-time
+                 (->> source-tokens
+                   (map #(some-> % :token token->token-parameters (get "last-update-time")))
+                   (reduce max))]
```
DaoWen, line 1240: Should we use `keep` instead of `map` here? The `(reduce max)` is going to NPE if there are `nil`s in the sequence.

waiter/src/waiter/service_description.clj
```diff
@@ -1222,21 +1223,21 @@
       (> (+ auth-timestamp (-> consent-expiry-days t/days t/in-millis))
          (.getMillis ^DateTime (clock))))))
 
-(defn source-tokens->idle-timeout
-  "Computes the idle timeout of the service using token parameters and defaults."
+(defn- source-tokens->gc-time
+  "Computes the time when a service created using provided source tokens should be GC-ed.
+   It ignores whether GC has been disabled by configuring idle-timeout-mins=0."
```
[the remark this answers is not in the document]
shamsimam: Added in docstring that it is the caller's responsibility to handle that case.

waiter/src/waiter/service_description.clj
shamsimam: Changed to use `keep` and check for empty sequence.
This can happen when the token has been deleted.
Also added unit test.

PR #1118, opened 2020-06-10
work-stealing changes: enable for specific services and configurable offer timeouts

shamsimam:

- enables work-stealing only for simple distribution services
- allows configuring the idle timeout on work-stealing offers

Our experience has show work-stealing to provide limited value when instance slots are distributed across all clusters. Work-stealing is required when instance slots may be unavailable by the partitioning algorithm and is especially helpful for long requests.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -824,7 +824,7 @@
                  :offer-help-interval-ms 100
 
                  ;; The duration (milliseconds) after which Waiter times out work-stealing offers:
-                 :offer-idle-timeout-ms 1800000
+                 :offer-idle-timeout-ms 900000
 
                  ;; The timeout (milliseconds) used internally by the
                  ;; Waiter router to reserve an instance for offering:
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -825,7 +825,10 @@
 
                  ;; The timeout (milliseconds) used internally by the
                  ;; Waiter router to reserve an instance for offering:
-                 :reserve-timeout-ms 1000}
+                 :reserve-timeout-ms 1000
+
+                 ;; The distribution schemes of services for which to enable work-stealing support:
+                 :supported-distribution-schemes #{"balanced" "simple"}}
 
  ; ---------- Metrics - StatsD ----------
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -823,6 +823,9 @@
                  ;; The interval (milliseconds) on which Waiter makes work-stealing offers:
                  :offer-help-interval-ms 100
 
+                 ;; The duration (milliseconds) after which Waiter times out work-stealing offers:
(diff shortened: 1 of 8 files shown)


waiter/src/waiter/core.clj
```diff
@@ -1085,21 +1090,26 @@
...
+                                            (let [{:keys [service-id->router-id->metrics]} router-metrics-helpers]
+                                              (work-stealing/start-work-stealing-balancer
+                                                populate-maintainer-chan! reserve-timeout-ms offer-help-interval-ms
+                                                offer-idle-timeout-ms offers-allowed-semaphore service-id->router-id->metrics
+                                                make-inter-router-requests-async-fn router-id service-id))))))
    :stop-work-stealing-balancer-fn (pc/fnk []
-                                     (fn stop-work-stealing-balancer [service-id work-stealing-chan-map]
-                                       (log/info "stopping work-stealing balancer for" service-id)
-                                       (async/go
-                                         (when-let [exit-chan (get work-stealing-chan-map [:exit-chan])]
```
shamsimam, line 1101: This is a bug where work-stealing go-blocks are not being terminated correctly.
This may explain high CPU usage over time.

waiter/config-full.edn
```diff
@@ -823,9 +823,15 @@
                  ;; The interval (milliseconds) on which Waiter makes work-stealing offers:
                  :offer-help-interval-ms 100
 
+                 ;; The duration (milliseconds) after which Waiter times out work-stealing offers:
+                 :offer-idle-timeout-ms 1800000
```
sradack, line 827: This seems like an too large idle timeout.  We keep the connection open the entire time the offer is being made?

waiter/src/waiter/core.clj
sradack: Nice find

waiter/src/waiter/handler.clj
```diff
@@ -819,14 +819,17 @@
               responder-state-chan (service/query-maintainer-channel-map-with-timeout!
                                      populate-maintainer-chan! service-id timeout-ms :query-state)
               _ (log/info "waiting for response from query-work-stealing channel...")
-              work-stealing-state-chan (service/query-maintainer-channel-map-with-timeout!
-                                         populate-maintainer-chan! service-id timeout-ms :query-work-stealing)
+              work-stealing-state (if (enable-work-stealing-support? service-id)
+                                    (service/query-maintainer-channel-map-with-timeout!
+                                      populate-maintainer-chan! service-id timeout-ms :query-work-stealing)
+                                    {"state" "disabled"})
```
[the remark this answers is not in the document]
shamsimam: Changed, I wanted to avoid making changes later in the function.

waiter/config-full.edn
shamsimam: Yes, the connection stays open for the duration of the offer.
Changed it back to 15 minutes.

PR #1119, opened 2020-06-10
configures the idle timeout on api requests in the integration tests

shamsimam:

- configures the idle timeout on api requests in the integration tests

Tests should not hang indefinitely in the case that the server becomes unresponsive.


shamsimam:
There is an issue with cookie support and current version of Jetty we are using. We need to upgrade Jetty (the current latest Jetty version has an NPE issue) before enabling this PR.

shamsimam:
waiter-architecture

PR #1128, opened 2020-06-22
Formats imports

shamsimam:

- formats imports (using amethyst intellij plugin)
- removes unused metrics.core imports

Cleaning up the imports.


waiter/src/waiter/core.clj
```diff
@@ -74,17 +73,16 @@
...
-           org.apache.curator.framework.api.CuratorListener
-           org.apache.curator.framework.recipes.leader.LeaderLatch
-           org.apache.curator.retry.BoundedExponentialBackoffRetry
-           org.eclipse.jetty.client.HttpClient
-           org.eclipse.jetty.client.util.BasicAuthentication$BasicResult
-           org.eclipse.jetty.websocket.client.WebSocketClient
-           (org.eclipse.jetty.websocket.servlet ServletUpgradeResponse ServletUpgradeRequest)))
+           (org.apache.curator.framework.api CuratorEventType CuratorListener)
+           (org.apache.curator.framework CuratorFrameworkFactory)
+           (org.apache.curator.framework.recipes.leader LeaderLatch)
```
DaoWen, line 80: This doesn't look right... The `framework` import should come before both `framework.*` imports.

shamsimam: Updated, seems like a bug in Amethyst formatting :(

waiter/src/waiter/core.clj
```diff
@@ -74,17 +73,16 @@
             [waiter.websocket :as ws]
             [waiter.work-stealing :as work-stealing])
   (:import (java.net InetAddress URI)
-           java.util.concurrent.Executors
+           (java.util.concurrent Executors)
```
DaoWen, line 76: `java.util.concurrent.Executors` → `(java.util.concurrent Executors)` doesn't seem like it would preserve semantics...

Is this just unused?

DaoWen: Never mind. I just didn't understand the Clojure semantics:

> The `:import` clause declares this lib’s use of `java.util.Date` and makes it available to code in this lib using its unqualified name.
https://clojure.org/reference/libs

waiter/src/waiter/util/http_utils.clj
```diff
@@ -24,17 +24,15 @@
             [slingshot.slingshot :as ss]
             [waiter.status-codes :refer :all])
   (:import (java.net URI)
-           (java.net URI)
            (java.util ArrayList)
            (org.apache.commons.codec.binary Base64)
-           (org.eclipse.jetty.client HttpClient)
            (org.eclipse.jetty.client.api Authentication$Result Request)
+           (org.eclipse.jetty.client HttpClient)
```
DaoWen, line 30: Same ordering bug here

waiter/src/waiter/util/http_utils.clj
```diff
@@ -24,17 +24,15 @@
...
            (org.eclipse.jetty.http HttpField HttpHeader)
-           (org.eclipse.jetty.http2.client HTTP2Client)
            (org.eclipse.jetty.http2.client.http HttpClientTransportOverHTTP2)
+           (org.eclipse.jetty.http2.client HTTP2Client)
```
DaoWen, line 33: Same ordering bug here

waiter/src/waiter/util/http_utils.clj
shamsimam: Done.

waiter/src/waiter/util/http_utils.clj
shamsimam: Done.

PR #1134, opened 2020-06-26
reduces test-s3-logs flakiness by allowing more time for second instance to become healthy

shamsimam:

- reduces test-s3-logs flakiness by allowing more time for second instance to become healthy

Reduces flakiness in `test-s3-logs`.


DaoWen:
Unrelated test flaked in Travis. Merged anyway.

PR #1140, opened 2020-06-30
uses https urls for redirects during OIDC auth

shamsimam:

- uses https urls for redirects during OIDC auth

We wish to enforce use of https in OIDC authentication workflow.

### DIFF
--- a/waiter/src/waiter/auth/jwt.clj
+++ b/waiter/src/waiter/auth/jwt.clj
@@ -194,12 +194,10 @@
     (when (str/blank? oidc-authorize-uri)
       (throw (ex-info "OIDC authorize endpoint not configured!" {})))
     (counters/inc! (metrics/waiter-counter "core" "jwt" "authorize-url"))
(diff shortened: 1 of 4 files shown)


shamsimam:
> Have you considered redirecting via HTTPS to back to Waiter before issuing the redirect?

The HTTPS redirect would eventually end-up triggering the same redirect to the auth server (since x-waiter-auth cookie or headers would not change on the request). We are saving the extra round-trip by continuing the redirect to the OIDC auth server.

sradack:
Agreed but then the redirect to IdP is secured and we avoid special casing
inside the OIDC code.

shamsimam:
Updated the code to trigger SSL redirect to Waiter before triggering OIDC auth redirect

shamsimam:
Hopefully, the request comes in with the `upgrade-insecure-requests` header and gets SSL redirected even before the OIDC auth workflow is triggered.

PR #1147, opened 2020-07-04
test-basic-priority-support: reduces flakiness due to threads not being scheduled

shamsimam:

- reduces flakiness due to threads not being scheduled

Launching too many threads can keep the individual requests at the client from being processed and hence received by the backend server. Thus higher priority requests may reach a server after a lower priority request has been scheduled causing the test to become flaky. Determining the priority when the thread runs helps avoid this issue.

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -820,7 +820,8 @@
 
 (deftest ^:parallel ^:integration-fast test-basic-priority-support
   (testing-using-waiter-url
-    (let [headers {:x-waiter-name (rand-name)
+    (let [headers {:x-waiter-concurrency-level 1
+                   :x-waiter-name (rand-name)
                    :x-waiter-distribution-scheme "simple" ;; disallow work-stealing interference from balanced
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -820,7 +820,8 @@
 
 (deftest ^:parallel ^:integration-fast test-basic-priority-support
   (testing-using-waiter-url
-    (let [headers {:x-waiter-name (rand-name)
+    (let [headers {:x-waiter-concurrency-level 1
```
shamsimam, line 823: Avoids any issues due to concurrent processing of requests if we ever change the defaults.

waiter/integration/waiter/basic_test.clj
```diff
@@ -829,7 +830,7 @@
         service-id
         (let [router-url (some-router-url-with-assigned-slots waiter-url service-id)
               response-priorities-atom (atom [])
-              num-threads 15
+              num-threads 8
```
shamsimam, line 833: Reduce the number of threads needed by the test to function correctly.

PR #1149, opened 2020-07-07
adds support for disabling kerberos authentication

shamsimam:

- adds support for disabling kerberos authentication

We wish to allow services to rely only upon provided authentication schemes (i.e. cookie support, JWT or OIDC) and not use any of the other methods (e.g. Keberos, SAML). With this change, users can choose `standard` authentication and explicitly disable kerberos support via an environment variable. Long-term we can alias this behavior to `standard` authentication when we have migrated existing Kerberos auth users.


DaoWen:
Would it be hard to add an additional assertion to the integration test that asserts we get a `WWW-Authenticate: Bearer` response back when token-auth is enabled?

shamsimam:
added assertion for Bearer www-authentication header. Ready for another round of review.

PR #1150, opened 2020-07-07
avoids updating caches for empty responses from prestash cache query

shamsimam:

- avoids updating caches for empty responses from prestash cache query

We wish to avoid flushing our cache when we get an empty response from the prestash query.

### DIFF
--- a/waiter/src/waiter/auth/kerberos.clj
+++ b/waiter/src/waiter/auth/kerberos.clj
@@ -33,7 +33,13 @@
   (try
     (let [{:keys [exit out err]} (shell/sh "krb5_prestash" "query" "host" host)]
       (if (zero? exit)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/auth/kerberos.clj
```diff
@@ -33,7 +33,13 @@
   (try
     (let [{:keys [exit out err]} (shell/sh "krb5_prestash" "query" "host" host)]
       (if (zero? exit)
-        (set (map #(utils/principal->username %) (str/split-lines out)))
+        (some->> out
+          (str/split-lines)
+          (remove str/blank?) ;; remove empty lines
+          (map #(utils/principal->username %))
```
DaoWen, line 39:
```suggestion
          (map utils/principal->username)
```

PR #1152, opened 2020-07-08
Envoy sidecar integration

pandashadopan:

- Envoy sidecar is added

- To allow integration with envoy sidecar reverse-proxy


containers/reverse-proxy/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM envoyproxy/envoy-alpine:v1.14.4
+
+ARG SERVICE_PORT
+ARG PORT0
```
DaoWen, line 4: We should just assume these are in the container environment rather than passing them in as args at image-build time.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -916,7 +916,29 @@
            :resources {:limits {:memory memory}
                        :requests {:cpu cpu :memory memory}}
            :volumeMounts [{:mountPath "/srv/www"
-                           :name "user-home"}]})))))
+                           :name "user-home"}]}))
+
+      ;; Optional envoy sidecar container
+      (pos-int? offset)
```
DaoWen, line 922: This should probably be predicated on `has-reverse-proxy?` rather than a positive port offset.

containers/reverse-proxy/create-envoy-config.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
+
+PORT0=$1
+PORT1=$2
```
DaoWen, line 4: We should get rid of these and just assume SERVICE_PORT and PORT0 are in the environment.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -916,7 +916,29 @@
...
+      (update-in
+        [:spec :template :spec :container]
+        conj
+        (let [cmd ""]
+             {:command cmd
+              :env (into [{:name "SERVICE-PORT" :value service-port}
+                          {:name "PORT0" :value port0}
+                          {:name "PORT" :value port0}]
```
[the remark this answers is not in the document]
pandashadopan: Done

containers/reverse-proxy/create-envoy-config.sh
```diff
@@ -0,0 +1,61 @@
...
+              socket_address:
+                address: 0.0.0.0
+                port_value: $PORT1
+admin:
+  access_log_path: "/dev/stdout"
+  address:
+    socket_address:
+      address: 0.0.0.0
+      port_value: 15000
+EOF
```
DaoWen, line 61: We want this script to also `exec envoy`. We should rename the script to something like `run-envoy-sidecar.sh` since it no longer just creates the config file and exits.

containers/reverse-proxy/Dockerfile
pandashadopan: Done

containers/reverse-proxy/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM envoyproxy/envoy-alpine:v1.14.4
+
+ARG SERVICE_PORT
+ARG PORT0 
+
+COPY create-envoy-config.sh /
+RUN sh -x create-envoy-config.sh ${SERVICE_PORT} ${PORT0} > /etc/envoy/envoy.yaml
```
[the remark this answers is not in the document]
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

containers/reverse-proxy/create-envoy-config.sh
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +923,29 @@
...
+      ;; Optional envoy sidecar container
+      has-reverse-proxy?
+      (update-in
+        [:spec :template :spec :container]
+        conj
+        (let [cmd ["/bin/sh" "-x"]]
+             {:command cmd
+              :args ["./opt/waiter/envoy/bin/run-envoy-sidecar.sh"
+                     "$(SERVICE_PORT)"
+                     "$(PORT0)"]
```
pandashadopan, line 937: I still have to refactor this so that the whole setup would be just command "/bin/run-envoy" I'm using this :args setup currently since this is one of the few configurations that works.

pandashadopan: Fixed

containers/reverse-proxy/create-envoy-config.sh
pandashadopan: Done

containers/reverse-proxy/envoy-config-template.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
+
+cat << EOF
```
DaoWen, line 3: Let's inline this in the envoy-start script. You can do the output redirect like this:

```
cat > /etc/envoy/envoy.yaml << EOF
...
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +923,26 @@
...
+        [:spec :template :spec :container]
+        conj
+        (let [cmd ["/bin/sh" "/opt/waiter/envoy/bin/envoy-start"]]
+             {:command cmd
+              :env (into [{:name "SERVICE-PORT" :value service-port}
+                          {:name "PORT0" :value port0}]
+                         (concat
+                           (for [[k v] base-env]
+                                {:name k :value v})))
+              :image "twosigma/waiter-envoy"
```
DaoWen, line 940: This needs to be parameterized so we can set it in `config.edn`, similar to the file server's image.

containers/reverse-proxy/envoy-config-template.sh
pandashadopan: Resolved

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,19 @@
...
         (log/warn "skipping test as INTEGRATION_TEST_BAD_IMAGE is not specified")))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (let [{:keys [service-id] :as response}
+            (make-request-with-debug-info
+              {:x-waiter-distribution-scheme "simple"
+               :x-waiter-name (rand-name)
+               :x-waiter-env-GRPC_TRANSCODER "yes"}
```
DaoWen, line 348: Since this environment variable is configurable, can we pull it from settings?

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,19 @@
...
+              #(make-kitchen-request waiter-url % :method :get :path "/"))
```
DaoWen, line 349: You should be hitting the `/request-info` endpoint to get the request headers coming from envoy.

DaoWen:
You need to `chmod +x` your new docker image script:

```
./bin/build-docker-images.sh: line 32: ./reverse-proxy/bin/build-docker-image.sh: Permission denied
```

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,36 @@
...
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar-port
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (let [{:keys [service-id] :as response}
+            (make-request-with-debug-info
+              {:x-waiter-distribution-scheme "simple"
+               :x-waiter-name (rand-name)
+               :x-waiter-env-GRPC_TRANSCODER "yes"}
+              #(make-kitchen-request waiter-url % :method :get :path "/environment"))]
+        (println (with-out-str (clojure.pprint/pprint response)))
```
DaoWen, line 366: Remove println before final version

containers/reverse-proxy/envoy-start
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
+
+cat > /etc/envoy/envoy.yaml << EOF
```
shamsimam, line 3: Is using `envsubst` an option, then the config below can be in a separate file?
https://stackoverflow.com/questions/14155596/how-to-substitute-shell-variables-in-complex-text-files

Else we can use `sed` to selectively replace environment variables from a file.
e.g. sed -i -e "s/SERVICE_PORT/${SERVICE_PORT}/g" config.yaml

waiter/config-full.edn
```diff
@@ -382,7 +382,9 @@
                                              :kubernetes {;; a factory function must be provided
                                                           ;; see the :kubernetes entry for other configuration options
                                                           :factory-fn waiter.scheduler.kubernetes/kubernetes-scheduler
-                                                          :todo :additional-config}}
+                                                          :todo :additional-config
+                                                          :proxy-options {:reverse-proxy-flag "GRPC_TRANSCODER"
+                                                                          :reverse-proxy-offset 1}}}
```
shamsimam, line 387: This entry can be removed, the comment above is referring the user to the segment below.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +927,26 @@
...
+
+      ;; Optional envoy sidecar container
+      has-reverse-proxy?
+      (update-in
+        [:spec :template :spec :containers]
+        conj
+        (let [cmd ["/opt/waiter/envoy/bin/envoy-start"]]
+             {:command cmd
+              :env (into [{:name "SERVICE_PORT" :value service-port}
+                          {:name "PORT0" :value port0}]
```
DaoWen, line 940:
```suggestion
              :env (into [{:name "SERVICE_PORT" :value (str service-port)}
                          {:name "PORT0" :value (str port0)}]
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +927,26 @@
...
+                         (concat
+                           (for [[k v] base-env]
+                                {:name k :value v})))
+              :image "twosigma/waiter-envoy"
+              :imagePullPolicy "IfNotPresent"
+              :name "waiter-envoy-sidecar"
+              :ports [{:containerPort service-port}]
+              :resources {:limits {:memory "256 Mi"}
+                          :requests {:cpu "0.1" :memory "256 Mi"}}})))))
```
DaoWen, line 949:
```suggestion
              :resources {:limits {:memory "256Mi"}
                          :requests {:cpu "0.1" :memory "256Mi"}}})))))
```

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,36 @@
...
+        (with-service-cleanup
+          service-id
+          (assert-response-status response http-200-ok)
+          (testing "Port value is correctly offset compared to instance value"
+            (is (= (+ 1 (Integer/parseInt (get (:headers response) "x-waiter-backend-port")))
+                   (Integer/parseInt (get (:body response) "PORT0"))))))))))
```
DaoWen, line 372: You need to parse the body as JSON, otherwise it's just a string. Use the `try-parse-json` utility function (already pulled into this namespace from `waiter.util.client-tools`).

DaoWen: Also, this would be more readable if you used a `let` to create local bindings for `backend-port` and `port0`.
Clojure's threading macro is also nice for this type of thing:

```
(let [backend-port (-> response :headers (get "x-waiter-backend-port") Integer/parseInt)
      ...
```

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/config-full.edn
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +923,26 @@
...
            :volumeMounts [{:mountPath "/srv/www"
-                           :name "user-home"}]})))))
+                           :name "user-home"}]}))
+
+      ;; Optional envoy sidecar container
+      has-reverse-proxy?
+      (update-in
+        [:spec :template :spec :containers]
+        conj
+        (let [cmd ["/opt/waiter/envoy/bin/envoy-start"]]
```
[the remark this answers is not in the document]
shamsimam: Alternatively, we could have:
```
:proxy-options {:reverse-proxy-parameters {"cmd" ..., "cpus" ..., "env" ..., "image" ..., "mem" ...}}
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +923,26 @@
...
+             {:command cmd
+              :env (into [{:name "SERVICE_PORT" :value (str service-port)}
+                          {:name "PORT0" :value (str port0)}]
+                         (concat
+                           (for [[k v] base-env]
+                                {:name k :value v})))
+              :image "twosigma/waiter-envoy"
```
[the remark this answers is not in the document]
shamsimam: Alternatively, we could have:
```
:proxy-options {:reverse-proxy-parameters {"cmd" ..., "cpus" ..., "env" ..., "image" ..., "mem" ...}}
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +923,26 @@
...
+              :imagePullPolicy "IfNotPresent"
+              :name "waiter-envoy-sidecar"
+              :ports [{:containerPort service-port}]
+              :resources {:limits {:memory "256Mi"}
+                          :requests {:cpu "0.1" :memory "256Mi"}}})))))
```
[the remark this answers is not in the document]
shamsimam: Alternatively, we could have:
```
:proxy-options {:reverse-proxy-parameters {"cmd" ..., "cpus" ..., "env" ..., "image" ..., "mem" ...}}
```

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
                     :interval 15
                     :timeout timeout-secs)))))
         (log/warn "skipping test as INTEGRATION_TEST_BAD_IMAGE is not specified")))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar-headers
```
shamsimam, line 341: We should combine the two tests `test-kubernetes-reverse-proxy-sidecar-headers` and `test-kubernetes-reverse-proxy-sidecar-env` into a single test.

We can then send a canary request and then have separate assertion blocks for the `/request-info` and `/environment` requests.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
...
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (when-let [proxy-options (get-k8s-proxy-options waiter-url)]
+        (let [{:keys [service-id] :as response}
+              (make-request-with-debug-info
+                {:x-waiter-distribution-scheme "simple"
```
shamsimam, line 347: Why do we need `simple` distribution for this test?

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
...
+                (let [response-header-backend-port (-> response-headers
+                                                       (get "x-waiter-backend-port")
+                                                       Integer/parseInt
+                                                       (+ reverse-proxy-offset))
+                      env-response-port0 (-> response-body
+                                             (get "PORT0")
+                                             Integer/parseInt)]
+                  (is (= response-header-backend-port env-response-port0))))
+              (testing "Reverse proxy flag environment variable is present"
+                (is (contains? response-body reverse-proxy-flag))))))))))
```
shamsimam, line 386: We should validate the value as well.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
...
+                  response-headers (:headers response)]
+              (testing "Port value is correctly offset compared to instance value"
+                (let [response-header-backend-port (-> response-headers
+                                                       (get "x-waiter-backend-port")
+                                                       Integer/parseInt
+                                                       (+ reverse-proxy-offset))
+                      env-response-port0 (-> response-body
+                                             (get "PORT0")
+                                             Integer/parseInt)]
+                  (is (= response-header-backend-port env-response-port0))))
```
shamsimam, line 384: We should do the addition of the proxy offset here, else the name `response-header-backend-port` is misleading.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
...
+                (is (contains? response-headers "x-envoy-upstream-service-time"))
+                (is (= "envoy" (get response-headers "server")))))))))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar-env
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (when-let [{:keys [reverse-proxy-flag reverse-proxy-offset]} (get-k8s-proxy-options waiter-url)]
+        (let [{:keys [service-id] :as response}
+              (make-request-with-debug-info
+                {:x-waiter-distribution-scheme "simple"
```
shamsimam, line 367: Why do we need simple distribution for this test?

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,50 @@
...
+            (testing "Expected envoy specific headers are present in both request and response"
+              (let [response-body (try-parse-json (:body response))
+                    response-headers (:headers response)]
+                (is (contains? (get response-body "headers") "x-envoy-expected-rq-timeout-ms"))
+                (is (contains? response-headers "x-envoy-upstream-service-time"))
+                (is (= "envoy" (get response-headers "server")))))))))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar-env
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
```
shamsimam, line 363: Please change the `when`s to `if`s with appropriate logs about skipping the tests.

waiter/config-k8s.edn
```diff
@@ -58,7 +58,9 @@
                                                            :default-namespace "waiter"
                                                            :image-aliases {"alias/integration" "twosigma/integration"}}
                                  :restart-expiry-threshold 2
-                                 :url "http://localhost:8001"}}
+                                 :url "http://localhost:8001"
+                                 :proxy-options {:reverse-proxy-flag "GRPC_TRANSCODER"
```
shamsimam, line 62: Maybe change `GRPC_TRANSCODER` to `REVERSE_PROXY`?

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Resolved

containers/reverse-proxy/envoy-start
pandashadopan: Done

containers/bin/build-docker-images.sh
```diff
@@ -27,3 +27,7 @@ cd "$DIR"
 
 # the fileserver container is independent
 ./fileserver/bin/build-docker-image.sh
+
+# the envoy reverse-proxy container is independent
+chmod +x ./reverse-proxy/bin/build-docker-image.sh
```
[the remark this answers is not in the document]
pandashadopan: Resolved

containers/reverse-proxy/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM envoyproxy/envoy-alpine:v1.14.4
+
+COPY envoy-start /opt/waiter/envoy/bin/
+
+RUN chmod +x /opt/waiter/envoy/bin/envoy-start
```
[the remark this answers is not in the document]
pandashadopan: Resolved

containers/reverse-proxy/envoy-start
```diff
@@ -0,0 +1,59 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
pandashadopan: Added documentation.

waiter/src/waiter/scheduler/kubernetes.clj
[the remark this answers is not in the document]
pandashadopan: Changed the :reverse-proxy-parameters keys from strings to keywords because edn reader was complaining and validate-k8s-config were complaining

waiter/src/waiter/scheduler/kubernetes.clj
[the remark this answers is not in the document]
pandashadopan: Changed the :reverse-proxy-parameters keys from strings to keywords because edn reader was complaining and validate-k8s-config were complaining

waiter/src/waiter/scheduler/kubernetes.clj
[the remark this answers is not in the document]
pandashadopan: Changed the :reverse-proxy-parameters keys from strings to keywords because edn reader was complaining and validate-k8s-config were complaining

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Done

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Resolved

waiter/config-k8s.edn
pandashadopan: Resolved

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: You are right. There is no need for that here

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Resolved above

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Resolved

containers/reverse-proxy/envoy-start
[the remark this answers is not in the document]
DaoWen: You should add a comment right after the shebang line that explains the usage (i.e., that it takes no arguments, but needs SERVICE_PORT and PORT0 set).

Adding a log line saying "<timestamp> Starting envoy sidecar" would be useful just to see when the script started running.

waiter/src/waiter/scheduler/kubernetes.clj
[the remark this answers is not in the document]
shamsimam: I prefer strings because they match-up with the service description parameters.
We can change the schema to support keys instead of keywords.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,48 @@
                     :interval 15
                     :timeout timeout-secs)))))
         (log/warn "skipping test as INTEGRATION_TEST_BAD_IMAGE is not specified")))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (if-let [{:strs [flag offset]} (-> waiter-url
```
DaoWen, line 344: This should just be a `let`. This will never be false.

waiter/config-full.edn
```diff
@@ -544,7 +544,13 @@
...
                                  ;; https://github.com/kubernetes/kubernetes/blob/207e9d1/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L45
                                  :pod-suffix-length 5
-
+                                 :proxy-options {:reverse-proxy-parameters {:cmd "/opt/waiter/envoy/bin/envoy-start"
+                                                                            :cpus "0.1"
+                                                                            :env {}
+                                                                            :flag "REVERSE_PROXY"
+                                                                            :image "twosigma/waiter-envoy"
+                                                                            :mem "256Mi"
+                                                                            :offset 1}}
```
shamsimam, line 553: I thought of parameters as the equivalent for service parameters with string keys.
As such, `cpus` and `mem` are floating point values.
We could reuse the service description schema to validate this (we should already have a function for it).

`:offset` and `:flag` do not fit as those parameters and should be in a different location.

@DaoWen what do you think?

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,48 @@
...
+                      response-headers (:headers response)]
+                  (is (contains? (get response-body "headers") "x-envoy-expected-rq-timeout-ms"))
+                  (is (contains? response-headers "x-envoy-upstream-service-time"))
+                  (is (= "envoy" (get response-headers "server")))))))
+          (let [{:keys [service-id] :as response}
+                (make-request-with-debug-info
+                  {:x-waiter-name (rand-name)
+                   (keyword (str "x-waiter-env-" flag)) "yes"}
+                  #(make-kitchen-request waiter-url % :method :get :path "/environment"))]
+            (with-service-cleanup
```
DaoWen, line 367: Instead of creating+deleting this same service twice, we should reuse the service across both requests.

waiter/config-full.edn
DaoWen: @shamsimam - I feel strongly that this should be formatted exactly as the `:fileserver` entry above for consistency. No `:proxy-options :reverse-proxy` nesting, it should have a `:resources` map, etc.

I agree that `offset` does not belong here. I don't even think that should be a configuration option—we should just know that the offset is 1 based on the fact that we have 1 proxy.

I think the `flag` makes sense here since the `flag` is what controls whether this sidecar is turned on or off.

shamsimam: sgtm

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Resolved

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,54 @@
                     :interval 15
                     :timeout timeout-secs)))))
         (log/warn "skipping test as INTEGRATION_TEST_BAD_IMAGE is not specified")))))
+
+(deftest ^:parallel ^:integration-fast test-kubernetes-reverse-proxy-sidecar
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (if (contains? (get-kubernetes-scheduler-settings waiter-url) :reverse-proxy)
+        (let [reverse-proxy-flag "REVERSE_PROXY"
```
shamsimam, line 345: Nitpick: We should read this from an environment variable with a default to `REVERSE_PROXY` since the value depends on the configured predicate function.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,3 +337,54 @@
...
+              reverse-proxy-offset 1
+              x-waiter-name (rand-name)
+              _ (log/info "making canary request")
+              {:keys [cookies service-id] :as response} (make-request-with-debug-info
```
shamsimam, line 349: Nitpick: you can get the `headers` from the response here and re-use them below in your requests instead of having to reconstruct them. Alternatively, you can extract the header-map into `request-headers` variable and use them in all the three places.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -782,12 +782,20 @@
    :periodSeconds health-check-interval-secs
    :timeoutSeconds 1})
 
+(defn envoy-sidecar-enabled?
+  "Returns true if the envoy sidecar is enabled, and false otherwise"
+  [_ _ {:strs [env]} _]
+  (let [reverse-proxy-flag "REVERSE_PROXY"]
+    (if (contains? env reverse-proxy-flag)
```
shamsimam, line 789: Nitpick: `if` is unnecessary, the `contains?` will return true/false.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -805,9 +813,11 @@
         ;; delay iff the log-bucket-url setting was given the scheduler config.
         log-bucket-sync-secs (if log-bucket-url (:log-bucket-sync-secs context) 0)
         total-sigkill-delay-secs (+ pod-sigkill-delay-secs log-bucket-sync-secs)
-        {:keys [reverse-proxy-flag reverse-proxy-offset]} proxy-options
-        has-reverse-proxy? (contains? env reverse-proxy-flag)
-        offset (if has-reverse-proxy? reverse-proxy-offset 0)
+        envoy-sidecar-check-fn (:predicate-fn reverse-proxy)
+        has-reverse-proxy? (if envoy-sidecar-check-fn
+                             (envoy-sidecar-check-fn scheduler service-id service-description context)
+                             false)
```
shamsimam, line 819:
```suggestion
        has-reverse-proxy? (when envoy-sidecar-check-fn 
                             (envoy-sidecar-check-fn scheduler service-id service-description context))
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +933,29 @@
            :resources {:limits {:memory memory}
                        :requests {:cpu cpu :memory memory}}
            :volumeMounts [{:mountPath "/srv/www"
-                           :name "user-home"}]})))))
+                           :name "user-home"}]}))
+
+      ;; Optional envoy sidecar container
+      has-reverse-proxy?
+      (update-in
```
shamsimam, line 940: Nitpick: this update could be done by a configurable function like we configure the `predicate-fn`. We can do that in a separate PR.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -923,7 +933,29 @@
...
+        (let [{:keys [cmd resources image]} reverse-proxy
+              env (:env service-description)]
+             {:command cmd
+              :env (into [{:name "SERVICE_PORT" :value (str service-port)}
+                          {:name "PORT0" :value (str port0)}]
+                         (concat
+                           (for [[k v] base-env]
+                                {:name k :value v})
+                           (for [[k v] env]
+                                {:name k :value v})))
```
shamsimam, line 952: Nitpick: We should merge the maps into a single environment map and loop through it to avoid conflicting values (esp. user overriding SERVICE_PORT) from getting through and leading to an unexpected configuration.

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
pandashadopan: Since we are not storing the reverse-proxy-flag anywhere now, shouldn't this be hardcoded?

pandashadopan: I can reintroduce the :flag in :reverse-proxy or I can do something like this do not add any extra keys to reverse-proxy:

```
(defn envoy-sidecar-enabled?
  "Returns true if the envoy sidecar is enabled, and false otherwise"
  [_ _ {:strs [env]} _]
  (let [reverse-proxy-flag "REVERSE_PROXY"]
    (contains? env reverse-proxy-flag))
  []
  (let [reverse-proxy-flag "REVERSE_PROXY"]
    reverse-proxy-flag))
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -786,9 +786,30 @@
...
+  "Attaches envoy sidecar to replicaset"
+  [replicaset reverse-proxy service-description base-env service-port port0]
+  (update-in replicaset
+    [:spec :template :spec :containers]
+    conj
+    (let [{:keys [cmd resources image]} reverse-proxy
+          user-env (:env service-description)
+          env-vector (-> user-env
+                         (merge base-env)
+                         (assoc "PORT0" (str port0) "SERVICE_PORT" (str service-port)))
```
pandashadopan, line 801: this shouldn't be env-vector but env-map since env-vector is a map

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
shamsimam: We can revisit this later in another PR if needed.

waiter/src/waiter/util/client_tools.clj
```diff
@@ -502,6 +502,7 @@
                       (ensure-cid-in-headers)
                       (walk/stringify-keys))
                   debug (assoc :x-waiter-debug true))]
+    (println headers)
```
shamsimam, line 505: Please remove this `println`.

waiter/src/waiter/schema.clj
```diff
@@ -105,6 +105,16 @@
   [{:keys [kind] :as config}]
   (nil? (s/check {(s/required-key :factory-fn) s/Symbol, s/Keyword s/Any} (get config kind))))
 
+(defn valid-reverse-proxy-config
```
shamsimam, line 108: This function should return the schema and we can invoke the `check` as a precondition in `kubernetes-scheduler`

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -805,9 +834,14 @@
...
         log-bucket-sync-secs (if log-bucket-url (:log-bucket-sync-secs context) 0)
         total-sigkill-delay-secs (+ pod-sigkill-delay-secs log-bucket-sync-secs)
-        {:keys [reverse-proxy-flag reverse-proxy-offset]} proxy-options
-        has-reverse-proxy? (contains? env reverse-proxy-flag)
-        offset (if has-reverse-proxy? reverse-proxy-offset 0)
+        envoy-sidecar-check-fn (when reverse-proxy
+                                 (-> reverse-proxy
+                                     :predicate-fn
+                                     utils/resolve-symbol
+                                     deref))
```
shamsimam, line 841:
```suggestion
        envoy-sidecar-check-fn (some-> reverse-proxy
                                 :predicate-fn
                                 utils/resolve-symbol
                                 deref)
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -782,12 +782,41 @@
...
+
+(defn attach-envoy-sidecar
+  "Attaches envoy sidecar to replicaset"
+  [replicaset reverse-proxy service-description base-env service-port port0]
+  (update-in replicaset
+    [:spec :template :spec :containers]
+    conj
+    (let [{:keys [cmd resources image]} reverse-proxy
+          user-env (:env service-description)
+          env-vector (-> user-env
```
shamsimam, line 799: `env-vector` is actually a map.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -782,12 +782,41 @@
...
+                         (merge base-env)
+                         (assoc "PORT0" (str port0) "SERVICE_PORT" (str service-port)))
+          env (into []
```
shamsimam, line 802: `env` is a vector, maybe we should flip the names.

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Resolved

waiter/src/waiter/schema.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Resolved

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Resolved

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/scheduler/kubernetes.clj
pandashadopan: Done

waiter/src/waiter/util/client_tools.clj
pandashadopan: Removed

PR #1161, opened 2020-07-23
adds support for PDBs for ReplicaSets requiring greater than one min instance

shamsimam:

- makes the ReplicaSet uid available on the Service
- adds support for PDBs for ReplicaSets requiring greater than one min instance

For services configured to run more than one min-instance, we want to limit the number of disruptions that the service experiences. Configuring Pod Disruption Budgets allows for higher availability while permitting the cluster administrator to manage the cluster's nodes.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -109,9 +109,8 @@
   (try
     (pc/letk
       [[spec
-        [:metadata name namespace [:annotations waiter/service-id]]
-        [:status {replicas 0} {availableReplicas 0} {readyReplicas 0} {unavailableReplicas 0}]]
-       replicaset-json
+        [:metadata name namespace uid [:annotations waiter/service-id]]
+        [:status {replicas 0} {availableReplicas 0} {readyReplicas 0} {unavailableReplicas 0}]] replicaset-json
        requested (get spec :replicas 0)
        staged (- replicas (+ availableReplicas unavailableReplicas))]
       (scheduler/make-Service
@@ -119,6 +118,7 @@
          :instances requested
          :k8s/app-name name
          :k8s/namespace namespace
+         :k8s/replicaset-uid uid
          :task-count replicas
          :task-stats {:healthy readyReplicas
                       :running (- replicas staged)
@@ -516,7 +516,7 @@
 (defn create-service
   "Reify a Waiter Service as a Kubernetes ReplicaSet."
   [{:keys [service-description service-id]}
-   {:keys [api-server-url replicaset-api-version replicaset-spec-builder-fn] :as scheduler}]
+   {:keys [api-server-url pdb-api-version pdb-spec-builder-fn replicaset-api-version replicaset-spec-builder-fn] :as scheduler}]
   (let [{:strs [cmd-type]} service-description]
     (when (= "docker" cmd-type)
       (throw (ex-info "Unsupported command type on service"
@@ -525,16 +525,23 @@
                        :service-id service-id}))))
   (let [rs-spec (replicaset-spec-builder-fn scheduler service-id service-description)
         request-namespace (k8s-object->namespace rs-spec)
-        request-url (str api-server-url "/apis/" replicaset-api-version "/namespaces/"
-                         request-namespace "/replicasets")
+        request-url (str api-server-url "/apis/" replicaset-api-version "/namespaces/" request-namespace "/replicasets")
         response-json (api-request request-url scheduler
                                    :body (utils/clj->json rs-spec)
                                    :request-method :post)
-        service (some-> response-json replicaset->Service)]
+        {:keys [k8s/replicaset-uid] :as service} (some-> response-json replicaset->Service)]
     (when-not service
       (throw (ex-info "failed to create service"
                       {:service-description service-description
                        :service-id service-id})))
+    (let [{:strs [min-instances]} service-description]
+      (when (> min-instances 1)
+        (let [pdb-spec (pdb-spec-builder-fn pdb-api-version rs-spec replicaset-uid)
+              request-url (str api-server-url "/apis/" pdb-api-version "/namespaces/" request-namespace "/poddisruptionbudgets")]
+          (log/info "creating pod disruption budget with min available instance of one for" service-id)
+          (api-request request-url scheduler
+                       :body (utils/clj->json pdb-spec)
+                       :request-method :post))))
     service))
 
 (defn- delete-service
@@ -576,6 +583,8 @@
                                 log-bucket-url
                                 max-patch-retries
                                 max-name-length
+                                pdb-api-version
+                                pdb-spec-builder-fn
                                 pod-base-port
                                 pod-sigkill-delay-secs
                                 pod-suffix-length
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -525,16 +525,23 @@
...
+        {:keys [k8s/replicaset-uid] :as service} (some-> response-json replicaset->Service)]
     (when-not service
       (throw (ex-info "failed to create service"
                       {:service-description service-description
                        :service-id service-id})))
+    (let [{:strs [min-instances]} service-description]
+      (when (> min-instances 1)
+        (let [pdb-spec (pdb-spec-builder-fn pdb-api-version rs-spec replicaset-uid)
+              request-url (str api-server-url "/apis/" pdb-api-version "/namespaces/" request-namespace "/poddisruptionbudgets")]
+          (log/info "creating pod disruption budget with min available instance of one for" service-id)
```
DaoWen, line 541: This log line belongs inside `pdb-spec-builder-fn`, or it should pull some keys out of the result, rather than assuming the function returns a pdb with min-available=1.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -964,6 +973,25 @@
...
+    {:apiVersion pdb-api-version
+     :kind "PodDisruptionBudget"
+     :metadata {:name (str rs-name "-pdb")
+                :ownerReferences [{:apiVersion rs-api-version
+                                   :blockOwnerDeletion true
+                                   :controller false
+                                   :kind rs-kind
+                                   :name rs-name
+                                   :uid replicaset-uid}]}
+     :spec {:minAvailable 1
```
DaoWen, line 992: Why min-available=1 rather than max-unavailable=1?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -525,16 +525,23 @@
...
                                    :body (utils/clj->json rs-spec)
                                    :request-method :post)
-        service (some-> response-json replicaset->Service)]
+        {:keys [k8s/replicaset-uid] :as service} (some-> response-json replicaset->Service)]
     (when-not service
       (throw (ex-info "failed to create service"
                       {:service-description service-description
                        :service-id service-id})))
+    (let [{:strs [min-instances]} service-description]
+      (when (> min-instances 1)
```
DaoWen, line 538: I believe `pdb-spec-builder-fn` is treated as optional in the Scheduler constructor below, and I agree with that decision (i.e., not breaking current configs). Based on that, shouldn't we similarly have a non-nil check for `pdb-spec-builder-fn` in this conditional?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -525,16 +525,23 @@
...
+        (let [pdb-spec (pdb-spec-builder-fn pdb-api-version rs-spec replicaset-uid)
+              request-url (str api-server-url "/apis/" pdb-api-version "/namespaces/" request-namespace "/poddisruptionbudgets")]
+          (log/info "creating pod disruption budget with min available instance of one for" service-id)
+          (api-request request-url scheduler
```
DaoWen, line 542: Do we want a try/catch around this? Right now the function will fail if we fail to create the PDB. Is that expected and/or desired?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -964,6 +973,25 @@
...
+(defn default-pdb-spec-builder
+  "Factory function which creates a Kubernetes PodDisruptionBudget spec for the given ReplicaSet."
+  [pdb-api-version rs-spec replicaset-uid]
+  (let [rs-api-version (get rs-spec :apiVersion)
+        rs-kind (get rs-spec :kind)
+        rs-name (get-in rs-spec [:metadata :name])
+        rs-selector (get-in rs-spec [:spec :selector])]
+    {:apiVersion pdb-api-version
+     :kind "PodDisruptionBudget"
+     :metadata {:name (str rs-name "-pdb")
```
DaoWen, line 985: Should we add something random to the name?

If an RS gets deleted and then a replacement is created before the k8s gc runs, then the old RS's PDB might still be around, and cause a conflict error when we try to post a new PDB with the same name.

shamsimam: My understanding of `blockOwnerDeletion true` means that the PDB needs to be deleted before the ReplicaSet deletion is complete and any recreation triggered.

> Once the garbage collector has deleted all "blocking" dependents (objects with ownerReference.blockOwnerDeletion=true), it deletes the owner object.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: I chose `minAvailable=1` as that is least disruptive for cluster operators. 

Changed it to `(quot (inc min-instances) 2)` now.
This formula works even if the function is called with `min-instances=1`.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Ah, I missed the `or` there. Sorry.

We need to have some way to turn this feature off. Either we have the default be `nil` (which means this whole block gets ignored), or we allow the user to specify a function that returns `nil`, and don't make the `api-request` below in that case.

I think we should definitely handle the second case as a way to disable this feature and make it more robust.

We might also want to change the default behavior back to no-op, but creating the PDBs is probably better.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: I don't think we want to fail to return the `service` just because the PDB creation failed.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -964,6 +972,33 @@
...
+        rs-selector (get-in rs-spec [:spec :selector])
+        service-id (k8s-object->service-id rs-spec)
+        min-available (quot (inc rs-replicas) 2)]
+    (log/info "creating pod disruption budget"
+              {:min-available min-available
+               :replicaset-name rs-name
+               :service-id service-id})
+    {:apiVersion pdb-api-version
+     :kind "PodDisruptionBudget"
+     :metadata {:annotations {:waiter/service-id service-id}
```
DaoWen, line 991: We should set the same `label` values on the PDB that we set on the ReplicaSet.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: We use Background deletion, not Foreground:

https://github.com/twosigma/waiter/blob/eeb0e6fb98a4e13c0e6ef2104f5390c0cf1cb57d/waiter/src/waiter/scheduler/kubernetes.clj#L553-L554

Your quote is from the Foreground section. The rules for Background are different:

>In background cascading deletion, Kubernetes deletes the owner object immediately and the garbage collector then deletes the dependents in the background.

https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#background-cascading-deletion

PR #1166, opened 2020-07-30
Envoy sidecar integration test is not consistent with internal design decision

zaurbek:

- Removal of an integration test that relied on a fact that we can request to create an Envoy sidecar without gRPC transcoding

- We addressed feedback that we only want to include PROTO_DESCRIPTOR to turn on envoy sidecar with transcoding
- All Envoy sidecars now are using gRPC transcoding and do not support simple port forwarding.
- As a result of this change, this integration test can be considered deprecated


DaoWen:
I don't remember the context on this. Since this hasn't been causing problems we either already fixed it or the integration test is fine as-is.

PR #1175, opened 2020-08-10
renames blacklist to eject for internal uses

shamsimam:

- renames blacklist to eject for internal uses

Renaming to use inclusive language.


waiter/config-full.edn
```diff
@@ -777,14 +777,14 @@
...
- ; elapsed since their last use. This is referred to as blacklisting an instance:
- :blacklist-config {
-                    ;; Erroneous instances are blacklisted using an exponential delay based on the number of successive
-                    ;; failures and blacklist-backoff-base-time-ms (milliseconds):
-                    :blacklist-backoff-base-time-ms 10000
-
-                    ;; Killed instances are blacklisted for max-blacklist-time-ms (milliseconds):
-                    :max-blacklist-time-ms 300000}
+ ; elapsed since their last use. This is referred to as ejecting an instance:
+ :eject-config {
```
sradack, line 781: `eject` is the verb, `ejection` is the process by which something becomes `ejected`, but after discussion, use of `eject` instead of `ejection` is probably good enough and avoids the annoying choice of whether to use `ejection` or `eject` in a bunch of places.

waiter/src/waiter/core.clj
```diff
@@ -98,8 +98,8 @@
                              ["/" :service-id "/refresh"] :service-refresh-handler-fn
                              ["/" :service-id "/resume"] :service-resume-handler-fn
                              ["/" :service-id "/suspend"] :service-suspend-handler-fn}
-                     "blacklist" {"" :blacklist-instance-handler-fn
-                                  ["/" :service-id] :blacklisted-instances-list-handler-fn}
+                     "eject" {"" :eject-instance-handler-fn
+                              ["/" :service-id] :ejected-instances-list-handler-fn}
```
sradack, line 102: Feels like maybe this should be `/ejected/<service-id>`

PR #1182, opened 2020-08-19
add support for cookie keep-alive (transparent auth cookie refresh)

shamsimam:

- add the `x-auth-expires-at cookie` in the `x-waiter-auth` set-cookie response
- add `expires-at` metadata to the `x-waiter-auth` cookie
- x-waiter-auth cookie validation expects the `expires-at` metadata and fails to validate when it is missing (i.e. old cookie will no longer validate)
- two new endpoints (described below)

Add support for cookie keep-alive triggered via daemon javascript threads. 
The change includes adding an `x-auth-expires-at` cookie in the authenticated response that is visible to Javascript snippets (i.e. not setting the `HttpOnly` attribute).
This PR also introduces an `expires-at` metadata field into the `x-waiter-auth` cookie to allow easily determining when the auth cookie is set to expire.
As a result of this change, **all existing `x-waiter-auth` cookies will no longer validate** (`decoded-auth-valid?` returns false for old cookies) and trigger authentication flows after deployment.

This PR introduces two endpoints:

- `/.well-known/auth/expires-at` that returns a JSON response explicitly confirming the expiry time of the cookie in UTC epoch seconds.
-  `/.well-known/auth/keep-alive` that accepts an offset query parameter and triggers an authentication flow if the auth cookie is set to expire inside the offset period.


waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +219,62 @@
...
+(defn process-auth-expires-at-request
+  "Handler to allow a client to update it's knowledge of when user's cookie-based credentials expire.
+   Returns a json response containing the expires-at key containing the expiration time in UTC epoch seconds.
+   Relies on the metadata in the x-waiter-auth cookie."
+  [password {:keys [headers]}]
+  (let [decoded-auth-cookie (get-and-decode-auth-cookie-value headers password)
+        [auth-principal _ auth-metadata] decoded-auth-cookie
+        {:keys [expires-at]} auth-metadata
+        expires-at (or (when (decoded-auth-valid? decoded-auth-cookie) expires-at) 0)]
+    (log/info auth-principal "cookie expires at" expires-at)
```
DaoWen, line 232: Should this be debug level?

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +219,62 @@
...
+
+(defn process-auth-keep-alive-request
+  "Handler to eagerly trigger authentication workflow even if cookie has not yet expired.
+   This allows clients to pre-emptively refresh credentials before they expire.
+   Presence of done parameter is used to avoid infinite auth redirect loops and will return 204 No Content.
+   Invalid offset parameters result in a 400 error.
+   Missing offset or soon to expire cookie (based on offset) will trigger in the authentication workflow.
+   If cookie is expected to be live longer than offset value, return 204 No Content."
+  [password auth-handler {:keys [headers] :as request}]
+  (let [{:strs [done offset]} (-> request ru/query-params-request :query-params)
```
DaoWen, line 244: Can we call this `delta` instead of `offset`? I was confused about what this meant because I usually see "offset" used for string offsets, or for UTC offsets (i.e., timezones).

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +219,62 @@
...
+                                :status http-400-bad-request)
+
+      ;; offset parameter must be positive when provided
+      (and offset-parsed (not (pos? offset-parsed)))
+      (utils/clj->json-response {:message "Invalid offset parameter"
+                                 :parameter {:offset offset-parsed}}
+                                :status http-400-bad-request)
+
+      ;; trigger auth workflow if cookie has already expired or adding offset will cause it to expire
+      (or (nil? offset)
```
DaoWen, line 271: The logic here doesn't match the comment. Why does eliding the offset/delta force a refresh?

waiter/src/waiter/cookie_support.clj
```diff
@@ -78,6 +79,12 @@
               (assoc-in response [:headers "set-cookie"] new-header)))]
     (ru/update-response response add-cookie-into-response)))
 
+(defn add-encoded-cookie
+  "Inserts the provided name-value pair as a Set-Cookie header in the response after encoding the value."
+  [response password name value age-in-seconds http-only?]
+  (let [encoded-value (encode-cookie value password)]
```
DaoWen, line 85: Are we potentially leaking our password in an easy-to-brute-force manner here? Should we be using a per-token salt or something?

https://github.com/twosigma/waiter/blob/e0ca0902c2ecd763fdd844d4675d799c5de2424b/waiter/src/waiter/cookie_support.clj#L60-L64

https://github.com/ptaoussanis/nippy#encryption-v2

waiter/src/waiter/util/utils.clj
```diff
@@ -554,6 +554,16 @@
...
   (Boolean/parseBoolean (str (get params flag "false"))))
 
+(defn parse-int
+  "Returns either the input as an integer or nil if there was an error in parsing."
+  [header-value]
+  (try
+    (when header-value
+      (Integer/parseInt (str header-value)))
+    (catch Exception _
+      (log/warn "cannot convert header to an int:" header-value)
```
DaoWen, line 564: Why is this a warning-level log, whereas other places that we fail to parse things it just emits an info-level log?

Also, why is the parameter here called `header-value` and the log line also says "header"? Why is this function specific to headers? It seems like there are other places we might want to use this generic-sounding `parse-int` helper.

waiter/test/waiter/state/responder_test.clj
```diff
@@ -412,6 +413,7 @@
                                      (:instance-id->state expected-state)
                                      (merge (let [[slots-assigned slots-used slots-available] (compute-slots-values (:instance-id->state expected-state))]
                                               {"slots-assigned" slots-assigned "slots-available" slots-available "slots-in-use" slots-used})))]
+          (log/info "expected counters" expected-counter-map)
```
DaoWen, line 416: What is this?

waiter/src/waiter/auth/authentication.clj
DaoWen: 😢

waiter/src/waiter/auth/authentication.clj
DaoWen: The issue wasn't with the docstring, it's with the comment on line 270. The `(nil? offset)` case doesn't seem to follow from the comment.

shamsimam: Oops, updated the comment.

waiter/src/waiter/cookie_support.clj
DaoWen: I agree that if this is an issue, it was already an issue before refactoring. I'm just wondering if this is something we need to address (not necessarily in this PR).

@sradack, thoughts?

waiter/src/waiter/auth/authentication.clj
```diff
@@ -162,20 +188,21 @@
   Authenticator
   (wrap-auth-handler [_ request-handler]
     (let [single-user-prefix "SingleUser "]
-      (fn anonymous-handler [request]
+      (fn anonymous-handler [{:keys [headers] :as request}]
         (let [auth-header (select-auth-header request #(str/starts-with? % single-user-prefix))
+              {:strs [x-waiter-single-user]} headers
```
sradack, line 193: Please document the purpose of `x-waiter-single-user` and the general reason for changes to this function.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +228,110 @@
...
+(defn process-auth-expires-at-request
+  "Handler to allow a client to update its knowledge of when a user's cookie-based credentials expire.
+   Returns a json response containing the expires-at key containing the expiration time in UTC epoch seconds.
+   Relies on the metadata in the x-waiter-auth cookie."
+  [password {:keys [headers]}]
+  (let [decoded-auth-cookie (get-and-decode-auth-cookie-value headers password)
+        [auth-principal _ auth-metadata] decoded-auth-cookie
+        {:keys [expires-at]} auth-metadata
+        cookie-valid? (decoded-auth-valid? decoded-auth-cookie)
+        sanitized-expires-at (or (when cookie-valid? expires-at) 0)]
```
sradack, line 241: In what situations is it possible to hit this endpoint without a valid cookie?  Should we avoid emitting anything if the cookie is not valid?

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +228,110 @@
...
+          ;; remove existing auth cookies
+          (update-in [:headers "cookie"] remove-auth-cookie)
+          ;; trigger auth as if it is a proxy request
+          (assoc :waiter-api-call? false)
+          (auth-handler)
+          (ru/update-response
+            (fn [response]
+              (-> response
+                (remove-done-param-from-keep-alive-https-redirect)
+                (attach-authorization-headers)
```
sradack, line 330: Please document why these always attached as opposed to only when `x-waiter-debug=true`.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +228,110 @@
...
+  [password auth-handler {:keys [headers query-string] :as request}]
+  (let [{:strs [done offset]} (-> request ru/query-params-request :query-params)
+        offset-parsed (utils/parse-int offset)
+        decoded-auth-cookie (get-and-decode-auth-cookie-value headers password)
+        [auth-principal _ cookie-metadata] decoded-auth-cookie
+        current-epoch-time (tc/to-epoch (t/now))
+        {:keys [expires-at]} cookie-metadata
+        ;; handle legacy cookies which will not have this value set
+        expires-at (or expires-at current-epoch-time)
+        waiter-token (some-> request utils/request->host utils/authority->host)]
```
sradack, line 286: This isn't a waiter token until it's been validated as such.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +228,110 @@
...
+      (or (nil? offset)
+          (not (decoded-auth-valid? decoded-auth-cookie))
+          (>= (+ current-epoch-time offset-parsed) expires-at))
+      (do
+        (log/info "initiating authentication flow for keep-alive")
+        (-> request
+          ;; ensure loop back to this endpoint terminates
+          (assoc :query-string (str query-string (when query-string "&") auth-keep-alive-done-parameter))
+          ;; remove existing auth cookies
+          (update-in [:headers "cookie"] remove-auth-cookie)
```
sradack, line 322: Please document that this is necessary to force reauthentication, since when an unexpired cookie is valid a new cookie will not be generated.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +228,110 @@
...
+          ;; trigger auth as if it is a proxy request
+          (assoc :waiter-api-call? false)
```
sradack, line 324: This certainly feels like an API call.  Please document why this should not be considered API call.

waiter/src/waiter/auth/authentication.clj
shamsimam: Renamed variable, but it is treated as a Waiter token later on in the code.

waiter/src/waiter/auth/authentication.clj
shamsimam: Added comment

waiter/src/waiter/auth/authentication.clj
shamsimam: Added comment, OIDC and JWT auth behavior may be different for Proxy and Waiter api requests

waiter/src/waiter/auth/authentication.clj
shamsimam: Changed to add them only for x-waiter-debug header is provided.

waiter/src/waiter/cookie_support.clj
```diff
@@ -78,6 +79,12 @@
               (assoc-in response [:headers "set-cookie"] new-header)))]
     (ru/update-response response add-cookie-into-response)))
 
+(defn add-encoded-cookie
+  "Inserts the provided name-value pair as a Set-Cookie header in the response after encoding the value."
+  [response password name value age-in-seconds http-only?]
```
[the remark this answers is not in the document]
shamsimam: Not sure if this answers your question, but the `x-auth-expires-at` does not have `HttpOnly=true` as we want Javascript code to be able to view the cookie.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -109,20 +127,26 @@
...
-  (let [well-formed? (and decoded-auth-cookie
-                          (<= 2 (count decoded-auth-cookie) 3)
-                          (integer? auth-time)
-                          (string? auth-principal)
-                          (or (nil? auth-metadata) (map? auth-metadata)))
-        one-day-in-millis (-> 1 t/days t/in-millis)]
-    (and well-formed? (> (+ auth-time one-day-in-millis) (System/currentTimeMillis)))))
+  (let [expires-at (when (map? auth-metadata)
+                     (get auth-metadata :expires-at))]
+    (log/debug "well-formed?" decoded-auth-cookie
```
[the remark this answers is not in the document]
shamsimam: removed the debug log, we log when the cookie is non-nil and not valid.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -52,12 +64,18 @@
...
+  (let [creation-time (t/now)
+        creation-time-millis (tc/to-long creation-time)
+        creation-time-secs (tc/to-epoch creation-time)
+        cookie-age-in-seconds (or age-in-seconds (-> 1 t/days t/in-seconds))
+        expiry-time-secs (+ creation-time-secs cookie-age-in-seconds)
+        cookie-metadata (assoc auth-metadata :expires-at expiry-time-secs)
+        cookie-value [principal creation-time-millis cookie-metadata]]
+    (-> response
+      (cookie-support/add-cookie AUTH-COOKIE-EXPIRES-AT (str expiry-time-secs) cookie-age-in-seconds false)
+      (cookie-support/add-encoded-cookie password AUTH-COOKIE-NAME cookie-value cookie-age-in-seconds true))))
```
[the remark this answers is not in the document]
shamsimam: Added comment.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -21,10 +21,22 @@
...
 
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
 
+(def ^:const auth-expires-at-uri "/.well-known/auth/expires-at")
+
+(def ^:const auth-keep-alive-uri "/.well-known/auth/keep-alive")
+
+(def ^:const auth-keep-alive-done-parameter "done=true")
+
+(def ^:const auth-keep-alive-done-parameter-replacement "waiter-redirect=true")
```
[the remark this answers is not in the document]
shamsimam: Added docstring in `waiter.auth.authentication/remove-done-param-from-keep-alive-https-redirect`

waiter/src/waiter/auth/authentication.clj
```diff
@@ -21,10 +21,22 @@
...
+
+(def ^:const AUTH-COOKIE-EXPIRES-AT "x-auth-expires-at")
 
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
 
+(def ^:const auth-expires-at-uri "/.well-known/auth/expires-at")
+
+(def ^:const auth-keep-alive-uri "/.well-known/auth/keep-alive")
+
+(def ^:const auth-keep-alive-done-parameter "done=true")
```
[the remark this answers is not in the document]
shamsimam: Changde comment in `waiter.auth.authentication/process-auth-keep-alive-request` to:
`ensure loop back to this endpoint terminates instead of continuously triggering re-authentication`

waiter/integration/waiter/basic_test.clj
```diff
@@ -781,7 +784,76 @@
...
+                                :method :get
+                                :query-params {"offset" "10"})]
+              (assert-response-status response http-204-no-content)
+              (assert-waiter-response response)
+              (is (nil? (extract-cookie cookies "x-auth-expires-at")))
+              (is (nil? (extract-cookie cookies "x-waiter-auth"))))
+            (let [{:keys [cookies headers] :as response}
+                  (make-request waiter-url "/.well-known/auth/keep-alive"
+                                :cookies request-cookies
+                                :disable-auth false
```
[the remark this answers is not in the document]
shamsimam: Added comment, avoids eagerly sending kerberos authorization headers.

waiter/integration/waiter/basic_test.clj
```diff
@@ -781,7 +784,76 @@
...
+                                :headers {"x-waiter-debug" true}
+                                :method :get
+                                :query-params {"offset" "100000000"})
+                  response-auth-expires-at-cookie (extract-cookie cookies "x-auth-expires-at")
+                  response-waiter-auth-cookie (extract-cookie cookies "x-waiter-auth")]
+              (assert-response-status response http-204-no-content)
+              (assert-waiter-response response)
+              (assert-waiter-authentication-cookies cookies)
+              (is (contains? headers "x-waiter-auth-method") (str headers))
```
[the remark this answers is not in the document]
shamsimam: Changed code to send these headers only when x-waiter-debug header is provided.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -254,7 +255,9 @@
               user (assoc "x-waiter-auth-user" (str user))))))
 
 (defn remove-done-param-from-keep-alive-https-redirect
-  "Removes the done=true query parameter from the response when it is an https redirect to the keep-alive endpoint."
+  "Removes the done=true query parameter from the response when it is an https redirect to the keep-alive endpoint.
+   Keeping the done=true parameter would prevent authentication to be triggered for soon to expire cookies after
+   the requests comes back with https."
```
sradack, line 260:
```suggestion
   Keeping the done=true parameter would prevent authentication from being triggered for soon-to-expire cookies after
   the requests comes back with https."
```

waiter/src/waiter/auth/authentication.clj
```diff
@@ -283,28 +286,28 @@
...
-        waiter-token (some-> request utils/request->host utils/authority->host)]
+        request-host (some-> request utils/request->host utils/authority->host)]
     (log/info auth-principal "cookie expires at" expires-at "offset is" offset-parsed)
     (cond
       ;; avoid infinite redirect loop
       done
       (-> {:status http-204-no-content}
         (utils/attach-waiter-source)
-        (assoc :waiter/token waiter-token))
+        (assoc :waiter/token request-host))
```
sradack, line 296: Contexts outside of this function may reasonably believe `:waiter/token` to always contain valid (extant) Waiter tokens.  We should validate that this is, in fact, a Waiter token and return an error code if it is not.

waiter/src/waiter/auth/authentication.clj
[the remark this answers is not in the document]
sradack: Nitpick: Extracting constants gives them scope outside of the original functions from which they were extracted and as such they deserve their own comments instead of relying on future contributors to ascertain their meaning from their use within functions.

waiter/src/waiter/auth/authentication.clj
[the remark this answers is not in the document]
sradack: See previous nitpick

waiter/src/waiter/auth/authentication.clj
sradack: The comment was helpful.  Thank you.

waiter/src/waiter/auth/authentication.clj
sradack: If a cookie is expired, we should not emit an expired principal.

waiter/src/waiter/cookie_support.clj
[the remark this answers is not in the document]
sradack: I indeed saw that.  `add-cookie`, not `add-encoded-cookie`, is used for `x-auth-expires-at`.  I'd like to avoid implying that  exposing non `http-only?`, encrypted cookies is an OK thing to do unless we have a compelling use case.

waiter/src/waiter/cookie_support.clj
sradack: As a separate PR, we should explore adding a random value into the map for encrypted cookies.

waiter/src/waiter/auth/authentication.clj
sradack: Is this correct?  I still don't follow entirely.  Why would there be a redirect inside the auth handler to HTTPS that also contains `done=true` as well?

waiter/src/waiter/auth/authentication.clj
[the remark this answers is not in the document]
sradack: From the comment, it's not obvious why the replacement value needs to be `waiter-redirect=true`.  Could you please document?

waiter/src/waiter/auth/authentication.clj
shamsimam: Added check for valid tokens.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -201,3 +225,126 @@
...
+
+  (defn remove-done-param-from-keep-alive-https-redirect
+    "Removes the done=true query parameter from the response when it is an https redirect to the keep-alive endpoint.
+     Keeping the done=true parameter would prevent authentication from being triggered for soon-to-expire cookies after
+     the requests comes back with https.
+     To simplify processing of the query string, we replace the done=true parameter with another parameter,
+     waiter-redirect=true, reflecting that the request was redirected."
+    [{:keys [headers status] :as response}]
+    (let [{:strs [location]} headers
+          location-uri (when-not (str/blank? location)
```
sradack, line 267: Upstream handler adds the `location` header.  Sitting downstream, we're parsing that header to determine whether or not upstream actually is performing a redirect instead of performing auth.  It would be better to do either:

1. Move https redirection downstream and out of the auth layer which is currently upstream, avoiding https concerns here entirely, since the redirect will have already handled and this code can assume HTTPS, or
2. More explicitly indicate that https redirection is occurring with a response map entry like `:waiter/redirecting-to-https` or `:waiter/authentication-performed`, adding the done parameter only when `:waiter/authentication-performed` is present.

My preference is for the former.

waiter/src/waiter/auth/authentication.clj
```diff
@@ -290,27 +290,30 @@
...
+          waiter-token? (not-empty (token->token-parameters request-host))]
       (log/info auth-principal "cookie expires at" expires-at "offset is" offset-parsed)
       (cond
         ;; avoid infinite redirect loop
         done
-        (-> {:status http-204-no-content}
-          (utils/attach-waiter-source)
+        (cond-> (utils/attach-waiter-source {:status http-204-no-content})
+          waiter-token?
           (assoc :waiter/token request-host))
```
sradack, line 301: We should not be responding with success for arbitrary host headers.  My preference would be for 404.

waiter/src/waiter/auth/authentication.clj
shamsimam: Will move https redirects to a higher level filter, e.g. the current https redirect handler in a separate PR.

waiter/src/waiter/auth/authentication.clj
shamsimam: Added check for invalid token and 404 response.

sradack (approved):

PR #1184, opened 2020-08-26
adds support for filtering tokens by run-as-requester and parameterized services

shamsimam:

- adds support for filtering tokens by run-as-requester and parameterized services

We would like to be able to filter tokens by whether they are configured to run-as-requester or as parameterized services.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -589,12 +589,10 @@
       :get (let [{:strs [can-manage-as-user] :as request-params} (-> req ru/query-params-request :query-params)
                  include-deleted (utils/param-contains? request-params "include" "deleted")
                  show-metadata (utils/param-contains? request-params "include" "metadata")
-                 include-run-as-requester? (when (contains? request-params "run-as-requester")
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/token.clj
```diff
@@ -621,7 +627,13 @@
                                                         kv-store token
                                                         :error-on-missing false
                                                         :include-deleted include-deleted)]
-                                 (every? #(% token-parameters) parameter-filter-predicates))))
+                                 (and (every? #(% token-parameters) parameter-filter-predicates)
+                                      (or (nil? include-run-as-requester?)
+                                          (boolean-xnor include-run-as-requester?
+                                                        (sd/run-as-requester? token-parameters)))
```
nsinkov, line 633: is it better to write
(or (nil? include-run-as-requester?)
      (= (boolean include-run-as-requester?) (sd/run-as-requester? token-parameters)))
?

DaoWen: Yes, I also prefer `=` over `boolean-xnor` :grimacing:

shamsimam: Done.

PR #1185, opened 2020-08-31
adds support for relaxed and strict mode for OIDC auth via environment variable

shamsimam:

- adds support for relaxed and strict mode for OIDC auth env

The strict mode instructs Waiter to use the expiry time from the access token during OIDC authentication for the x-waiter-auth cookie. The relaxed mode instructs Waiter to use the default 24 hour expiry on the x-waiter-auth cookie. The relaxed mode allows clients to easily migrate between authentication schemes since the expiry duration is similar.

### DIFF
--- a/waiter/src/waiter/auth/oidc.clj
+++ b/waiter/src/waiter/auth/oidc.clj
@@ -20,11 +20,9 @@
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [digest]
-            [schema.core :as s]
             [waiter.auth.authentication :as auth]
             [waiter.auth.jwt :as jwt]
             [waiter.cookie-support :as cookie-support]
-            [waiter.schema :as schema]
             [waiter.status-codes :refer :all]
             [waiter.util.http-utils :as hu]
             [waiter.util.ring-utils :as ru]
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/auth/oidc.clj
```diff
@@ -20,9 +20,11 @@
             [clojure.string :as str]
             [clojure.tools.logging :as log]
             [digest]
+            [schema.core :as s]
```
sradack, line 23: Schema validation should be reserved for configs.  I understand the desire to not repeat "strict" and "relaxed" but the dependency should flow the other way, if at all.  I would prefer to respecify the list of valid options and remove the schema imports.

waiter/src/waiter/auth/oidc.clj
```diff
@@ -206,29 +213,35 @@
...
-  [allow-oidc-auth-services? request]
+(defn retrieve-oidc-mode-on-service
+  "Returns the OIDC mode, one of :disabled :relaxed or :strict, if OIDC auth is enabled for the service.
+   Result depends on whether OIDC auth is configured explicitly based on env variable or
+   computed using allow-oidc-auth-services? and oidc-default-mode."
+  [allow-oidc-auth-services? oidc-default-mode request]
   ;; service requests will enable OIDC auth based on env variable or when absent, allow-oidc-auth-services?
   (let [use-oidc-auth-env (get-in request [:waiter-discovery :service-description-template "env" "USE_OIDC_AUTH"])]
-    (if (some? use-oidc-auth-env) (= "true" use-oidc-auth-env) allow-oidc-auth-services?)))
+    (if (some? use-oidc-auth-env)
```
sradack, line 232: For the config/defaults, you use two parameters to control behavior: `allow-oidc-auth-services?` and `oidc-default-mode`.  For services, you use one.  They should mirror each other in function.  Either 2 for both or 1 for both.

waiter/src/waiter/auth/oidc.clj
shamsimam: Removed dependency schema packages.

waiter/src/waiter/auth/oidc.clj
shamsimam: Do you mind rephrasing the concern?

sradack (approved):

PR #1186, opened 2020-08-31
makes the fileserver configurable per service

shamsimam:

- makes the fileserver configurable per service

We wish to make the configurable per service. For example, this can be useful while trying to support limited access to backend instance logs.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -486,9 +486,6 @@
                                                            :default-container-image "twosigma/waiter-test-apps:latest"
                                                            ;; The default namespace to use for Kubernetes objects created by Waiter
                                                            :default-namespace "waiter"
-                                                           ;; Function that determines if the fileserver container should be enabled for a service.
-                                                           ;; The function takes the service description as input.
-                                                           :fileserver-container-enabled? waiter.some.package/fileserver-enabled?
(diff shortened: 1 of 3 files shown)


waiter/config-full.edn
```diff
@@ -482,6 +482,9 @@
                                                            :default-container-image "twosigma/waiter-test-apps:latest"
                                                            ;; The default namespace to use for Kubernetes objects created by Waiter
                                                            :default-namespace "waiter"
+                                                           ;; Function that determines if the fileserver container should be enabled for a service.
+                                                           ;; The function takes the service description as input.
+                                                           :fileserver-container-enabled? waiter.some.package/fileserver-enabled?
```
DaoWen, line 487: I strongly prefer that this be implemented as a `:predicate-fn` inside the `:fileserver` config like we did with the `:reverse-proxy` for consistency.

DaoWen: We would also want this function to have the same signature as the `:reverse-proxy :predicate-fn` for consistency and flexibility.

shamsimam: sgtm, will change

shamsimam (commented):
Addressed requested changes

DaoWen (approved):

PR #1188, opened 2020-09-02
Account for sidecar in pod health checks

DaoWen:

The `health?` property of Waiter service instances should be the conjunction of all container-ready results for all containers in the pod with a readiness probe defined. Currently we only check the readiness of the primary (waiter-app) container.

Other containers (e.g., envoy proxy sidecar) may need to be ready before we can safely forward requests to a given instance. Reporting that the service-instance is healthy based on the primary container's readiness will result in a 502 error if the sidecar proxy isn't also ready when the requests start coming.

Note that we only include containers with a readiness-probe defined in the health check. A non-critical sidecar container may still define a custom liveness-probe (to control that container's restarts), or simply define neither.

### DIFF
--- a/waiter/test/waiter/scheduler/kubernetes_test.clj
+++ b/waiter/test/waiter/scheduler/kubernetes_test.clj
@@ -551,7 +551,8 @@
                                       :waiter/user "myself"}
                              :annotations {:waiter/port-count "1"
                                            :waiter/service-id "test-app-1234"}}
-                  :spec {:containers [{:ports [{:containerPort 8080 :protocol "TCP"}]}]
+                  :spec {:containers [{:ports [{:containerPort 8080 :protocol "TCP"}]
+                                       :readinessProbe {:path "/health"}}]
                          :nodeName "node-0.k8s.com"}
                   :status {:phase "Pending"
                            :startTime "2014-09-13T00:24:46Z"
@@ -567,7 +568,8 @@
                                       :waiter/user "myself"}
                              :annotations {:waiter/port-count "1"
                                            :waiter/service-id "test-app-1234"}}
-                  :spec {:containers [{:ports [{:containerPort 8080 :protocol "TCP"}]}]}
+                  :spec {:containers [{:ports [{:containerPort 8080 :protocol "TCP"}]
+                                       :readinessProbe {:path "/health"}}]}
                   :status {:phase "Running"
                            :podIP "10.141.141.11"
                            :startTime "2014-09-13T00:24:46Z"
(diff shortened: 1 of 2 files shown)


waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -1887,14 +1923,6 @@
             instance (pod->ServiceInstance dummy-scheduler pod)]
         (is (= (scheduler/make-ServiceInstance expired-instance-map) instance))))
 
-    (testing "pod with envoy sidecar to instance"
```
DaoWen, line 1890: I moved this code so that the expiry tests are continuous again.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -1887,14 +1923,6 @@
...
-      (let [dummy-scheduler (assoc base-scheduler :restart-expiry-threshold 10)
-            pod' (merge
```
DaoWen, line 1892: This `merge` shouldn't have been merged. :wink: 

Changed to `(-> pod ...)`.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -1887,14 +1923,6 @@
...
-                   (assoc-in pod [:metadata :annotations :waiter/service-port] "8080")
-                   (assoc-in pod [:spec :containers 0 :ports 0 :containerPort] 8081))
-            instance (pod->ServiceInstance dummy-scheduler pod)]
```
DaoWen, line 1895: Note that I also fixed a nasty typo here: `pod` should have been `pod'`; i.e., this test wasn't testing anything. :facepalm:

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -1822,7 +1839,8 @@
 (deftest test-pod->ServiceInstance
   (let [api-server-url "https://k8s-api.example/"
         base-scheduler {:api-server-url api-server-url
-                        :container-running-grace-secs 120}
+                        :container-running-grace-secs 120
+                        :restart-expiry-threshold 10}
```
DaoWen, line 1843: I added `:restart-expiry-threshold 10` to the base-scheduler here because I found it confusing in the tests below how every one had this value explicitly set, even if the test doesn't really care about the expiry threshold. I think treating 10 as the "default" makes the intent more clear, especially with the sidecar-related tests.

DaoWen:
Closing in favor of a new solution in #1193.

PR #1191, opened 2020-09-03
add --admin as option

kevo1ution:
- add "--admin" flag in CLI for whenever attempting to create or update token

- this allows admins to do token create and update operations as admins through CLI

### DIFF
--- a/cli/integration/tests/waiter/test_cli.py
+++ b/cli/integration/tests/waiter/test_cli.py
@@ -1288,6 +1288,7 @@ class WaiterCliTest(util.WaiterTest):
         try:
             token_data = util.load_token(self.waiter_url, token_name)
             self.assertIn('update-mode=admin', cli.stderr(cp))
+            self.assertIn(f'Attempting to {action} token in ADMIN mode', cli.stdout(cp))
             self.assertEqual(cpus, token_data['cpus'])
             self.assertEqual(run_as_user, token_data['run-as-user'])
         finally:
--- a/cli/integration/tests/waiter/test_cli.py
+++ b/cli/integration/tests/waiter/test_cli.py
@@ -1274,17 +1274,22 @@ class WaiterCliTest(util.WaiterTest):
             util.delete_token(self.waiter_url, token_name_2, kill_services=True)
 
     def __test_create_update_token_admin_mode(self, action):
+        cpus = 0.1
+        run_as_user = 'FAKE_USERNAME'
         token_name = self.token_name()
         temp_env = os.environ.copy()
         temp_env["WAITER_ADMIN"] = 'true'
-        action_flags = f'--cpus 0.1 --admin --run-as-user FAKE_USERNAME'
+        action_flags = f'--cpus {cpus} --admin --run-as-user {run_as_user}'
         if action == 'create':
             cp = cli.create(self.waiter_url, token_name, flags="-v", create_flags=action_flags, env=temp_env)
         elif action == 'update':
             cp = cli.update(self.waiter_url, token_name, flags="-v", update_flags=action_flags, env=temp_env)
         self.assertEqual(0, cp.returncode, cp.stderr)
         try:
+            token_data = util.load_token(self.waiter_url, token_name)
             self.assertIn('update-mode=admin', cli.stderr(cp))
+            self.assertEqual(cpus, token_data['cpus'])
+            self.assertEqual(run_as_user, token_data['run-as-user'])
         finally:
             util.delete_token(self.waiter_url, token_name)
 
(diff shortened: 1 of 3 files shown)


dposada:
In the Cook CLI, we enabled admin mode with an environment variable:

https://github.com/twosigma/Cook/blob/master/cli/cook/subcommands/admin.py#L145

Should we consider using the same approach here (e.g. `WAITER_ADMIN=true waiter ...`)?

shamsimam:
@dposada The `admin` flag feels like a better fit. It does not introduce any new subcommand or new arguments to existing subcommand. The flag is discoverable from the help. 
One drawback is non-admin users can see the flag, but their admin operations should be denied when they try and use that flag.

Does the above change your mind overall about using environment variables?

dposada:
I should have been more clear. I'm not suggesting we need a `waiter admin ...` sub-command just for this. We may want it for other stuff, but it's not needed here in my opinion. I was just suggesting that you could avoid the `--admin` flag by instead making "admin mode" enabled when the `WAITER_ADMIN` environment variable is set to truthy. That makes it invisible to non-admin users, and if/when we do add a `waiter admin ...` sub-command, it would use the same thing instead of having different ways to enable "admin mode" in different parts of the CLI.

cli/integration/tests/waiter/test_cli.py
```diff
@@ -1272,3 +1272,23 @@ def test_show_service_current(self):
...
             util.delete_token(self.waiter_url, token_name_2, kill_services=True)
+
+    def __test_create_update_token_admin_mode(self, action):
+        token_name = self.token_name()
+        temp_env = os.environ.copy()
+        temp_env["WAITER_ADMIN"] = 'true'
+        if action == 'create':
+            cp = cli.create(self.waiter_url, token_name, flags="-v", create_flags='--cpus 0.1 --admin', env=temp_env)
+        elif action == 'update':
+            cp = cli.update(self.waiter_url, token_name, flags="-v", update_flags='--cpus 0.1 --admin', env=temp_env)
```
shamsimam, line 1283: Can we include configuring the `run-as-user` field too, it triggers the authorization logic in the Waiter API.

cli/waiter/token_post.py
```diff
@@ -43,18 +44,21 @@ def post_failed_message(cluster_name, reason):
     return f'Token post {terminal.failed("failed")} on {cluster_name}:\n{terminal.reason(reason)}'
 
 
-def create_or_update(cluster, token_name, token_fields, action):
+def create_or_update(cluster, token_name, token_fields, action, update_mode_admin=False):
```
shamsimam, line 47: Can we rename the variable to `admin_mode`? 
Also, I'd prefer it to not have a default value and be placed before `action` in the parameter list.

cli/waiter/token_post.py
```diff
@@ -43,18 +44,21 @@ def post_failed_message(cluster_name, reason):
...
     """Creates (or updates) the given token on the given cluster"""
     cluster_name = cluster['name']
     cluster_url = cluster['url']
 
     existing_token_data, existing_token_etag = get_token(cluster, token_name)
     try:
         print_info(f'Attempting to {action} token on {terminal.bold(cluster_name)}...')
```
shamsimam, line 54: It'd be nice to include admin mode information in this log output too.

cli/integration/tests/waiter/test_cli.py
```diff
@@ -1272,3 +1272,23 @@ def test_show_service_current(self):
...
+        self.assertEqual(0, cp.returncode, cp.stderr)
+        try:
+            self.assertIn('update-mode=admin', cli.stderr(cp))
```
shamsimam, line 1286: We should retrieve the token and assert on the fields too.

cli/integration/tests/waiter/test_cli.py
```diff
@@ -1272,3 +1272,30 @@ def test_show_service_current(self):
...
+            self.assertEqual(cpus, token_data['cpus'])
+            self.assertEqual(run_as_user, token_data['run-as-user'])
+        finally:
+            util.delete_token(self.waiter_url, token_name)
+
+    def test_create_token_admin_mode(self):
+        self.__test_create_update_token_admin_mode('create')
+
+    def test_update_token_admin_mode(self):
+        self.__test_create_update_token_admin_mode('update')
```
shamsimam, line 1301: The update should have an already created token from some other user and then try to update it.

containers/test-apps/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM twosigma/mesos-agent:1.0.0-1.0.1604
+FROM openjdk:11-slim
```
shamsimam, line 1: Please rebase on master so that these changes do not show up on the diff?

cli/integration/tests/waiter/test_cli.py
shamsimam: We should also assert that the create/update of the token without admin mode fails.

shamsimam (approved):

PR #1193, opened 2020-09-03
Proxy health check through envoy sidecar

DaoWen:

When an Envoy (reverse-proxy) sidecar is present on k8s, proxy the health check port through the sidecar, and use that proxied port for readiness checks as well as Waiter's own health check request.

The envoy proxy sidecar needs to be ready before we can safely forward requests to a given instance. Reporting that the service-instance is healthy based on the primary container's readiness will result in a 502 error if the sidecar proxy isn't also ready when the requests start coming.

Note that we only change the readiness probe to proxy through the sidecar; the liveness probe (which triggers container restarts) is not proxied.

### DIFF
--- a/containers/reverse-proxy/Dockerfile
+++ b/containers/reverse-proxy/Dockerfile
@@ -1,7 +1,8 @@
 FROM envoyproxy/envoy-alpine:v1.14.4
 
 COPY envoy-start /opt/waiter/envoy/bin/
-COPY envoy.yaml /etc/envoy/
+COPY envoy1.yaml /etc/envoy/
+COPY envoy2.yaml /etc/envoy/
 
 
 ENTRYPOINT /opt/waiter/envoy/bin/envoy-start
--- a/containers/reverse-proxy/Dockerfile
+++ b/containers/reverse-proxy/Dockerfile
@@ -1,8 +1,8 @@
 FROM envoyproxy/envoy-alpine:v1.14.4
 
 COPY envoy-start /opt/waiter/envoy/bin/
-COPY envoy1.yaml /etc/envoy/
-COPY envoy2.yaml /etc/envoy/
+COPY envoy-one-port.yaml /etc/envoy/
+COPY envoy-two-ports.yaml /etc/envoy/
 
 
 ENTRYPOINT /opt/waiter/envoy/bin/envoy-start
--- a/containers/reverse-proxy/envoy-start
+++ b/containers/reverse-proxy/envoy-start
@@ -19,12 +19,29 @@ else
   exit 1
 fi
 
-echo "$(date -Is) Changing the port values in etc/envoy/envoy.yaml to ${SERVICE_PORT} and ${PORT0}"
+if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
+    echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
+    PROXY_PORT_COUNT=2
+else
+    PROXY_PORT_COUNT=1
+fi
+
+CONFIG_YAML=/etc/envoy/envoy${PROXY_PORT_COUNT}.yaml
+
+echo "$(date -Is) Substituting variables in configuration template"
 
 # change the SERVICE_PORT and PORT0 values in /etc/envoy/envoy.yaml to respective env variables
-sed -i -e "s/SERVICE_PORT/${SERVICE_PORT}/g" -e "s/PORT0/${PORT0}/g" /etc/envoy/envoy.yaml
+(
+set -x
+sed -i \
+    -e "s/{{APP_PORT}}/${PORT0}/g" \
+    -e "s/{{HC_PORT}}/$(( PORT0 + HEALTH_CHECK_PORT_INDEX ))/g" \
+    -e "s/{{PROXY_APP_PORT}}/${SERVICE_PORT}/g" \
+    -e "s/{{PROXY_HC_PORT}}/$(( SERVICE_PORT + HEALTH_CHECK_PORT_INDEX ))/g" \
+    ${CONFIG_YAML}
+)
 
 echo "$(date -Is) Starting envoy"
 
 # start envoy with /etc/envoy/envoy.yaml config file
-exec envoy -c /etc/envoy/envoy.yaml --base-id 1
+exec envoy -c ${CONFIG_YAML} --base-id 1
--- a/containers/reverse-proxy/envoy-start
+++ b/containers/reverse-proxy/envoy-start
@@ -21,12 +21,12 @@ fi
 
 if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
     echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
-    PROXY_PORT_COUNT=2
+    PROXY_PORT_COUNT=two-ports
 else
-    PROXY_PORT_COUNT=1
+    PROXY_PORT_COUNT=one-port
 fi
 
-CONFIG_YAML=/etc/envoy/envoy${PROXY_PORT_COUNT}.yaml
+CONFIG_YAML=/etc/envoy/envoy-${PROXY_PORT_COUNT}.yaml
 
 echo "$(date -Is) Substituting variables in configuration template"
 
@@ -34,14 +34,19 @@ echo "$(date -Is) Substituting variables in configuration template"
 (
 set -x
 sed -i \
-    -e "s/{{APP_PORT}}/${PORT0}/g" \
-    -e "s/{{HC_PORT}}/$(( PORT0 + HEALTH_CHECK_PORT_INDEX ))/g" \
-    -e "s/{{PROXY_APP_PORT}}/${SERVICE_PORT}/g" \
-    -e "s/{{PROXY_HC_PORT}}/$(( SERVICE_PORT + HEALTH_CHECK_PORT_INDEX ))/g" \
+    -e "s/{{SERVICE_PORT}}/${PORT0}/g" \
+    -e "s/{{HEALTH_CHECK_PORT}}/$(( PORT0 + HEALTH_CHECK_PORT_INDEX ))/g" \
+    -e "s/{{PROXY_SERVICE_PORT}}/${SERVICE_PORT}/g" \
+    -e "s/{{PROXY_HEALTH_CHECK_PORT}}/$(( SERVICE_PORT + HEALTH_CHECK_PORT_INDEX ))/g" \
     ${CONFIG_YAML}
 )
 
-echo "$(date -Is) Starting envoy"
+# disable http2 upstream for non-http2 backends
+( echo "${SERVICE_PROTOCOL}" | grep -iq h2 ) || (
+set -x
+sed -i '/http2_protocol_options/d' ${CONFIG_YAML}
+)
 
 # start envoy with /etc/envoy/envoy.yaml config file
+echo "$(date -Is) Starting envoy"
 exec envoy -c ${CONFIG_YAML} --base-id 1
--- a/containers/reverse-proxy/envoy-start
+++ b/containers/reverse-proxy/envoy-start
@@ -2,25 +2,31 @@
 # This script takes no argument and uses SERVICE_PORT and PORT0 environment variables
 # to setup /etc/envoy/envoy.yaml file.
 # SERVICE_PORT and PORT0 must be set
+# SERVICE_POROTOCOL is optional (http|h2c, default http)
+# HEALTH_CHECK_PORT_INDEX is optional ([0-9], default 0)
+
+now() {
+    date -u +%FT%TZ
+}
 
 # check if SERVICE_PORT is present
 if [ "$SERVICE_PORT" ]; then
-  echo "$(date -Is) SERVICE_PORT is set"
+  echo "$(now) SERVICE_PORT is set"
 else
-  echo "$(date -Is) SERVICE_PORT is not set. Exiting envoy-start"
+  echo "$(now) SERVICE_PORT is not set. Exiting envoy-start"
   exit 1
 fi
 
 # check if PORT0 is present
 if [ "$PORT0" ]; then
-  echo "$(date -Is) PORT0 is set"
+  echo "$(now) PORT0 is set"
 else
-  echo "$(date -Is) PORT0 is not set. Exiting envoy-start"
+  echo "$(now) PORT0 is not set. Exiting envoy-start"
   exit 1
 fi
 
 if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
-    echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
+    echo "$(now) HEALTH_CHECK_PORT_INDEX is positive: ${HEALTH_CHECK_PORT_INDEX}"
     PROXY_PORT_COUNT=two-ports
 else
     PROXY_PORT_COUNT=one-port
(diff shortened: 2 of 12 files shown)


containers/reverse-proxy/envoy-start
```diff
@@ -19,12 +19,29 @@ else
...
 
-echo "$(date -Is) Changing the port values in etc/envoy/envoy.yaml to ${SERVICE_PORT} and ${PORT0}"
+if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
+    echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
+    PROXY_PORT_COUNT=2
+else
+    PROXY_PORT_COUNT=1
+fi
+
+CONFIG_YAML=/etc/envoy/envoy${PROXY_PORT_COUNT}.yaml
```
shamsimam, line 29: As mentioned above, I'd prefer to use more descriptive names for the envoy YAML files.

containers/reverse-proxy/envoy-start
```diff
@@ -19,12 +19,29 @@ else
...
 
 # change the SERVICE_PORT and PORT0 values in /etc/envoy/envoy.yaml to respective env variables
-sed -i -e "s/SERVICE_PORT/${SERVICE_PORT}/g" -e "s/PORT0/${PORT0}/g" /etc/envoy/envoy.yaml
+(
+set -x
+sed -i \
+    -e "s/{{APP_PORT}}/${PORT0}/g" \
+    -e "s/{{HC_PORT}}/$(( PORT0 + HEALTH_CHECK_PORT_INDEX ))/g" \
+    -e "s/{{PROXY_APP_PORT}}/${SERVICE_PORT}/g" \
+    -e "s/{{PROXY_HC_PORT}}/$(( SERVICE_PORT + HEALTH_CHECK_PORT_INDEX ))/g" \
```
shamsimam, line 40: 👍

shamsimam: Can we use the following names: `PROXY_SERVICE_PORT` and `PROXY_ HEALTH_CHECK_PORT`?

waiter/integration/waiter/grpc_test.clj
```diff
@@ -1085,3 +1085,24 @@
...
+                                     (str "x-waiter-env-" reverse-proxy-flag) "yes")
+              _ (log/info "making canary request")
+              {:keys [cookies service-id] :as response} (make-request-with-debug-info
+                                                          request-headers
+                                                          #(make-shell-request waiter-url % :method :get :path "/"))]
+          (with-service-cleanup
+            service-id
+            (assert-service-on-all-routers waiter-url service-id cookies)
+            (assert-response-status response http-200-ok)
+            (testing "Expected envoy specific headers are present in both request and response"
```
shamsimam, line 1106: This is testing only response headers.
Why is the server changed to Envoy?

shamsimam: Maybe test for `x-envoy-upstream-service-time` instead?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -865,12 +872,13 @@
...
-        offset (if has-reverse-proxy? 1 0)
         ;; Make $PORT0 value pseudo-random to ensure clients can't hardcode it.
         ;; Helps maintain compatibility with Marathon, where port assignment is dynamic.
-        service-port (-> service-id hash (mod 100) (* 10) (+ pod-base-port))
-        port0 (+ service-port offset)
-        health-check-port (+ port0 health-check-port-index)
+        service-id-hash (hash service-id)
+        service-port (service-id-hash->port0 service-id-hash pod-base-port)
+        port0 (if has-reverse-proxy?
+                (service-id-hash->port0 (inc service-id-hash) pod-base-port)
```
shamsimam, line 880: Instead of `inc` can we pass an additional argument to `service-id-hash->port0`?
The incremented value is not a service hash.

Alternatively, we limit to 10 ports, so we can add a reasonable larger number after keeping sufficient buffer. E.g. `50` or `100`.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -914,7 +922,7 @@
        :spec {:replicas min-instances
               :selector {:matchLabels {:app k8s-name
                                        :waiter/user run-as-user}}
-              :template {:metadata {:annotations {:waiter/port-count (str (+ ports offset))
+              :template {:metadata {:annotations {:waiter/port-count (str ports)
```
shamsimam, line 925: Can we add in the top-level `[:metadata :labels]` a `:waiter/envoy-sidecar` entry similar to the `:waiter/fileserver` entry?

Then we can make assumptions about the ports being used based on the pod labels.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -814,15 +814,17 @@
...
   (update-in replicaset
     [:spec :template :spec :containers]
     conj
     (let [{:keys [cmd resources image]} reverse-proxy
           user-env (:env service-description)
           env-map (-> user-env
-                  (merge base-env)
-                  (assoc "PORT0" (str port0) "SERVICE_PORT" (str service-port)))
+                      (merge base-env)
+                      (assoc "HEALTH_CHECK_PORT_INDEX" (str health-check-port-index)
```
shamsimam, line 825: Are we planning on support health-check-proto too?

containers/reverse-proxy/envoy1.yaml
```diff
@@ -43,8 +43,8 @@ static_resources:
               - endpoint:
                   address:
                     socket_address:
-                      address: 0.0.0.0
-                      port_value: PORT0
+                      address: 127.0.0.1
```
shamsimam, line 46: 👍

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: I'm also renaming the function to `get-port-range` since `service-id-hash->port0` isn't very accurate.

containers/reverse-proxy/envoy-start
DaoWen: Renamed all four variables used in the templates:
```
SERVICE_PORT
HEALTH_CHECK_PORT
PROXY_SERVICE_PORT
PROXY_HEALTH_CHECK_PORT
```

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: The existing proof-of-concept doesn't support backend-proto (it just assumes plain-text). Adding that was outside the scope of this particular patch, so I didn't add the health-check-backend-proto either.

containers/reverse-proxy/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM envoyproxy/envoy-alpine:v1.14.4
 
 COPY envoy-start /opt/waiter/envoy/bin/
-COPY envoy.yaml /etc/envoy/
+COPY envoy1.yaml /etc/envoy/
+COPY envoy2.yaml /etc/envoy/
```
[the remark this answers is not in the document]
DaoWen: Renamed `envoy-one-port.yaml` and `envoy-two-ports.yaml`

waiter/integration/waiter/grpc_test.clj
DaoWen: I updated this test to send an actual gRPC (binary) payload and test the response. We also check for the `x-envoy-upstream-service-time` header.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Added new label `waiter/proxy-sidecar` with "enabled" / "disabled" value (just like the `waiter/fileserver` label).

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -848,6 +853,16 @@
...
 
+(def ^:const service-ports-index 0)
+(def ^:const proxied-ports-index 1)
+
+(defn get-port-range
+  "0th port in a range of up to 10 for a given service-id's hash.
+   Note that each container is limited to a contiguous 100 ranges of 10 ports,
+   therefore the return values are only unique for range-index [0, 99]."
+  [service-id-hash range-index base-port]
+  (-> service-id-hash (+ range-index) (mod 100) (* 10) (+ base-port)))
```
shamsimam, line 864: Nitpick: we should extract the `10` in a constant, preferably use the same value in port value constraints.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -91,7 +91,7 @@
      map->KubernetesScheduler)))
 
 (def dummy-service-description
-  {"backend-proto" "http"
+  {"backend-proto" "HTTP"
```
shamsimam, line 94: Backend proto is always one of http and https. If this needs to be capitalized, that processing needs to happen elsewhere.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -826,15 +826,20 @@
...
     (let [{:keys [cmd resources image]} reverse-proxy
           user-env (:env service-description)
           env-map (-> user-env
-                  (merge base-env)
-                  (assoc "PORT0" (str port0) "SERVICE_PORT" (str service-port)))
+                      (merge base-env)
+                      (assoc "HEALTH_CHECK_PORT_INDEX" (str health-check-port-index)
+                             "PORT0" (str port0)
+                             "SERVICE_PORT" (str service-port)
+                             "SERVICE_PROTOCOL" backend-proto))
```
shamsimam, line 842: The `backend-proto` may need to be capitalized here.

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -164,49 +165,151 @@
...
+    (let [service-id "proxy-health-test-service-id"
+          scheduler (make-dummy-scheduler [service-id] {:reverse-proxy {:cmd ["/opt/waiter/envoy/bin/envoy-start"]
+                                                                               :image "twosigma/waiter-envoy"
+                                                                               :predicate-fn 'waiter.scheduler.kubernetes/envoy-sidecar-enabled?
+                                                                               :resources {:cpu 0.1 :mem 256}
+                                                                               :scheme "http"}})
+          service-description (assoc dummy-service-description
+                                     "env" {"REVERSE_PROXY" "yes"
+                                            "PORT0" "to-be-overwritten"
+                                            "SERVICE_PORT" "to-be-overwritten"}
```
shamsimam, line 261: Nitpick: lexicographic order

waiter/integration/waiter/grpc_test.clj
```diff
@@ -1095,3 +1095,33 @@
...
+                                     "content-type" "application/grpc"
+                                     (str "x-waiter-env-" reverse-proxy-flag) "yes")
+              _ (log/info "making canary request")
+              ;; grpc post-data = {id: "x", from: "y", message: "z"}
+              ;; see the Proto3 Language Guide for details on the binary encoding:
+              ;; https://developers.google.com/protocol-buffers/docs/proto3
+              post-data (byte-array [0x00 0x00 0x00 0x00 0x09 0x0a 0x01 0x78 0x12 0x01 0x79 0x1a 0x01 0x7a])
+              {:keys [cookies service-id] :as response} (make-request-with-debug-info
+                                                          request-headers
+                                                          #(make-shell-request waiter-url % :body post-data :path "/courier.Courier/SendPackage"))]
```
shamsimam, line 1115: Nitpick: should we use `start-courier-instance` instead to start the service and then make grpc requests.

containers/reverse-proxy/envoy-start
```diff
@@ -19,12 +19,34 @@ else
...
+if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
+    echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
+    PROXY_PORT_COUNT=two-ports
+else
+    PROXY_PORT_COUNT=one-port
+fi
+
+CONFIG_YAML=/etc/envoy/envoy-${PROXY_PORT_COUNT}.yaml
+
+echo "$(date -Is) Substituting variables in configuration template"
```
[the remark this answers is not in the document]
DaoWen: You can install GNU `date` on your mac: https://stackoverflow.com/a/38856066/1427124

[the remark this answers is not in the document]
DaoWen: Lifted `date` invocation into a new `now` function, and used `date -u +%FT%TZ` instead (which I think will work on the default `date` included on your mac).

containers/reverse-proxy/envoy-start
```diff
@@ -19,12 +19,34 @@ else
   exit 1
 fi
 
-echo "$(date -Is) Changing the port values in etc/envoy/envoy.yaml to ${SERVICE_PORT} and ${PORT0}"
+if [ ${HEALTH_CHECK_PORT_INDEX:-0} -gt 0 ]; then
```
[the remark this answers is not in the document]
DaoWen: Added `SERVICE_POROTOCOL` and `HEALTH_CHECK_PORT_INDEX` to docstring.

containers/reverse-proxy/envoy-start
```diff
@@ -19,12 +19,34 @@ else
...
+    echo "$(date -Is) HEALTH_CHECK_PORT_INDEX is positive"
```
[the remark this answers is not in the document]
DaoWen: Added value of `HEALTH_CHECK_PORT_INDEX` to this `echo`.

waiter/test/waiter/scheduler/kubernetes_test.clj
DaoWen: I can't remember why I thought this was necessary... I was probably thinking that Waiter forces this to caps in production. Reverted to lower-case.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: No longer needed since we're consistently using lower-case again.

waiter/test/waiter/scheduler/kubernetes_test.clj
DaoWen: Fixed indirectly since "REVERSE_PROXY" has been replaced by a variable.

waiter/integration/waiter/grpc_test.clj
DaoWen: Fixed

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -343,7 +343,6 @@
     (when (using-k8s? waiter-url)
       (if (contains? (get-kubernetes-scheduler-settings waiter-url) :reverse-proxy)
         (let [reverse-proxy-flag "REVERSE_PROXY"
```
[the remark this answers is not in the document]
DaoWen: The flag (environment variable name) is now read from a `WAITER_TEST_...` environment variable, which I think is consistent with how we set up other test parameters like this.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Since I'm not going to audit all of the uses of `10` right now, I opened an issue for this: https://github.com/twosigma/waiter/issues/1204

shamsimam (approved):

PR #1203, opened 2020-09-11
Service delete wait to ensure deletion across routers

kevo1ution:
## Changes proposed in this PR

- added internal route /apps/service-id/ensure-delete which polls atom for service deletion
- modify route DELETE /apps/service-id to take a query parameter "timeout", and polls other routers to ensure deletion of service
- default timeout is 0, to not release any breaking changes

## Why are we making these changes?

- DELETE /apps/service-id should ensure that the service is deleted across the other routers or it will respond prematurely

PR #1205, opened 2020-09-17
CLI kill waits for routers to agree with timeout query parameter

kevo1ution:

- for `waiter kill service-id --timeout TIMEOUT` command: use the timeout query parameter instead of the client side polling method

Here is timing evidence where server sleeps and client is forced to timeout:
```
$ time waiter kill service-id --timeout 5
Killing service w9091-499285cd09532196626361355fe0699d in WAITER1...
Encountered error while killing w9091-499285cd09532196626361355fe0699d in WAITER1.
waiter kill service-id --timeout 5  
0.18s user 
0.02s system 3% cpu 
5.204 total
```

- removes the need to poll on the client side and ensures that the command returns when all routers agree that a service has been deleted

depends on #1203

### DIFF
--- a/cli/waiter/http_util.py
+++ b/cli/waiter/http_util.py
@@ -72,7 +72,7 @@ def __delete(url, params=None, headers=None, read_timeout=None):
     delete_timeouts = timeouts
     if read_timeout is not None:
         delete_timeouts = (timeouts[0], read_timeout)
-    return session.delete(url, params=params, timeout=get_timeouts, headers=headers)
+    return session.delete(url, params=params, timeout=delete_timeouts, headers=headers)
 
 
 def __make_url(cluster, endpoint):
--- a/cli/waiter/http_util.py
+++ b/cli/waiter/http_util.py
@@ -69,9 +69,9 @@ def __get(url, params=None, read_timeout=None, **kwargs):
 def __delete(url, params=None, headers=None, read_timeout=None):
     """Sends a DELETE with params to the given url"""
     logging.debug(f'DELETE {url} with params {params} and headers {headers}')
-    get_timeouts = timeouts
+    delete_timeouts = timeouts
     if read_timeout is not None:
-        get_timeouts = (timeouts[0], read_timeout)
+        delete_timeouts = (timeouts[0], read_timeout)
     return session.delete(url, params=params, timeout=get_timeouts, headers=headers)
 
 
--- a/cli/waiter/http_util.py
+++ b/cli/waiter/http_util.py
@@ -66,10 +66,13 @@ def __get(url, params=None, read_timeout=None, **kwargs):
     return session.get(url, params=params, timeout=get_timeouts, **kwargs)
 
 
-def __delete(url, params=None, headers=None):
+def __delete(url, params=None, headers=None, read_timeout=None):
     """Sends a DELETE with params to the given url"""
     logging.debug(f'DELETE {url} with params {params} and headers {headers}')
-    return session.delete(url, params=params, timeout=timeouts, headers=headers)
+    get_timeouts = timeouts
+    if read_timeout is not None:
+        get_timeouts = (timeouts[0], read_timeout)
+    return session.delete(url, params=params, timeout=get_timeouts, headers=headers)
 
 
 def __make_url(cluster, endpoint):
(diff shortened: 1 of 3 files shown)


shamsimam (commented):
I'm wondering whether we should continue to have the timeout on the CLI in case the backend server misbehaves.

cli/waiter/http_util.py
```diff
@@ -66,10 +66,13 @@ def __get(url, params=None, read_timeout=None, **kwargs):
     return session.get(url, params=params, timeout=get_timeouts, **kwargs)
 
 
-def __delete(url, params=None, headers=None):
+def __delete(url, params=None, headers=None, read_timeout=None):
     """Sends a DELETE with params to the given url"""
     logging.debug(f'DELETE {url} with params {params} and headers {headers}')
-    return session.delete(url, params=params, timeout=timeouts, headers=headers)
+    get_timeouts = timeouts
```
shamsimam, line 72: Let's name this `delete_timeouts`

cli/waiter/subcommands/kill.py
```diff
@@ -5,29 +5,29 @@
...
-            if killed:
+            if resp.json().get('routers-agree', False):
                 print(f'Successfully killed {service_id} in {cluster_name}.')
                 return True
             else:
-                print_error('Timeout waiting for service to die.')
-                return False
+                print(f'Successfully killed {service_id} in {cluster_name}, but routers may not be updated yet!')
+                return True
         else:
```
shamsimam, line 31: We should differentiate between a failed delete (non-200 response) reported by the server and us timing out and output different error messages.

cli/waiter/subcommands/kill.py
```diff
@@ -122,8 +122,11 @@ def register(add_parser):
...
     parser.add_argument('--force', '-f', help='kill all services, never prompt', dest='force', action='store_true')
-    parser.add_argument('--timeout', '-t', help='timeout (in seconds) for kill to complete',
-                        type=check_positive, default=30)
     parser.add_argument('--service-id', '-s', help='kill by service id instead of token',
                         dest='is-service-id', action='store_true')
+    timeout_group = parser.add_mutually_exclusive_group()
+    timeout_group.add_argument('--timeout', '-t', help='timeout (in seconds) for kill to complete',
+                        type=check_positive, default=30)
+    timeout_group.add_argument('--no-wait', help="does not wait for all routers to confirm deletion", dest='no-wait',
+                        action='store_true')
```
shamsimam, line 131: Can this flag be configured to set value of timeout to `0` when configured directly instead of us handling it manually?

cli/waiter/subcommands/kill.py
```diff
@@ -1,36 +1,44 @@
...
-    def service_is_killed():
-        return get_service(cluster, service_id)['status'] == 'Inactive'
-
     cluster_name = cluster['name']
+    http_util.set_retries(0)
     try:
         print(f'Killing service {terminal.bold(service_id)} in {terminal.bold(cluster_name)}...')
-        resp = http_util.delete(cluster, f'/apps/{service_id}')
+        params = {'timeout': timeout_seconds * 1000}
+        read_timeout = 5 if timeout_seconds == 0 else timeout_seconds
```
shamsimam, line 20: Why default to 5 if user chose 0 seconds? Avoid passing it (i.e. read_timeout=None) and rely on default global timeout values in that situation?

cli/waiter/subcommands/kill.py
```diff
@@ -1,36 +1,44 @@
...
-                print_error('Timeout waiting for service to die.')
-                return False
+                print(f'Successfully killed {service_id} in {cluster_name}. '
+                      f'Server-side timeout waiting for routers to update.')
+                return True
         else:
             print_error(response_message(resp.json()))
             return False
+    except requests.exceptions.ReadTimeout:
+        message = f'Encountered request read timeout while killing {service_id} in {cluster_name}.'
```
shamsimam, line 39: Maybe rephrase to: 
```suggestion
        message = f'Request timed out while killing {service_id} in {cluster_name}.'
```

cli/waiter/subcommands/kill.py
kevo1ution: I saw this default being used in ping.py and wasn't sure
https://github.com/twosigma/waiter/blob/02470066f5037bec706b2258adaa046f30e9c6a2/cli/waiter/subcommands/ping.py#L16

cli/waiter/subcommands/kill.py
```diff
@@ -1,36 +1,40 @@
...
+            routers_agree = resp.json().get('routers-agree')
+            if routers_agree:
                 print(f'Successfully killed {service_id} in {cluster_name}.')
                 return True
             else:
-                print_error('Timeout waiting for service to die.')
-                return False
+                print(f'Successfully killed {service_id} in {cluster_name}. '
+                      f'Server-side timeout waiting for routers to update.')
+                return True
```
shamsimam, line 30: This should return false since the operation timed-out.

shamsimam (approved):

PR #1210, opened 2020-09-22
fixes computation of service default parameters while registering tokens

shamsimam:

- fixes computation of service default parameters while registering tokens

Users should be able to create tokens where defaults for required parameters are provided by the profile.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -391,7 +391,7 @@
   "Validates that the user is the creator of the token if it already exists.
    Then, updates the configuration for the token in the database using the newest password."
   [clock synchronize-fn kv-store cluster-calculator token-root history-length limit-per-owner waiter-hostnames
-   entitlement-manager make-peer-requests-fn validate-service-description-fn attach-token-defaults-fn
+   entitlement-manager make-peer-requests-fn validate-service-description-fn attach-service-defaults-fn
    {:keys [headers] :as request}]
   (let [request-params (-> request ru/query-params-request :query-params)
         admin-mode? (= "admin" (get request-params "update-mode"))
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/token.clj
```diff
@@ -434,7 +434,8 @@
       (when (not-empty unknown-keys)
         (throw (ex-info (str "Unsupported key(s) in token: " (str (vec unknown-keys)))
                         {:status http-400-bad-request :token token :log-level :warn}))))
-    (let [service-parameter-with-token-defaults (attach-token-defaults-fn new-service-parameter-template)]
+    (let [service-parameter-with-token-defaults (attach-service-defaults-fn new-service-parameter-template)
+          missing-parameters (->> sd/service-required-keys (remove #(contains? new-service-parameter-template %1)) seq)]
```
DaoWen, line 438: Why is `missing-parameters` defined in terms of `new-service-parameter-template` and not `service-parameter-with-token-defaults`?

waiter/src/waiter/token.clj
```diff
@@ -434,7 +434,8 @@
       (when (not-empty unknown-keys)
         (throw (ex-info (str "Unsupported key(s) in token: " (str (vec unknown-keys)))
                         {:status http-400-bad-request :token token :log-level :warn}))))
-    (let [service-parameter-with-token-defaults (attach-token-defaults-fn new-service-parameter-template)]
+    (let [service-parameter-with-token-defaults (attach-service-defaults-fn new-service-parameter-template)
```
DaoWen, line 437: Did we audit all other usages of `attach-token-defaults` to make sure this mistake isn't repeated elsewere?

DaoWen (changes requested):

waiter/src/waiter/token.clj
DaoWen: Also, this should be renamed <i>service-parameter-with-<b>service</b>-defaults</i>.

shamsimam: The [previous PR which wanted to fix the bug](https://github.com/twosigma/waiter/pull/1197/files) introduced `attach-token-defaults` into the file. It has now been replaced with uses of `attach-service-defaults` in this file.

Also, renamed variable to `service-parameter-with-service-defaults`.

waiter/src/waiter/token.clj
shamsimam: Oops, copy-paste error. Fixed and added assertions in tests to validate the reported missing parameters.

DaoWen (approved):

waiter/src/waiter/token.clj
shamsimam: > Did we audit all other usages of attach-token-defaults to make sure this mistake isn't repeated elsewhere?

I audited all uses in this file where we introduced the function as part of a [recent PR](https://github.com/twosigma/waiter/pull/1197/files).

I also checked the other uses. In all but one usage, the function is passed around as a pair to `attach-service-defaults`. The only single pass of `attach-token-defaults` is to compute token GC time which relies on token only parameters: `fallback-period-secs` and `stale-timeout-mins` in `waiter.service-description/source-tokens->gc-time-secs`.

DaoWen (approved):

PR #1211, opened 2020-09-22
adds support for replicaset version label

shamsimam:

- adds support for ReplicaSet revision timestamps
- marks instances from stale ReplicaSet revisions as expired

Adds an explicit version label on the ReplicaSet to make it easy to identify the versions of a ReplicaSet that a pod maps to.
It makes it possible to identify stale pods if the ReplicaSet version ever changes.

Marking instance as stakes cause the instance expiry machinery to kick in and replace expired instances (pods).

### Example service and instance field value:

```
"k8s/revision-timestamp": "2020-09-23T20:35:40.581Z"
```

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -224,32 +224,41 @@
    - it has restarted too many times (reached the restart-expiry-threshold threshold)
    - the primary container (waiter-apps) has not transitioned to running state in container-running-grace-secs seconds
    - the pod has the waiter/pod-expired=true annotation."
-  [{:keys [container-running-grace-secs restart-expiry-threshold]}
-   instance-id restart-count {:keys [waiter/pod-expired]} primary-container-status pod-started-at]
-  (cond
-    (>= restart-count restart-expiry-threshold)
-    (do
-      (log/info "instance expired as it reached the restart threshold"
-                {:instance-id instance-id
-                 :restart-count restart-count})
-      true)
-    (and pod-started-at
-         (pos? container-running-grace-secs)
-         (empty? (:lastState primary-container-status))
-         (not (contains? (:state primary-container-status) :running))
-         (<= container-running-grace-secs (t/in-seconds (t/interval pod-started-at (t/now)))))
-    (do
-      (log/info "instance expired as it took too long to transition to running state"
-                {:instance-id instance-id
-                 :primary-container-status primary-container-status
-                 :started-at pod-started-at})
-      true)
-    (= "true" pod-expired)
-    (do
-      (log/info "instance expired based on pod annotation"
-                {:instance-id instance-id
-                 :waiter/pod-expired pod-expired})
-      true)))
+  [{:keys [container-running-grace-secs restart-expiry-threshold watch-state]}
+   service-id instance-id restart-count {:keys [waiter/pod-expired waiter/revision-timestamp]}
+   primary-container-status pod-started-at]
+  (let [rs-revision-timestamp (get-in @watch-state [:service-id->service service-id :k8s/revision-timestamp])]
+    (cond
+      (>= restart-count restart-expiry-threshold)
+      (do
+        (log/info "instance expired as it reached the restart threshold"
+                  {:instance-id instance-id
+                   :restart-count restart-count})
+        true)
+      (and pod-started-at
+           (pos? container-running-grace-secs)
+           (empty? (:lastState primary-container-status))
+           (not (contains? (:state primary-container-status) :running))
+           (<= container-running-grace-secs (t/in-seconds (t/interval pod-started-at (t/now)))))
+      (do
+        (log/info "instance expired as it took too long to transition to running state"
+                  {:instance-id instance-id
+                   :primary-container-status primary-container-status
+                   :started-at pod-started-at})
+        true)
+      (= "true" pod-expired)
+      (do
+        (log/info "instance expired based on pod annotation"
+                  {:instance-id instance-id
+                   :waiter/pod-expired pod-expired})
+        true)
+
+      (pos? (compare rs-revision-timestamp revision-timestamp))
+      (do
+        (log/info "instance expired based on pod and replicaset revision timestamps"
+                  {:instance-id instance-id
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -115,19 +120,21 @@
       [[spec
         [:metadata name namespace uid [:annotations waiter/service-id]]
         [:status {replicas 0} {availableReplicas 0} {readyReplicas 0} {unavailableReplicas 0}]] replicaset-json
+       replicaset-version (get-in replicaset-json [:metadata :labels :waiter/replicaset-version] nil)
```
DaoWen, line 123: Since we expect this to be a timestamp, we should probably make that explicit in the name... E.g., `:waiter/revision-timestamp`, or `:waiter/service-patch-timestamp`.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -59,6 +60,10 @@
   "Kubernetes reports dates in ISO8061 format, sans the milliseconds component."
   (DateTimeFormat/forPattern "yyyy-MM-dd'T'HH:mm:ss'Z'"))
 
+(def rs-version-time-format
+  "The time format used while creating ReplicaSet version labels, only alphanumeric characters."
```
DaoWen, line 64: Let's make the new field an annotation rather than a label so that we can use our default iso8601 format. I don't think having it as a label will be very useful anyway since the values are so granular and unpredictable.

DaoWen (changes requested):

DaoWen (commented):
Should we also include the new stale-pod expiration logic in this patch?

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Changed the name to `revision-timestamp`

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Switched to using annotations.

shamsimam:
> Should we also include the new stale-pod expiration logic in this patch?

Initially I was planning on making a separate follow-up PR, but added it in this PR since you're okay with it :)

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -222,39 +224,49 @@
...
+                   :started-at pod-started-at})
+        true)
+      (= "true" pod-expired)
+      (do
+        (log/info "instance expired based on pod annotation"
+                  {:instance-id instance-id
+                   :waiter/pod-expired pod-expired})
+        true)
+
+      (pos? (compare rs-revision-timestamp revision-timestamp))
```
DaoWen, line 256: I thought through all of the different combinations of `nil` that are possible in the `compare` here, and I think it works correctly in all cases! 👍

waiter/test/waiter/scheduler/kubernetes_test.clj
```diff
@@ -620,7 +626,8 @@
                                :labels {:app "test-app-9999"
                                         :waiter/cluster "waiter"
                                         :waiter/service-hash "test-app-9999"}
-                               :annotations {:waiter/service-id "test-app-9999"}
+                               :annotations {:waiter/revision-timestamp "v20200922T202222Z"
```
DaoWen, line 629: Should we switch all of these strings to ISO 8601 format? You should be able to do this automatically with `sed`:
```
$ sed -E 's/v([0-9]{4})([0-9]{2})([0-9]{2})T([0-9]{2})([0-9]{2})([0-9]{2})Z/\1-\2-\3T\4:\5:\6Z/g' <<< v20200922T202222Z
2020-09-22T20:22:22Z
```

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: I'm also glad to see that you covered all of those cases in the unit tests.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -931,7 +946,8 @@
                   ;; but store the full service-id as an annotation.
                   ;; https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
                   ;; https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
-                  :annotations {:waiter/service-id service-id}
+                  :annotations {:waiter/revision-timestamp revision-timestamp
```
DaoWen, line 949: You injected this new binding between the comment about storing the service-id as an annotation and the service-id binding.

DaoWen (dismissed):

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -115,19 +115,21 @@
...
+                 :instances requested
+                 :k8s/app-name name
+                 :k8s/namespace namespace
+                 :k8s/replicaset-uid uid
+                 :task-count replicas
+                 :task-stats {:healthy readyReplicas
+                              :running (- replicas staged)
+                              :staged staged
+                              :unhealthy (- replicas readyReplicas staged)}}
+          revision-timestamp (assoc :k8s/revision-timestamp revision-timestamp))))
```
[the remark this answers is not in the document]
shamsimam: Changed.

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Thanks for cross-checking. Spent some time thinking about the cases too and using as simple an expression as possible, happy that compare handles `nil` :)

waiter/src/waiter/scheduler/kubernetes.clj
shamsimam: Moved the comment closer to the `service-id` annotation.

waiter/test/waiter/scheduler/kubernetes_test.clj
shamsimam: Changed to ISO format.

shamsimam (commented):
Addressed feedback.

DaoWen (approved):

PR #1214, opened 2020-09-28
avoids instance ejection on no_error responses

shamsimam:

- avoids instance ejection on no_error responses

```
   NO_ERROR (0x0):  The associated condition is not a result of an
      error.  For example, a GOAWAY might include this code to indicate
      graceful shutdown of a connection.
```

https://tools.ietf.org/html/rfc7540#section-7

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -98,7 +98,7 @@
                              ;; no_error is used to indicate that the stream is no longer needed
                              ;; HTTP2 spec: The associated condition is not a result of an error...indicate graceful shutdown of a connection.
                              (and (instance? IOException error) (= "no_error" error-message))
-                             [:generic-error "Bad request, server closed connection" http-400-bad-request error-class]
+                             [:generic-error "Server closed connection as stream no longer needed" http-400-bad-request error-class]
                              ;; connection has already been closed by the client
                              (and (instance? EofException error) (= "reset" error-message))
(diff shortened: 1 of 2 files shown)


shamsimam:
@DaoWen fyi, we encountered this for gRPC large message responses.

waiter/src/waiter/process_request.clj
```diff
@@ -95,6 +95,10 @@
                              ;; cancel_stream_error is used to indicate that the stream is no longer needed
                              (and (instance? IOException error) (= "cancel_stream_error" error-message))
                              [:client-error "Client action means stream is no longer needed" http-400-bad-request error-class]
+                             ;; no_error is used to indicate that the stream is no longer needed
+                             ;; HTTP2 spec: The associated condition is not a result of an error...indicate graceful shutdown of a connection.
+                             (and (instance? IOException error) (= "no_error" error-message))
+                             [:generic-error "Bad request, server closed connection" http-400-bad-request error-class]
```
sradack, line 101: If this is a graceful shutdown, why are we calling this a "Bad request"

shamsimam: I classified this as a 400 generic error as we've observed this error so far for scenarios where some constraint has been violated, e.g. max size on grpc payload.

shamsimam: I changed the error message to: `Server closed connection as stream no longer needed`

sradack (approved):

PR #1215, opened 2020-09-28
/waiter-ping ensures that all routers see service before responding

kevo1ution:
- /waiter-ping command should wait for routers to agree that the service is healthy and exists

- users expect the command to return when all routers have a consistent state

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1504,20 +1504,15 @@
                                   (oidc/oidc-enabled-request-handler oidc-authenticator waiter-hostnames request))))
    :not-found-handler-fn (pc/fnk [] handler/not-found-handler)
    :ping-service-handler (pc/fnk [[:daemons router-state-maintainer]
-                                  [:state fallback-state-atom user-agent-version]
+                                  [:state fallback-state-atom router-id user-agent-version]
+                                  [:routines make-inter-router-requests-async-fn]
                                   process-request-handler-fn process-request-wrapper-fn wrap-secure-request-fn]
                            (let [{{:keys [query-state-fn]} :maintainer} router-state-maintainer
                                  user-agent (str "waiter-ping/" user-agent-version)
                                  handler (process-request-wrapper-fn
                                            (fn inner-ping-service-handler [request]
-                                             (let [service-state-fn
-                                                   (fn [service-id]
-                                                     (let [fallback-state @fallback-state-atom
-                                                           global-state (query-state-fn)]
-                                                       {:exists? (descriptor/service-exists? fallback-state service-id)
-                                                        :healthy? (descriptor/service-healthy? fallback-state service-id)
-                                                        :service-id service-id
-                                                        :status (service/retrieve-service-status-label service-id global-state)}))]
+                                             (let [retrieve-service-status-label-fn #(service/retrieve-service-status-label % (query-state-fn))
+                                                   service-state-fn (partial descriptor/extract-service-state router-id retrieve-service-status-label-fn fallback-state-atom make-inter-router-requests-async-fn)]
                                                (pr/ping-service user-agent process-request-handler-fn service-state-fn request))))]
                              (wrap-secure-request-fn
                                (fn ping-service-handler [request]
@@ -1623,10 +1618,10 @@
                                      (sd/service-id->suspended-state kv-store service-id :refresh true)
                                      (sd/service-id->overrides kv-store service-id :refresh true))))
    :service-await-handler-fn (pc/fnk [[:state fallback-state-atom]
-                                               wrap-router-auth-fn]
-                                        (wrap-router-auth-fn
-                                          (fn service-await-handler-fn [request]
-                                            (handler/service-await-handler fallback-state-atom request))))
+                                      wrap-router-auth-fn]
+                               (wrap-router-auth-fn
+                                 (fn service-await-handler-fn [request]
+                                   (handler/service-await-handler fallback-state-atom request))))
    :service-resume-handler-fn (pc/fnk [[:routines allowed-to-manage-service?-fn make-inter-router-requests-sync-fn]
                                        [:state kv-store]
                                        wrap-secure-request-fn]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1504,8 +1504,8 @@
                                   (oidc/oidc-enabled-request-handler oidc-authenticator waiter-hostnames request))))
    :not-found-handler-fn (pc/fnk [] handler/not-found-handler)
    :ping-service-handler (pc/fnk [[:daemons router-state-maintainer]
-                                  [:state fallback-state-atom router-id user-agent-version]
                                   [:routines make-inter-router-requests-async-fn]
+                                  [:state fallback-state-atom router-id user-agent-version]
                                   process-request-handler-fn process-request-wrapper-fn wrap-secure-request-fn]
                            (let [{{:keys [query-state-fn]} :maintainer} router-state-maintainer
                                  user-agent (str "waiter-ping/" user-agent-version)
--- a/waiter/src/waiter/descriptor.clj
+++ b/waiter/src/waiter/descriptor.clj
@@ -443,9 +443,8 @@
                                             <?
                                             vals
                                             reduce-results)))
-          fallback-state @fallback-state-atom
           final-fallback-state (case ping-result
-                                 :timed-out (do
+                                 :timed-out (let [fallback-state @fallback-state-atom]
                                               (log/info "skipping inter router checks because ping timed out")
                                               {:healthy? (service-healthy? fallback-state service-id)
                                                :exists? (service-exists? fallback-state service-id)})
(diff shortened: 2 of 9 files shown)


waiter/src/waiter/core.clj
```diff
@@ -1504,20 +1504,15 @@
                                   (oidc/oidc-enabled-request-handler oidc-authenticator waiter-hostnames request))))
    :not-found-handler-fn (pc/fnk [] handler/not-found-handler)
    :ping-service-handler (pc/fnk [[:daemons router-state-maintainer]
-                                  [:state fallback-state-atom user-agent-version]
+                                  [:state fallback-state-atom router-id user-agent-version]
+                                  [:routines make-inter-router-requests-async-fn]
```
shamsimam, line 1508: Nitpick: lexicographic order
```suggestion
                                  [:routines make-inter-router-requests-async-fn]
                                  [:state fallback-state-atom router-id user-agent-version]
```

waiter/test/waiter/core_test.clj
```diff
@@ -510,6 +510,7 @@
           (is (= http-500-internal-server-error status))
           (is (= expected-json-response-headers headers)))))
 
+    (println "START")
```
shamsimam, line 513: please remove println

waiter/test/waiter/descriptor_test.clj
```diff
@@ -1111,4 +1113,35 @@
...
+    (testing (str fn-name ":ping-result-timed-out")
+      (let [ping-result :timed-out
+            make-inter-router-requests-async-fn (fn [& _]
+                                                  (throw (Exception. "Should not call this function")))
+            fallback-state-atom (atom {:available-service-ids #{"s1"}
+                                       :healthy-service-ids #{"s1"}})
+            service-state (<?? (extract-service-state router-id retrieve-service-status-label-fn fallback-state-atom make-inter-router-requests-async-fn service-id ping-result))]
+        (is (= (:service-id service-state) service-id))
+        (is (true? (:exists? service-state)))
+        (is (true? (:healthy? service-state)))))))
```
shamsimam, line 1147: Can we also make it that the result doesn't report the service is healthy?

waiter/test/waiter/descriptor_test.clj
```diff
@@ -1111,4 +1113,35 @@
...
+                                       :healthy-service-ids #{"s1"}})
+            service-state (<?? (extract-service-state router-id retrieve-service-status-label-fn fallback-state-atom make-inter-router-requests-async-fn service-id ping-result))]
+        (is (= (:service-id service-state) service-id))
+        (is (true? (:exists? service-state)))
+        (is (true? (:healthy? service-state)))))
+
+    (testing (str fn-name ":ping-result-timed-out")
+      (let [ping-result :timed-out
+            make-inter-router-requests-async-fn (fn [& _]
+                                                  (throw (Exception. "Should not call this function")))
```
shamsimam, line 1141: Can we have a failing assertion instead of/in addition to throwing the exception?
That way the test will fail if some other piece of code is handling the exception.

waiter/test/waiter/handler_test.clj
```diff
@@ -846,6 +846,22 @@
         (is (get (json/read-str body) "success?"))
         (is (<= update-delay (- timeout 50) (- end-time start-time) (+ timeout 50)))))
 
+    (testing (str handler-name ":success-with-update")
+      (let [fallback-state-atom (atom {:available-service-ids #{"s1"}})
+            timeout 10000
+            update-delay 2000
+            request-query (assoc request :query-string (str "timeout=" timeout))
+            start-time (System/currentTimeMillis)
```
shamsimam, line 854: Start time should be after the go block.
This test is sensitive to scheduling delays, is there some other way we can test the running time?

waiter/src/waiter/descriptor.clj
```diff
@@ -419,3 +419,33 @@
...
+(defn extract-service-state
+  "Extract the service state which is consistent on all routers. The ping-result determines what checks are made."
+  [router-id retrieve-service-status-label-fn fallback-state-atom make-inter-router-requests-async-fn service-id ping-result]
+  (go-try
+    (let [fallback-state @fallback-state-atom
+          router-comm-timeout-ms 5000
+          sleep-duration-ms 100
+          ping-timeout? (= ping-result :timed-out)
+          ping-success? (= ping-result :received-response)
+          _ (when ping-timeout? (log/info (:ping-result ping-result)
```
shamsimam, line 432: Nitpick: move the log into the next line.

waiter/src/waiter/descriptor.clj
```diff
@@ -419,3 +419,33 @@
...
+                                          "is :timed-out; therefore skipping inter router checks and defaulting to local service exists? and healthy? checks"))
```
shamsimam, line 433:
```suggestion
          _ (when ping-timeout? 
              (log/info "skipping inter router checks as ping timed out"))
```

waiter/src/waiter/descriptor.clj
```diff
@@ -419,3 +419,33 @@
...
+          service-healthy? (if ping-timeout?
+                             (service-healthy? fallback-state service-id)
+                             (and ping-success? (every?
+                                                  true?
+                                                  (vals
+                                                    (<? (await-service-goal-fallback-state fallback-state-atom make-inter-router-requests-async-fn router-id service-id router-comm-timeout-ms sleep-duration-ms "healthy"))))))
```
shamsimam, line 439:
```suggestion
                             (and ping-success?
                                  (->> (await-service-goal-fallback-state fallback-state-atom make-inter-router-requests-async-fn router-id service-id router-comm-timeout-ms sleep-duration-ms "healthy")
                                    (<?)
                                    (vals)
                                    (every? true?))
```

waiter/src/waiter/descriptor.clj
```diff
@@ -419,3 +419,33 @@
...
+          service-exists? (or
+                            (and ping-success? service-healthy?)
```
shamsimam, line 441:
```suggestion
                            service-healthy?
```

waiter/src/waiter/descriptor.clj
```diff
@@ -419,3 +419,33 @@
...
+                            (if ping-timeout?
+                              (service-exists? fallback-state service-id)
+                              (every?
+                                true?
+                                (vals
+                                  (<? (await-service-goal-fallback-state fallback-state-atom make-inter-router-requests-async-fn router-id service-id router-comm-timeout-ms sleep-duration-ms "exist"))))))]
```
shamsimam, line 447: Can we avoid two rounds of inter-router calls and instead do with a single call?

waiter/test/waiter/handler_test.clj
kevo1ution: I couldn't find a better way for time measurement. Because the timeout is much larger than the update-delay by 8 seconds, the ending assertion `(<= update-delay (- end-time start-time) timeout)` should still hold in the scenario that the update thread starts late or sleeps over 2 seconds. 

Also, I don't think there would be a scenario where the `service-await-handler` could finish before 2 seconds (violating inequality: update-delay <= end-time - start-time) unless the thread can sleep for less than 2 seconds?

waiter/src/waiter/descriptor.clj
```diff
@@ -405,17 +412,44 @@
...
+          final-fallback-state (case ping-result
+                                 :timed-out (do
+                                              (log/info "skipping inter router checks because ping timed out")
+                                              {:healthy? (service-healthy? @fallback-state-atom service-id)
+                                               :exists? (service-exists? @fallback-state-atom service-id)})
+                                 :received-response (<? (get-final-fallback-state "healthy"))
+                                 :else (<? (get-final-fallback-state "exist")))]
+      (merge final-fallback-state
+             {:service-id service-id
+              :status (retrieve-service-status-label-fn service-id)}))))
```
shamsimam, line 455: Nitpick:
```suggestion
      (assoc final-fallback-state
        :service-id service-id
        :status (retrieve-service-status-label-fn service-id}))))
```

waiter/src/waiter/descriptor.clj
```diff
@@ -405,17 +412,44 @@
...
+                                     (go-try
+                                       (->> (await-service-goal-fallback-state fallback-state-atom make-inter-router-requests-async-fn router-id service-id router-comm-timeout-ms sleep-duration-ms event)
+                                            <?
+                                            vals
+                                            reduce-results)))
+          final-fallback-state (case ping-result
+                                 :timed-out (do
+                                              (log/info "skipping inter router checks because ping timed out")
+                                              {:healthy? (service-healthy? @fallback-state-atom service-id)
+                                               :exists? (service-exists? @fallback-state-atom service-id)})
```
shamsimam, line 450: We should use the same value for the fallback state to avoid any race issues.
Might be painful while debugging if the values do not agree.
```suggestion
                                              (let [fallback-state @fallback-state-atom]
                                                {:healthy? (service-healthy? fallback-state service-id)
                                                 :exists? (service-exists? fallback-state service-id)}))
```

PR #1218, opened 2020-09-30
Refactor await endpoint

kevo1ution:
- Refactor await deletion endpoint to just await different fallback state scenarios
- move functions to descriptor.clj

- this function is planning to be reused by waiter-ping endpoint

### DIFF
--- a/waiter/src/waiter/descriptor.clj
+++ b/waiter/src/waiter/descriptor.clj
@@ -396,7 +396,7 @@
               (recur (- time-left-ms sleep-duration)))))))))
 
 (defn await-service-goal-fallback-state
-  "Polls local router and other routers and returns true if goal state is reached by all routers"
+  "Polls local router and other routers and returns map router-id->success?"
   [fallback-state-atom make-inter-router-requests-fn router-id service-id timeout sleep-duration goal]
   (go-try
     (let [router-id->response-chan (assoc
@@ -407,17 +407,16 @@
                                                  {:body (async/go
                                                           (json/write-str
                                                             {:success? (async/<! (await-service-goal-fallback-state-locally
-                                                                                  fallback-state-atom service-id timeout sleep-duration goal))}))}))
-          router-id->success? (loop [result {}
-                                    [[router-id response-chan] & remaining] (seq router-id->response-chan)]
-                               (if (and router-id response-chan)
-                                 (recur
-                                   (assoc result router-id (some-> response-chan
-                                                                   async/<!
-                                                                   :body
-                                                                   async/<!
-                                                                   utils/try-parse-json
-                                                                   (get "success?")))
-                                   remaining)
-                                 result))]
-      (every? true? (vals router-id->success?)))))
+                                                                                  fallback-state-atom service-id timeout sleep-duration goal))}))}))]
+      (loop [result {}
+             [[router-id response-chan] & remaining] (seq router-id->response-chan)]
+        (if (and router-id response-chan)
+          (recur
+            (assoc result router-id (some-> response-chan
+                                            async/<!
+                                            :body
+                                            async/<!
+                                            utils/try-parse-json
+                                            (get "success?")))
+            remaining)
+          result)))))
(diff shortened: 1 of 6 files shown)


waiter/src/waiter/descriptor.clj
```diff
@@ -374,3 +376,47 @@
...
+    (let [router-id->response-chan (assoc
+                                     (make-inter-router-requests-fn (str "apps/" service-id "/await/" goal)
+                                                                    :method :get
+                                                                    :config {:query-string (str "timeout=" timeout)})
+                                     router-id (async/go
+                                                 {:body (async/go
+                                                          (json/write-str
+                                                            {:success? (async/<! (await-service-goal-fallback-state-locally
+                                                                                  fallback-state-atom service-id timeout sleep-duration goal))}))}))]
+      (loop [result {}
```
shamsimam, line 411: Nitpick: please rename to `router-id->success?`

waiter/src/waiter/descriptor.clj
```diff
@@ -374,3 +376,47 @@
...
+          (if (or goal-reached? (not (pos? time-left-ms)))
+            goal-reached?
+            (do
+              (async/<! (async/timeout (min sleep-duration time-left-ms)))
+              (recur (- time-left-ms sleep-duration)))))))))
+
+(defn await-service-goal-fallback-state
+  "Polls local router and other routers and returns map router-id->success?"
+  [fallback-state-atom make-inter-router-requests-fn router-id service-id timeout sleep-duration goal]
+  (go-try
```
shamsimam, line 401: Where can this function throw an exception?

waiter/src/waiter/handler.clj
```diff
@@ -402,40 +390,24 @@
...
-                                             remaining)
-                                           result))
+                    should-poll? (and (= http-200-ok response-status)
+                                      (pos? timeout))
+                    routers-agree (when should-poll?
+                                    (every?
+                                      true?
+                                      (vals
+                                        (<?
+                                          (descriptor/await-service-goal-fallback-state fallback-state-atom make-inter-router-requests-fn router-id service-id timeout sleep-duration-ms goal)))))
```
shamsimam, line 407: If ` await-service-goal-fallback-state` can't rthow an exception, then we should avoid `fa/<?`

waiter/src/waiter/handler.clj
```diff
@@ -402,40 +390,24 @@
...
                     response-body-map (cond-> {:service-id service-id,
                                                :success (= http-200-ok response-status)}
-                                              router-id->response-chan (assoc :routers-agree
-                                                                              (every?
-                                                                                not
-                                                                                (vals router-id->exists?)))
+                                              should-poll? (assoc :routers-agree routers-agree)
```
shamsimam, line 410:
```suggestion
                                              (some? routers-agree) (assoc :routers-agree routers-agree)
```

waiter/src/waiter/handler.clj
```diff
@@ -697,33 +669,34 @@
     (catch Exception ex
       (utils/exception->response ex request))))
 
-(defn service-await-deletion-handler
+(defn service-await-handler
   "Polls fallback-state-atom until timeout is reached or service does not exist"
-  [fallback-state-atom {{:keys [service-id]} :route-params
+  [fallback-state-atom {{:keys [service-id goal-state]} :route-params
```
shamsimam, line 674: Nitpick: lexicographic order

waiter/src/waiter/handler.clj
```diff
@@ -697,33 +669,34 @@
...
                         {:keys [src-router-id]} :basic-authentication
                         :keys [request-method]
                         :as request}]
   (async/go
     (try
-      (log/info service-id "refresh-delete triggered by router" src-router-id)
+      (log/info service-id "service-await-handler triggered by router" src-router-id)
```
shamsimam, line 680: Nitpick: we can also log the goal here

waiter/src/waiter/descriptor.clj
kevo1ution: The function doesn't explicitly throw an exception, but this test case https://github.com/twosigma/waiter/blob/2b21e3c99b7196974e7a6aa184bf5c29c371363d/waiter/test/waiter/core_test.clj#L554
depends on the JSON error being thrown, but the error gets lost in the async/go without a try catch

waiter/src/waiter/handler.clj
kevo1ution: see above

shamsimam (approved):

PR #1220, opened 2020-10-03
handles multiple www-authenticate headers in the response

shamsimam:

- allows passing authorization header from the environment
- handles multiple www-authenticate headers in the response

The token syncer should be able to sync using the Waiter API even if Waiter changes its authentication schemes (or starts returning multiple `www-authenticate` headers on 401 responses.

In addition, being able to explicitly use the authorization header to use allows us to decide the authentication method used by the tokensyncer client.

### DIFF
--- a/token-syncer/src/token_syncer/waiter.clj
+++ b/token-syncer/src/token_syncer/waiter.clj
@@ -45,11 +45,20 @@
     (throw error))
   response)
 
+(defn contains-header?
+  "Returns true if the target value if present in the provided header value(s)."
+  [header-value target-value]
+  (if (string? header-value)
+    (= target-value header-value)
+    (some #(= target-value %) header-value)))
+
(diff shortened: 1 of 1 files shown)


token-syncer/src/token_syncer/waiter.clj
```diff
@@ -45,11 +45,20 @@
     (throw error))
   response)
 
+(defn contains-header?
+  "Returns true if the target value if present in the provided header value(s)."
```
dposada, line 49:
```suggestion
  "Returns true if the target value is present in the provided header value(s)."
```

dposada (approved):

token-syncer/src/token_syncer/waiter.clj
shamsimam: Included suggestion.

PR #1238, opened 2020-10-30
Remove Unecessary Service Id

kevo1ution:

- maintenance middleware will never have service id available (no descriptor is provided until later middlewares)
- screenshot of html and txt output to confirm no changes to users
Screenshot from 2020-10-30 09-26-53
Screenshot from 2020-10-30 09-26-47

- smaller maintenance txt and html file
- users should not notice these changes (no changes to html and txt output for users)

### DIFF
--- a/waiter/resources/web/maintenance.txt
+++ b/waiter/resources/web/maintenance.txt
@@ -12,8 +12,7 @@ Request Info
   Query String: <%= (or query-string "") %>
         Method: <%= request-method %>
            CID: <%= cid %>
-          Time: <%= timestamp %><% (when principal %>
-     Principal: <%= principal %><% ) %>
+          Time: <%= timestamp %>
 
 <% (when (or name token token-owner) %>
 Additional Info
--- a/waiter/resources/web/maintenance.txt
+++ b/waiter/resources/web/maintenance.txt
@@ -13,8 +13,7 @@ Request Info
         Method: <%= request-method %>
            CID: <%= cid %>
           Time: <%= timestamp %><% (when principal %>
-     Principal: <%= principal %><% ) %><% (when service-id %>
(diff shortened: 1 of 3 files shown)


shamsimam (approved):

waiter/resources/web/maintenance.txt
```diff
@@ -13,8 +13,7 @@ Request Info
         Method: <%= request-method %>
            CID: <%= cid %>
           Time: <%= timestamp %><% (when principal %>
-     Principal: <%= principal %><% ) %><% (when service-id %>
-    Service Id: <%= service-id %><% ) %>
+     Principal: <%= principal %><% ) %>
```
shamsimam, line 16: I suspect principal will also be missing?

kevo1ution: The principal should also be missing
This populates the `authorization/principal` key:
https://github.com/twosigma/waiter/blob/adbabdb86508b973481411a1183f0026a0c2a743/waiter/src/waiter/auth/authentication.clj#L39
https://github.com/twosigma/waiter/blob/adbabdb86508b973481411a1183f0026a0c2a743/waiter/src/waiter/core.clj#L961
https://github.com/twosigma/waiter/blob/adbabdb86508b973481411a1183f0026a0c2a743/waiter/src/waiter/core.clj#L1915
https://github.com/twosigma/waiter/blob/adbabdb86508b973481411a1183f0026a0c2a743/waiter/src/waiter/core.clj#L1550

The `wrap-auth-secure-fn` middleware eventually calls the `auth/wrap-auth-handler` function which populates the principal, and because the middleware is called after the maintenance middleware, `authorization/principal` is not populated when the request is going through the maintenance middleware.

I'll remove principal field from the html and txt

shamsimam (approved):

PR #1242, opened 2020-10-31
allow namespaced services to specify larger min-instances

shamsimam:

- allow namespaced services to specify larger min-instances

Allow service owners to determine their own min/max limits (within reason) when running in their own namespace.


DaoWen (changes requested):
The changes look good, but I would like integration tests too. We should test that tokens _can_ be created with large min-instances in the non-default namespace, but _can't_ in the default namespace. We probably just want to test creation, since spinning up 5+ instances might take a while and result in flaky tests in some environments (e.g., local minikube).

DaoWen (approved):

PR #1247, opened 2020-11-06
GitHub Actions CI Checks

kevo1ution:

- GitHub Actions for various CI Checks
- GitHub secret created for this repo at GDRIVE_LOG_POST_URL

- Travis CI has been slower recently. GitHub Actions provides more workers.


.github/workflows/cli-integration.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          cd cli
+          pip install -e .
+          pip install -r integration/requirements.txt
+      - name: Run cli-integration tests
+        run: |
+          cd cli
+          ./bin/ci/run-integration-tests.sh
+        env:
+          GDRIVE_LOG_POST_URL: ${{ secrets.GDRIVE_LOG_POST_URL }}
```
DaoWen, line 50: Using a secret for this value means it won't work for PRs opened from our forks.

It looks like there are ways to get GitHub to trigger builds with access to secrets even if they came from a fork — but it's not automatic. The example I found required a slash command:

https://github.com/imjohnbo/ok-to-test

We could probably have something where we have a list of core devs that should have builds with secrets.

However, I'm not sure it's worth the effort. I set up this URL specifically because it's low risk to expose. I think it's fine to configure it like we did in Travis.

kevo1ution: Are you suggesting we explicitly put the url in the configuration yaml files?

DaoWen: I think on Travis we were able to configure it as an environment variable in their UI — but GitHub doesn't seem to have anything like that, so I guess sticking it in the YAML is the best option.

shamsimam (approved):

DaoWen (approved):

shamsimam (approved):

PR #1250, opened 2020-11-24
Escape HTML for Maintenance Mode Message

kevo1ution:

- Escape HTML for Maintenance Mode Message

- prevent XSS vulnerability

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -237,11 +237,12 @@
 (defn escape-html
   "Change special characters into HTML character entities to prevent XSS."
   [text]
-  (-> text
-      (str/replace #"&" "&amp;")
-      (str/replace #"<" "&lt;")
-      (str/replace #">" "&gt;")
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -332,6 +342,7 @@
   "Converts the error-context to the response body html string."
   [error-context render-fn]
   (-> error-context
+    (update :message escape-html)
```
shamsimam, line 345: This needs to exclude changing the `&` in links.

kevo1ution: We should still HTML encode the `&` in the links, it still works even if something like `<a href"http://google.com/&amp;>"`. (source [https://stackoverflow.com/questions/3705591/do-i-encode-ampersands-in-a-href](https://stackoverflow.com/questions/3705591/do-i-encode-ampersands-in-a-href)). I'll add a unit test for this case

shamsimam (approved):

PR #1252, opened 2020-12-14
add preference for application/json in health request

kevo1ution:

- prefer application/json in health request checks

- Deployment errors are given in the ping response

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -990,7 +990,7 @@
 
  :health-check-config {
                        ;; HTTP accept header that is sent to a service when making health check request through /waiter-ping
-                       :health-check-headers-accept "application/json;q=1.0, */*;q=0.9"
+                       :health-check-accept-headers "application/json;q=1.0, */*;q=0.9"
 
                        ;; The HTTP connect timeout and idle timeout (milliseconds) for instance health checks:
                        :health-check-timeout-ms 200
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -990,7 +990,7 @@
 
  :health-check-config {
                        ;; HTTP accept header that is sent to a service when making health check request through /waiter-ping
-                       :health-check-accept-headers "application/json;q=1.0, */*;q=0.9"
+                       :health-check-accept-header "application/json;q=1.0, */*;q=0.9"
 
                        ;; The HTTP connect timeout and idle timeout (milliseconds) for instance health checks:
                        :health-check-timeout-ms 200
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -989,6 +989,9 @@
                            :min-hosts 2}
 
  :health-check-config {
+                       ;; HTTP accept header that is sent to a service when making health check request through /waiter-ping
+                       :health-check-headers-accept "application/json;q=1.0, */*;q=0.9"
+
                        ;; The HTTP connect timeout and idle timeout (milliseconds) for instance health checks:
                        :health-check-timeout-ms 200
 
--- a/waiter/integration/waiter/deployment_errors_test.clj
+++ b/waiter/integration/waiter/deployment_errors_test.clj
@@ -15,8 +15,8 @@
 ;;
 (ns waiter.deployment-errors-test
   (:require [clojure.pprint :as pprint]
-            [clojure.string :as str]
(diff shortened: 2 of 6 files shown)


shamsimam (commented):
Can you add an integration test that confirms the `/waiter-ping` is sending back the json response from the backend for a deployment error?

waiter/integration/waiter/health_check_test.clj
```diff
@@ -252,3 +252,25 @@
                 (is (= (str "Hello " (retrieve-username)) (-> backend-response :body str))))))
           (finally
             (delete-token-and-assert waiter-url token)))))))
+
+(deftest ^:parallel ^:integration-fast test-ping-deployment-error
```
shamsimam, line 256: Can we introduce assertions in the tests here: https://github.com/twosigma/waiter/blob/master/waiter/integration/waiter/deployment_errors_test.clj

That way we do not have to worry about the overhread of launching a new service and detecting it is failing.

waiter/integration/waiter/deployment_errors_test.clj
```diff
@@ -49,18 +49,19 @@
...
   [response deployment-error]
   `(let [response# ~response
-         body# (:body response#)
-         service-id# (response->service-id response#)]
-     (assert-response-status response# http-503-service-unavailable)
-     (is (str/includes? body# (deployment-error->str ~'waiter-url ~deployment-error))
-         (formatted-service-state ~'waiter-url service-id#))
-     (testing "status is reported as failing"
-       (is
-         (wait-for
```
shamsimam, line 59: Are we relying on recent `/waiter-ping` behavior to make sure the `Failing` is returned consistently?

waiter/src/waiter/settings.clj
```diff
@@ -280,7 +281,8 @@
                              :min-hosts 2}
    :entitlement-config {:kind :simple
                         :simple {:factory-fn 'waiter.authorization/->SimpleEntitlementManager}}
-   :health-check-config {:health-check-timeout-ms 200
+   :health-check-config {:health-check-headers-accept "application/json;q=1.0, */*;q=0.9"
```
shamsimam, line 284: Nitpick: I prefer `health-check-accept-header`

waiter/integration/waiter/deployment_errors_test.clj
kevo1ution: @shamsimam yes, should I not be relying on `/waiter-ping`?

kevo1ution: It seems we are relying on `/waiter-ping` endpoint for the json response in these test cases as well

waiter/config-full.edn
```diff
@@ -989,6 +989,9 @@
                            :min-hosts 2}
 
  :health-check-config {
+                       ;; HTTP accept header that is sent to a service when making health check request through /waiter-ping
+                       :health-check-accept-headers "application/json;q=1.0, */*;q=0.9"
```
shamsimam, line 993: Nitpick: `health-check-accept-header` (not plural)

waiter/integration/waiter/deployment_errors_test.clj
shamsimam: Previously there used to be a delay in the failing state being reflected. I was checking whether we feel whether the recent /waiter-ping changes mean that we will consistently see the `Failing` state.

shamsimam: This assertion is also testing output from some other endpoint `/apps/<service-id>`
I would check the git history to see why/when this change was introduced.

kevo1ution: Looking at the git history it seems that these were also testing `/apps/<service-id>` endpoint for adding support of the service "status" https://github.com/twosigma/waiter/pull/519 . I'll include this assertion :+1:

kevo1ution: also /waiter-ping change will wait for "healthy" state with a default timeout of 5 seconds

shamsimam: okay, that works.

shamsimam (approved):

shamsimam (approved):

PR #1254, opened 2020-12-21
add maintenance custom filter

kevo1ution:

- allows "maintenance=true" and "maintenance=false" to be query parameters on "/tokens" endpoint

- makes it easier to see all tokens in maintenance mode

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -616,10 +616,15 @@
                                                (let [search-parameter-values (cond
                                                                                (string? raw-param) #{raw-param}
                                                                                :else (set raw-param))]
-                                                 (fn [token-parameters]
-                                                   (and (contains? token-parameters parameter-name)
-                                                        (contains? search-parameter-values
-                                                                   (str (get token-parameters parameter-name)))))))]
+                                                 (case parameter-name
+                                                   "maintenance"
+                                                   (fn [token-parameters]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/token.clj
```diff
@@ -601,6 +601,8 @@
       :get (let [{:strs [can-manage-as-user] :as request-params} (-> req ru/query-params-request :query-params)
                  include-deleted (utils/param-contains? request-params "include" "deleted")
                  show-metadata (utils/param-contains? request-params "include" "metadata")
+                 filter-maintenance (when (contains? request-params "maintenance")
+                                       (utils/request-flag request-params "maintenance"))
```
shamsimam, line 605: Nitpick: The when is not necessary, it defaults to `false` when the parameter is not provided.

shamsimam (dismissed):

waiter/src/waiter/token.clj
kevo1ution: I make a distinction between nil and false values because you can filter for "maintenance=true" or "maintenance=false" or not filter at all. I will split this into two separate variables

shamsimam (approved):

waiter/src/waiter/token.clj
shamsimam: Oops didn't realize the use of `nil` to indicate parameter being provided.

PR #1256, opened 2020-12-21
add logging for changes in maintenance mode of a token

kevo1ution:

- Logs whenever maintenance mode starts, stops, or updates.
- Does not log if there were not changes to maintenance mode.

Log screen shot examples:
Screenshot from 2020-12-21 16-13-15
Screenshot from 2020-12-21 16-14-39
Screenshot from 2020-12-21 16-16-52

- better visibility on maintenance mode and metrics of its usage

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -554,12 +554,12 @@
                               nil)
                             limit-per-owner)]
           (if
-            (contains? overriding-token-data :maintenance)
-            (if (contains? overridden-token-data :maintenance)
-              (log/info "updating maintenance mode for token" {})
-              (log/info "starting maintenance mode for token" {}))
-            (when (contains? overriden-token-data :maintenance)
-              (log/info "stopping maintenance mode for token" {})))
+            (contains? overriding-token-data "maintenance")
+            (if (contains? overridden-token-data "maintenance")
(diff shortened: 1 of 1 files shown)


shamsimam (commented):
The log logic can be much simpler, it is just letting us know whether mainenace is enabled or not, doesn;t have to necessarily detect that maintenance mode is changing. But upto you.

kevo1ution:
> The log logic can be much simpler, it is just letting us know whether mainenace is enabled or not, doesn;t have to necessarily detect that maintenance mode is changing. But upto you.

I wanted to be able to see the usage of each operation. We can use /tokens already for seeing all the tokens currently in maintenance mode.

waiter/src/waiter/token.clj
```diff
@@ -551,6 +553,13 @@
                               (log/info "will not enforce count limit on owner tokens in admin mode" {:owner owner})
                               nil)
                             limit-per-owner)]
+          (if
+            (contains? overriding-token-data "maintenance")
```
shamsimam, line 557:
```suggestion
          (if (contains? overriding-token-data "maintenance")
```

shamsimam (dismissed):

shamsimam (approved):

PR #1264, opened 2021-01-06
adds support for killing frequently restarting pods

shamsimam:

- adds support for killing frequently restarting pods

A frequently restarting pod may be a signal of the host going bad. Killing the pod allows K8s to schedule the pod on a different host allowing the pod to start successfully. We process killing of the pods asynchronously in a separate dedicated thread pool. We do not adjust the replica count during the pod delete to allow a replacement pod to be scheduled. 

Ideally, we would rely on the expiration logic to replace a pod that is starting too many times. However, there is a the bug where the scaler gets stuck and does not scale down to replace an expired unhealthy instance. This change helps us work around the issue.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -39,6 +39,7 @@
             [waiter.util.http-utils :as hu]
             [waiter.util.utils :as utils])
   (:import (java.io InputStreamReader)
+           (java.util.concurrent Executors)
            (org.joda.time.format DateTimeFormat)))
 
 (defn authorization-from-environment
@@ -264,12 +265,35 @@
                      :revision-timestamp {:pod revision-timestamp :rs rs-revision-timestamp}})
           true))))
 
+;; forward declaration of the hard-delete-service-instance function
+(declare hard-delete-service-instance)
+
+(let [kill-frequently-restarting-pods-thread-pool (Executors/newFixedThreadPool 4)]
+  (defn kill-frequently-restarting-pods
+    "Processes killing of a frequently restarting pod (service instance).
+     On non-leader routers it returns immediately without doing anything.
+     On the leader it triggers killing of the instance asynchronously without adjusting ReplicaSet replica count."
+    [{:keys [leader?-fn] :as scheduler} {:keys [k8s/pod-name k8s/restart-count service-id] :as instance}]
+    (if (leader?-fn)
+      (au/execute
+        (fn kill-frequently-restarting-pods-task []
+          (try
+            (log/info "deleting frequently restarting pod"
+                      {:pod-name pod-name :restart-count restart-count :service-id service-id})
+            (hard-delete-service-instance scheduler instance)
+            (catch Exception ex
+              (log/error ex "error in deleting frequently restarting pod" pod-name))))
+        kill-frequently-restarting-pods-thread-pool)
+      (log/info "skipping deleting frequently restarting pod on non-leader"
+                {:pod-name pod-name :restart-count restart-count :service-id service-id}))))
+
 (defn pod->ServiceInstance
   "Convert a Kubernetes Pod JSON response into a Waiter Service Instance record."
-  [{:keys [api-server-url] :as scheduler} pod]
+  [{:keys [api-server-url restart-kill-threshold] :as scheduler} pod]
   (try
     (let [;; waiter-app is the first container we register
-          primary-container-restart-count (get-in pod [:status :containerStatuses 0 :restartCount] 0)
+          primary-container-restart-count (or (get-in pod [:status :containerStatuses 0 :restartCount]) 0)
+          frequently-restarting? (>= primary-container-restart-count restart-kill-threshold)
           service-id (k8s-object->service-id pod)
           instance-id (pod->instance-id pod primary-container-restart-count)
           node-name (get-in pod [:spec :nodeName])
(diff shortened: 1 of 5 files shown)


DaoWen:
@shamsimam - Could you please add some clarification to the PR description about why the Expiration logic tied to restarts is insufficient? It seems like this patch should really be framed as a fix for the Unhealthy+Expired+Scaling interaction bug.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -264,12 +265,35 @@
...
+
 (defn pod->ServiceInstance
   "Convert a Kubernetes Pod JSON response into a Waiter Service Instance record."
-  [{:keys [api-server-url] :as scheduler} pod]
+  [{:keys [api-server-url restart-kill-threshold] :as scheduler} pod]
   (try
     (let [;; waiter-app is the first container we register
-          primary-container-restart-count (get-in pod [:status :containerStatuses 0 :restartCount] 0)
+          primary-container-restart-count (or (get-in pod [:status :containerStatuses 0 :restartCount]) 0)
+          frequently-restarting? (>= primary-container-restart-count restart-kill-threshold)
```
DaoWen, line 296: I think `frequently` doesn't match the semantics here since we're not checking frequency (i.e., we don't care how long the pod was live between restarts). Since we only care how many times it restarted, `exceeded-restart-kill-threshold?` seems more accurate.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -294,37 +318,41 @@
...
-                                                     (some? state) (assoc :state state))))
-                                               container-statuses)))))
+          {:keys [waiter/revision-timestamp]} (get-in pod [:metadata :annotations])
+          instance (scheduler/make-ServiceInstance
+                     (cond-> {:extra-ports (->> pod-annotations :waiter/port-count Integer/parseInt range next (mapv #(+ port0 %)))
+                              :flags (cond-> #{}
+                                       (check-expired scheduler service-id instance-id pod-restart-count pod-annotations primary-container-status pod-started-at)
+                                       (conj :expired))
+                              :healthy? (and (true? (get primary-container-status :ready))
+                                             (not frequently-restarting?))
```
DaoWen, line 328: Why are we changing the logic for `healthy?`? I think we eject instances before killing them when we do a normal scale-down operation, so shouldn't we do the same thing here?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -264,12 +265,35 @@
                      :revision-timestamp {:pod revision-timestamp :rs rs-revision-timestamp}})
           true))))
 
+;; forward declaration of the hard-delete-service-instance function
+(declare hard-delete-service-instance)
+
+(let [kill-frequently-restarting-pods-thread-pool (Executors/newFixedThreadPool 4)]
+  (defn kill-frequently-restarting-pods
```
DaoWen, line 272: Since this function only accepts a single target instance (pod), it should be named `kill-frequently-restarting-pod` (no trailing `s`).

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -507,10 +535,40 @@
...
+   Does not adjust ReplicaSet replica count; preventing scheduling of a replacement pod must be ensured by the callee.
+   Uses locking to prevent concurrent calls for killing instances from the same service.
+   Returns nil on success, but throws on failure."
+  [{:keys [api-server-url] :as scheduler} {:keys [service-id] :as instance}]
+  (locking
+    (service-id->deletion-lock service-id)
+    (let [pod-url (instance->pod-url api-server-url instance)]
+      (try
+        ; "hard" delete the pod (i.e., actually kill, allowing the pod's default grace period expires)
+        ; (note that the pod's default grace period is different from the 300s period set above)
```
DaoWen, line 561: This comment no longer makes sense because you moved this logic out of and _above_ the original function.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -507,10 +535,40 @@
           (<= attempt max-patch-retries)
           (recur (inc attempt) (get-replica-count scheduler service-id)))))))
 
-(defn- kill-service-instance
+(let [num-deletion-locks 100
+      deletion-locks (repeatedly num-deletion-locks (fn [] (Object.)))]
+  (defn- service-id->deletion-lock
+    "Returns the deletion lock striped by a hash of the service ID."
+    [service-id]
+    (nth deletion-locks (mod (hash service-id) num-deletion-locks))))
```
DaoWen, line 543: `repeatedly` returns a lazy sequence, which means this `nth` operation has to traverse the 100-element lazy list to get the specified item each time. 😱 If we don't delete the locks altogether (see comment below asking what the lock is accomplishing), then we should convert the sequence to a vector to get fast indexing.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -507,10 +535,40 @@
...
+  [api-server-url {:keys [k8s/namespace k8s/pod-name]}]
+  (str api-server-url "/api/v1/namespaces/" namespace "/pods/" pod-name))
+
+(defn hard-delete-service-instance
+  "Force kill the Kubernetes pod corresponding to the given Waiter Service Instance.
+   Does not adjust ReplicaSet replica count; preventing scheduling of a replacement pod must be ensured by the callee.
+   Uses locking to prevent concurrent calls for killing instances from the same service.
+   Returns nil on success, but throws on failure."
+  [{:keys [api-server-url] :as scheduler} {:keys [service-id] :as instance}]
+  (locking
```
DaoWen, line 556: What problem does this lock solve? The comment says it's to avoid "concurrent calls to kill an instance" — but it's not at all clear to me why it's not OK to do this concurrently while it is OK to do it twice immediately back-to-back (serialized by the lock).

I don't feel like this lock is actually accomplishing anything. Unless we have a concrete scenario where it solves a problem, then I'd prefer to just drop the locks from this change.

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -1292,12 +1343,13 @@
...
-  [{:keys [authentication authorizer cluster-name container-running-grace-secs custom-options http-options log-bucket-sync-secs
+  [{:keys [authentication authorizer cluster-name container-running-grace-secs custom-options http-options leader?-fn log-bucket-sync-secs
            log-bucket-url max-patch-retries max-name-length pdb-api-version pdb-spec-builder pod-base-port pod-sigkill-delay-secs
-           pod-suffix-length replicaset-api-version replicaset-spec-builder restart-expiry-threshold reverse-proxy scheduler-name
-           scheduler-state-chan scheduler-syncer-interval-secs service-id->service-description-fn
+           pod-suffix-length replicaset-api-version replicaset-spec-builder restart-expiry-threshold restart-kill-threshold
+           reverse-proxy scheduler-name scheduler-state-chan scheduler-syncer-interval-secs service-id->service-description-fn
            service-id->password-fn start-scheduler-syncer-fn url watch-connect-timeout-ms watch-retries watch-socket-timeout-ms]
     {fileserver-port :port fileserver-scheme :scheme :as fileserver} :fileserver
+    :or {restart-kill-threshold 8}
```
DaoWen, line 1352: How about `(+ 2 restart-expiry-threshold)` instead of `8`? That would avoid breaking existing configurations where restart-expiry-threshold > 8.

DaoWen (changes requested):

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -264,12 +265,35 @@
...
+      (au/execute
+        (fn kill-frequently-restarting-pods-task []
+          (try
+            (log/info "deleting frequently restarting pod"
+                      {:pod-name pod-name :restart-count restart-count :service-id service-id})
+            (hard-delete-service-instance scheduler instance)
+            (catch Exception ex
+              (log/error ex "error in deleting frequently restarting pod" pod-name))))
+        kill-frequently-restarting-pods-thread-pool)
+      (log/info "skipping deleting frequently restarting pod on non-leader"
```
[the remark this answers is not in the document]
DaoWen: OK, this is fine then.

waiter/src/waiter/scheduler/kubernetes.clj
DaoWen: Since we mark the instance _unhealthy_ as soon as we see the container restart count go up, it doesn't have a chance to actually become _ready_, so I don't think there's no chance of a race across the routers. Based on that plus the fact that it sounds like getting Waiter to do ejection in this case (via the auto-scaler) sounds very complex, I guess I'm OK with this for now. Maybe just add a comment like the following above the new clause on line 328:
```
;; Note that when frequently-restarting? becomes true, the container *just* restarted and is non-ready,
;; therefore it's impossible to observe a healthy? status between the time the container restarted
;; and the time that the instance was marked permanently unhealthy due to the high restart count.
```

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -264,12 +265,35 @@
                      :revision-timestamp {:pod revision-timestamp :rs rs-revision-timestamp}})
           true))))
 
+;; forward declaration of the hard-delete-service-instance function
```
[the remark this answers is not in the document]
DaoWen: Okay, yeah, I see that now. Thanks for clarifying. I'm fine with just leaving the forward declaration as long as the comment is changed to note that it's only there to avoid reordering a bunch of functions (which also makes the git history messy for operations like git-blame).

DaoWen (approved):

PR #1268, opened 2021-01-08
CLI: choose correct cluster when updating token

kevo1ution:

- using the most up to date token description, use the "cluster" field to get the appropriate cluster configuration for sending the POST request to.
- This modifies the behavior of maintenance, create, and update subcommands
- ping was left out because it would break current test cases "test_federated_ping", which expects to ping multiple cluster tokens

- users may have their token in a secondary cluster but the current code will update the token as the first cluster configured.

### DIFF
--- a/cli/waiter/querying.py
+++ b/cli/waiter/querying.py
@@ -172,7 +172,7 @@ def _get_latest_cluster(clusters, query_result):
         provided_cluster_names.append(cluster_config_name)
         if cluster_name_goal.upper() == cluster_config_name.upper():
             return c
-    raise Exception(f'The token is configured in cluster {cluster_name_goal} which is not provided.' +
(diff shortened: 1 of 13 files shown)


cli/waiter/querying.py
```diff
@@ -154,10 +154,57 @@ def query_tokens(clusters, user):
...
+        for cluster in list(query_result['clusters'].keys()):
+            cluster_config = next(c for c in clusters if c['name'] == cluster)
+            sync_group = cluster_config.get('sync-group', False)
+            # consider clusters that don't have a configured sync-group as in their own unique group
+            if not sync_group:
+                sync_group = sync_group_count
+                sync_group_count += 1
+            sync_groups_set.add(sync_group)
+            cluster_names.add(cluster)
+        if len(sync_groups_set) > 1:
```
shamsimam, line 205: 👍

shamsimam (approved):

PR #1276, opened 2021-01-27
Token Watch

kevo1ution:

- add daemon process to maintain local map of token->index and broker messages to listeners
- add handler for `/state/tokens-watch-maintainer` to query the state of the maintainer
- example (no query parameters)
Screenshot from 2021-02-02 11-28-16
- example with include query parameters
Screenshot from 2021-02-02 11-29-21
- request log for `/tokens?watch-true`
Screenshot from 2021-02-11 08-48-33

- This token watch can be used by clients who want to maintain a local replica of the overall token index state

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1050,6 +1050,10 @@
                                           ;; or removed
                                           :channels-update-chan-buffer-size 1024
 
+                                          ;; Buffer size of the tokens-update-chan. The channel receives updates when
+                                          ;; tokens are created, updated, deleted (including index refresh endpoint)
+                                          :tokens-update-chan-buffer-size 1024
+
                                           ;; Time the tokens-watch-maintainer waits before fetching
                                           ;; the kv-store for up to date indexes and sending possible
                                           ;; missed events to listening watches
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1040,7 +1040,20 @@
                                  ;; Stale services are detected by looking at the version of the
                                  ;; token used to create them, such services will have their idle timeout
                                  ;; reduced to stale-timeout-mins + fallback-period-secs.
-                                 "stale-timeout-mins" 15}}
+                                 "stale-timeout-mins" 15}
+
+                ;; Configures tokens-watch-maintainer which handles and updates watches
+                ;; for token events
+                :tokens-watch-maintainer {
+                                          ;; Buffer size of the tokens-watch-channels-update-chan
+                                          ;; The channel receives updates whenever a watch is added
+                                          ;; or removed
+                                          :channels-update-chan-buffer-size 1024
+
+                                          ;; Time the tokens-watch-maintainer waits before fetching
+                                          ;; the kv-store for up to date indexes and sending possible
+                                          ;; missed events to listening watches
+                                          :watch-refresh-timeout-ms 10000}}
 
  ;; Waiter supports making websocket connections to backends.
  ;; The maximum binary and text message sizes in the websocket request or response can be configured
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1040,24 +1040,27 @@
                                  ;; Stale services are detected by looking at the version of the
                                  ;; token used to create them, such services will have their idle timeout
                                  ;; reduced to stale-timeout-mins + fallback-period-secs.
-                                 "stale-timeout-mins" 15}
+                                 "stale-timeout-mins" 15}}
 
-                ;; Configures tokens-watch-maintainer which handles and updates watches
+ ;; Waiter provides watch endpoints to listen on changes of a resource
+ ;; This configures various watch providers for different resources
+ :watch-config {
+                ;; Configures token-watch-maintainer which handles and updates watches
                 ;; for token events
-                :tokens-watch-maintainer {
-                                          ;; Buffer size of the tokens-watch-channels-update-chan
-                                          ;; The channel receives updates whenever a watch is added
-                                          ;; or removed
-                                          :channels-update-chan-buffer-size 1024
+                :tokens {
+                         ;; Buffer size of the tokens-watch-channels-update-chan
+                         ;; The channel receives updates whenever a watch is added
+                         ;; or removed
+                         :channels-update-chan-buffer-size 1024
 
-                                          ;; Buffer size of the tokens-update-chan. The channel receives updates when
-                                          ;; tokens are created, updated, deleted (including index refresh endpoint)
-                                          :tokens-update-chan-buffer-size 1024
+                         ;; Buffer size of the tokens-update-chan. The channel receives updates when
+                         ;; tokens are created, updated, deleted (including index refresh endpoint)
+                         :tokens-update-chan-buffer-size 1024
 
-                                          ;; Time the tokens-watch-maintainer waits before fetching
-                                          ;; the kv-store for up to date indexes and sending possible
-                                          ;; missed events to listening watches
-                                          :watch-refresh-timeout-ms 10000}}
+                         ;; Time the token-watch-maintainer waits before fetching
+                         ;; the kv-store for up to date indexes and sending possible
+                         ;; missed events to listening watches
+                         :watch-refresh-timeout-ms 10000}}
 
  ;; Waiter supports making websocket connections to backends.
  ;; The maximum binary and text message sizes in the websocket request or response can be configured
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1826,3 +1826,24 @@
                 (is (= (-> token-description (dissoc :token) (assoc :owner current-user)) parsed-description))))))
         (finally
           (delete-token-and-assert waiter-url token))))))
+
+(deftest ^:parallel ^:integration-fast test-tokens-watch-maintainer
+  (testing-using-waiter-url
+    (let [{:keys [body] :as response} (get-tokens-watch-maintainer-state waiter-url)
+          {:strs [router-id state]} (try-parse-json body)
+          router-url (router-endpoint waiter-url router-id)]
+      (testing "no query parameters provide entire state"
+        (assert-response-status response 200)
+        (is (= (set (keys state))
+               #{"token->token-index" "watches-count"})))
+      (testing "include only watches-count"
+        (let [{:keys [body]} (get-tokens-watch-maintainer-state router-url :query-params "include=watches-count")
+              {:strs [state]} (try-parse-json body)]
+          (is (= (set (keys state))
+                 #{"watches-count"}))))
+      (testing "include multiple fields"
+        (let [{:keys [body]} (get-tokens-watch-maintainer-state
+                               router-url :query-params "include=watches-count&include=token->token-index")
+              {:strs [state]} (try-parse-json body)]
+          (is (= (set (keys state))
+                 #{"token->token-index" "watches-count"})))))))
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1826,30 +1826,3 @@
                 (is (= (-> token-description (dissoc :token) (assoc :owner current-user)) parsed-description))))))
         (finally
           (delete-token-and-assert waiter-url token))))))
-
-(deftest ^:parallel ^:integration-fast test-tokens-watch-maintainer
-  (testing-using-waiter-url
-    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
-          {:keys [body] :as response} (get-tokens-watch-maintainer-state waiter-url)
-          {:strs [router-id state]} (try-parse-json body)
-          router-url (router-endpoint waiter-url router-id)]
-      (testing "no query parameters provide entire state"
-        (assert-response-status response 200)
-        (is (= (set (keys state))
-               #{"token->token-index" "watches-count"})))
-      (testing "include only watches-count"
-        (let [{:keys [body] :as response}
-              (get-tokens-watch-maintainer-state router-url :query-params "include=watches-count" :cookies cookies)
-              {:strs [state]} (try-parse-json body)]
-          (assert-response-status response 200)
-          (is (= (set (keys state))
-                 #{"watches-count"}))))
-      (testing "include multiple fields"
-        (let [{:keys [body] :as response}
-              (get-tokens-watch-maintainer-state router-url
-                                                 :query-params "include=watches-count&include=token->token-index"
-                                                 :cookies cookies)
-              {:strs [state]} (try-parse-json body)]
-          (assert-response-status response 200)
-          (is (= (set (keys state))
-                 #{"token->token-index" "watches-count"})))))))
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1829,7 +1829,8 @@
 
 (deftest ^:parallel ^:integration-fast test-tokens-watch-maintainer
   (testing-using-waiter-url
-    (let [{:keys [body] :as response} (get-tokens-watch-maintainer-state waiter-url)
+    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
+          {:keys [body] :as response} (get-tokens-watch-maintainer-state waiter-url)
           {:strs [router-id state]} (try-parse-json body)
           router-url (router-endpoint waiter-url router-id)]
       (testing "no query parameters provide entire state"
@@ -1837,13 +1838,18 @@
         (is (= (set (keys state))
                #{"token->token-index" "watches-count"})))
       (testing "include only watches-count"
-        (let [{:keys [body]} (get-tokens-watch-maintainer-state router-url :query-params "include=watches-count")
+        (let [{:keys [body] :as response}
+              (get-tokens-watch-maintainer-state router-url :query-params "include=watches-count" :cookies cookies)
               {:strs [state]} (try-parse-json body)]
+          (assert-response-status response 200)
           (is (= (set (keys state))
                  #{"watches-count"}))))
       (testing "include multiple fields"
-        (let [{:keys [body]} (get-tokens-watch-maintainer-state
-                               router-url :query-params "include=watches-count&include=token->token-index")
+        (let [{:keys [body] :as response}
+              (get-tokens-watch-maintainer-state router-url
+                                                 :query-params "include=watches-count&include=token->token-index"
+                                                 :cookies cookies)
               {:strs [state]} (try-parse-json body)]
+          (assert-response-status response 200)
           (is (= (set (keys state))
                  #{"token->token-index" "watches-count"})))))))
--- a/waiter/integration/waiter/token_watch_integration_test.clj
+++ b/waiter/integration/waiter/token_watch_integration_test.clj
@@ -1,10 +1,14 @@
 (ns waiter.token-watch-integration-test
-  (:require [clojure.core.async :as async]
+  (:require [cheshire.core :as cheshire]
+            [clojure.core.async :as async]
             [clojure.set :as set]
             [clojure.test :refer :all]
             [waiter.status-codes :refer :all]
-            [waiter.util.client-tools :refer :all])
-  (:import (clojure.lang ExceptionInfo)))
+            [waiter.util.client-tools :refer :all]
+            [waiter.util.utils :as utils]
+            [clojure.tools.logging :as log])
+  (:import (java.io SequenceInputStream InputStreamReader ByteArrayInputStream)
+           (java.util Collections)))
 
 (defn- await-goal-response-for-all-routers
   "Returns true if the goal-response-fn was satisfied with the response from request-fn for all routers before
@@ -111,59 +115,59 @@
 (defn- start-watch
   [router-url cookies & {:keys [query-params] :or {query-params {"include" ["metadata", "deleted"]
                                                                  "watch" "true"}}}]
-  (let [{:keys [body error]} (make-request router-url "/tokens"
-                                           :cookies cookies
-                                           :fold-chunked-response? false
-                                           :query-params query-params)
+  (let [{:keys [body error] :as request} (make-request router-url "/tokens"
+                                                       :async? true
+                                                       :cookies cookies
+                                                       :query-params query-params)
+        _ (assert-response-status request 200)
+        json-objects (->> body
+                          utils/chan-to-seq!!
+                          (map (fn [chunk] (-> chunk .getBytes ByteArrayInputStream.)))
+                          Collections/enumeration
+                          SequenceInputStream.
+                          InputStreamReader.
+                          cheshire/parsed-seq)
         token->index-atom (atom {})
         query-state-fn (fn [] @token->index-atom)
-        exit-chan (async/promise-chan)
+        exit-fn (fn []
+                  (async/close! body))
         go-chan
-        (async/go-loop [token->index @token->index-atom
-                        prefix ""]
-          (reset! token->index-atom token->index)
-          (let [[msg chan] (async/alts! [body exit-chan] :priority true)
-                [next-token->index next-prefix]
-                (condp = chan
-                  body
-                  (when (some? msg)
-                    (try
-                      (let [{:strs [object type]} (try-parse-json (str prefix msg))]
+        (async/go
+          (try
+            (doseq [msg json-objects
+                    :when (some? msg)]
+              (reset!
+                token->index-atom
+                (let [{:strs [object type]} msg
+                      token->index @token->index-atom]
+                  (case type
+                    "INITIAL"
+                    (reduce
+                      (fn [token->index index]
+                        (assoc token->index (get index "token") index))
+                      {}
+                      object)
+
+                    "EVENTS"
+                    (reduce
+                      (fn [token->index {:strs [object type]}]
                         (case type
-                          "INITIAL"
-                          [(reduce
-                             (fn [token->index index]
-                               (assoc token->index (get index "token") index))
-                             {}
-                             object)
-                           ""]
-
-                          "EVENTS"
-                          [(reduce
-                             (fn [token->index {:strs [object type]}]
-                               (case type
-                                 "UPDATE"
-                                 (assoc token->index (get object "token") object)
-                                 "DELETE"
-                                 (dissoc token->index (get object "token") object)
-                                 (throw (ex-info "Unknown event type received in EVENTS object" {:event object}))))
-                             token->index
-                             object)
-                           ""]
-                          (throw (ex-info "Unknown event type received from watch" {:event msg}))))
-                      (catch ExceptionInfo _
-                        [token->index (str prefix msg)])))
-
-                  exit-chan
-                  (do
-                    (async/close! body)
-                    [false ""]))]
-            (when next-token->index
-              (recur next-token->index next-prefix))))]
-    {:exit-chan exit-chan
+                          "UPDATE"
+                          (assoc token->index (get object "token") object)
+                          "DELETE"
+                          (dissoc token->index (get object "token") object)
+                          (throw (ex-info "Unknown event type received in EVENTS object" {:event object}))))
+                      token->index
+                      object)
+                    (throw (ex-info "Unknown event type received from watch" {:event msg}))))))
+            (catch Exception e
+              (exit-fn)
+              (log/error e "Error in test watch-chan" {:router-url router-url}))))]
+    {:exit-fn exit-fn
      :error-chan error
      :go-chan go-chan
-     :query-state-fn query-state-fn}))
+     :query-state-fn query-state-fn
+     :router-url router-url}))
 
 (defn- start-watches
   [router-urls cookies]
@@ -172,8 +176,8 @@
     router-urls))
 
 (defn- stop-watch
-  [{:keys [exit-chan go-chan]}]
-  (async/close! exit-chan)
+  [{:keys [exit-fn go-chan]}]
+  (exit-fn)
   (async/<!! go-chan))
 
 (defn- stop-watches
@@ -186,11 +190,30 @@
   `(let [watch# ~watch
          token# ~token
          entry# ~entry
-         query-state-fn# (:query-state-fn watch#)]
+         router-url# (:router-url watch#)
+         query-state-fn# (:query-state-fn watch#)
+         get-current-token-entry-fn# #(get (query-state-fn#) token#)]
      (is (wait-for
-           #(= (get (query-state-fn#) token#)
+           #(= (get-current-token-entry-fn#)
                entry#)
-           :interval 1 :timeout 5))))
+           :interval 1 :timeout 5)
+         (str "watch for " router-url# " token->index entry for token '" token# "' was '" (get-current-token-entry-fn#)
+              "' instead of '" entry# "'"))))
+
+(defmacro assert-watch-token-index-entry-does-not-change
+  [watch token entry]
+  `(let [watch# ~watch
+         token# ~token
+         entry# ~entry
+         router-url# (:router-url watch#)
+         query-state-fn# (:query-state-fn watch#)
+         get-current-token-entry-fn# #(get (query-state-fn#) token#)]
+     (is (not (wait-for
+                #(not= (get-current-token-entry-fn#)
+                       entry#)
+                :interval 1 :timeout 5))
+         (str "watch for " router-url# " token->index entry for token '" token# "' was '" (get-current-token-entry-fn#)
+              "' instead of '" entry# "'"))))
 
 (defmacro assert-watches-token-index-entry
   [watches token entry]
@@ -278,8 +301,7 @@
               (assert-watch-token-index-entry watch token-1 nil)
               (assert-watch-token-index-entry watch token-2 entry)
               (delete-token-and-assert waiter-url token-2 :hard-delete false)
-              (async/<!! (async/timeout 4000))
-              (assert-watch-token-index-entry watch token-2 entry)
+              (assert-watch-token-index-entry-does-not-change watch token-2 entry)
               (stop-watches [watch]))
             (finally
               (delete-token-and-assert waiter-url token-1)
@@ -301,13 +323,12 @@
                                                              "owner" (retrieve-username)
                                                              "maintenance" false})
               (post-token waiter-url (assoc (kitchen-params) :token token-1 :version "update-1"))
-              (async/<!! (async/timeout 4000))
-              (assert-watch-token-index-entry watch token-1 {"token" token-1
-                                                             "owner" (retrieve-username)
-                                                             "maintenance" false})
-              (assert-watch-token-index-entry watch token-2 {"token" token-2
-                                                             "owner" (retrieve-username)
-                                                             "maintenance" false})
+              (assert-watch-token-index-entry-does-not-change watch token-1 {"token" token-1
+                                                                             "owner" (retrieve-username)
+                                                                             "maintenance" false})
+              (assert-watch-token-index-entry-does-not-change watch token-2 {"token" token-2
+                                                                             "owner" (retrieve-username)
+                                                                             "maintenance" false})
               (stop-watch watch))
             (finally
               (delete-token-and-assert waiter-url token-1)
@@ -329,11 +350,10 @@
                                                              "maintenance" false})
               (assert-watch-token-index-entry watch token-2 nil)
               (post-token waiter-url (assoc (kitchen-params) :token token-1 :maintenance {:message "maintenance message"}))
-              (async/<!! (async/timeout 4000))
-              (assert-watch-token-index-entry watch token-1 {"token" token-1
-                                                             "owner" (retrieve-username)
-                                                             "maintenance" false})
-              (assert-watch-token-index-entry watch token-2 nil)
+              (assert-watch-token-index-entry-does-not-change watch token-1 {"token" token-1
+                                                                             "owner" (retrieve-username)
+                                                                             "maintenance" false})
+              (assert-watch-token-index-entry-does-not-change watch token-2 nil)
               (stop-watch watch))
             (finally
               (delete-token-and-assert waiter-url token-1)
@@ -356,13 +376,12 @@
                                                              "maintenance" true})
               (post-token waiter-url (assoc (kitchen-params) :token token-1 :maintenance {:message "maintenance message"}))
               (post-token waiter-url (assoc (kitchen-params) :token token-2))
-              (async/<!! (async/timeout 4000))
-              (assert-watch-token-index-entry watch token-1 {"token" token-1
-                                                             "owner" (retrieve-username)
-                                                             "maintenance" true})
-              (assert-watch-token-index-entry watch token-2 {"token" token-2
-                                                             "owner" (retrieve-username)
-                                                             "maintenance" true})
+              (assert-watch-token-index-entry-does-not-change watch token-1 {"token" token-1
+                                                                             "owner" (retrieve-username)
+                                                                             "maintenance" true})
+              (assert-watch-token-index-entry-does-not-change watch token-2 {"token" token-2
+                                                                             "owner" (retrieve-username)
+                                                                             "maintenance" true})
               (stop-watch watch))
             (finally
               (delete-token-and-assert waiter-url token-1)
--- a/waiter/integration/waiter/token_watch_integration_test.clj
+++ b/waiter/integration/waiter/token_watch_integration_test.clj
@@ -302,7 +302,7 @@
               (assert-watch-token-index-entry watch token-2 entry)
               (delete-token-and-assert waiter-url token-2 :hard-delete false)
               (assert-watch-token-index-entry-does-not-change watch token-2 entry)
-              (stop-watches [watch]))
+              (stop-watch watch))
             (finally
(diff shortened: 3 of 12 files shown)


waiter/config-full.edn
```diff
@@ -1040,7 +1040,24 @@
                                  ;; Stale services are detected by looking at the version of the
                                  ;; token used to create them, such services will have their idle timeout
                                  ;; reduced to stale-timeout-mins + fallback-period-secs.
-                                 "stale-timeout-mins" 15}}
+                                 "stale-timeout-mins" 15}
+
+                ;; Configures tokens-watch-maintainer which handles and updates watches
+                ;; for token events
+                :tokens-watch-maintainer {
```
shamsimam, line 1047: Can we nest this as `:watch-configuration :token`? Then we can add additional components inside `:watch-configuration` as we add more watches (e.g. new instances showing up).

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+
+                      query-chan
+                      (let [response-chan msg]
+                        (async/put! response-chan current-state)
+                        [current-state watch-refresh-timeout-chan]))]
+                (if (and next-state next-watch-refresh-timeout-chan)
+                  (recur next-state next-watch-refresh-timeout-chan)
+                  (log/info "Stopping tokens-watch-maintainer as next loop values are nil"))))
+            (catch Exception e
+              (log/error e "Fatal error in tokens-watch-maintainer"))))]
```
shamsimam, line 832: This is not fatal since we are not exiting the router. We need to invoke the JVM exit method.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+    {:go-chan go-chan
```
shamsimam, line 833: We should be returning `exit-chan` instead.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1826,3 +1826,30 @@
...
+                 #{"watches-count"}))))
+      (testing "include multiple fields"
+        (let [{:keys [body] :as response}
+              (get-tokens-watch-maintainer-state router-url
+                                                 :query-params "include=watches-count&include=token->token-index"
+                                                 :cookies cookies)
+              {:strs [state]} (try-parse-json body)]
+          (assert-response-status response 200)
+          (is (= (set (keys state))
+                 #{"token->token-index" "watches-count"})))))))
```
shamsimam, line 1855: We should include a few token updates/deletes and confirm they show up on this state too.
The test should be resilient to concurrently running tests and presence of other token entries.

waiter/src/waiter/core.clj
```diff
@@ -131,6 +131,7 @@
                               ["/service-description-builder" :state-service-description-builder-handler-fn]
                               ["/service-maintainer" :state-service-maintainer-handler-fn]
                               ["/statsd" :state-statsd-handler-fn]
+                              ["/tokens-watch-maintainer" :state-tokens-watch-maintainer]
```
shamsimam, line 134: Can we use `/token-watch-maintainer` (singular) instead.
Also, a link to this page should be included in the `/state` output.

waiter/src/waiter/core.clj
```diff
@@ -1354,7 +1361,16 @@
                (let [{:keys [sync-instances-interval-ms]} statsd
                      {{:keys [query-state-fn]} :maintainer} router-state-maintainer]
                  (statsd/start-service-instance-metrics-publisher
-                   service-id->service-description-fn query-state-fn sync-instances-interval-ms))))})
+                   service-id->service-description-fn query-state-fn sync-instances-interval-ms))))
+   :tokens-watch-maintainer (pc/fnk [[:settings [:token-config [:tokens-watch-maintainer watch-refresh-timeout-ms]]]
```
shamsimam, line 1365: Singular name please.

waiter/src/waiter/handler.clj
```diff
@@ -891,6 +891,21 @@
...
+(defn get-tokens-watch-maintainer-state
+  "Using the query-state-fn, pass include-flags to get the tokens-watch-maintainer-state and return
+  streaming response"
+  [router-id query-state-fn request]
+  (try
+    (let [{:strs [include]} (-> request ru/query-params-request :query-params)
+          include-flags (if (string? include) #{include} (set include))
+          state (if (nil? include)
+                  (query-state-fn)
+                  (query-state-fn include-flags))]
```
shamsimam, line 903: We should not need to support two variants of the `query-state-fn`. We can support one variant: `(query-state-fn include-flags)`

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
                                              :status http-405-method-not-allowed})))
     (catch Exception ex
       (utils/exception->response ex req))))
+
+(defn make-index-event
```
shamsimam, line 729: Missing docstring.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+  [type object]
+  {:object object :type type})
+
+(defn start-tokens-watch-maintainer
```
shamsimam, line 733: Missing docstring.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+  [kv-store tokens-update-chan tokens-watch-channels-update-chan watch-refresh-timeout-ms exit-chan]
```
shamsimam, line 734: We should try and avoid passing in channels and instead create the channels inside this function in the `let` block if possible.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+                      (do
+                        (log/warn "Stopping tokens-watch-maintainer")
+                        (when (not= :exit msg)
+                          (throw (ex-info "Stopping router-state maintainer" {:time (t/now) :reason msg}))))
+
+                      tokens-update-chan
+                      (timers/start-stop-time!
+                        (metrics/waiter-timer "core" "tokens-watch-maintainer" "process-tokens-update")
+                        (let [{:keys [token owner]} msg
+                              token-index-entry (some-> (list-index-entries-for-owner kv-store owner :refresh true)
```
shamsimam, line 774: Can we have a helper function, `token->index-entry`, instead? It might be useful in `get-token->token-index-fn` also.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+                                                        (get token)
+                                                        (assoc :owner owner :token token))
+                              [event next-state]
+                              (when (not= token-index-entry (get token->token-index token))
+                                (if token-index-entry
+                                  [(make-index-event "UPDATE" token-index-entry)
```
shamsimam, line 780: Please use keywords for `:UPDATE` and `:DELETE` event types?

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+                          (doseq [[token {:keys [owner]}] only-old-indexes]
+                            (when-not (contains? only-next-indexes token)
+                              (async/>! tokens-event-chan (make-index-event "DELETE" {:owner owner :token token}))))
+                          ; if token in only-next-indexes, it has been updated (soft-delete included)
+                          (doseq [[token _] only-next-indexes]
+                            (async/>! tokens-event-chan (make-index-event "UPDATE" (get next-token->index token))))
+                          (if (and (nil? only-old-indexes) (nil? only-next-indexes))
+                            (log/info "Tokens-watch-maintainer found no differences in kv-store and current-state")
+                            (log/info "Tokens-watch-maintainer found some differences in kv-store and current-state"
+                                      {:only-in-current only-old-indexes :only-in-next only-next-indexes}))
```
shamsimam, line 820: We can also include a count of the total number of tokens

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+                        [current-state watch-refresh-timeout-chan]))]
+                (if (and next-state next-watch-refresh-timeout-chan)
+                  (recur next-state next-watch-refresh-timeout-chan)
+                  (log/info "Stopping tokens-watch-maintainer as next loop values are nil"))))
+            (catch Exception e
+              (log/error e "Fatal error in tokens-watch-maintainer"))))]
+    {:go-chan go-chan
+     :query-chan query-chan
+     :query-state-fn (fn tokens-watch-query-state-fn
+                       ([] @state-atom)
```
shamsimam, line 836: Please remove the overloaded no-args variant.

waiter/test/waiter/token_test.clj
```diff
@@ -2721,3 +2723,241 @@
...
+                                     (async/<!! go-chan))
+      auth-user "auth-user"
+      token1-metadata {"cluster" "c1" "last-update-time" 1000 "owner" "owner1"}
+      token1-service-desc {"cpus" 1}
+      token1-index {:deleted false
+                    :last-update-time (get token1-metadata "last-update-time")
+                    :maintenance false
+                    :owner (get token1-metadata "owner")
+                    :token "token1"}]
+  (deftest test-start-tokens-watch-maintainer-empty-starting-state
```
shamsimam, line 2776: Please add new lines between different `deftest` blocks.

waiter/test/waiter/token_test.clj
```diff
@@ -2721,3 +2723,241 @@
...
+        (add-watch-chans tokens-watch-channels-update-chan watch-chans)
+        (is (= {:token->token-index {} :watches-count 10} (get-latest-state query-chan)))
+        (assert-channels-next-message watch-chans '()))
+      (stop-tokens-watch-maintainer go-chan exit-chan)))
+  (deftest test-start-tokens-watch-maintainer-non-empty-starting-state
+    (let [kv-store (kv/->LocalKeyValueStore (atom {}))
+          watch-chans (create-watch-chans)
+          exit-chan (async/chan 1)
+          tokens-update-chan (au/latest-chan)
+          tokens-watch-channels-update-chan (async/chan)
```
shamsimam, line 2796: We can make the tests simpler, avoiding repeatedly initalizing the channels, if the channels are created inside the  `start-tokens-watch-maintainer`.

waiter/test/waiter/token_test.clj
```diff
@@ -2721,3 +2723,241 @@
...
+     (is (every? msg-fn# res#))))
+
+(defmacro assert-channels-next-message
+  "Assert that list of channels next message"
+  [chans msg]
+  `(let [chans# ~chans
+         msg# ~msg]
+     (assert-channels-next-message-with-fn chans# #(= % msg#))))
+
+(let [get-token-hash (fn [kv-store token] (sd/token-data->token-hash (kv/fetch kv-store token)))
```
shamsimam, line 2751: We should create a new file for the tokens watch unit tests, e.g. `token_watch_test.clj`.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +725,119 @@
...
+                                               current-token-index-map
+                                               (list-index-entries-for-owner kv-store owner :refresh true)))
+                                           {})))
+        state-atom (atom {:token->token-index (get-token->token-index-fn)
+                          :watches-count 0})
+        go-chan
+        (async/go
+          (try
+            (loop [{:keys [token->token-index watches-count] :as current-state} @state-atom
+                   watch-refresh-timeout-chan (async/timeout watch-refresh-timeout-ms)]
```
shamsimam, line 756: The way we are using `watch-refresh-timeout-chan`, I would consider using `async-utils/timer-chan`.
Also, this channel is a good candidate to pass in as an argument as it can simplify the unit tests and avoid the `Thread/sleep` calls.
In the unit tests we can pass in a regular channel as the `watch-refresh-timeout-chan` and write data to it to trigger processing of the branch we want.

waiter/src/waiter/token.clj
kevo1ution: Should we return both? In the unit tests, I wait for the go-chan to respond so it ensures the process is finished executing

waiter/src/waiter/core.clj
```diff
@@ -1354,7 +1361,16 @@
...
+                                     [:state kv-store tokens-update-chan tokens-watch-channels-update-chan]]
+                              (let [exit-chan (async/chan)
```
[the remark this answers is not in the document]
kevo1ution: looks like promise-chan is the way to go here, Only allows one message and the promise chan always provides readers the exit message

waiter/src/waiter/token.clj
shamsimam: sgtm

waiter/src/waiter/token.clj
```diff
@@ -719,3 +736,117 @@
...
+                        (metrics/waiter-timer "core" "token-watch-maintainer" "refresh")
+                        (let [next-token->index (get-token->token-index-fn)
+                              [only-old-indexes only-next-indexes _] (data/diff token->token-index next-token->index)]
+                          ; if token in old-indexes and not in only-next-indexes, then those token indexes were deleted
+                          (doseq [[token {:keys [owner]}] only-old-indexes]
+                            (when-not (contains? only-next-indexes token)
+                              (async/>! tokens-event-chan (make-index-event "DELETE" {:owner owner :token token}))))
+                          ; if token in only-next-indexes, it has been updated (soft-delete included)
+                          (doseq [[token _] only-next-indexes]
+                            (async/>! tokens-event-chan (make-index-event "UPDATE" (get next-token->index token))))
```
shamsimam, line 824: Please consider sending a single message aggregating the updates and deletes on `tokens-event-chan`.
The handler can break this message down into individual pieces if it wants to.

waiter/src/waiter/handler.clj
kevo1ution: do we want the default behavior (include-flags is nil) to provide the entire state?

waiter/config-full.edn
```diff
@@ -1042,6 +1042,26 @@
                                  ;; reduced to stale-timeout-mins + fallback-period-secs.
                                  "stale-timeout-mins" 15}}
 
+ ;; Waiter provides watch endpoints to listen on changes of a resource
+ ;; This configures various watch providers for different resources
+ :watch-config {
+                  ;; Configures token-watch-maintainer which handles and updates watches
+                  ;; for token events
+                  :token-watch-maintainer {
```
shamsimam, line 1050: Indentation feels off.

waiter/src/waiter/core.clj
```diff
@@ -1738,13 +1748,21 @@
                               (wrap-secure-request-fn
                                 (fn state-statsd-handler-fn [request]
                                   (handler/get-statsd-state router-id request))))
+   :state-token-watch-maintainer-fn (pc/fnk [[:daemons token-watch-maintainer]
+                                             [:state router-id]
+                                             wrap-secure-request-fn]
+                                    (let [{:keys [query-state-fn]} token-watch-maintainer]
```
shamsimam, line 1754: Indentation feels off.

waiter/src/waiter/core.clj
```diff
@@ -1754,7 +1772,7 @@
                            (token/handle-token-request
                              clock synchronize-fn kv-store token-cluster-calculator token-root history-length limit-per-owner
                              waiter-hostnames entitlement-manager make-inter-router-requests-sync-fn validate-service-description-fn
-                             attach-service-defaults-fn request))))
+                             attach-service-defaults-fn (:tokens-update-chan token-watch-maintainer) request))))
```
shamsimam, line 1775: Nitpick, extract this in a let-block, `{:keys [tokens-update-chan]} token-watch-maintainer`.

waiter/src/waiter/core.clj
```diff
@@ -1765,11 +1783,13 @@
...
                                 (fn token-owners-handler-fn [request]
                                   (token/handle-list-token-owners-request kv-store request))))
-   :token-refresh-handler-fn (pc/fnk [[:state kv-store]
+   :token-refresh-handler-fn (pc/fnk [[:daemons token-watch-maintainer]
+                                      [:state kv-store]
                                       wrap-router-auth-fn]
                                (wrap-router-auth-fn
                                  (fn token-refresh-handler-fn [request]
-                                   (token/handle-refresh-token-request kv-store request))))
+                                   (token/handle-refresh-token-request kv-store (:tokens-update-chan token-watch-maintainer)
```
shamsimam, line 1791: Nitpick, extract this in a let-block, {:keys [tokens-update-chan]} token-watch-maintainer.

waiter/src/waiter/process_request.clj
```diff
@@ -365,7 +365,7 @@
                                                  (log/debug "closing request input body" message)
                                                  (async/close! body-ch))))))))]
     (counters/inc! request-body-streaming-counter)
-    (when ctrl-ch
+    (when ctrl-ch                                           ; NOTE ctrl-chan example
```
shamsimam, line 368: Please delete

waiter/src/waiter/process_request.clj
```diff
@@ -523,7 +523,7 @@
                                       (timers/start-stop-time!
                                         stream-onto-resp-chan
                                         ;; don't wait forever to write to server
-                                        (au/timed-offer! resp-chan buffer streaming-timeout-ms)))
+                                        (au/timed-offer! resp-chan buffer streaming-timeout-ms))) ; NOTE: read later
```
shamsimam, line 526: Please delete

waiter/src/waiter/settings.clj
```diff
@@ -179,6 +179,12 @@
                                    (s/required-key :token-defaults) {(s/required-key "fallback-period-secs") schema/non-negative-int
                                                                      (s/required-key "https-redirect") s/Bool
                                                                      (s/required-key "stale-timeout-mins") schema/non-negative-int}}
+   (s/required-key :watch-config) {(s/required-key :token-watch-maintainer) {(s/required-key :channels-update-chan-buffer-size)
```
shamsimam, line 182: Let's keep the values in the same line as the keys.

shamsimam: Maybe rename `token-watch-maintainer` to `tokens`, it is already nested inside `watch-config`?

waiter/src/waiter/settings.clj
```diff
@@ -471,6 +477,9 @@
                   :token-defaults {"fallback-period-secs" (-> 5 t/minutes t/in-seconds)
                                    "https-redirect" false
                                    "stale-timeout-mins" 15}}
+   :watch-config {:token-watch-maintainer {:channels-update-chan-buffer-size 1024
+                                              :tokens-update-chan-buffer-size 1024
+                                              :watch-refresh-timeout-ms 10000}}
```
shamsimam, line 482: Indentation is off

waiter/src/waiter/token.clj
```diff
@@ -719,3 +760,128 @@
...
+                                    :when (not (contains? only-next-indexes token))]
+                                (make-index-event :DELETE {:owner owner :token token}))
+                              ; if token in only-next-indexes, it has been updated (soft-delete, and creation included)
+                              update-events
+                              (for [[token _] only-next-indexes]
+                                (make-index-event :UPDATE (get next-token->index token)))
+                              events (concat delete-events update-events)]
+                          (if (empty? events)
+                            (log/info "token-watch-maintainer found no differences in kv-store and current-state")
+                            (do
```
shamsimam, line 855: we should add a counter and meter to be able to check how often this even finds differences.
It gives us signals to what events might be missed.

waiter/src/waiter/token.clj
```diff
@@ -74,6 +79,16 @@
...
 
+(defn make-index-event
+  "Create an event for watch endpoints"
+  [type object]
+  {:object object :type type})
+
+(defn send-internal-index-event
+  "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
+  [tokens-update-chan token owner]
+  (async/put! tokens-update-chan {:owner owner :token token}))
```
shamsimam, line 90: Let's add a log here, so we can track which events get picked up from the logs while debugging.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +760,128 @@
                                              :status http-405-method-not-allowed})))
     (catch Exception ex
       (utils/exception->response ex req))))
+
+; TODO:
+; aggregate events,
+; possibly aggregate from tokens-event-chan,
+; finish migrating token-watch-tests
+(defn start-token-watch-maintainer
```
shamsimam, line 768: I'm thinking whether we should create a new `token_watch.clj` file for the watch maintainer.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +760,128 @@
...
+                                           :add (do
+                                                  (async/put! watch-chan
+                                                              (make-index-event :INITIAL
+                                                                                (or (vals token->token-index) '())))
+                                                  (async/tap tokens-event-mult watch-chan)
+                                                  (update current-state :watch-count inc))
+                                           :remove (update current-state :watch-count dec)
+                                           (throw (ex-info "Invalid tokens-watch-channels-update-chan event" {:event msg})))]
+                          (log/info "token-watch-maintainer watch count changed!" {:previous-count watches-count
+                                                                                    :new-count (get next-state :watch-count)})
```
shamsimam, line 835: Indentation is off.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +760,128 @@
...
+            (catch Exception e
+              (clojure.stacktrace/print-stack-trace e)
+              (log/error e "Fatal error in token-watch-maintainer")
+              ;(System/exit 1)
+              )))]
+    {:exit-chan exit-chan
+     :go-chan go-chan
+     :tokens-update-chan tokens-update-chan
+     :tokens-watch-channels-update-chan tokens-watch-channels-update-chan
+     :query-chan query-chan
```
shamsimam, line 879: lexicographic ordering

waiter/test/waiter/state/token_watch_test.clj
```diff
@@ -0,0 +1,268 @@
...
+                                                    (= :INITIAL
+                                                       (get % :type)))))
+      (stop-token-watch-maintainer go-chan exit-chan)))
+
+  (deftest test-start-token-watch-maintainer-watch-channels-updates
+    (let [kv-store (kv/->LocalKeyValueStore (atom {}))
+          watch-chans (create-watch-chans)
+          {:keys [exit-chan go-chan tokens-update-chan tokens-watch-channels-update-chan query-chan]}
+          (start-token-watch-maintainer kv-store 1 1 (async/chan))]
+      (testing "watch-channels get UPDATE event for added tokens"
```
shamsimam, line 132: Let's have newlines between individual `testing` clauses.

waiter/src/waiter/token.clj
```diff
@@ -15,19 +15,26 @@
...
             [waiter.kv :as kv]
+            [waiter.metrics :as metrics]
             [waiter.service-description :as sd]
             [waiter.status-codes :refer :all]
             [waiter.util.date-utils :as du]
             [waiter.util.ring-utils :as ru]
-            [waiter.util.utils :as utils])
+            [waiter.util.utils :as utils]
+            [waiter.util.async-utils :as au]
+            [metrics.meters :as meters])
```
shamsimam, line 37: Lexicographic ordering

waiter/src/waiter/token.clj
```diff
@@ -74,6 +81,17 @@
...
 
+(defn make-index-event
+  "Create an event for watch endpoints"
+  [type object]
+  {:object object :type type})
+
+(defn send-internal-index-event
+  "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
+  [tokens-update-chan token owner]
+  (log/info "Sending internal index event" {:token token :owner owner})
```
shamsimam, line 92: Nitpick: please use lowercase letter for first words in the logs

waiter/src/waiter/token.clj
```diff
@@ -260,7 +279,29 @@
               (doseq [[_ owner-key] existing-owner->owner-key]
                 ;; defensive check to avoid deleting duplicated keys after reindex
                 (when-not (contains? new-owner-keys owner-key)
-                  (kv/delete kv-store owner-key))))))))))
+                  (kv/delete kv-store owner-key)))))))))
+
+  (defn get-token->token-index
+    "Return a fresh map of ALL token to token index entry. The token index entries also include the owner and token"
```
shamsimam, line 285: `fresh` depends on the `refresh` flag passed in?

waiter/src/waiter/token.clj
```diff
@@ -719,3 +763,125 @@
...
+
+                      tokens-watch-channels-update-chan
+                      (timers/start-stop-time!
+                        (metrics/waiter-timer "core" "token-watch-maintainer" "process-tokens-channels-update")
+                        (let [{:keys [channel-event watch-chan]} msg
+                              next-state (case channel-event
+                                           :add (do
+                                                  (async/put! watch-chan
+                                                              (make-index-event :INITIAL
+                                                                                (or (vals token->token-index) '())))
```
shamsimam, line 828: Please use an empty vector `[]` instead of a list. Seeing the use of list literal is unusual in the codebase.

waiter/src/waiter/token.clj
```diff
@@ -719,3 +763,125 @@
...
+              (log/error e "Fatal error in token-watch-maintainer")
+              (System/exit 1))))]
+    {:exit-chan exit-chan
+     :go-chan go-chan
+     :query-chan query-chan
+     :query-state-fn (fn tokens-watch-query-state-fn
+                       [include-flags]
+                       (let [state @state-atom]
+                         (cond-> {}
+                                 (contains? include-flags "token->token-index")
```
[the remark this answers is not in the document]
kevo1ution: maybe `token-index-map` is better? Ideally, I want to have consistent field naming internally with the include and json response we expose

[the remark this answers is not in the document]
shamsimam: Okay.

Resolve the name issue by explicitly providing the support parameters.
e.g. https://github.com/twosigma/waiter/blob/6401cdd70fe779a48073f699286fb629d2f48476/waiter/src/waiter/scheduler/kubernetes.clj#L854-L855

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+                              current-state
+                              ; If index-entry retrieved from kv-store exists then treat as UPDATE (includes token creation)
+                              (some? token-index-entry)
+                              (let [closed-chans (send-event-to-channels watch-chans (tkn/make-index-event :UPDATE token-index-entry))]
+                                (-> current-state
+                                    (assoc-in [:token-index-map token] token-index-entry)
+                                    (assoc :watch-chans (apply disj watch-chans closed-chans))))
+                              :else
+                              (let [event-obj {:owner owner :token token}
+                                    closed-chans (send-event-to-channels watch-chans (tkn/make-index-event :DELETE event-obj))]
```
shamsimam, line 89: Nitpick: Can we avoid repeated code and pull out the `send-event-to-channels` call and the `:watch-chans` update?

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+                                (-> current-state
+                                    (assoc :token-index-map (dissoc token-index-map token))
```
shamsimam, line 91: We can use `dissoc-in` here.
https://github.com/twosigma/waiter/blob/6401cdd70fe779a48073f699286fb629d2f48476/waiter/src/waiter/util/utils.clj#L475

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+                      (condp = current-chan
+                        exit-chan
+                        (do
+                          (log/warn "Stopping token-watch-maintainer")
+                          (when (not= :exit msg)
+                            (throw (ex-info "Stopping router-state maintainer" {:time (clock) :reason msg}))))
+
+                        tokens-update-chan
+                        (timers/start-stop-time!
+                          (metrics/waiter-timer "core" "token-watch-maintainer" "process-tokens-update")
```
shamsimam, line 73: Nitpick: change name to mimic the `refresh` name below.
```suggestion
                          (metrics/waiter-timer "core" "token-watch-maintainer" "process")
```

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+                                [only-old-indexes only-next-indexes _] (data/diff token-index-map next-token-index-map)
+                                ; if token in old-indexes and not in only-next-indexes, then those token indexes were deleted
+                                delete-events
+                                (for [[token {:keys [owner]}] only-old-indexes
+                                      :when (not (contains? only-next-indexes token))]
+                                  (tkn/make-index-event :DELETE {:owner owner :token token}))
+                                ; if token in only-next-indexes, it has been updated (soft-delete, and creation included)
+                                update-events
+                                (for [[token _] only-next-indexes]
+                                  (tkn/make-index-event :UPDATE (get next-token-index-map token)))
```
shamsimam, line 114: Should `make-index-event` belong to this file too?

waiter/src/waiter/token.clj
```diff
@@ -74,6 +75,17 @@
    :last-update-time last-update-time
    :maintenance (some? maintenance)})
 
+(defn make-index-event
```
shamsimam, line 78: We should move this to token_watch.clj

waiter/src/waiter/token.clj
```diff
@@ -74,6 +75,17 @@
...
+  "Create an event for watch endpoints"
+  [type object]
+  {:object object :type type})
+
+(defn send-internal-index-event
```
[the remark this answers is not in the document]
kevo1ution: I think this causes a dependency cycle, token-watch depends on some functions in `token.clj`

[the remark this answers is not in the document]
shamsimam: Okay.

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+            [waiter.correlation-id :as cid]
+            [waiter.metrics :as metrics]
+            [waiter.token :as tkn]
+            [waiter.util.async-utils :as au]))
+
+(defn send-event-to-channels
+  "Given a list of watch channels and the event to send to each channel, send the event in a non blocking fashion and
+  close channels that error the async/put! operation. Returns a list of closed channels."
+  [watch-chans event]
+  (for [watch-chan watch-chans
```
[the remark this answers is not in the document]
kevo1ution: https://clojuredocs.org/clojure.core/filter
The doc states that the predicate should be free of side effects, and it returns a lazy sequence. I saw some articles discussing that you shouldn't mix side effects with lazy sequences, so I decided to use a `reduce` here instead.

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,151 @@
...
+        :when
+        (try
+          (not (async/put! watch-chan event))
+          (catch AssertionError e
+            (log/error e "closing watch-chan due to error")
+            (async/close! watch-chan)
+            true))]
+    watch-chan))
```
shamsimam, line 25: Returns the list of active chans, so ew do not have to run a remove/`disj` later:
```suggestion
  (->> watch-chans
    (map (fn send-event! [watch-chan]
           (try
              (when (async/put! watch-chan event)
                watch-chan)
              (catch AssertionError e
                (log/error e "closing watch-chan due to error")
                (async/close! watch-chan)
                nil))
    (remove nil?))  
```

waiter/src/waiter/token_watch.clj
kevo1ution: `Any empty maps that result will not be present in the new structure.` in the function description, so that means the token->index-map will be nil instead of an empty map. Do we distinguish between empty maps and nil?

waiter/src/waiter/token_watch.clj
kevo1ution: I went with reduce because of the side effects, as `map` and `remove` also returns a lazy sequence.

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -0,0 +1,362 @@
...
+    {:exit-chan exit-chan
+     :error-chan error
+     :go-chan go-chan
+     :query-state-fn query-state-fn}))
+
+(defn- start-watches
+  [router-urls cookies]
+  (mapv
+    #(start-watch % cookies)
+    router-urls))
```
shamsimam, line 165: If we maintain a router-id/url->watch map, we can provide more explanations in the failed assertions about what failed and on which router.

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -0,0 +1,362 @@
...
+(defmacro assert-watch-token-index-entry
+  [watch token entry]
+  `(let [watch# ~watch
+         token# ~token
+         entry# ~entry
+         query-state-fn# (:query-state-fn watch#)]
+     (is (wait-for
+           #(= (get (query-state-fn#) token#)
+               entry#)
+           :interval 1 :timeout 5))))
```
shamsimam, line 186: `is` accepts a string as the second argument, we should use it to explain which assertion failed and on which router.

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -0,0 +1,362 @@
...
+          (is (await-goal-response-for-all-routers goal-fn watch-state-request-fn router_urls)))))))
+
+(defn- start-watch
+  [router-url cookies & {:keys [query-params] :or {query-params {"include" ["metadata", "deleted"]
+                                                                 "watch" "true"}}}]
+  (let [{:keys [body error]} (make-request router-url "/tokens"
+                                           :cookies cookies
+                                           :fold-chunked-response? false
+                                           :query-params query-params)
+        token->index-atom (atom {})
```
shamsimam, line 117: Nitpick: Can we assert the response status here?

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -0,0 +1,362 @@
...
+   {:keys [cookies headers query-params] :or {cookies [] headers {} query-params {}}}]
+  (let [{:keys [body] :as index-response}
+        (make-request waiter-url "/tokens" :query-params {"include" ["metadata", "deleted"]} :cookies cookies)]
+    (assert-response-status index-response 200)
+    (->> body
+         try-parse-json
+         (filter #(= token (get % "token")))
+         first)))
+
+(deftest ^:parallel ^:integration-fast test-token-watch-maintainer-watches
```
shamsimam, line 207: ❤️

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -0,0 +1,362 @@
...
+          (assert-response-status res-1 200)
+          (assert-response-status res-2 200)
+          (try
+            (delete-token-and-assert waiter-url token-1 :hard-delete false)
+            (let [entry (get-token-index waiter-url token-2 :cookies cookies)
+                  watch (start-watch waiter-url cookies :query-params {"include" "metadata" "watch" "true"})]
+              (assert-watch-token-index-entry watch token-1 nil)
+              (assert-watch-token-index-entry watch token-2 entry)
+              (delete-token-and-assert waiter-url token-2 :hard-delete false)
+              (async/<!! (async/timeout 4000))
```
shamsimam, line 274: Why do we have these sleeps?

waiter/src/waiter/token.clj
```diff
@@ -589,26 +623,74 @@
...
+                                          (map metadata-transducer-fn)))
+                :EVENTS
+                (assoc event :object (->> object
+                                          (filter (fn [{:keys [object]}]
+                                                    (index-filter-fn object)))
+                                          (map (fn [{:keys [type] :as entry}]
+                                                 (if (= type :UPDATE)
+                                                   (update entry :object metadata-transducer-fn)
+                                                   entry)))))
+                (throw (ex-info "Invalid event type provided" {:event event})))))
```
shamsimam, line 659: Does throwing the exception perform the necessary cleanup? Are we relying on `ctrl` chan to be triggered when the exception is thrown?

waiter/src/waiter/token_watch.clj
```diff
@@ -0,0 +1,158 @@
...
+
+                        query-chan
+                        (let [{:keys [response-chan include-flags]} msg]
+                          (async/put! response-chan (query-state-fn include-flags))
+                          current-state))]
+                  (if next-state
+                    (recur (assoc next-state :last-update-time (clock)))
+                    (log/info "Stopping token-watch-maintainer as next loop values are nil"))))
+              (catch Exception e
+                (clojure.stacktrace/print-stack-trace e)
```
shamsimam, line 146: Please remove.

waiter/src/waiter/util/client_tools.clj
```diff
@@ -373,11 +373,12 @@
 
 (defn make-request
   ([waiter-url path &
-    {:keys [body client cookies content-type disable-auth form-params headers
+    {:keys [body client cookies content-type disable-auth fold-chunked-response? form-params headers
             method multipart protocol query-params scheme trailers-fn verbose]
      :or {body nil
           cookies []
           disable-auth nil
+          fold-chunked-response? true
```
shamsimam, line 381: Instead of `fold-chunked-response?` , should we call it `async?` as basically we want to differentiate between synchronous and asynchronous response. And for asynchronous responses we disable folding.

waiter/integration/waiter/token_watch_integration_test.clj
kevo1ution: The server can possibly send an event during the `delete-token-and-assert`. If I do the assertion too soon (without the sleeps), it might be a false positive as the event can be process after the assertion.

shamsimam: Can we use `wait-for` instead?

kevo1ution: I could assert that wati-for returns false

waiter/src/waiter/token.clj
kevo1ution: These exceptions trigger the `watch-chan-ex-handler-fn` https://clojuredocs.org/clojure.core.async/chan

waiter/src/waiter/token.clj
```diff
@@ -589,26 +623,74 @@
...
+        (fn watch-chan-ex-handler [e]
+          (log/error e "Error during transformation of a token watch event")
+          (async/go (async/>! ctrl e)))
+        watch-chan-buffer (async/buffer 1000)
+        watch-chan (async/chan watch-chan-buffer watch-chan-xform watch-chan-ex-handler-fn)]
+    (async/go
+      (let [data (async/<! ctrl)]
+        (log/info "closing watch-chan, as ctrl channel in request has been triggered" {:data data})
+        (async/close! watch-chan)))
+    (async/put! tokens-watch-channels-update-chan watch-chan)
```
shamsimam, line 680: We should handle `false` return value from `put!`

waiter/src/waiter/token.clj
```diff
@@ -589,26 +623,74 @@
...
+                (throw (ex-info "Invalid event type provided" {:event event})))))
+          (filter
+            (fn empty-aggregate-events [{:keys [object type] :as event}]
+              (case type
+                :INITIAL true
+                :EVENTS (not-empty object)
+                (throw (ex-info "Invalid event type provided" {:event event})))))
+          (map
+            (fn [event]
+              (log/info "forwarding tokens event to client" {:event event})
```
shamsimam, line 668: This may be very large, should we have an terser info and a more verbose debug?

waiter/src/waiter/token.clj
```diff
@@ -589,26 +623,74 @@
...
       :get (handle-get-token-request kv-store cluster-calculator token-root waiter-hostnames request)
       :post (handle-post-token-request clock synchronize-fn kv-store cluster-calculator token-root history-length limit-per-owner
                                        waiter-hostnames entitlement-manager make-peer-requests-fn validate-service-description-fn
-                                       attach-service-defaults-fn request)
+                                       attach-service-defaults-fn tokens-update-chan request)
       (throw (ex-info "Invalid request method" {:log-level :info :request-method request-method :status http-405-method-not-allowed})))
     (catch Exception ex
       (utils/exception->response ex request))))
 
+(defn- handle-list-tokens-watch
```
shamsimam, line 639: Should we unit test this?

kevo1ution: good idea, we can test certain behaviors that are difficult to replicate in integration tests

waiter/src/waiter/token.clj
```diff
@@ -665,19 +666,23 @@
...
         watch-chan (async/chan watch-chan-buffer watch-chan-xform watch-chan-ex-handler-fn)]
     (async/go
       (let [data (async/<! ctrl)]
         (log/info "closing watch-chan, as ctrl channel in request has been triggered" {:data data})
         (async/close! watch-chan)))
-    (async/put! tokens-watch-channels-update-chan watch-chan)
+    (when-not (async/put! tokens-watch-channels-update-chan watch-chan)
+      (let [e (ex-info "tokens-watch-channels-update-chan is closed!" {})]
+        (async/put! ctrl e)
+        (throw e)))
```
shamsimam, line 685: we can use `if-not` and render the error response directly using `exception->response`  and avoid using the ctrl chan altogether. We should set up the `async/go` above only if the put succeeds in the else block before returning the response map.

shamsimam (approved):

PR #1278, opened 2021-01-28
adds support for overriding value from file in arguments

shamsimam:

- adds support for overriding value from file in arguments during token create and update

Not being able to override values from one provided ina file is sometimes surprising to users. We preserve the current behavior as the default and introduce a new `--override` flag to allow users to pass in values they intend to override.

#### Help text
```
$ waiter create -h
usage: waiter create [-h] [--name NAME] [--owner OWNER] [--version VERSION] [--cmd CMD] [--cmd-type CMD-TYPE] [--cpus CPUS] [--mem MEM] [--ports PORTS] [--json JSON | --yaml YAML | --input INPUT] [--override] [token]
...
  --input INPUT         provide the data in a JSON/YAML file
  --override            Allow overriding values in file with values from other arguments. Overriding values is disallowed by default.
```

#### Usage
```
$ waiter create test-token --cpus 2 --json integration/foo.json
You cannot specify the same parameter in both an input file and token field flags at the same time (cpus) without specifying the --override flag.

$ waiter create test-token --no-override --cpus 2 --json integration/foo.json
You cannot specify the same parameter in both an input file and token field flags at the same time (cpus) without specifying the --override flag.

$ waiter create test-token --no-override --override --cpus 2 --json integration/foo.json
...
waiter create: error: argument --override: not allowed with argument --no-override

$ waiter create test-token --override --cpus 2 --json integration/foo.json
Successfully created test-token.

$ waiter create test-token --cpus 2 --json integration/foo.json
You cannot specify the token name both as an argument and in the input file without specifying the --override flag.

$ waiter --verbose create test-token --override --cpus 2 --json integration/foo.json
...
2021-01-28 08:34:38,670 [DEBUG] [root] Will use token name (test-token) from args and skip token specified in file(my-token)
...
No changes detected for test-token.
2021-01-28 08:34:38,692 [DEBUG] [root] result: 0
```

### DIFF
--- a/cli/waiter/token_post.py
+++ b/cli/waiter/token_post.py
@@ -1,15 +1,12 @@
-import argparse
-import json
 import logging
-import sys
 import os
 from enum import Enum
 
 import requests
 
 from waiter import terminal, http_util
-from waiter.querying import get_token, query_token, get_target_cluster_from_token
 from waiter.data_format import load_data
+from waiter.querying import get_token, query_token, get_target_cluster_from_token
 from waiter.util import FALSE_STRINGS, print_info, response_message, TRUE_STRINGS, guard_no_cluster, str2bool
 
 BOOL_STRINGS = TRUE_STRINGS + FALSE_STRINGS
@@ -166,11 +163,9 @@ def add_token_flags(parser):
 
 def add_override_flags(parser):
     """Adds the arguments override file values flags to the given parser"""
-    override_group = parser.add_mutually_exclusive_group(required=False)
-    override_group.add_argument('--override', action='store_true', dest='override',
-                                help='Allow overriding values in input file with values from CLI arguments. '
-                                     'Overriding values is disallowed by default.')
-    override_group.add_argument('--no-override', action='store_false', dest='override', help=argparse.SUPPRESS)
+    parser.add_argument('--override', action='store_true', default=False, dest='override',
+                        help='Allow overriding values in input file with values from CLI arguments. '
+                             'Overriding values is disallowed by default.')
 
 
 def register_argument_parser(add_parser, action):
--- a/cli/waiter/token_post.py
+++ b/cli/waiter/token_post.py
@@ -168,7 +168,7 @@ def add_override_flags(parser):
     """Adds the arguments override file values flags to the given parser"""
     override_group = parser.add_mutually_exclusive_group(required=False)
     override_group.add_argument('--override', action='store_true', dest='override',
-                                help='Allow overriding values in file with values from other arguments. '
+                                help='Allow overriding values in input file with values from CLI arguments. '
                                      'Overriding values is disallowed by default.')
     override_group.add_argument('--no-override', action='store_false', dest='override', help=argparse.SUPPRESS)
 
--- a/cli/waiter/token_post.py
+++ b/cli/waiter/token_post.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


cli/waiter/token_post.py
```diff
@@ -152,6 +164,15 @@ def add_token_flags(parser):
     parser.add_argument('--ports', help='number of ports to reserve for service', type=int)
 
 
+def add_override_flags(parser):
+    """Adds the arguments override file values flags to the given parser"""
+    override_group = parser.add_mutually_exclusive_group(required=False)
+    override_group.add_argument('--override', action='store_true', dest='override',
+                                help='Allow overriding values in file with values from other arguments. '
+                                     'Overriding values is disallowed by default.')
+    override_group.add_argument('--no-override', action='store_false', dest='override', help=argparse.SUPPRESS)
```
kevo1ution, line 173: Is this necessary if the default is to not do an override?

cli/waiter/token_post.py
```diff
@@ -93,13 +95,22 @@ def create_or_update_token(clusters, args, _, enforce_cluster, action):
...
+                            f'and token field flags at the same time ({", ".join(shared_keys)}) '
+                            f'without specifying the --override flag.')
+        else:
+            logging.debug(f'Following parameters have specified values in both file and flags: {shared_keys}')
 
     token_fields = {**token_fields_from_json, **token_fields_from_args}
     token_name_from_json = token_fields.pop('token', None)
     if token_name_from_args and token_name_from_json:
-        raise Exception('You cannot specify the token name both as an argument and in the input file.')
+        if not allow_override:
```
kevo1ution, line 108: Can there be a test for overriding the token name behavior?

cli/waiter/token_post.py
```diff
@@ -152,6 +164,15 @@ def add_token_flags(parser):
     parser.add_argument('--ports', help='number of ports to reserve for service', type=int)
 
 
+def add_override_flags(parser):
+    """Adds the arguments override file values flags to the given parser"""
+    override_group = parser.add_mutually_exclusive_group(required=False)
+    override_group.add_argument('--override', action='store_true', dest='override',
+                                help='Allow overriding values in file with values from other arguments. '
```
kevo1ution, line 171: nitpick
```suggestion
                                help='Allow overriding values in input file with values from CLI arguments. '
```

kevo1ution (changes requested):

cli/waiter/token_post.py
shamsimam: Added it for completeness for boolean arguments, I'm okay dropping it if you do not like it.

cli/waiter/token_post.py
shamsimam: Will add.

shamsimam: Added test, ready for another round of review.

cli/waiter/token_post.py
shamsimam: Removed `--no-override` argument.

kevo1ution (approved):

cli/waiter/token_post.py
twosigmajab: FWIW, Click (a popular alternative CLI library to argparse) recommends keeping both options. This allows a caller who wants to be explicit about their `--no-override` preference to do so today, and their invocation will continue to behave the same way, even if the Waiter CLI changes its default behavior to `--override` in the future.

https://click.palletsprojects.com/en/7.x/options/#boolean-flags
> Click always wants you to provide an enable and disable flag so that you can change the default later.

That said, if there are any usage metrics about how often the "You cannot specify the same parameter in both..." error is occurring, it would be worth consulting those. If the only occurrences in the past N months correspond to our own usage, then I think we could safely flip the default behavior to `--override` today without breaking anyone.

Sorry for the late comment, I only just saw this!

twosigmajab: And thanks for getting this improvement in so quickly! Shams ftw!

shamsimam: Adding both the options back, `--override` and `--no-override`.

PR #1281, opened 2021-01-31
CORS access-control-request-headers is optional

DaoWen:

Make `access-control-request-headers` header optional on CORS preflight requests.

Requiring the header was a bug. The MDN documentation is not very clear on this, but a careful reading of the linked spec shows that this header is not required if no custom headers are being sent with the CORS request.

### DIFF
--- a/waiter/src/waiter/cors.clj
+++ b/waiter/src/waiter/cors.clj
@@ -42,14 +42,13 @@
 (defn preflight-request?
   "Determines if a request is a CORS preflight request.
    A CORS preflight request is a CORS request that checks to see if the CORS protocol is understood.
-   It is an OPTIONS request, using three HTTP request headers:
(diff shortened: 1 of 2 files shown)


shamsimam (approved):

waiter/src/waiter/cors.clj
```diff
@@ -81,7 +80,7 @@
             (let [{:strs [access-control-request-headers]} headers]
               (utils/attach-waiter-source
                 {:headers {"access-control-allow-credentials" "true"
-                           "access-control-allow-headers" access-control-request-headers
+                           "access-control-allow-headers" (or access-control-request-headers "*")
```
shamsimam, line 83: Nitpick: we should have a test that confirms this new behavior.

PR #1308, opened 2021-03-15
Token Watch Maintainer uses token from KV-store for event generation

kevo1ution:

- use token for event generation instead of owner-index

- fixes possible race condition when user changes owner index very quickly

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -248,7 +248,7 @@
 
   (defn get-token-index
     "Given a token, determine the correct index entry of the token. Returns nil if token doesn't exist."
-    [kv-store token & {:keys [refresh] :or {refresh false}}]
+    [kv-store token & {:keys [include refresh] :or {include #{} refresh false}}]
     (timers/start-stop-time!
       (metrics/waiter-timer "core" "token" "get-token-index" (get-refresh-metric-name refresh))
       (let [{:strs [deleted last-update-time maintenance owner] :as token-data} (kv/fetch kv-store token :refresh refresh)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/token.clj
```diff
@@ -264,10 +276,7 @@
               owner->index-entries (pc/map-vals
                                      (fn [tokens]
                                        (pc/map-from-keys
-                                         (fn [token]
-                                           (let [{:strs [deleted last-update-time maintenance] :as token-data} (kv/fetch kv-store token)
-                                                 token-hash (sd/token-data->token-hash token-data)]
-                                             (make-index-entry token-hash deleted last-update-time maintenance)))
+                                         (partial get-token-index kv-store)
```
shamsimam, line 279: Please uses inline function instead of `partial`. We have found performance issues with use of `partial` in the past.
```suggestion
                                         #(get-token-index kv-store % :include #{})
```

waiter/src/waiter/token.clj
```diff
@@ -246,6 +246,18 @@
     [kv-store]
     (into {} (kv/fetch kv-store token-owners-key)))
 
+  (defn get-token-index
+    "Given a token, determine the correct index entry of the token. Returns nil if token doesn't exist."
+    [kv-store token & {:keys [include refresh] :or {include #{} refresh false}}]
```
shamsimam, line 251:
```suggestion
    [kv-store token & {:keys [include refresh] :or {include #{:token :owner} refresh false}}]
```
Keeps the changes backward compatible with previous function.

shamsimam: An index, by default, contains the `owner` and `token`. Not sure what we would label something that doesn't have these.
Maybe drop making `include` configurable?

shamsimam (approved):

PR #1310, opened 2021-03-16
provide better validation error message with user-metadata

kevo1ution:

- Add validation error for token user metadata (e.g. "maintenance", "cors-rules", "owner")
- example cli output when putting invalid user metadata:
```
{ // token config
    "cors-rules": [{
        "target-path-regex": "",
        "methods": [], 
        "random-key": "testing" 
    }],
    "fallback-period-secs": 86401,
    "owner": "",
    "stale-timeout-mins": 241
}
```
Screenshot from 2021-03-16 17-09-55

- Error message provided to user makes it unclear why validation failed when creating their token

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -229,8 +229,7 @@
                    (conj error-messages
                          (str key " "
                               (case key
-                                "origin-regex" "must be a valid regular expression"
-                                "target-path-regex" "must be a valid regular expression"
+                                ("origin-regex" "target-path-regex") "must be a valid regular expression"
                                 "methods" "must be a non-empty list of http methods (in upper-case)"
                                 "is not a valid key"))))
                  []
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -478,6 +478,7 @@
        (assoc parameter-key# (do ~error-message-body)))))
 
 (defn validate-user-metadata-schema
+  "Validates provided user-metadata of a token and throws an error with user readable validation issues."
   [user-metadata]
   (try
     (s/validate user-metadata-schema user-metadata)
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -445,6 +477,40 @@
        (contains? parameter->issues# parameter-name#)
        (assoc parameter-key# (do ~error-message-body)))))
 
+(defn validate-user-metadata-schema
```
shamsimam, line 480: Missing docstring.

shamsimam (approved):

waiter/src/waiter/service_description.clj
```diff
@@ -445,6 +477,40 @@
...
+                                           :maintenance
+                                           (generate-friendly-maintenance-error-message parameter->issues))
+                                         (attach-error-message-for-parameter
+                                           parameter->issues
+                                           :owner
+                                           "owner must be a non-empty string")
+                                         (attach-error-message-for-parameter
+                                           parameter->issues
+                                           :stale-timeout-mins
+                                           "stale-timeout-mins must be an integer between 0 240"))]
```
shamsimam, line 510:
```suggestion
                                           "stale-timeout-mins must be an integer between 0 and 240 (includisve)"))]
```

waiter/src/waiter/service_description.clj
```diff
@@ -216,6 +216,38 @@
...
+  (when-let [cors-rules (get issue "cors-rules")]
+    (if (coll? cors-rules)
+      (->> cors-rules
+           (map
+             (fn [rule]
+               (reduce
+                 (fn [error-messages [key _]]
+                   (conj error-messages
+                         (str key " "
+                              (case key
```
shamsimam, line 231: Nitpick: use `cond` to collapse the origin-regex and target-apth-regex cases.

waiter/src/waiter/service_description.clj
```diff
@@ -216,6 +216,38 @@
...
+                                "origin-regex" "must be a valid regular expression"
+                                "target-path-regex" "must be a valid regular expression"
+                                "methods" "must be a non-empty list of http methods"
```
shamsimam, line 234:
```suggestion
                                "methods" "must be a non-empty list of http methods (in upper-case)"
```

shamsimam (approved):

PR #1311, opened 2021-03-17
Token Watch API: Remove owner from internal index events and delete event

kevo1ution:

- Remove owner from internal index events and delete events

- Decrease tech debt

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -87,7 +87,7 @@
   "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
   [tokens-update-chan token]
   (log/info "sending internal index event" {:token token})
-  (async/put! tokens-update-chan token))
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/token.clj
```diff
@@ -85,9 +85,9 @@
 
 (defn send-internal-index-event
   "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
-  [tokens-update-chan token owner]
-  (log/info "sending internal index event" {:token token :owner owner})
-  (async/put! tokens-update-chan {:owner owner :token token}))
+  [tokens-update-chan token]
+  (log/info "sending internal index event" {:token token})
+  (async/put! tokens-update-chan token))
```
shamsimam, line 90: Let's continue to forward a map to `tokens-update-chan`. We can add in more items later if we need to easily then.

shamsimam (approved):

PR #1313, opened 2021-03-24
adds support for editor parameter in tokens

shamsimam:

- adds support for editor parameter in tokens

Allows giving editing privileges to another account other than the owner.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -519,11 +519,11 @@
       (let [existing-editor (get existing-token-metadata "editor")
             existing-owner (get existing-token-metadata "owner")
             creating-token? (empty? existing-token-metadata)
-            owner? (and existing-owner
-                        (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
+            existing-owner? (and existing-owner
+                                 (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
             editor? (and (not creating-token?)
+                         (not existing-owner?)
                          existing-editor
-                         (not owner?)
                          (authz/run-as? entitlement-manager authenticated-user existing-editor))]
         (when editor?
           (log/info "applying editor privileges to operation" {:editor authenticated-user :owner existing-owner})
@@ -538,7 +538,7 @@
                                    :parameter parameter-name
                                    :parameter-exiting-value existing-value
                                    :parameter-new-value new-value
-                                   :privileges {:editor? editor? :owner? owner?}
+                                   :privileges {:current-owner? existing-owner? :editor? editor?}
                                    :status http-403-forbidden
                                    :log-level :warn})))))))
         ;; only check run-as-user rules when not running as editor, editor cannot change run-as-user from previous check
@@ -557,7 +557,7 @@
                               {:authenticated-user authenticated-user
                                :existing-owner existing-owner
                                :new-user owner
-                               :privileges {:editor? editor? :owner? owner?}
+                               :privileges {:editor? editor? :owner? existing-owner?}
                                :status http-403-forbidden
                                :log-level :warn}))))
           ;; new token creation
@@ -565,7 +565,6 @@
             (throw (ex-info "Cannot create token as user"
                             {:authenticated-user authenticated-user
                              :owner owner
-                             :privileges {:editor? editor? :owner? owner?}
                              :status http-403-forbidden
                              :log-level :warn}))))
         ;; Neither owner nor editor may modify system metadata fields
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -518,11 +518,15 @@
       nil
       (let [existing-editor (get existing-token-metadata "editor")
             existing-owner (get existing-token-metadata "owner")
-            owner? (and existing-owner (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
-            editor? (and existing-editor (not owner?)  (authz/run-as? entitlement-manager authenticated-user existing-editor))]
+            creating-token? (empty? existing-token-metadata)
+            owner? (and existing-owner
+                        (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
+            editor? (and (not creating-token?)
+                         existing-editor
+                         (not owner?)
+                         (authz/run-as? entitlement-manager authenticated-user existing-editor))]
         (when editor?
-          (log/info "applying editor privileges to operation" {:editor authenticated-user :owner existing-owner}))
-        (if editor?
+          (log/info "applying editor privileges to operation" {:editor authenticated-user :owner existing-owner})
           (let [existing-token-parameters (sd/token->token-parameters kv-store token :include-deleted false)]
             (doseq [parameter-name ["editor" "owner" "run-as-user"]]
               (let [existing-value (get existing-token-parameters parameter-name)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,61 @@
...
 
       nil
-      (do
+      (let [existing-editor (get existing-token-metadata "editor")
+            existing-owner (get existing-token-metadata "owner")
+            owner? (and existing-owner (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
+            editor? (and existing-editor (not owner?)  (authz/run-as? entitlement-manager authenticated-user existing-editor))
+            delegated-user (if editor?
+                             (do ;; allow editor to run with owner privileges
+                               (log/info "promoting editor to owner privileges while editing token"
```
sradack, line 525: Can we update this log message?  Editors don't have the same permissions as owners, so the editor doesn't actually have owner privileges.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,61 @@
                            :log-level :warn}))))
 
       nil
-      (do
+      (let [existing-editor (get existing-token-metadata "editor")
+            existing-owner (get existing-token-metadata "owner")
```
sradack, line 520: We used to have code that allowed for tokens to not have an existing owner.  Does this break that backward compatibility?

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,61 @@
...
+            owner? (and existing-owner (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
+            editor? (and existing-editor (not owner?)  (authz/run-as? entitlement-manager authenticated-user existing-editor))
+            delegated-user (if editor?
```
sradack, line 523: A weird situation can happen when setting an editor on a token that doesn't have an owner.  e.g. If run-as-user is set to me, and owner is null, and I set editor to me as well, I can no longer change the owner, editor, or run-as-user.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,61 @@
...
+                                  {:authenticated-user authenticated-user
+                                   :existing-token-description existing-token-parameters
+                                   :parameter parameter-name
+                                   :parameter-exiting-value existing-value
+                                   :parameter-new-value new-value
+                                   :status http-403-forbidden
+                                   :log-level :warn})))))))
         (when (and run-as-user (not= "*" run-as-user))
-          (when-not (authz/run-as? entitlement-manager authenticated-user run-as-user)
+          (when-not (authz/run-as? entitlement-manager delegated-user run-as-user)
```
sradack, line 544: Previously we were checking if the authenticated user could `run-as?`. Now we're checking if the owner can `run-as?` It's not clear why we are changing the original behavior.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,61 @@
...
             (throw (ex-info "Cannot run as user"
                             {:authenticated-user authenticated-user
+                             :delegated-user delegated-user
```
sradack, line 547: The `:delegated-user` key and `delegated-user` symbol are named curiously in the case of editing as an owner.  If I'm editing the token as the owner, I'm not delegating.

sradack (changes requested):

waiter/src/waiter/token.clj
sradack: What if a token doesn't have a `run-as-user` or `run-as-user` is `*` or ...

waiter/src/waiter/token.clj
sradack: Do we always have a `run-as-user`?

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,62 @@
...
-          (throw (ex-info "Cannot modify root token metadata"
-                          {:status http-400-bad-request
-                           :token-metadata new-token-metadata
-                           :log-level :warn})))
-        (when (contains? new-token-metadata "previous")
-          (throw (ex-info "Cannot modify previous token metadata"
-                          {:status http-400-bad-request
-                           :token-metadata new-token-metadata
-                           :log-level :warn}))))
+        (if-not (str/blank? existing-owner)
```
sradack, line 549: Isn't your claim that `existing-owner` is always available?

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,62 @@
...
+                                  {:authenticated-user authenticated-user
+                                   :existing-token-description existing-token-parameters
+                                   :parameter parameter-name
+                                   :parameter-exiting-value existing-value
+                                   :parameter-new-value new-value
+                                   :privileges {:editor? editor? :owner? owner?}
+                                   :status http-403-forbidden
+                                   :log-level :warn}))))))
+          ;; only check run-as-user rules when not running as editor
+          (when (and run-as-user (not= "*" run-as-user))
```
sradack, line 541: Nitpick: I think it reads more cleanly if move this out of the else clause into it's own high level "rule".  It's unrelated to the branch taken in the `true` case.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,62 @@
...
+            (when-not (authz/run-as? entitlement-manager authenticated-user run-as-user)
+              (throw (ex-info "Cannot run as user"
                               {:authenticated-user authenticated-user
-                               :owner owner
+                               :privileges {:editor? editor? :owner? owner?}
```
sradack, line 545: Can `editor?` ever be anything other than `false` here?  Should we revert back to the old log message?

waiter/src/waiter/token.clj
sradack: If this check for existing owner is really checking whether the token exists or not, can we make that a bit more explicit?

waiter/src/waiter/token.clj
shamsimam: Oops!

I changed the code to account for a missing owner. This should not break backward compatibility now.

> A weird situation can happen when setting an editor on a token that doesn't have an owner. e.g. If run-as-user is set to me, and owner is null, and I set editor to me as well, I can no longer change the owner, editor, or run-as-user.
> https://github.com/twosigma/waiter/pull/1313#discussion_r602824305

Newly created/modified tokens always have an owner. A token with an editor, will not have a `null` owner.

waiter/src/waiter/token.clj
shamsimam: Responded here: https://github.com/twosigma/waiter/pull/1313#discussion_r603327556

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,65 @@
...
-                          {:status http-400-bad-request
-                           :token-metadata new-token-metadata
-                           :log-level :warn}))))
+                               :log-level :warn}))))
+          ;; new token creation
+          (when-not (authz/run-as? entitlement-manager authenticated-user owner)
+            (throw (ex-info "Cannot create token as user"
+                            {:authenticated-user authenticated-user
+                             :owner owner
+                             :privileges {:editor? editor? :owner? owner?}
```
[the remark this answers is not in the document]
shamsimam: Removed the `privileges` section from the map to return to the state where we were before introducing editor changes.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,65 @@
                            :log-level :warn}))))
 
       nil
-      (do
-        (when (and run-as-user (not= "*" run-as-user))
+      (let [existing-editor (get existing-token-metadata "editor")
+            existing-owner (get existing-token-metadata "owner")
+            creating-token? (empty? existing-token-metadata)
+            owner? (and existing-owner
```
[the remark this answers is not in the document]
shamsimam: Renamed to `existing-owner?` to avoid confusion.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,64 @@
                            :log-level :warn}))))
 
       nil
-      (do
-        (when (and run-as-user (not= "*" run-as-user))
+      (let [existing-editor (get existing-token-metadata "editor")
+            existing-owner (get existing-token-metadata "owner")
+            creating-token? (empty? existing-token-metadata)
+            existing-owner? (and existing-owner
```
sradack, line 522: Nit-pick: One might confuse the meaning of `existing-owner?` with the meaning "token has an existing owner".  Here it means "token has an existing owner and the authenticated-user can be that owner"

sradack (changes requested):

waiter/src/waiter/token.clj
shamsimam: Changed to `current-owner?`

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,64 @@
...
+                                 (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata))
+            editor? (and (not creating-token?)
```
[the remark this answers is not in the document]
shamsimam: Renamed to `editing?`

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,64 @@
...
           (when-not (authz/run-as? entitlement-manager authenticated-user run-as-user)
             (throw (ex-info "Cannot run as user"
                             {:authenticated-user authenticated-user
                              :run-as-user run-as-user
                              :status http-403-forbidden
                              :log-level :warn}))))
-        (let [existing-service-description-owner (get existing-token-metadata "owner")]
-          (if-not (str/blank? existing-service-description-owner)
-            (when-not (authz/manage-token? entitlement-manager authenticated-user token existing-token-metadata)
+        (if-not creating-token?
```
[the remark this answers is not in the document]
shamsimam: Swapped the clauses, I hadn't done so earlier to make the diff easier to read.

waiter/src/waiter/token.clj
```diff
@@ -516,48 +516,64 @@
...
+          ;; editing token
+          (let [delegated-user (or (when editor? existing-owner) authenticated-user)]
+            (when-not (authz/manage-token? entitlement-manager delegated-user token existing-token-metadata)
               (throw (ex-info "Cannot change owner of token"
```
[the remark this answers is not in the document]
shamsimam: Changed to `Cannot update token`

sradack (approved):

PR #1318, opened 2021-03-30
Drop managedFields from k8s objects

DaoWen:

Drop managedFields metadata from k8s objects

The managedFields metadata is not useful to us, and it's very verbose.

https://github.com/kubernetes/kubernetes/issues/90066#issuecomment-626828512


kevo1ution (approved):

DaoWen:
Note that `dissoc-in` is a no-op if the target key does not exist.

PR #1320, opened 2021-03-30
Cli Fix SSH Subcommand not getting instances of service from multiple clusters

kevo1ution:

- There was a wrong assumption that you can't the same service id exist in multiple clusters, so the cli would pick the first cluster that responded 200 on `/apps/<service-id>`.
- concatenate instances from all the clusters that provide information for the `service-id`
- when choosing a `service-id` (i.e. `waiter ssh token`), set the possible clusters to the exact cluster that `service-id` choice corresponds to.

### DIFF
--- a/cli/integration/tests/waiter/test_cli_multi_cluster.py
+++ b/cli/integration/tests/waiter/test_cli_multi_cluster.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import unittest
 import uuid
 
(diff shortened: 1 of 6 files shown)


cli/integration/tests/waiter/test_cli_multi_cluster.py
```diff
@@ -519,3 +520,38 @@ def test_federated_tokens(self):
...
 
+    def test_ssh_token_service_id_on_multiple_clusters(self):
+        token_name = self.token_name()
+        service1_desc = util.minimal_service_description(cluster=self.waiter_1_cluster)
+        service2_desc = {**service1_desc, 'cluster': self.waiter_2_cluster}
+        util.post_token(self.waiter_url_1, token_name, service1_desc)
+        util.post_token(self.waiter_url_2, token_name, service2_desc)
+        try:
+            # service ids should be the same as their service descriptions are the same
+            service1_id = util.ping_token(self.waiter_url_1, token_name)
```
shamsimam, line 531: We can retrieve service_id without pinging the token.
We should add assertions for the scenario when there no active instances and also when only one of the clusters has active instances.

kevo1ution: > We can retrieve service_id without pinging the token.
How can we determine the service_id without pinging the token? Won't the service get created when there are new requests queued for the token. Also, there doesn't seem like a util function for doing the service-id calculation in the cli tests locally.

shamsimam: There is `/service-id` endpoint that returns the service ID given the appropriate request headers. The Waiter integration tests use it quite a bit.

shamsimam (approved):

PR #1324, opened 2021-04-05
Add more token watch logs and correlate them throughout event flow

kevo1ution:

- internal index event uses cid (this comes from inter router request, or local token edit request)
- token-watch-maintainer uses the cid from the internal index event for logging and puts the cid in the outer event object
- token-watch-maintainer creates a cid when doing refresh events, or initial event
- handler-lists-tokens-watch only logs the id before processing and logs extra event information when forwarding the event to a client

- This gives better visibility into what tokens are in what events
- correlates the initial token edit request with the event being sent to clients makes it easier to relate logs while debugging

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1371,7 +1371,7 @@
                               [:state clock kv-store]]
                              (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)]
                                (token-watch/start-token-watch-maintainer
-                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan utils/unique-identifier)))})
+                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan)))})
 
 (def request-handlers
   {:app-name-handler-fn (pc/fnk [service-id-handler-fn]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1369,9 +1369,11 @@
                                 [:tokens channels-update-chan-buffer-size tokens-update-chan-buffer-size
                                  watch-refresh-timeout-ms]]]
                               [:state clock kv-store]]
-                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)]
+                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)
+                                   cid-factory-fn (fn create-watch-cid []
+                                                    (str "token-watch-maintainer" "." (utils/unique-identifier)))]
                                (token-watch/start-token-watch-maintainer
-                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan)))})
+                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan cid-factory-fn)))})
 
 (def request-handlers
   {:app-name-handler-fn (pc/fnk [service-id-handler-fn]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1369,11 +1369,9 @@
                                 [:tokens channels-update-chan-buffer-size tokens-update-chan-buffer-size
                                  watch-refresh-timeout-ms]]]
                               [:state clock kv-store]]
-                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)
-                                   cid-factory-fn (fn create-watch-cid []
-                                                    (str "token-watch-maintainer" "." (utils/unique-identifier)))]
+                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)]
                                (token-watch/start-token-watch-maintainer
-                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan cid-factory-fn)))})
+                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan utils/unique-identifier)))})
 
 (def request-handlers
   {:app-name-handler-fn (pc/fnk [service-id-handler-fn]
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -674,9 +674,8 @@
             watch-chan-xform
             (comp
               (map
-                (fn event-filter [{:keys [object type] :as event}]
-                  (cid/cinfo correlation-id "received event from token-watch-maintainer daemon" {:type (:type event)
-                                                                                                 :token-sample (first (:object event))})
+                (fn event-filter [{:keys [id object type] :as event}]
+                  (cid/cinfo correlation-id "received event from token-watch-maintainer daemon" {:id id})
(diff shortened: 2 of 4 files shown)


waiter/src/waiter/token.clj
```diff
@@ -80,7 +80,7 @@
   "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
   [tokens-update-chan token]
   (log/info "sending internal index event" {:token token})
-  (async/put! tokens-update-chan {:token token}))
+  (async/put! tokens-update-chan {:token token :x-cid (cid/get-correlation-id)}))
```
shamsimam, line 83: We use `:cid` in other places, e.g. the functions in `process-request`

waiter/src/waiter/token_watch.clj
```diff
@@ -7,12 +7,15 @@
...
+            [waiter.util.utils :as utils]))
 
 (defn make-index-event
   "Create an event for watch endpoints"
-  [type object]
-  {:object object :type type})
+  [type object & {:keys [id]}]
+  (cond->
+    {:object object :type type}
+    id (assoc :id id)))
```
shamsimam, line 18: Every event should have an id, so this should not be conditional.

waiter/src/waiter/token_watch.clj
```diff
@@ -36,7 +39,7 @@
 (defn start-token-watch-maintainer
   "Starts daemon thread that maintains token watches and process/filters internal token events to be streamed to
   clients through the watch handlers. Returns map of various channels and state functions to control the daemon."
-  [kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan]
+  [kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan cid-factory-fn]
```
shamsimam, line 42: Instead of passing a `cid-factory-fn` we can construct the CID internally in the code here using an iteration index or a timestamp. Not sure how passing the CID factory is helping.

waiter/test/waiter/token_watch_test.clj
```diff
@@ -22,7 +22,8 @@
             [waiter.status-codes :refer :all]
             [waiter.test-helpers :refer :all]
             [waiter.token :refer :all]
-            [waiter.token-watch :refer :all])
+            [waiter.token-watch :refer :all]
+            [waiter.correlation-id :as cid])
```
shamsimam, line 26: Nitpick: lexicographic order

waiter/src/waiter/token_watch.clj
```diff
@@ -79,72 +82,79 @@
...
+                                        (if (some? token-index-entry)
+                                          ; If index-entry retrieved from kv-store exists then treat as UPDATE
+                                          ; (includes token creation and soft deletion)
+                                          [(make-index-event :UPDATE token-index-entry)
+                                           (assoc-in current-state [:token->index token] token-index-entry)]
+                                          ; index-entry doesn't exist then treat as DELETE
+                                          [(make-index-event :DELETE {:token token})
+                                           (assoc current-state :token->index (dissoc token->index token))])
+                                        _ (log/info "token-watch-maintainer sending a token event to watches" {:event index-event})
+                                        open-chans (->> (make-index-event :EVENTS [index-event] :id x-cid)
```
shamsimam, line 104: CIDs are not expected to be unique, I think we want the events to have their own unique identifier, e.g. by invoking `utils/unique-identifier`. We can link external CIDs to the id via the info log above.

waiter/src/waiter/core.clj
```diff
@@ -1369,9 +1369,11 @@
                                 [:tokens channels-update-chan-buffer-size tokens-update-chan-buffer-size
                                  watch-refresh-timeout-ms]]]
                               [:state clock kv-store]]
-                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)]
+                             (let [watch-refresh-timer-chan (au/timer-chan watch-refresh-timeout-ms)
+                                   cid-factory-fn (fn create-watch-cid []
+                                                    (str "token-watch-maintainer" "." (utils/unique-identifier)))]
                                (token-watch/start-token-watch-maintainer
-                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan)))})
+                                 kv-store clock tokens-update-chan-buffer-size channels-update-chan-buffer-size watch-refresh-timer-chan cid-factory-fn)))})
```
shamsimam, line 1376: I don't think passing a `cid-factory-fn` is necessary, users can invoke `utils/unique-identifier` directly.

waiter/src/waiter/token_watch.clj
kevo1ution: For aggregate events (e.g. on refresh), this may be repetitive because all the ids will be the same. The current state only has top level events having an id guaranteed (INITIAL and EVENTS)

waiter/src/waiter/token_watch.clj
kevo1ution: I was doing this originally to make it easier to assert events in unit tests. The ids are random and I'm not sure if there is a way to get the id value that utils/unique-identifier will return to make the test assertions

shamsimam: You can mock functions with `with-redefs` for unit tests.

waiter/src/waiter/token_watch.clj
shamsimam: Okay.

waiter/src/waiter/token_watch.clj
kevo1ution: start-token-watch-maintainer calls the `utils/unique-identifier` in an `async/go` block, which the `with-redef` can't mock functions. I can set a variable to equal the function outside the async go block and call that var which will be changed by the `with-redef`

shamsimam: Is asserting on the exact value of the id important, or just asserting its presence sufficient?

kevo1ution: asserting the presence should be sufficient now that we are solely using unique ids for the event id instead of a request cid provided to us

waiter/src/waiter/token.clj
```diff
@@ -80,7 +80,7 @@
   "Send an internal event to be processed by the tokens-watch-maintainer daemon process"
   [tokens-update-chan token]
   (log/info "sending internal index event" {:token token})
-  (async/put! tokens-update-chan {:token token}))
+  (async/put! tokens-update-chan {:token token :cid (cid/get-correlation-id)}))
```
shamsimam, line 83: Nitpick: lexicographic order

waiter/src/waiter/token_watch.clj
```diff
@@ -79,72 +83,79 @@
...
+                                  (let [[index-event next-state]
+                                        (if (some? token-index-entry)
+                                          ; If index-entry retrieved from kv-store exists then treat as UPDATE
+                                          ; (includes token creation and soft deletion)
+                                          [(make-index-event :UPDATE token-index-entry)
+                                           (assoc-in current-state [:token->index token] token-index-entry)]
+                                          ; index-entry doesn't exist then treat as DELETE
+                                          [(make-index-event :DELETE {:token token})
+                                           (assoc current-state :token->index (dissoc token->index token))])
+                                        _ (log/info "token-watch-maintainer sending a token event to watches" {:event index-event})
```
shamsimam, line 104: Nitpick: maybe drop `token-watch-maintainer` in the log message, the CID already has it

shamsimam (approved):

shamsimam (approved):

PR #1326, opened 2021-04-08
avoid reading entire request body into memory

DaoWen:

avoid reading entire request body into memory

avoid allocating huge strings for request bodies that we don't actually use

### DIFF
--- a/waiter/src/waiter/main.clj
+++ b/waiter/src/waiter/main.clj
@@ -69,7 +69,11 @@
                  (not (instance? ManyToManyChannel response))
                  (not (hu/http2? internal-protocol)))
         (try
-          (slurp body)
+          (let [body-stream ^ServletInputStream body
+                buffer (byte-array 1024)]
+            ;; read until the request payload is fully consumed
+            (while (not (.isFinished body-stream))
+              (.read body-stream buffer)))
           (catch IOException e
             (log/error e "Unable to consume request stream"))))
       response)))


waiter/src/waiter/main.clj
```diff
@@ -69,7 +69,11 @@
                  (not (instance? ManyToManyChannel response))
                  (not (hu/http2? internal-protocol)))
         (try
-          (slurp body)
+          (let [body-stream ^ServletInputStream body
+                buffer (byte-array 1024)]
+            ;; read until the request payload is fully consumed
+            (while (not (.isFinished body-stream))
+              (.read body-stream buffer)))
```
shamsimam, line 76: Let's add a timer around this body and also get a meter for the number of iterations we run to consume the body.

shamsimam: We should also report request bytes streamed like we do at:
https://github.com/twosigma/waiter/blob/b0667031c7356822e792cf9132cb662344ad2fea/waiter/src/waiter/process_request.clj#L320-L336

shamsimam: Instead of this loop, we need to read the bytes asynchronously to prevent the Jetty threads from being blocked:
https://github.com/twosigma/waiter/blob/b0667031c7356822e792cf9132cb662344ad2fea/waiter/src/waiter/process_request.clj#L374-L401

DaoWen:
Superseded by #1327

PR #1327, opened 2021-04-08
consumes request stream asynchronously with reduced young gen memory pressure

shamsimam:

- consumes request stream asynchronously with reduced young gen memory pressure

The existing `slurp` is bad on two counts:
- uses blocking reads to read from the input stream
- generates a lot of short-lived objects and creates an unused string from the bytes consumed from the input stream

### DIFF
--- a/waiter/src/waiter/main.clj
+++ b/waiter/src/waiter/main.clj
@@ -16,6 +16,7 @@
 (ns waiter.main
   (:gen-class)
   (:require [clj-time.core :as t]
+            [clojure.core.async :as async]
             [clojure.java.io :as io]
             [clojure.java.shell :as sh]
             [clojure.string :as str]
@@ -23,6 +24,7 @@
             [clojure.tools.logging :as log]
             [metrics.core :as mc]
             [metrics.jvm.core :as jvm-metrics]
+            [metrics.meters :as meters]
             [plumbing.core :as pc]
             [plumbing.graph :as graph]
             [qbits.jet.server :as server]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/main.clj
```diff
@@ -64,15 +65,47 @@
...
-      response)))
+        (let [response-ch (async/promise-chan)
+              input-stream ^ServletInputStream body
+              bytes-counter-atom (atom 0)
+              throughput-meter-global (metrics/waiter-meter "streaming" "request-bytes")
+              throughput-iterations-meter-global (metrics/waiter-meter "streaming" "request-iterations")
+              complete-request-streaming (fn complete-request-streaming [throwable]
+                                           (if throwable
+                                             (log/error throwable "error after consuming" @bytes-counter-atom "bytes from request stream")
+                                             (log/info "successfully consumed" @bytes-counter-atom "bytes from request stream"))
```
DaoWen, line 80: How do we know which request these log lines are talking about? Don't we need the cid?

DaoWen:
How hard would it be to add an integration test for this?

It seems like we could force a 400 error response from Waiter by submitting a bad token or something, have a streaming request body, send part of the request body, sleep for a couple seconds, assert that the response hasn't come yet, send the rest of the body, then assert that the response comes back.

waiter/src/waiter/main.clj
shamsimam: Will add the CID.

DaoWen (approved):

PR #1328, opened 2021-04-08
makes draining request bytes configurable

shamsimam:

- makes draining request bytes configurable

Draining request bytes which we are going to discard is wasteful. Discarding such request bytes is actually preferable as it avoids keeping the server busy doing wasted work.


sradack (approved):

shamsimam:
Closing as some clients do not tolerate streaming response before consuming request body.

PR #1333, opened 2021-04-14
log rotation every hour

kevo1ution:

- Set log rotation to every hour
- changes are based on these docs https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/DailyRollingFileAppender.html

- Many request logs cause logs to use up lots of disk space.

### DIFF
--- a/waiter/resources/log4j.properties
+++ b/waiter/resources/log4j.properties
@@ -6,7 +6,7 @@ log4j.category.waiter=INFO
 log4j.appender.InfoAppender=org.apache.log4j.DailyRollingFileAppender
 log4j.appender.InfoAppender.Threshold=INFO
 log4j.appender.InfoAppender.File=log/${waiter.logFilePrefix}waiter.log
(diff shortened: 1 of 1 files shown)


waiter/resources/log4j.properties
```diff
@@ -6,15 +6,15 @@ log4j.category.waiter=INFO
 log4j.appender.InfoAppender=org.apache.log4j.DailyRollingFileAppender
 log4j.appender.InfoAppender.Threshold=INFO
 log4j.appender.InfoAppender.File=log/${waiter.logFilePrefix}waiter.log
-log4j.appender.InfoAppender.DatePattern='.'yyyy-MM-dd
```
shamsimam, line 9: Should it be only `HH` for hourly?

kevo1ution: Yep, doing on minutes was for testing purposes

shamsimam (approved):

PR #1337, opened 2021-04-14
always includes service-id in service-state for waiter-ping output

shamsimam:

- always includes service-id in service-state for waiter-ping output

We would like to identify the service by service-id from ping output.


DaoWen (approved):

DaoWen:
Approved. Feel free to merge once the build is green.

PR #1339, opened 2021-04-18
tags waiter error responses with an error class for the request-log

shamsimam:

- tags waiter error responses with an error class for the request-log

Allows identifying sources of various error responses from the `waiter-error-class` in the request log.

### DIFF
--- a/waiter/src/waiter/auth/spnego.clj
+++ b/waiter/src/waiter/auth/spnego.clj
@@ -70,6 +70,7 @@
   [request]
   (log/info "triggering 401 negotiate for spnego authentication")
   (-> (response-http-401-unauthorized request "for negotiation")
+    (assoc :error-class "waiter.KerberosNegotiate")
     (assoc-in [:headers "www-authenticate"] (str/trim negotiate-prefix))))
 
 (defn response-http-401-unauthorized-spnego-disabled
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/auth/spnego.clj
```diff
@@ -84,7 +85,8 @@
   (log/info "triggering 503 unavailable for spnego authentication")
   (counters/inc! (metrics/waiter-counter "core" "response-status" "503"))
   (meters/mark! (metrics/waiter-meter "core" "response-status-rate" "503"))
-  (-> {:message "Too many Kerberos authentication requests"
+  (-> {:details {:error-class "waiter.KerberosQueueLength"}
```
kevo1ution, line 88: Should these error classes be defined constants. For example, the "http-503-service-unavailable" is defined in status_codes.clj?

shamsimam: Good idea, will change.

shamsimam: The status codes file is for well-defined status codes, so didn't put these in that file.
Instead, I introduced constants in each file for the error classes referenced in each file.

kevo1ution (approved):

PR #1341, opened 2021-04-21
Event-maintainer for sending external events

kevo1ution:

- adds a components that allows tracking failed instances across all services

We would like to configure custom logic around discovery of new failed instances, e.g. notification to interested parties.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1043,13 +1043,13 @@
  ;; Configure event maintainer that handles internal events to be sent externally
  :instance-tracker-config {
                            ;; component that handles new instance failures
-                           {:instance-failure-handler {:kind :default
-                                                       :default {:factory-fn 'waiter.instance-tracker/create-instance-failure-event-handler
-                                                                 :config {
-                                                                          ;; configuration for recent-failed-instance-cache
-                                                                          ;; the cache is used for integration and debugging purposes
-                                                                          :recent-failed-instance-cache {:threshold 5000
-                                                                                                         :ttl 900000}}}}}}
+                           :instance-failure-handler {:kind :default
+                                                      :default {:factory-fn 'waiter.instance-tracker/create-instance-failure-event-handler
+                                                                :config {
+                                                                         ;; configuration for recent-failed-instance-cache
+                                                                         ;; the cache is used for integration and debugging purposes
+                                                                         :recent-failed-instance-cache {:threshold 5000
+                                                                                                        :ttl 900000}}}}}
 
  :token-config {
                 ;; Define the calculator that compute the root assigned to a token when they are created.
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1040,6 +1040,17 @@
                        ;; flagged as never-passed-health-checks upon failure
                        :failed-check-threshold 5}
 
+ ;; Configure event maintainer that handles internal events to be sent externally
+ :instance-tracker-config {
+                           ;; component that handles new instance failures
+                           {:instance-failure-handler {:kind :default
+                                                       :default {:factory-fn 'waiter.instance-tracker/create-instance-failure-event-handler
+                                                                 :config {
+                                                                          ;; configuration for recent-failed-instance-cache
+                                                                          ;; the cache is used for integration and debugging purposes
+                                                                          :recent-failed-instance-cache {:threshold 5000
+                                                                                                         :ttl 900000}}}}}}
+
  :token-config {
                 ;; Define the calculator that compute the root assigned to a token when they are created.
                 :cluster-calculator {;; :kind :configured create a calculator that looks up the cluster name using a configured map.
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -66,13 +66,13 @@
               (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
                 (log/info "The failed instances should be tracked by the instance-tracker" {:failed-instances failed-instances})
                 (is (pos? (count failed-instances)))
-                (let [query-params "include=instance-failure-handler&include=recent-id->failed-instance-date"
+                (let [query-params "include=instance-failure-handler&include=id->failed-date"
                       {{:keys [last-update-time]
-                        {:keys [last-error-time recent-id->failed-instance-date type]} :instance-failure-handler} :state}
+                        {:keys [last-error-time id->failed-date type]} :instance-failure-handler} :state}
                       (get-instance-tracker-state router-url
                                                   :cookies cookies
                                                   :query-params query-params)
-                      default-event-handler-ids (set (keys recent-id->failed-instance-date))]
+                      default-event-handler-ids (set (keys id->failed-date))]
                   (is (t/before? start-time (du/str-to-date last-update-time)))
                   (when (= type "DefaultInstanceFailureHandler")
                     ; assert failed instances are tracked by DefaultInstanceFailureHandler cache of new failed instances
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -60,10 +60,10 @@
             (is (wait-for
                   (fn []
                     (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
-                      (< 0 (count failed-instances))))))
+                      (pos? (count failed-instances))))))
             (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
               (log/info "The failed instances should be tracked by the instance-tracker" {:failed-instances failed-instances})
-              (is (< 0 (count failed-instances)))
+              (is (pos? (count failed-instances)))
               (let [query-params "include=instance-failure-handler&include=recent-id->failed-instance-date&include=id->failed-instance"
                     {{:keys [id->failed-instance]
                       {:keys [last-error-time recent-id->failed-instance-date type]} :instance-failure-handler} :state}
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -45,7 +45,7 @@
     (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
           routers (routers waiter-url)
           router-urls (vals routers)]
-      (testing "new failing instances appear in instance-tracker state on same router"
+      (testing "new failing instances appear in instance-tracker state on all routers"
         (let [start-time (t/now)
               {:keys [service-id] :as response}
               (make-request-with-debug-info
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -42,43 +42,42 @@
 
 (deftest ^:parallel ^:integration-fast ^:resource-heavy test-instance-tracker-failing-instance
   (testing-using-waiter-url
-    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")]
-
-      ; TODO: When instance tracking becomes more reliable in multi router scenarios this test should not be
-      ; on the same router
+    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
+          routers (routers waiter-url)
+          router-urls (vals routers)]
       (testing "new failing instances appear in instance-tracker state on same router"
         (let [start-time (t/now)
-              router-url (rand-router-url waiter-url)
               {:keys [service-id] :as response}
               (make-request-with-debug-info
                 {:x-waiter-cmd "invalidcmd34sdfadsve"
                  :x-waiter-name (rand-name)}
-                #(make-kitchen-request router-url % :cookies cookies))]
+                #(make-kitchen-request waiter-url % :cookies cookies))]
           (with-service-cleanup
             service-id
             (assert-response-status response http-503-service-unavailable)
+            ; wait for all routers to have positive number of failed instances
             (is (wait-for
-                  (fn []
-                    (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
-                      (pos? (count failed-instances))))))
-            (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
-              (log/info "The failed instances should be tracked by the instance-tracker" {:failed-instances failed-instances})
-              (is (pos? (count failed-instances)))
-              (let [query-params "include=instance-failure-handler&include=recent-id->failed-instance-date&include=id->failed-instance"
-                    {{:keys [id->failed-instance]
-                      {:keys [last-error-time recent-id->failed-instance-date type]} :instance-failure-handler} :state}
-                    (get-instance-tracker-state router-url
-                                                :cookies cookies
-                                                :query-params query-params)
-                    daemon-failed-ids (set (keys id->failed-instance))
-                    default-event-handler-ids (set (keys recent-id->failed-instance-date))]
-                ; assert failed instances are tracked by daemon
-                (doseq [{:keys [id]} failed-instances]
-                  (is (contains? daemon-failed-ids
-                                 (keyword id)))
-                  ; assert failed instances are tracked by DefaultInstanceFailureHandler cache of new failed instances
+                  (fn every-router-has-failed-instances? []
+                    (every? (fn has-failed-instances? [router-url]
+                              (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
+                                (pos? (count failed-instances))))
+                            router-urls))))
+            (doseq [router-url router-urls]
+              (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
+                (log/info "The failed instances should be tracked by the instance-tracker" {:failed-instances failed-instances})
+                (is (pos? (count failed-instances)))
+                (let [query-params "include=instance-failure-handler&include=recent-id->failed-instance-date"
+                      {{:keys [last-update-time]
+                        {:keys [last-error-time recent-id->failed-instance-date type]} :instance-failure-handler} :state}
+                      (get-instance-tracker-state router-url
+                                                  :cookies cookies
+                                                  :query-params query-params)
+                      default-event-handler-ids (set (keys recent-id->failed-instance-date))]
+                  (is (t/before? start-time (du/str-to-date last-update-time)))
                   (when (= type "DefaultInstanceFailureHandler")
-                    (is (contains? default-event-handler-ids
-                                   (keyword id)))))
-                ; assert that the error time is recent
-                (is (t/before? start-time (du/str-to-date last-error-time)))))))))))
+                    ; assert failed instances are tracked by DefaultInstanceFailureHandler cache of new failed instances
+                    (doseq [{:keys [id]} failed-instances]
+                      (is (contains? default-event-handler-ids
+                                     (keyword id))))
+                    ; assert that the error time is recent
+                    (is (t/before? start-time (du/str-to-date last-error-time)))))))))))))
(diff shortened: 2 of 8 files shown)


waiter/src/waiter/state/maintainer.clj
```diff
@@ -583,18 +584,19 @@
...
-                                       :service-id->unhealthy-instances service-id->unhealthy-instances'
-                                       :service-id->expired-instances service-id->expired-instances'
-                                       :service-id->starting-instances service-id->starting-instances'
-                                       :service-id->failed-instances service-id->failed-instances'
-                                       :service-id->instance-counts service-id->instance-counts'
-                                       :service-id->deployment-error service-id->deployment-error'
-                                       :service-id->instability-issue service-id->instability-issue'
-                                       :service-id->killed-instances service-id->killed-instances'
-                                       :time scheduler-sync-time))
+                                     [(assoc loop-state
```
shamsimam, line 587: We already have `service-id->failed-instances`, I do not think this needs to change.

waiter/src/waiter/state/maintainer.clj
```diff
@@ -689,8 +699,12 @@
                                              (select-keys new-state relevant-keys))
                                    ; propagate along router-state-push-chan only when state changes
                                    (async/put! router-state-push-chan new-state))
+                                 ; handle sending events asynchronously
+                                 (when (not (empty? all-new-failed-instances'))
+                                   (log/info "propagating failed instances events to internal handler" {:new-failed-instances-count (count all-new-failed-instances')})
+                                   (async/put! event-chan {:type :instance-failure :object all-new-failed-instances'}))
```
shamsimam, line 705: Instead of pushing this out, I think we should leverage the `router-state-push-mult` and create a new go-block (what you currently call `start-event-maintainer`) that listens on state from here and does the failed instance diff calculation. The autoscaler is another example of a reader of this state.

waiter/src/waiter/core.clj
```diff
@@ -1208,6 +1211,13 @@
                                                   :resolved-factory-fn resolved-factory-fn})))
                                reporter-instance))
                            codahale-reporters))
+   :event-maintainer (pc/fnk
```
shamsimam, line 1214: I think `instance-tracker` is a better name, `event` is too generic. It will need additional configuration for a predicate to trigger the notification. By default, the trigger can be a simple logging of instance counts. The configured method can do any filtering if it wants to skip notification events.

waiter/config-full.edn
```diff
@@ -1001,6 +1001,14 @@
  ;; Components also get passed in the :kv-store-factory, :leader?-fn and :synchronize-fn for use in the factory.
  :custom-components {:in-memory-kv-store {:factory-fn waiter.kv/new-local-kv-store}}
 
+ ;; Configure event maintainer that handles internal events to be sent externally
+ :event-maintainer-config {
+                           ;; Buffer size of the event-chan where other threads can push internal events to
+                           :event-chan-buffer-size 1000
+
+                           ;; The fn that handles new instance-failure events
+                           :instance-failure-event-handler-fn 'waiter.event-handler/default-instance-failure-event-handler}
```
shamsimam, line 1010: If we listen on `router-state-push-mult` then we do not need a separate `event-chan`
```suggestion
 ;; Configure instance notifier that handles notifications for instances
 :instance-tracker-config {
                           ;; function that triggers notification, it needs access to service-id, service-description, and map of current and delat of different kinds of instances, we can start with current failed instances and delta of failed instances.
                           :notify-service-fn ...}
```

waiter/src/waiter/core.clj
```diff
@@ -632,6 +633,8 @@
                        router-id]
                 (discovery/register router-id curator name (str base-path "/" discovery-relative-path)
                                     {:host host :port (primary-port port)}))
+   :event-chan (pc/fnk [[:settings [:event-maintainer-config event-chan-buffer-size]]]
```
shamsimam, line 636: This should not be necessary to expose. The state can be exposed in a `query-state-fn` and we will use a new state endpoint to display the state.

waiter/src/waiter/event.clj
```diff
@@ -0,0 +1,81 @@
...
+(defn start-event-maintainer
+  "Starts daemon thread that ingests various internal events and routes them to an external source"
+  [clock event-chan service-id->service-description-fn instance-failure-event-handler-fn]
+  (cid/with-correlation-id
+    "event-maintainer"
+    (let [exit-chan (async/promise-chan)
+          query-chan (async/chan)
+          error-chan (async/chan)
+          state-atom (atom {:last-error-time nil
+                            :last-update-time (clock)})
```
shamsimam, line 27: Using current clock time for last update time feels wrong.

waiter/src/waiter/state/maintainer.clj
```diff
@@ -517,7 +517,7 @@
    Exposes the state of the router via a `query-state-fn` no-args function that is returned."
   [scheduler-state-chan router-chan router-id exit-chan service-id->service-description-fn
    refresh-service-descriptions-fn service-id->deployment-error-config-fn default-deployment-error-config
-   ejection-expiry-tracker]
+   ejection-expiry-tracker event-chan]
```
shamsimam, line 520: I do not think this function needs to change.

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,89 @@
...
+                        (timers/start-stop-time!
+                          (metrics/waiter-timer "core" "event-maintainer" "router-state-chan")
+                          (let [{:keys [service-id->failed-instances]} msg
+                                all-failed-instances' (reduce (fn [cur-set cur-failed-instances]
+                                                                (set/union cur-set (set cur-failed-instances)))
+                                                              #{}
+                                                              (vals service-id->failed-instances))
+                                new-failed-instances (filter (complement all-failed-instances) all-failed-instances')
+                                error-cb-fn! (fn error-cb! [errors]
+                                               (async/put! error-chan {:errors errors :instance-event-type :instance-failure}))]
```
shamsimam, line 70: Not sure we need an `error-cb-fn!`, we can handle it directly here and update state.

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,89 @@
+(ns waiter.instance-tracker
+  (:require [clojure.core.async :as async]
+            [waiter.correlation-id :as cid]
+            [clojure.tools.logging :as log]
+            [metrics.timers :as timers]
+            [waiter.metrics :as metrics]
+            [waiter.util.utils :as utils]
+            [clojure.set :as set]))
```
shamsimam, line 8: lexicographic order

waiter/src/waiter/handler.clj
```diff
@@ -810,9 +810,10 @@
       (utils/clj->streaming-json-response {:details (->> ["autoscaler" "autoscaling-multiplexer" "codahale-reporters"
                                                           "ejection-expiry" "entitlement-manager" "fallback"
                                                           "gc-broken-services" "gc-services" "gc-transient-metrics" "interstitial"
-                                                          "jwt-auth-server" "kv-store" "launch-metrics" "leader" "local-usage"
-                                                          "maintainer" "router-metrics" "scheduler" "service-description-builder"
-                                                          "service-maintainer" "statsd" "token-watch-maintainer" "work-stealing"]
+                                                          "instance-tracker" "jwt-auth-server" "kv-store" "launch-metrics" "leader"
```
shamsimam, line 813: lexicographic order, it can go on previous line and none of the other lines have to change.

waiter/src/waiter/core.clj
```diff
@@ -1911,9 +1931,9 @@
                                (fn wrap-secure-request-fn
                                  [handler]
                                  (let [handler (-> handler
-                                                 (cors/wrap-cors-request
-                                                   cors-validator waiter-request?-fn exposed-headers)
-                                                 authentication-method-wrapper-fn)]
+                                                   (cors/wrap-cors-request
+                                                     cors-validator waiter-request?-fn exposed-headers)
+                                                   authentication-method-wrapper-fn)]
```
shamsimam, line 1936: lots of unrelated whitespace changes :(

waiter/config-full.edn
```diff
@@ -1001,6 +1001,14 @@
  ;; Components also get passed in the :kv-store-factory, :leader?-fn and :synchronize-fn for use in the factory.
  :custom-components {:in-memory-kv-store {:factory-fn waiter.kv/new-local-kv-store}}
 
+ ;; Configure event maintainer that handles internal events to be sent externally
+ :event-maintainer-config {
```
shamsimam, line 1005: outdated name and configuration

waiter/config-full.edn
```diff
@@ -1001,6 +1001,14 @@
...
+                           ;; Buffer size of the event-chan where other threads can push internal events to
+                           :event-chan-buffer-size 1000
+
+                           ;; The fn that handles new instance-failure events
```
shamsimam, line 1009: We should document the signature. I'm also wondering if this should be a component with an interface (i.e. defprotocol) so it can be initialized with some nested configuration, e.g. how we use `kind` and initialize other components. E.g. a components that will make http-calls may need additional configuration.

waiter/src/waiter/core.clj
```diff
@@ -1228,6 +1230,17 @@
...
                                  metrics-gc-chans))
+   :instance-tracker (pc/fnk
+                       [[:routines service-id->service-description-fn]
+                        [:settings [:instance-tracker-config instance-failure-event-handler-fn!]]
+                        [:state clock]
+                        router-state-maintainer]
+                       (let [instance-failure-event-handler-fn! (-> instance-failure-event-handler-fn! utils/resolve-symbol!)
+                             {{:keys [router-state-push-mult]} :maintainer} router-state-maintainer
+                             router-state-chan (au/latest-chan)]
+                         (async/tap router-state-push-mult router-state-chan)
```
shamsimam, line 1241:
```suggestion
                             router-state-chan (async/tap router-state-push-mult (au/latest-chan))]
```

shamsimam (changes requested):

waiter/src/waiter/core.clj
```diff
@@ -1228,6 +1230,17 @@
...
+                       [[:routines service-id->service-description-fn]
+                        [:settings [:instance-tracker-config instance-failure-handler]]
+                        [:state clock]
+                        router-state-maintainer]
+                       (let [{{:keys [router-state-push-mult]} :maintainer} router-state-maintainer
+                             instance-failure-context {:clock clock :service-id->service-description-fn service-id->service-description-fn}
+                             instance-failure-handler-component (utils/create-component instance-failure-handler
+                                                                                        :context instance-failure-context)]
+                         (instance-tracker/start-instance-tracker
+                           clock (async/tap router-state-push-mult (au/latest-chan)) instance-failure-handler-component)))
```
shamsimam, line 1243: Nitpick: move the `tap` output to the `let` block.

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,99 @@
...
+  (state [this include-flags]
+    "returns the state of the handler:
+    {:last-error-time}"))
+
+(defrecord DefaultInstanceFailureHandler [clock
+                                          handler-state]
+
+  InstanceEventHandler
+
+  (handle-instances-event! [this instances]
```
shamsimam, line 25: `instances` should instead be a map and we can currently populate `:new-failed-instances` and `:all-failed-instances`

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,99 @@
...
+    (let [{:keys [clock handler-state]} this]
+      (log/info "newly failed instances" {:new-failed-instances instances})
+      (reset! handler-state {:last-error-time (clock)})))
```
shamsimam, line 28: Nice, this lets us write integration tests.
It would be good to also track a specific instance, maybe cache X most recent failed instance IDs mapped by timestamp of when we first learnt about it.

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,99 @@
...
+                          (let [{:keys [service-id->failed-instances]} msg
+                                all-failed-instances' (reduce (fn [cur-set cur-failed-instances]
+                                                                (set/union cur-set (set cur-failed-instances)))
+                                                              #{}
+                                                              (vals service-id->failed-instances))
+                                new-failed-instances (filter (complement all-failed-instances) all-failed-instances')]
+                            (when (not (empty? new-failed-instances))
+                              (log/info "new failed instances" {:new-failed-instances (map :id new-failed-instances)})
+                              (handle-instances-event! instance-failure-handler-component new-failed-instances))
+                            (assoc current-state :all-failed-instances all-failed-instances')))
```
shamsimam, line 84: It'd be nice to add a timestamp here too, `:last-update-time`

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,99 @@
...
+            (let [{:keys [all-failed-instances last-update-time]} @state-atom]
+              (cond-> {:last-update-time last-update-time}
+                      (contains? include-flags "all-failed-instances")
+                      (assoc :all-failed-instances all-failed-instances)
+                      (contains? include-flags "instance-failure-handler")
+                      (assoc :instance-failure-handler (state instance-failure-handler-component [])))))
+          go-chan
+          (async/go
+            (try
+              (loop [{:keys [all-failed-instances] :as current-state} @state-atom]
```
shamsimam, line 61: `all-failed-instances` can grow very large, there should be a way to clear out old entries which we are no longer care about.

kevo1ution: all failed instances should reset to the failed instances that are tracked by state-maintainer which caps 10 failed instances per service. So we should not be tracking any more instances than what is already tracked in the state-maintainer. Is this acceptable?

kevo1ution: The problem with comparing `recent-failed-instances` and the `service-id->failed-instances` provided with by the state-maintainer is that the tracker might think some failing instances are new but they are really an hour older

waiter/src/waiter/instance_tracker.clj
kevo1ution: last-update-time get changed during `recur` so it gets updated at every iteration of the loop

waiter/src/waiter/instance_tracker.clj
kevo1ution: good idea

waiter/src/waiter/util/http_utils.clj
```diff
@@ -40,7 +40,7 @@
 
 (defn spnego-authentication
   "Returns an Authentication$Result for endpoint which will use SPNEGO to generate an Authorization header"
-  [^URI endpoint]
+  [^URI endpoint & {:keys [proxy-auth] :or {proxy-auth false}}]
```
shamsimam, line 43: Nitpick: we can promote `proxy-auth?` (with question mark) to a required parameter.

shamsimam (approved):

waiter/src/waiter/core.clj
```diff
@@ -42,6 +42,7 @@
             [waiter.curator :as curator]
             [waiter.descriptor :as descriptor]
             [waiter.discovery :as discovery]
+            [waiter.instance-tracker :as instance-tracker]
```
shamsimam, line 45: lexicographic order

waiter/src/waiter/instance_tracker.clj
shamsimam: I missed that we are resetting the value to instances from `service-id->failed-instances`. Failed instances cannot grow indefinitely large.

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,99 @@
...
+
+                        router-state-chan
+                        (timers/start-stop-time!
+                          (metrics/waiter-timer "core" "instance-tracker" "router-state-chan")
+                          (let [{:keys [service-id->failed-instances]} msg
+                                all-failed-instances' (reduce (fn [cur-set cur-failed-instances]
+                                                                (set/union cur-set (set cur-failed-instances)))
+                                                              #{}
+                                                              (vals service-id->failed-instances))
+                                new-failed-instances (filter (complement all-failed-instances) all-failed-instances')]
```
shamsimam, line 80: Should we compare failed instances by their ids in case other fields have changed?

waiter/src/waiter/util/http_utils.clj
```diff
@@ -38,9 +38,8 @@
 
 (def ^Base64 base64 (Base64.))
 
-(defn spnego-authentication
-  "Returns an Authentication$Result for endpoint which will use SPNEGO to generate an Authorization header"
-  [^URI endpoint]
+(defn create-spnego-authentication-result-factory
```
shamsimam, line 41: Missing docstring

waiter/src/waiter/util/http_utils.clj
```diff
@@ -38,9 +38,8 @@
...
+  [^URI endpoint proxy-auth?]
```
shamsimam, line 42: I think we should pass the authorization header name instead of `proxy-auth?` boolean value from the caller and avoid the if-else branching in the code below.

waiter/src/waiter/instance_tracker.clj
kevo1ution: why do we need to provide all failed instances to the handler? The handler will only process the newly failed instances

waiter/config-full.edn
```diff
@@ -1040,6 +1040,17 @@
                        ;; flagged as never-passed-health-checks upon failure
                        :failed-check-threshold 5}
 
+ ;; Configure event maintainer that handles internal events to be sent externally
+ :instance-tracker-config {
```
shamsimam, line 1044: I think we need to add the default configuration in settings.clj

waiter/src/waiter/instance_tracker.clj
```diff
@@ -0,0 +1,125 @@
...
+      (doseq [inst new-failed-instances]
+        (cu/cache-put! recent-id->failed-instance-date-cache (:id inst) (clock)))))
+
+  (state [this include-flags]
+    (let [{:keys [handler-state]} this
+          {:keys [last-error-time recent-id->failed-instance-date-cache]} @handler-state]
+      (cond-> {:last-error-time last-error-time
+               :supported-include-params ["recent-id->failed-instance-date"]
+               :type "DefaultInstanceFailureHandler"}
+              (contains? include-flags "recent-id->failed-instance-date")
```
shamsimam, line 39: Nitpick: use a shorter name.

waiter/integration/waiter/instance_tracker_integration_test.clj
```diff
@@ -0,0 +1,80 @@
...
+                             default-inst-event-handler-state-fields)))))))
+
+(deftest ^:parallel ^:integration-fast ^:resource-heavy test-instance-tracker-failing-instance
+  (testing-using-waiter-url
+    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")]
+
+      ; TODO: When instance tracking becomes more reliable in multi router scenarios this test should not be
+      ; on the same router
+      (testing "new failing instances appear in instance-tracker state on same router"
+        (let [router-url (rand-router-url waiter-url)
```
shamsimam: we should check on all routers.

kitchen has a header that causes a healthy instance to die, we could be using that too.

waiter/integration/waiter/instance_tracker_integration_test.clj
```diff
@@ -0,0 +1,80 @@
...
+          (with-service-cleanup
+            service-id
+            (assert-response-status response http-503-service-unavailable)
+            (is (wait-for
+                  (fn []
+                    (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
+                      (< 0 (count failed-instances))))))
+            (let [{:keys [failed-instances]} (:instances (service-settings router-url service-id :cookies cookies))]
+              (log/info "The failed instances should be tracked by the instance-tracker" {:failed-instances failed-instances})
+              (is (< 0 (count failed-instances)))
```
shamsimam, line 66: `pos?` instead of `< 0`

shamsimam (approved):

PR #1344, opened 2021-04-29
add async version of exponential backoff retry strategy

kevo1ution:

- adds async version of exponential backoff retry strategy

- we are planning to use this in future request calls

### DIFF
--- a/waiter/src/waiter/util/utils.clj
+++ b/waiter/src/waiter/util/utils.clj
@@ -24,12 +24,12 @@
             [clojure.walk :as walk]
             [comb.template :as template]
             [digest]
+            [full.async :as fa]
             [taoensso.nippy :as nippy]
             [taoensso.nippy.compression :as compression]
             [waiter.status-codes :refer :all]
             [waiter.util.date-utils :as du]
-            [waiter.util.http-utils :as hu]
-            [full.async :as fa])
+            [waiter.util.http-utils :as hu])
   (:import (clojure.core.async.impl.channels ManyToManyChannel)
            (clojure.lang ExceptionInfo)
            (java.io OutputStreamWriter)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/util/utils.clj
```diff
@@ -28,7 +28,8 @@
             [taoensso.nippy.compression :as compression]
             [waiter.status-codes :refer :all]
             [waiter.util.date-utils :as du]
-            [waiter.util.http-utils :as hu])
+            [waiter.util.http-utils :as hu]
+            [full.async :as fa])
```
shamsimam, line 32: lexicographic order :(

waiter/src/waiter/util/utils.clj
```diff
@@ -527,6 +528,36 @@
...
+   `initial-delay-ms` the initial delay for the first retry.
+   `max-delay-ms` the delay cap for exponential backoff delay.
+   `max-retries`  limit the number of retries.
+   "
+  [{:keys [delay-multiplier initial-delay-ms max-delay-ms max-retries]
+    :or {delay-multiplier 1.0
+         initial-delay-ms 100
+         max-delay-ms 300000 ; 300k millis = 5 minutes
+         max-retries 10}}]
+  (fn [body-function]
```
shamsimam, line 545: Nitpick: This function forces the body to have to launch its own `async/go` block and do error handling to return an `Exception` during failure. We should mention this clearly in the documentation.

waiter/src/waiter/util/utils.clj
kevo1ution: Is there a good plugin that does this?

shamsimam: I used to use: https://plugins.jetbrains.com/plugin/12731-amethyst
But it is not compatible with my IntelliJ anymore :(

waiter/src/waiter/util/utils.clj
```diff
@@ -527,6 +528,39 @@
...
+    (async/go-loop [num-tries 1
+                    current-delay-ms initial-delay-ms]
+      (let [{:keys [success result]}
+            (try
+              {:success true, :result (fa/<? (body-function))}
```
shamsimam, line 553: I think we can read the result and check for it being an instance of a `Throwable` instead of throwing and catching the exception.

waiter/test/waiter/util/utils_test.clj
```diff
@@ -27,7 +27,8 @@
             [waiter.test-helpers :refer :all]
             [waiter.util.cache-utils :refer :all]
             [waiter.util.date-utils :refer :all]
-            [waiter.util.utils :refer :all])
+            [waiter.util.utils :refer :all]
+            [full.async :as fa])
```
shamsimam, line 31: lexicographic order

shamsimam (approved):

PR #1348, opened 2021-05-04
adds support for unique services per token

shamsimam:

- adds support for including token in service env when unique services are requested

We would like an option to avoid sharing services among tokens.


waiter/src/waiter/service_description.clj
```diff
@@ -996,7 +1008,11 @@
   [attach-service-defaults-fn attach-token-defaults-fn token-sequence token->token-data]
   (let [merged-token-data (attach-token-defaults-fn
                             (token-sequence->merged-data token->token-data token-sequence))
-        service-description-template (select-keys merged-token-data service-parameter-keys)
+        exclusive-mode? (= "exclusive" (get merged-token-data "service-mapping"))
+        service-description-template (cond-> (select-keys merged-token-data service-parameter-keys)
+                                       ;; each unique token permutation will create a unique service
+                                       ;; leverage WAITER_CONFIG_ prefixed environment variables being allowed
```
DaoWen, line 1014: I don't particularly like this hack (plumbing the token through in this quasi-reserved environment variable, and then dropping it later) — but I'm not sure it's worth the effort right now to figure out another way to propagate the token value. Since the `WAITER_CONFIG_` values should be considered quasi-reserved anyway, I don't think it would be a problem if we came back and decided to clean this up later (i.e., it wouldn't break any contracts that users should be depending on) for the same reason that reserving it for internal use in this PR is OK.

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1864,3 +1864,101 @@
...
+                  (assert-response-status response http-200-ok)
+                  (is (= service-id-1 (:service-id response)))
+                  (is (= service-id-1 (-> body (try-parse-json) (get "WAITER_SERVICE_ID"))) (str body))
+                  (is (nil? (-> body (try-parse-json) (get "WAITER_TOKEN"))) (str body))))))
+          (finally
+            (delete-token-and-assert waiter-url token-2)))
+        (finally
+          (delete-token-and-assert waiter-url token-1))))))
+
+(deftest ^:parallel ^:integration-fast test-service-mapping-exclusive
```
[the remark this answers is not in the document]
shamsimam: The current PR does not disallow multi-token support, I'll add an integration test and code for disallowing it.

waiter/src/waiter/service_description.clj
shamsimam: An alternate is to replace the `service-mapping` value with the token value. Would you prefer that approach?

shamsimam: Also, I can avoid dropping it and sticking to the `WAITER_CONFIG_TOKEN` name, but I prefer we use the `WAITER_TOKEN` environment variable.

kevo1ution (dismissed):

waiter/docs/parameters.md
```diff
@@ -65,6 +65,7 @@ Additional (optional) parameters that can be set:
...
 |`x-waiter-scale-up-factor`|0.01|(0-1]|The percentage amount, per second, to increase the current number of instances toward the target number of instances. See [How Waiter Autoscaling Works](autoscaling.md).|Don't change it unless you have a good reason to do so.|
 |`x-waiter-scheduler`|empty|<a supported scheduler id in the composite scheduler>|The id of the scheduler to use while creating this service. For composite schedulers, when provided, it must match the name of the component scheduler.|Don't configure it unless you are using composite scheduler and you know which specific component scheduler you want to use as default.|
+|`x-waiter-service-mapping`|"legacy"|"legacy" or "exclusive"|Determines whether tokens having the same configuration will map to the same service (legacy mode) or to different services (exclusive mode).|Don't change it unless you want services unique to your token.|
```
DaoWen, line 68: Maybe reword?

> Don't change unless your service needs access to the token (dns) name.

waiter/src/waiter/service_description.clj
DaoWen: I also prefer `WAITER_TOKEN`. Let's just leave it as-is.

shamsimam (commented):
Addressed PR feedback:
- updated documentation
- added new integration tests for multi-tokens.

DaoWen (approved):

PR #1350, opened 2021-05-16
moves maintenance mode handling inside the ping response

shamsimam:
- adds integration tests for waiter-ping and https-redirect and maintenance mode behavior
- moves maintenance mode handling inside the ping response

When a token is in maintenance mode, the ping request fails a 503 Maintenance mode error. Instead, we should reflect this error in the `ping-response` field of the `/waiter-ping` response. The ping itself is successful and should return a `2XX` response. 

There is one corner case for `3XX` redirects which we propagate in the ping response as existing clients already rely on the `3XX` behavior for HTTPS redirects.

**Note**: [This commit highlights](https://github.com/twosigma/waiter/pull/1350/commits/4f7f9617b9b9c164fb827a55dca8a0d885001b1e?diff=split&w=1) the behavior changes in `/waiter-ping` via changes in assertions of the integration tests.
Summary:
- eager `4XX` response without the HTTPS redirect
- `3XX` response on `/waiter-ping` includes a body
- Maintenance mode token report the `503` inside the ping body, the going response itself is a `200`.

### DIFF
--- a/waiter/integration/waiter/token_request_test.clj
+++ b/waiter/integration/waiter/token_request_test.clj
@@ -1092,39 +1092,41 @@
             url (URL. (str "http://" waiter-url))
             endpoint "/request-info"
             ping-endpoint "/waiter-ping"]
-        (doseq [request-method [:get :post]]
-          (let [redirect-status-code (if (= :get request-method) http-301-moved-permanently http-307-temporary-redirect)]
-            (testing "https redirects"
-              (let [token-response (post-token waiter-url token-description)]
+        (doseq [[request-method redirect-status-code]
+                [[:get http-301-moved-permanently]
+                 [:post http-307-temporary-redirect]]]
+          (testing "https redirects"
+            (let [token-response (post-token waiter-url token-description)]
+              (assert-response-status token-response http-200-ok))
+
+            (testing (str (name request-method) " request")
+              (let [{:keys [headers] :as response}
+                    (make-request waiter-url endpoint :headers request-headers :method request-method)]
+                (assert-response-status response redirect-status-code)
+                (assert-waiter-response response)
+                (is (= (str "https://" (.getHost url) endpoint) (get headers "location")))))
+
+            (testing (str (name request-method) " /waiter-ping request")
+              (let [{:keys [body headers] :as response}
+                    (make-request waiter-url ping-endpoint :headers request-headers :method request-method)]
+                (assert-response-status response redirect-status-code)
+                (assert-waiter-response response)
+                (is (= (str "https://" (.getHost url) ping-endpoint) (get headers "location")))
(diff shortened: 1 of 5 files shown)


waiter/integration/waiter/token_request_test.clj
```diff
@@ -1085,32 +1085,46 @@
...
-              (is (str/starts-with? (str (get headers "server")) "waiter") (str "headers:" headers))))))
+          token-description (assoc (kitchen-params) :https-redirect true :name token :token token)]
+
+      (let [request-headers {"accept" "application/json"
+                             :x-waiter-token token}
+            url (URL. (str "http://" waiter-url))
+            endpoint "/request-info"
+            ping-endpoint "/waiter-ping"]
+        (doseq [request-method [:get :post]]
+          (let [redirect-status-code (if (= :get request-method) http-301-moved-permanently http-307-temporary-redirect)]
```
DaoWen, line 1096: I think it's more clear to just iterate over pairs rather than having the `if` inside to determine the status code.

```suggestion
        (doseq [[request-method redirect-status-code] [[:get http-301-moved-permanently] [:post http-307-temporary-redirect]]]
```

waiter/integration/waiter/token_request_test.clj
```diff
@@ -1453,6 +1467,20 @@
...
+            (assert-response-status response http-200-ok)
+            (assert-waiter-response response)
+            (let [ping-body-json (try-parse-json body)
+                  ping-response-body-json (try-parse-json (get-in ping-body-json ["ping-response" "body"]))]
+              (is (= "waiter.Maintenance" (str (get-in ping-response-body-json ["waiter-error" "details" "error-class"]))))
+              (is (str/includes? (str (get-in ping-response-body-json ["waiter-error" "message"])) custom-maintenance-message))
+              (is (= "waiter-response" (get-in ping-body-json ["ping-response" "result"])))
+              (is (false? (get-in ping-body-json ["service-state" "exists?"])))
+              (is (false? (get-in ping-body-json ["service-state" "healthy?"])))
+              (is (= "Inactive" (get-in ping-body-json ["service-state" "status"]))))))
```
DaoWen, line 1482: We should also assert that the `ping-body-json` `:status` is a 503.

DaoWen (dismissed):

waiter/integration/waiter/token_request_test.clj
shamsimam: Done.

waiter/integration/waiter/token_request_test.clj
shamsimam: Done.

kevo1ution (approved):

DaoWen (approved):

PR #1353, opened 2021-05-19
Support service overriding s3 log bucket url

DaoWen:

- Support service overriding s3 log bucket url.
- Include `WAITER_`-prefixed environment variables in the fileserver sidecar.

- Allow service owners to use their own S3 buckets rather than the default.
- Make it easier to override other fileserver behaviors via `WAITER_CONFIG_`-prefixed environment variables.

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -189,7 +189,7 @@
                           :interval 2 :timeout 45)
                 (str "no killed instances found for " service-id))
 
-            ;; Test that the killed instances' logs were persisted to S3.
+            ;; Test that the killed instance's logs were persisted to S3.
             ;; This portion of the test logic was modified from the active-instances tests above.
             (let [killed-instances (killed-instances router-url service-id :cookies cookies)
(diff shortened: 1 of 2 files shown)


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -214,6 +214,28 @@
...
                   (log/warn "test-s3-logs did not verify file link:" file-link))))))))))
 
+(deftest ^:parallel ^:integration-fast test-s3-custom-bucket
+  (testing-using-waiter-url
+    (when (using-k8s? waiter-url)
+      (when-let [log-bucket-url (-> waiter-url get-kubernetes-scheduler-settings :log-bucket-url)]
+        (let [bucket-subpath "/my/custom/path"
+              custom-bucket-url (str log-bucket-url bucket-subpath)
+              service-headers {:x-waiter-name (rand-name)
+                               :x-waiter-env-WAITER_CONFIG_LOG_BUCKET_URL custom-bucket-url}
```
shamsimam, line 224: nitpick: lexicographic order

shamsimam (approved):

PR #1354, opened 2021-05-28
adds support for opting out of token syncing when metadata is configured

shamsimam:

- adds support for opting out of token syncing when metadata is configured

We wish to allow service owners to explicitly opt-out of token syncing across clusters.

### DIFF
--- a/token-syncer/src/token_syncer/commands/syncer.clj
+++ b/token-syncer/src/token_syncer/commands/syncer.clj
@@ -75,7 +75,7 @@
   "Syncs a given token description on all clusters.
    If the cluster-url->token-data says that a given token was not successfully loaded, it is skipped.
    Token sync-ing is also skipped if the tokens are active and the roots are different."
-  [{:keys [store-token]} cluster-urls token latest-token-description cluster-url->token-data]
+  [{:keys [store-token]} cluster-urls token latest-token-description metadata-name cluster-url->token-data]
   (pc/map-from-keys
     (fn [cluster-url]
       (let [ignored-root-mismatch-equality-comparison-keys ["cluster" "last-update-time" "last-update-user" "previous" "root"]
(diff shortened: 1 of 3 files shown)


token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -271,28 +275,46 @@
...
-                          (let [sync-result (sync-tokens waiter-api cluster-urls-set limit)
+                          (let [{:keys [details summary] :as sync-result} (sync-tokens waiter-api cluster-urls-set limit metadata-name)
                                 exit-code (-> (get-in sync-result [:summary :sync :failed])
-                                              empty?
-                                              (if 0 1))]
-                            (log/info (-> sync-result pp/pprint with-out-str str/trim))
+                                            empty?
+                                            (if 0 1))]
+                            (log/info "sync details:" (into (sorted-map) details))
+                            (log/info "sync summary:" (into (sorted-map) summary))
```
shamsimam, line 303: Note: Unrelated change to this PR to reduce number if lines output by the syncer.

token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -75,7 +75,7 @@
   "Syncs a given token description on all clusters.
    If the cluster-url->token-data says that a given token was not successfully loaded, it is skipped.
    Token sync-ing is also skipped if the tokens are active and the roots are different."
-  [{:keys [store-token]} cluster-urls token latest-token-description cluster-url->token-data]
+  [{:keys [store-token]} cluster-urls token latest-token-description metadata-name cluster-url->token-data]
```
DaoWen, line 78: Would it make sense to rename `metadata-name` to `opt-out-metadata-name`? I think that would make it more immediately clear what that metadata key-value pair is being used for.

shamsimam: Renamed to `opt-out-metadata-name`

DaoWen (approved):

PR #1355, opened 2021-05-28
add deep merge option in waiter cli update

kevo1ution:
- Allow users to deep merge when providing json/yaml input or potentially nested fields flag arguments

- This is useful when editing "metadata" or "env" fields

### DIFF
--- a/cli/waiter/subcommands/update.py
+++ b/cli/waiter/subcommands/update.py
@@ -1,3 +1,4 @@
+import argparse
 from functools import partial
 
 from waiter import token_post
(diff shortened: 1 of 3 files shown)


cli/waiter/subcommands/update.py
```diff
@@ -11,6 +12,12 @@ def register(add_parser):
     action = token_post.Action.UPDATE
     parser = token_post.register_argument_parser(add_parser, action)
     token_post.add_arguments(parser)
+    deep_merge_group = parser.add_mutually_exclusive_group(required=False)
```
shamsimam, line 15: Why not use `a.b=c` as always adding to the entry `b=c` in the dictionary under `a`?
In `waiter create` this stores a new entry in a fresh `a` dictionary whereas in `waiter update` it updates the `b=c` entry in the existing `a` dictionary.

kevo1ution: That also works, in order to support the existing override logic with `waiter update --json file` (which does a shallow merge), I'll have to apply the flags after shallow merging the file json and the existing token config. I'll close this PR and make a separate one.

PR #1357, opened 2021-05-28
Cli add update and create nested fields support

kevo1ution:

- allows for nested fields: ex. `waiter create [token] --field1.field2.field3 50`
- for update command, these fields will be applied to existing token configuration (deep-merge) if there is no override/json/yaml input. Overrides will result in a shallow merge behavio
Screenshot from 2021-05-28 11-24-27

- makes modifying nested fields much easier for users

### DIFF
--- a/cli/waiter/util.py
+++ b/cli/waiter/util.py
@@ -12,14 +12,12 @@ TRUE_STRINGS = ('yes', 'true', 'y')
 FALSE_STRINGS = ('no', 'false', 'n')
 
 
-def assoc_in(obj, keys, value):
+def update_in(obj, keys, value):
     """Given a list of keys and a value, return a new object with that value set"""
-    obj = obj.copy()
     cur_node = obj
     for key in keys[:-1]:
         cur_node = cur_node.setdefault(key, {})
     cur_node[keys[-1]] = value
-    return obj
 
 
 def deep_merge(a, b):
--- a/cli/waiter/util.py
+++ b/cli/waiter/util.py
@@ -12,6 +12,16 @@ TRUE_STRINGS = ('yes', 'true', 'y')
 FALSE_STRINGS = ('no', 'false', 'n')
 
 
+def assoc_in(obj, keys, value):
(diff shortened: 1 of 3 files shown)


cli/waiter/util.py
```diff
@@ -12,6 +12,16 @@
 FALSE_STRINGS = ('no', 'false', 'n')
 
 
+def assoc_in(obj, keys, value):
+    """Given a list of keys and a value, return a new object with that value set"""
+    obj = obj.copy()
```
DaoWen, line 17: It's not clear to me why you need to make a copy of the object here. If you don't really need the copy, maybe we just rename this `set_in` or `update_in` or something like that and not bother with the copy?

Otherwise, you need to copy at every level if you're trying to avoid mutating the original (nested) map:
```
>>> def assoc_in(obj, keys, value):
...     """Given a list of keys and a value, return a new object with that value set"""
...     obj = obj.copy()
...     cur_node = obj
...     for key in keys[:-1]:
...         cur_node = cur_node.setdefault(key, {})
...     cur_node[keys[-1]] = value
...     return obj
...
>>> assoc_in(m, ['e'], 30)
{'a': {'b': {'c': 1}, 'd': 20}, 'e': 30}
>>> m
{'a': {'b': {'c': 1}, 'd': 20}, 'e': 3}
>>> m = dict(a=dict(b=dict(c=1), d=2), e=3)
>>> m
{'a': {'b': {'c': 1}, 'd': 2}, 'e': 3}
>>> assoc_in(m, ['a', 'd'], 20)
{'a': {'b': {'c': 1}, 'd': 20}, 'e': 3}
>>> m
{'a': {'b': {'c': 1}, 'd': 20}, 'e': 3}
```

DaoWen (changes requested):

cli/waiter/util.py
kevo1ution: good point, I didn't realize that .copy() doesn't make a deep copy. Going to make this change

DaoWen (approved):

PR #1358, opened 2021-05-29
adds templating with variable substitution support to input file

shamsimam:

- adds templating with variable substitution support to input file

Templating simplifies the management of multiple tokens that differ in certain values but otherwise share a common template. Such tokens want to have different names, which is why they cannot be treated as parameterized services. 

We limit the support to variable substitution and intentionally avoid full-fledged Jinja templating support, e.g. conditional and looping, to provide a simple interface to users. In addition, the simple substitution scheme ensures that all tokens managed by a common template always have the same shape.

### Example output:
```
$ cat template.json 
{"mem": 1024, "cmd":"/opt/bin/run-service ${version} ${name} $${WAITER_CLUSTER}", "token":"my-token"}

$ cat context.json 
{"version": "f1137960-aa23-488f-aff4-66ba6a744882"}

$ waiter create --context context.json --env.FOO bar 
The --context file can only be used when a data file is specified via --input, --json, or --yaml.

$ waiter create --env.FOO bar --context.name john.doe
The --context.xyz overrides can only be used when a data file is specified via --input, --json, or --yaml.

$ waiter create --input template.json --context invalid.json --env.FOO bar --context.name john.doe
Unable to load yaml from invalid.json.

$ waiter create --input template.json --context context.json --env.FOO bar
Error when processing template: missing variable 'name'

$ waiter create --input template.json --context context.json --env.FOO bar --context.name john.doe
Successfully created my-token.

$ curl -s 'http://localhost:9091/token?token=my-token' | jq -rS .
{
  "cmd": "/opt/bin/run-service f1137960-aa23-488f-aff4-66ba6a744882 john.doe ${WAITER_CLUSTER}",
  "env": {
    "FOO": "bar"
  },
  "mem": 1024,
  "owner": "shamsimam"
}
```

### DIFF
--- a/cli/waiter/data_format.py
+++ b/cli/waiter/data_format.py
@@ -1,3 +1,4 @@
+import jinja2
 import json
 import logging
 import os
@@ -157,6 +158,21 @@ def load_data(options):
         content = load_file(input_file)
         if not content:
             raise Exception(f'Unable to load {input_format} from {input_file}.')
+
+        context_file = options.get('context')
+        if context_file:
+            logging.debug(f'reading context as yaml from {context_file}')
+            context_content = load_file(context_file)
+            if not context_content:
+                raise Exception(f'Unable to load yaml from {context_file}.')
+
+            context_obj = YAML.parse(context_content)
+            if not isinstance(context_obj, dict):
+                raise Exception(f'Provided context file must evaluate to a dictionary, instead it is {context_obj}')
+
+            logging.debug(f'applying jinja templating to input using context {context_obj}')
+            jinja_template = jinja2.Template(content)
+            content = jinja_template.render(**context_obj)
 
     content = input_format.parse(content)
     if type(content) is dict:
--- a/cli/waiter/data_format.py
+++ b/cli/waiter/data_format.py
@@ -164,10 +164,10 @@ def load_data(options):
         context_file = options.get('context_file')
         if context_file:
             context_provided = True
-            logging.debug(f'reading context as yaml from {context_file}')
+            logging.debug(f'reading context from {context_file}')
             context_content = load_file(context_file)
             if not context_content:
-                raise Exception(f'Unable to load yaml from {context_file}.')
+                raise Exception(f'Unable to load context from {context_file}.')
 
             context_file_obj = YAML.parse(context_content)
             if not isinstance(context_file_obj, dict):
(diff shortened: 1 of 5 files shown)


cli/waiter/data_format.py
```diff
@@ -158,6 +159,36 @@ def load_data(options):
...
 
+        context_dict = {}
+        context_provided = False
+        context_file = options.get('context_file')
+        if context_file:
+            context_provided = True
+            logging.debug(f'reading context as yaml from {context_file}')
+            context_content = load_file(context_file)
+            if not context_content:
+                raise Exception(f'Unable to load yaml from {context_file}.')
```
kevo1ution, line 170: Is the format always yaml? Can we make the error message: "Unable to load yaml or json from..."

cli/waiter/data_format.py
```diff
@@ -158,6 +159,36 @@ def load_data(options):
         if not content:
             raise Exception(f'Unable to load {input_format} from {input_file}.')
 
+        context_dict = {}
+        context_provided = False
+        context_file = options.get('context_file')
+        if context_file:
+            context_provided = True
+            logging.debug(f'reading context as yaml from {context_file}')
```
kevo1ution, line 167: Same question as below. Is the format always yaml, and if not, can we mention "reading context as yaml or json from..."

cli/waiter/token_post.py
```diff
@@ -177,6 +199,10 @@ def add_arguments(parser):
     format_group.add_argument('--json', help='provide the data in a JSON file', dest='json')
     format_group.add_argument('--yaml', help='provide the data in a YAML file', dest='yaml')
     format_group.add_argument('--input', help='provide the data in a JSON/YAML file', dest='input')
+    parser.add_argument('--context', dest='context',
+                        help='can be used only when a JSON or YAML data file has been provided; '
```
kevo1ution, line 203:
```suggestion
                        help='can be used only when a data file has been provided via --input, --json, or --yaml; '
```

kevo1ution (commented):
I'm confused by the example which is context.json but is not yaml. What is the format of context that we are supporting?

cli/waiter/token_post.py
shamsimam: Included change.

cli/waiter/data_format.py
shamsimam: Fixed messaging. The context file is parsed as yaml, yaml being a superset of json also parses json files :)

cli/waiter/data_format.py
shamsimam: Fixed messaging.

kevo1ution (approved):

PR #1359, opened 2021-05-29
fixes typo in CLI test name

shamsimam:

- fixes typo in CLI test name

Ensures we are now covering both json and yaml `env` update tests.


kevo1ution (approved):
good catch

PR #1367, opened 2021-06-04
Add unsupported headers

kevo1ution:

- We can configure added unsupported headers which will short circuit and waiter will respond with an error

- We can easily add different waiter headers that are not available for on-the-fly requests

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -1513,8 +1513,8 @@
 
       (testing "should give 400 if unsupported headers are specified"
         (doseq [waiter-header waiter-unsupported-headers]
-          (let [{:keys [service-id body] :as response} (make-request-with-debug-info
-                                              {:x-waiter-name (rand-name)
+          (let [{:keys [service-id body] :as response}
+                (make-request-with-debug-info {:x-waiter-name (rand-name)
                                                waiter-header "some value"}
                                               #(make-kitchen-request waiter-url % :path "/hello"))]
             (assert-response-status response http-400-bad-request)
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -1506,7 +1506,7 @@
               response (make-kitchen-request waiter-url request-headers :path "/hello")]
           (assert-response-status response http-503-service-unavailable))))))
 
-(deftest test-request-unsupported-headers
+(deftest  ^:parallel ^:integration-fast test-request-unsupported-headers
   (testing-using-waiter-url
     (let [settings (waiter-settings waiter-url)
           waiter-unsupported-headers (get-in settings [:instance-request-properties :unsupported-headers])]
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -1505,3 +1505,23 @@
                                 :x-waiter-queue-timeout 5000)
               response (make-kitchen-request waiter-url request-headers :path "/hello")]
           (assert-response-status response http-503-service-unavailable))))))
+
+(deftest test-request-unsupported-headers
+  (testing-using-waiter-url
+    (let [settings (waiter-settings waiter-url)
+          waiter-unsupported-headers (get-in settings [:instance-request-properties :unsupported-headers])]
+
+      (testing "should give 400 if unsupported headers are specified"
+        (doseq [waiter-header waiter-unsupported-headers]
+          (let [{:keys [service-id body] :as response} (make-request-with-debug-info
+                                              {:x-waiter-name (rand-name)
+                                               waiter-header "some value"}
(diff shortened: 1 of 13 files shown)


waiter/src/waiter/core.clj
```diff
@@ -1894,14 +1896,20 @@
                                                (fn waiter-request-interstitial-handler-fn [request]
                                                  (interstitial/display-interstitial-handler request))))
    :websocket-request-acceptor (pc/fnk [[:routines wrap-service-discovery-fn]
+                                        [:settings [:instance-request-properties unsupported-headers]]
                                         [:state server-name]
                                         websocket-secure-request-acceptor-fn]
                                  ; If adding new middleware for websocket upgrade requests, consider adding the same middleware to
                                  ; process-request-wrapper-fn
-                                 (let [handler (-> #(ws/request-subprotocol-acceptor (:upgrade-request %) (:upgrade-response %))
+                                 (let [temp-fn (fn [handler]
```
shamsimam, line 1904: Needs to be removed.

waiter/src/waiter/core.clj
```diff
@@ -1547,6 +1548,7 @@
                                      auth/wrap-auth-bypass
                                      handler/wrap-https-redirect
                                      pr/wrap-maintenance-mode
+                                     (pr/wrap-unsupported-waiter-headers unsupported-headers)
```
shamsimam, line 1551: How do you feel about moving this to `descriptor/compute-descriptor` (after the `headers/split-headers` invocation)? I think that function should be responsible for disallowing certain headers.

kevo1ution: That's a good idea, I agree. I'm going to confirm it has the same behavior for websockets because the `compute-descriptor function` will have called by the `:default-websocket-handler-fn` instead of `:websocket-request-acceptor`

kevo1ution: Okay the error needs to surface in the `websocket-request-acceptor` handler which doesn't call `descriptor/compute-descriptor` in its tree of function calls. I had to add a error handler and include this logic in the `discover-service-parameters` function

shamsimam: How does websocket request acceptor decide on the service ID?

waiter/integration/waiter/basic_test.clj
```diff
@@ -1505,3 +1505,23 @@
...
+
+(deftest  ^:parallel ^:integration-fast test-request-unsupported-headers
+  (testing-using-waiter-url
+    (let [settings (waiter-settings waiter-url)
+          waiter-unsupported-headers (get-in settings [:instance-request-properties :unsupported-headers])]
+
+      (testing "should give 400 if unsupported headers are specified"
+        (doseq [waiter-header waiter-unsupported-headers]
+          (let [{:keys [service-id body] :as response} (make-request-with-debug-info
+                                              {:x-waiter-name (rand-name)
```
shamsimam, line 1517: Indentation is off

waiter/src/waiter/service_description.clj
```diff
@@ -1571,7 +1571,14 @@
         token-metadata (when token
                          (-> (token->token-parameters kv-store token :error-on-missing false)
                            (attach-token-defaults-fn)
-                           (select-keys token-metadata-keys)))]
+                           (select-keys token-metadata-keys)))
+        unsupported-headers-in-request (set/intersection (set (keys waiter-headers)) unsupported-headers)
+        details {:unsupported-headers-in-request unsupported-headers-in-request}]
```
shamsimam, line 1576: Nitpick: `details` can be moved inside the `when` and more appropriately named to `error-details`.

waiter/src/waiter/service_description.clj
```diff
@@ -1571,7 +1571,14 @@
...
+    (when (not-empty unsupported-headers-in-request)
+      (log/info "Unsupported waiter headers found" details)
+      (throw (ex-info "Unsupported waiter headers found"
+                      {:status http-400-bad-request
+                       :details details})))
```
shamsimam, line 1581: Maybe set `:level :info`?

shamsimam: Nitpick: lexicographic order

waiter/test/waiter/auth/authenticator_test.clj
```diff
@@ -145,7 +145,7 @@
         handler-response (Object.)
         execute-request (fn execute-request-fn [{:keys [headers] :as in-request}]
                           (let [test-request (->> (sd/discover-service-parameters
-                                                    kv-store attach-service-defaults-fn attach-token-defaults-fn waiter-hostnames headers)
+                                                    kv-store attach-service-defaults-fn attach-token-defaults-fn waiter-hostnames headers [])
```
shamsimam, line 148:
```suggestion
                                                    kv-store attach-service-defaults-fn attach-token-defaults-fn waiter-hostnames headers #{})
```

shamsimam (approved):

PR #1369, opened 2021-06-09
Refactor handle-post-token-request function to use a validator component

kevo1ution:

- allows configuration of the token validator component

- we can now add custom token validators

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1074,7 +1074,8 @@
                 :limit-per-owner 1000
 
                 ;; Custom validator on post /token endpoint
-                :post-validator-fn 'waiter.token/simple-post-validator
+                :post-validator {:kind :configured
+                                 :configured {:factory-fn 'waiter.token/create-simple-post-validator}}
 
                 ;; Default values associated with tokens while resolving the service description
                 :token-defaults {
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1074,8 +1074,8 @@
                 :limit-per-owner 1000
 
                 ;; Custom validator on post /token endpoint
-                :post-validator {:kind :configured
-                                 :configured {:factory-fn 'waiter.token/create-simple-post-validator}}
+                :validator {:kind :configured
+                            :configured {:factory-fn 'waiter.token/create-default-token-validator}}
 
                 ;; Default values associated with tokens while resolving the service description
                 :token-defaults {
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1074,8 +1074,8 @@
                 :limit-per-owner 1000
 
                 ;; Custom validator on post /token endpoint
-                :validator {:kind :configured
-                            :configured {:factory-fn 'waiter.token/create-default-token-validator}}
+                :validator {:kind :default
+                            :default {:factory-fn 'waiter.token/create-default-token-validator}}
 
                 ;; Default values associated with tokens while resolving the service description
                 :token-defaults {
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1075,7 +1075,7 @@
 
                 ;; Custom validator on post /token endpoint
                 :validator {:kind :default
-                            :default {:factory-fn 'waiter.token/create-default-token-validator}}
+                            :default {:factory-fn 'waiter.token-validator/create-default-token-validator}}
 
                 ;; Default values associated with tokens while resolving the service description
                 :token-defaults {
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1073,6 +1073,9 @@
                 ;; It must be a positive integer.
                 :limit-per-owner 1000
 
+                ;; Custom validator on post /token endpoint
+                :post-validator-fn 'waiter.token/simple-post-validator
+
                 ;; Default values associated with tokens while resolving the service description
                 :token-defaults {
                                  ;; The default period (minutes) for which an instance from a fallback service can be requested:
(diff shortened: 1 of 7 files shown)


shamsimam (commented):
I think this change is too specific linking only the field and the entitlement manager. Instead, we should be able to plugin in a custom validator object (using the `kind` support in e.g. `:token-config :validator`), The entitlement manager and other custom components can be passed in the context to the custom validator object factory function.

waiter/config-full.edn
```diff
@@ -1073,6 +1073,9 @@
                 ;; It must be a positive integer.
                 :limit-per-owner 1000
 
+                ;; Custom validator on post /token endpoint
+                :post-validator-fn 'waiter.token/simple-post-validator
```
shamsimam, line 1077: Ideally, this is a component with a factory function that gets initialized.

waiter/src/waiter/token.clj
```diff
@@ -437,9 +443,9 @@
 (defn- handle-post-token-request
   "Validates that the user is the creator of the token if it already exists.
    Then, updates the configuration for the token in the database using the newest password."
-  [clock synchronize-fn kv-store cluster-calculator token-root history-length limit-per-owner waiter-hostnames
+  [clock custom-components synchronize-fn kv-store cluster-calculator token-root history-length limit-per-owner waiter-hostnames
    entitlement-manager make-peer-requests-fn validate-service-description-fn attach-service-defaults-fn
-   tokens-update-chan {:keys [headers] :as request}]
+   tokens-update-chan post-validator-fn {:keys [headers] :as request}]
```
shamsimam, line 448: When `post-validator` is a component, then the `custom-components` will not need to be passed as an argument to this function.

kevo1ution: Thanks, that will be a cleaner change

waiter/src/waiter/token.clj
```diff
@@ -525,6 +534,10 @@
                           (not current-owner?)
                           existing-editor
                           (authz/run-as? entitlement-manager authenticated-user existing-editor))]
+
+        ;; do any custom validation
+        (post-validator-fn request new-token-data existing-token-metadata)
```
shamsimam, line 539: I think these lines should be replaced by an invocation to `post-validator-fn`: https://github.com/twosigma/waiter/blob/7d2bef587cbdec269d7c34903e86a77f88bb0b34/waiter/src/waiter/token.clj#L474-L476

I'm not sure we need to pass `existing-token-metadata` to the validation function. It is responsible for validating the new inputs passed in the post argument.

Also, instead of passing the raw request, ew should pass a map with certain keys, e.g. the user.

waiter/src/waiter/core.clj
```diff
@@ -779,6 +779,10 @@
    :token-cluster-calculator (pc/fnk [[:settings [:cluster-config name] [:token-config cluster-calculator]]]
                                (utils/create-component
                                  cluster-calculator :context {:default-cluster name}))
+   :token-post-validator-fn (pc/fnk [[:settings [:token-config post-validator]]
+                                     custom-components]
+                              (utils/create-component
+                                post-validator :context {:custom-components custom-components}))
```
shamsimam, line 785: This component needs to be able to expose state on the `/state` endpoint.

waiter/src/waiter/token.clj
```diff
@@ -37,6 +37,15 @@
 
 (def ^:const ANY-USER "*")
 
+(defn create-simple-post-validator
```
shamsimam, line 40: Maybe make this return a protocol which also exposes state.

waiter/config-full.edn
```diff
@@ -1073,6 +1073,10 @@
                 ;; It must be a positive integer.
                 :limit-per-owner 1000
 
+                ;; Custom validator on post /token endpoint
+                :validator {:kind :configured
```
shamsimam, line 1077: Maybe name to `default`?

waiter/src/waiter/core.clj
```diff
@@ -615,6 +615,10 @@
                                               :context {:default-authentication (get service-description-defaults "authentication")
                                                         :hostname hostname
                                                         :password (first passwords)})))
+   :attach-service-defaults-fn (pc/fnk [[:settings metric-group-mappings service-description-defaults]
```
shamsimam, line 618: Please pass these in the context map to the `validate` function.

waiter/src/waiter/token.clj
```diff
@@ -37,6 +38,155 @@
 
 (def ^:const ANY-USER "*")
 
+(defprotocol TokenValidator
+  "A protocol for validating a new token configuration"
+
+  (state [this]
```
shamsimam, line 44: We should have `include-flags` parameter in this function.

waiter/src/waiter/token.clj
```diff
@@ -19,6 +19,7 @@
             [clojure.data :as data]
             [clojure.set :as set]
             [clojure.string :as str]
+            [clojure.test :as test]
```
shamsimam, line 22: Please remove this import and use `fn?` if necessary.

waiter/src/waiter/token.clj
```diff
@@ -37,6 +38,155 @@
...
+                         :status http-400-bad-request
+                         :log-level :warn})))
+
+      (when-let [previous (get new-token-metadata "previous")]
+        (when-not (map? previous)
+          (throw (ex-info (str "Token previous must be a map")
+                          {:previous previous :status http-400-bad-request :token token :log-level :warn})))))))
+
+(defn create-default-token-validator
+  "creates the default token validator and returns it"
```
shamsimam, line 182:
```suggestion
  "Creates the default token validator."
```

shamsimam (approved):

PR #1376, opened 2021-07-02
Fix CLI parsing non string values for "env" and "metadata" fields

kevo1ution:

- fixes improper parsing of env and metadata fields as non strings

- by setting these fields to a boolean value like "true", the CLI command would fail:
```
$ waiter create --cpus 0.1 --mem 1024 --name testing --env.example "true" --metadata.example "true" kitchen.localtest.me
Service description using waiter headers/token improperly configured
The following metadata keys did not have string values: example: true. Metadata values must be strings.
```

### DIFF
--- a/cli/waiter/token_post.py
+++ b/cli/waiter/token_post.py
@@ -14,6 +14,7 @@ from waiter.util import deep_merge, FALSE_STRINGS, is_admin_enabled, print_info,
 BOOL_STRINGS = TRUE_STRINGS + FALSE_STRINGS
 INT_PARAM_SUFFIXES = ['-failures', '-index', '-instances', '-length', '-level', '-mins', '-secs']
 FLOAT_PARAM_SUFFIXES = ['-factor', '-rate', '-threshold']
+STRING_PARAM_PREFIXES = ['env', 'metadata']
 
 
 class Action(Enum):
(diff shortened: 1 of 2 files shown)


cli/waiter/token_post.py
```diff
@@ -272,12 +272,15 @@ def add_implicit_arguments(unknown_args, parser):
     for i in range(num_unknown_args):
         arg = unknown_args[i]
         if arg.startswith(("-", "--")):
+            arg_dest = arg.lstrip('-')
             if any(arg.endswith(suffix) for suffix in INT_PARAM_SUFFIXES):
                 arg_type = possible_int
             elif any(arg.endswith(suffix) for suffix in FLOAT_PARAM_SUFFIXES):
                 arg_type = possible_float
+            elif any(arg_dest.startswith(prefix) for prefix in STRING_PARAM_PREFIXES):
```
shamsimam, line 280: Should we generalize this to nested field access instead?

kevo1ution: I wanted to be specific here to `env` and `metadata` because we are sure those values should be parsed as strings. There may be other nested fields that we add that have boolean/number values, and I didn't want to change the default behavior for other fields

shamsimam (approved):

PR #1385, opened 2021-07-12
Add protocol to async request parameters

DaoWen:

Add request protocol to async request parameters.

There is currently no way to determine the correct backend protocol for async requests when the backend is using a proxy with a different protocol from that in the service description. The async parameters should include *all* of the information to independently make the request on any router, which means we need the protocol.

### DIFF
--- a/waiter/src/waiter/async_request.clj
+++ b/waiter/src/waiter/async_request.clj
@@ -27,6 +27,7 @@
             [waiter.service :as service]
             [waiter.statsd :as statsd]
             [waiter.status-codes :refer :all]
+            [waiter.util.async-utils :as au]
             [waiter.util.http-utils :as hu]
             [waiter.util.utils :as utils])
   (:import (java.net ConnectException SocketTimeoutException URI URLEncoder)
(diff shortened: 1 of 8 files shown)


waiter/src/waiter/async_request.clj
```diff
@@ -160,11 +160,16 @@
...
   "Converts the route params to a uri.
-   The function expects prefix to end with a slash and location to begin with a slash.
-   Returns a formatted url: prefix/{request-id}/{router-id}/{service-id}/{host}/{port}{location}"
-  [prefix {:keys [host location port request-id router-id service-id]}]
-  (let [encode #(if %1 (URLEncoder/encode %1 "UTF-8") (str %1))]
-    (str prefix (encode request-id) "/" (encode router-id) "/" service-id "/" (str host) "/" (str port) location)))
+   The function expects location to begin with a slash.
+   Returns a uri in one of the following formats (depending on whether :proto is set):
+   - /waiter-async/action/{request-id}/{router-id}/{service-id}/{host}/{port}{location}
+   - /waiter-async/v2/action/{request-id}/{router-id}/{service-id}/{host}/{port}/{proto}{location}"
```
shamsimam, line 166: Should we change the `v2` version of the url to accept an encrypted dictionary like we once discussed? This may prevent the need for a `v3` version of the url later.

waiter/src/waiter/core.clj
```diff
@@ -148,10 +148,17 @@
                                "/owners" :token-owners-handler-fn
                                "/refresh" :token-refresh-handler-fn
                                "/reindex" :token-reindex-handler-fn}
-                     "waiter-async" {["/complete/" :request-id "/" :service-id] :async-complete-handler-fn
+                     "waiter-async" {;; async api version 1
```
shamsimam, line 151: We should update the unit tests in `waiter.core-test/test-waiter-request?-factory` , `waiter.core-test/test-routes-mapper` and `waiter.core-test/test-async-*`

DaoWen: Done

waiter/src/waiter/async_request.clj
DaoWen: Done

waiter/src/waiter/async_request.clj
```diff
@@ -225,7 +262,9 @@
    {:keys [request-id] :as reason-map} request-properties location query-string]
   (let [correlation-id (cid/get-correlation-id)
         {:strs [metric-group backend-proto]} service-description
-        status-endpoint (scheduler/end-point-url backend-proto host port location)
+        request-proto (get-async-request-protocol instance port service-description)
+        proto (when (not= request-proto backend-proto) request-proto)
```
DaoWen, line 266: ℹ️ Since the option for the Scheduler to pick a different request protocol is coming in a follow-up patch, there isn't a straightforward way to integration test it in this PR. To help verify that the `/waiter-async/v2/` api actually works, I manually ran the `waiter.async-request-integration-test` integration tests with the logic in this line inverted.

waiter/src/waiter/util/async_utils.clj
```diff
@@ -22,6 +22,11 @@
   (:import (clojure.core.async.impl.channels ManyToManyChannel)
            (java.util.concurrent ExecutorService)))
 
+(defn singleton-chan [v]
```
shamsimam, line 25: Nitpick: we  could use a `promise-chan`?

waiter/src/waiter/util/utils.clj
```diff
@@ -235,6 +236,21 @@
       (attach-waiter-namespace-keys data-map)
       (attach-waiter-source))))
 
+(let [b64-decoder (java.util.Base64/getUrlDecoder)]
+  (defn b64-url-json-decode [^String json-str]
+    "Decode a URL-safe Base64 JSON string into a Clojure data structure."
+    (-> (.decode b64-decoder json-str)
+        (String. java.nio.charset.StandardCharsets/UTF_8)
```
shamsimam, line 243: Nitpick: import the `StandardCharsets` class.

waiter/src/waiter/util/utils.clj
```diff
@@ -235,6 +236,21 @@
...
+        (try-parse-json keyword))))
+
+(let [b64-encoder (.withoutPadding (java.util.Base64/getUrlEncoder))]
+  (defn b64-url-json-encode [data]
+    "Encode a Clojure data structure as a JSON object in a URL-safe Base64 string."
+    (as-> data $
+      (clj->json $)
+      (.getBytes $ java.nio.charset.StandardCharsets/UTF_8)
```
shamsimam, line 251: Nitpick: import class (same as above)

shamsimam (approved):

waiter/src/waiter/util/async_utils.clj
DaoWen: OK, yeah, I checked and a promise-chan actually uses a special buffer for the single value. Switched.

waiter/src/waiter/util/utils.clj
DaoWen: It was already imported for a different function! 😅
Dropped the package name.

waiter/src/waiter/util/utils.clj
DaoWen: Dropped the package name.

shamsimam (approved):

PR #1392, opened 2021-07-15
Add streaming timeout to test instance watches

kevo1ution:

- set the streaming-timeout on integration test instance tracking watches to 1000 seconds

- the watches would get disconnected during the test due to the default 20-second streaming timeout. This would cause the watch to disconnect prematurely and no longer receive instance updates.

### DIFF
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -89,7 +89,8 @@
                     (is (t/before? start-time (du/str-to-date last-error-time)))))))))))))
 
 (defn- start-watch
-  [router-url cookies & {:keys [query-params] :or {query-params {"watch" "true"}}}]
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/instance_tracker_integration_test.clj
```diff
@@ -89,7 +89,8 @@
                     (is (t/before? start-time (du/str-to-date last-error-time)))))))))))))
 
 (defn- start-watch
-  [router-url cookies & {:keys [query-params] :or {query-params {"watch" "true"}}}]
+  [router-url cookies & {:keys [query-params] :or {query-params {"streaming-timeout" "1000000"
```
shamsimam, line 92: Feels ecessive, maybe a more realistic timeout of a few minutes, maybe `120000`?

shamsimam (approved):

PR #1405, opened 2021-07-27
Add request-protocol to ServiceScheduler interface

DaoWen:

Delegate protocol selection to the Scheduler implementation rather than assuming the backend-protocol on the service-description is correct. The assumption is that, although we can't just infer the correct protocol from the service-description, the Scheduler has enough data at request time to make that determination for a specific back-end instance.

Allows us to connect to backends via an intermediate proxy with a different protocol (e.g., enabling TLS via a proxy for a cleartext backend service).

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -864,7 +864,8 @@
                                      (f))))))))))})
 
 (def scheduler
-  {:scheduler-promise-chan (pc/fnk [] (async/promise-chan))
+  {;; promise used to break cyclic dependency between scheduler and start-scheduler-syncer-fn
+   :scheduler-promise-chan (pc/fnk [] (async/promise-chan))
    :scheduler (pc/fnk [[:settings scheduler-config scheduler-syncer-interval-secs]
                        [:state custom-components leader?-fn scheduler-state-chan service-id-prefix]
                        scheduler-promise-chan
@@ -909,8 +910,9 @@
                                                      :user-agent (str "waiter-syncer/" user-agent-version)})
                                       available? (fn scheduler-available?
                                                    [scheduler-name service-instance service-description]
-                                                   (scheduler/available? service-id->password-fn* http-client scheduler-promise-chan
-                                                                         scheduler-name service-instance service-description))]
+                                                   (let [scheduler (async/<!! scheduler-promise-chan)]
+                                                     (scheduler/available? service-id->password-fn* http-client scheduler
+                                                                           scheduler-name service-instance service-description)))]
                                   (fn start-scheduler-syncer-fn
                                     [scheduler-name get-service->instances-fn scheduler-state-chan trigger-chan]
                                     (scheduler/start-scheduler-syncer
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -864,8 +864,10 @@
                                      (f))))))))))})
 
 (def scheduler
-  {:scheduler (pc/fnk [[:settings scheduler-config scheduler-syncer-interval-secs]
+  {:scheduler-promise-chan (pc/fnk [] (async/promise-chan))
+   :scheduler (pc/fnk [[:settings scheduler-config scheduler-syncer-interval-secs]
                        [:state custom-components leader?-fn scheduler-state-chan service-id-prefix]
+                       scheduler-promise-chan
                        service-id->password-fn*
                        service-id->service-description-fn*
                        start-scheduler-syncer-fn]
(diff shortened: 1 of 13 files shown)


waiter/src/waiter/core.clj
```diff
@@ -864,8 +864,10 @@
                                      (f))))))))))})
 
 (def scheduler
-  {:scheduler (pc/fnk [[:settings scheduler-config scheduler-syncer-interval-secs]
+  {:scheduler-promise-chan (pc/fnk [] (async/promise-chan))
```
shamsimam, line 867: Nitpick: comment about needing this to break the cycle in initialization.

waiter/src/waiter/core.clj
```diff
@@ -897,15 +901,15 @@
...
+                                       scheduler-promise-chan service-id->password-fn* service-id->service-description-fn*]
                                 (let [http-client (hu/http-client-factory
                                                     {:client-name (str "waiter-syncer-" user-agent-version)
                                                      :conn-timeout health-check-timeout-ms
                                                      :socket-timeout health-check-timeout-ms
                                                      :user-agent (str "waiter-syncer/" user-agent-version)})
                                       available? (fn scheduler-available?
                                                    [scheduler-name service-instance service-description]
-                                                   (scheduler/available? service-id->password-fn* http-client
+                                                   (scheduler/available? service-id->password-fn* http-client scheduler-promise-chan
```
shamsimam, line 912: Can we extract the scheduler from the channel inside the lambda and only pass the scheduler instance to scheduler/available??

```suggestion
                                                   (let [scheduler (async/<!! scheduler-promise-chan)]
                                                     (scheduler/available? service-id->password-fn* http-client scheduler
```

waiter/src/waiter/scheduler.clj
```diff
@@ -149,6 +150,10 @@
      These options should be merged (to override) the global :deployment-error-config options
      in the waiter.state module when computing the deployment errors for the given service-id.")
 
+  (request-protocol [this instance port-index service-description]
```
shamsimam, line 153: Is passing the `service-description` necessary, most scheduler instances can retrieve it using the service ID?

waiter/src/waiter/scheduler.clj
```diff
@@ -247,7 +252,19 @@
     (->> port-index dec (nth extra-ports))
     port))
 
-(defn base-url
+(defn port-index-protocol
```
shamsimam, line 255: Nitpick:
```suggestion
(defn port-index->protocol
```

waiter/src/waiter/scheduler.clj
DaoWen: You're probably right, but why do the lookup again when we already have the service-description handy everywhere that we want to call this function?

shamsimam: That may be true today, but not in future call sites.

waiter/src/waiter/scheduler.clj
DaoWen: I don't like using the `->` syntax in places where it doesn't actually match the required arguments. This would need to be `port-index+service-description->protocol` or something like that to be accurate, but that's way too verbose.

waiter/src/waiter/core.clj
DaoWen: We could, but that would mean adding another go-block (or a blocking call) here just to extract the scheduler. If you feel like that's a good idea then I can make the change.

DaoWen: Since the scheduler is also the thing that invokes `available?`, I guess it's safe to assume that it's populated by the time this function is called. I can go ahead and switch it, and I'll add a comment to explain that the call shouldn't block.

waiter/src/waiter/scheduler.clj
DaoWen: You can make the same argument for `instance` vs `instance-id`. It seems unlikely that we'll be using this somewhere that we wouldn't also want to pull up the service-description for populating the request logs, right?

waiter/src/waiter/core.clj
shamsimam: The value is already available when this function is invoked, we could even use a promise and throw an exception if it is invoked too eagerly. It removes the need to extract the scheduler from the consumer (`scheduler/available?`)

waiter/src/waiter/scheduler.clj
shamsimam: We've refactored the signatures to remove the service description in the past, but we can do that later when needed if you are no onboard now.

waiter/src/waiter/scheduler.clj
shamsimam: Then maybe `retrieve-protocol` ?

DaoWen: Renamed.

waiter/src/waiter/core.clj
DaoWen: Added comment.

waiter/src/waiter/core.clj
DaoWen: Yes, I'd prefer to defer this change and wait to see if we actually end up needing it.

shamsimam (approved):

PR #1417, opened 2021-08-05
adds support for CLI start and stop commands

shamsimam:

- adds support for CLI start and stop commands

### DIFF
--- a/cli/waiter/cli.py
+++ b/cli/waiter/cli.py
@@ -44,6 +44,12 @@ actions = {
     'ssh': {
         'run-function': ssh.register(subparsers.add_parser)
     },
+    'start': {
+        'run-function': maintenance.register_stop('start', subparsers.add_parser)
+    },
+    'stop': {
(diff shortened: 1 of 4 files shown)


cli/waiter/cli.py
```diff
@@ -44,6 +44,12 @@
     'ssh': {
         'run-function': ssh.register(subparsers.add_parser)
     },
+    'start': {
```
kevo1ution, line 47: If a user runs `waiter start` on a token that isn't in maintenance mode then the token won't get pinged (service won't be started), which can be confusing for the user. I think the behavior here should be user runs `waiter start`
1. if token is in maintenance mode then disable maintenance mode
2. always ping token

kevo1ution (approved):

cli/waiter/cli.py
kevo1ution: we decided to make this a separate follow up PR

PR #1419, opened 2021-08-11
fixes reference to default namespace in k8s scheduler factory function

shamsimam:

- fixes reference to default namespace in k8s scheduler factory function

Resolves bug in code which is referencing the default namespace from the wrong location.


DaoWen:
Kelechi and I noticed this bug when he was testing his intern project stuff. Hadn't got around to fixing it though. Thanks for doing this!

DaoWen (approved):

PR #1420, opened 2021-08-11
adds support for star default namespace configuration

shamsimam:

- adds support for star default namespace configuration

Allows us to launch replica sets and pods in the same namespace as the `run-as-user` by default.


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -333,9 +333,9 @@
   (testing-using-waiter-url
     (when (using-k8s? waiter-url)
       (let [current-user (retrieve-username)]
-        (testing "No service account with default namespace"
```
shamsimam, line 336: @DaoWen do you remember why we enforced this check?

DaoWen (dismissed):

waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
DaoWen: That's unfortunate. I think we just assumed this would be opt-in (i.e., add the `namespace` explicitly if you want it). If we're switching to "run as user by default", then we'll want to change this to just check if the namespace and run-as-user match.

PR #1423, opened 2021-08-14
fix grpc server cancel handling

DaoWen:

Gracefully handle eager-close by server via `RST_STREAM` frame

Fixes `test-grpc-bidi-streaming-server-cancellation` test response status being changed by Waiter when an Envoy proxy sits between the Waiter proxy and the backend grpc server.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -75,7 +75,7 @@
 
 (defn classify-error
   "Classifies the error responses from the backend into the following vector:
-   - error cause (:client-eagerly-closed, :client-error, :instance-error or :generic-error),
+   - error cause (:client-eagerly-closed, :client-error, :instance-error, :generic-error, or :server-eagerly-closed),
    - associated error message,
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -583,10 +589,13 @@
                         (recur bytes-streamed' bytes-reported-to-statsd'))))))))
           (catch Exception e
             (log/info e "exception occurred while streaming response for" service-id)
+            (log/info (let [cause (.getCause e)]
```
shamsimam, line 592: Let's remove this log?

shamsimam (dismissed):

waiter/src/waiter/process_request.clj
DaoWen: yikes! yes, I should have removed that

DaoWen: removed

shamsimam (approved):

PR #1425, opened 2021-08-16
Adds support for determining the creation time of a service

shamsimam:

- adds support for stats in local key-value store
- adds support for stats in file-based key-value store
- adds stats support to key-value store protocol
- adds `service-id->creation-time` function
- provides the `service-id->creation-time` function to the scheduler factory

Provides the scheduler the opportunity to make any validation and/or scheduler object creation decisions based on the service creation time.

### DIFF
--- a/waiter/src/waiter/kv.clj
+++ b/waiter/src/waiter/kv.clj
@@ -40,10 +40,14 @@
     copy is returned. This parameter is meant to be used by implementations
     that cache key-value pairs and is there to work around the Clojure
     limitation of not supporting variable arguments in defprotocol.")
+  (stats [this key]
+    "Fetches the stats map from the key value store corresponding to the provided key.
+     If the key is missing, nil will be returned.
+     Else a map containing the keys [:creation-time, :modified-time] is returned.")
   (store [this key value]
     "Store a value in the key value store.  If a key is already present,
     update the key's value with the supplied value.  There are no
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/kv.clj
```diff
@@ -69,13 +74,27 @@
...
   KeyValueStore
-  (retrieve [_ key _] (let [value (@store key)]
-                        (log/debug (str "(local) FETCH " key " => " (hashcode value)))
-                        value))
-  (store [_ key value] (do (log/debug (str "(local) STORE " key " => " (hashcode value)))
-                           (swap! store assoc key value)))
-  (delete [_ key] (do (log/debug (str "(local) DELETE " key))
-                      (swap! store dissoc key)))
+  (retrieve [_ key _]
+    (let [{:keys [value]} (@store key)]
```
kevo1ution, line 78: Is this a breaking change? When calling retrieve on previously stored values that weren't stored as `{:stats {}, :value}`.

Also separately, have we considered using zookeeper ctime and mtime of a znode?

kevo1ution: Disregard this, the zookeeper implementation uses the ctime and mtime, and we are okay with breaking changes to other kv-store implementations

kevo1ution (dismissed):

PR #1429, opened 2021-08-17
adds authorization to service logs display

shamsimam:

- adds authorization to service logs display

Allows restricting which users get to see the log URLs for a service.

### DIFF
--- a/waiter/integration/waiter/basic_test.clj
+++ b/waiter/integration/waiter/basic_test.clj
@@ -374,7 +374,35 @@
           (doseq [file-link [stderr-file-link stdout-file-link]]
             (if (str/starts-with? (str file-link) "http")
               (assert-response-status (make-request-fn file-link) http-200-ok)
-              (log/warn "test-basic-logs did not verify file link:" stdout-file-link))))))))
+              (log/warn "test-basic-logs did not verify file link:" stdout-file-link))))
+
+        (testing "x-view-as-user access"
+          (testing "restricted"
+            (let [request-headers {"x-view-as-user" "does-not-exist"}
+                  service-info (service-settings waiter-url service-id :cookies cookies :headers request-headers)
+                  active-instances (get-in service-info [:instances :active-instances 0])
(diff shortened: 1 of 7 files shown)


waiter/integration/waiter/basic_test.clj
```diff
@@ -374,7 +374,35 @@
           (doseq [file-link [stderr-file-link stdout-file-link]]
             (if (str/starts-with? (str file-link) "http")
               (assert-response-status (make-request-fn file-link) http-200-ok)
-              (log/warn "test-basic-logs did not verify file link:" stdout-file-link))))))))
+              (log/warn "test-basic-logs did not verify file link:" stdout-file-link))))
+
+        (testing "x-view-as-user access"
```
DaoWen, line 379: Would it be hard to add a couple more tests?

1. admins can read the logs
2. a user that does exist but isn't an admin or an owner can't the logs

shamsimam: The challenge is none of the current tests launch some as non-admin user.

> admins can read the logs

I have verified this manually. This is managed by the `allowed-to-manage-service?-fn` function: 
https://github.com/twosigma/waiter/blob/5a51d0639ead8e8bacea5b3182bd2f756e6bdb76/waiter/src/waiter/authorization.clj#L42-L45

If the underlying entitlement manager allows it (which at least one implementation does), an admin can see the logs.

We have unit tests for this: 
https://github.com/twosigma/waiter/blob/ed428d26f5de7c109f682407567d3d506b8a70c6/waiter/test/waiter/core_test.clj#L440-L450

> a user that does exist but isn't an admin or an owner can't the logs

This is similarly covered by the `allowed-to-manage-service?-fn` function.  Again, I've tested this manually, but hard to add tests for in our current setup.

We have unit tests for this: https://github.com/twosigma/waiter/blob/ed428d26f5de7c109f682407567d3d506b8a70c6/waiter/test/waiter/core_test.clj#L330-L341

DaoWen (approved):

PR #1431, opened 2021-08-18
Fix streaming timeout integration test to start timing from receiving initial event

kevo1ution:

- starts timing from when receiving initial event instead of when first making the watch request

- the streaming timeout on the server starts the timer from sending out the initial event

### DIFF
--- a/waiter/integration/waiter/instance_tracker_integration_test.clj
+++ b/waiter/integration/waiter/instance_tracker_integration_test.clj
@@ -102,7 +102,10 @@
                           InputStreamReader.
                           cheshire/parsed-seq)
         id->healthy-instance-atom (atom {})
-        query-state-fn (fn [] @id->healthy-instance-atom)
+        initial-event-time-epoch-ms-atom (atom nil)
+        query-state-fn (fn []
+                         {:initial-event-time-epoch-ms @initial-event-time-epoch-ms-atom
+                          :id->healthy-instance @id->healthy-instance-atom})
         exit-fn (fn []
                   (async/close! body))
         go-chan
@@ -114,6 +117,8 @@
                 id->healthy-instance-atom
                 (let [{:strs [object type]} msg
                       id->healthy-instance @id->healthy-instance-atom]
+                  (when (= "initial" type)
+                    (reset! initial-event-time-epoch-ms-atom (System/currentTimeMillis)))
                   (if (contains? #{"initial" "events"} type)
                     (let [updated-healthy-instances (get-in object ["healthy-instances" "updated"])
                           removed-healthy-instances (get-in object ["healthy-instances" "removed"])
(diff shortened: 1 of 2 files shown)


waiter/integration/waiter/instance_tracker_integration_test.clj
```diff
@@ -451,16 +456,17 @@
...
+          {:keys [id->healthy-instance initial-event-time-epoch-ms]} (query-state-fn)
           _ (exit-fn)
+          _ (is (some? initial-event-time-epoch-ms))
+          elapsed-time-ms (- end-time-epoch-ms initial-event-time-epoch-ms)
           assertion-message (str {:elapsed-time-ms elapsed-time-ms
                                   :headers headers})]
-      (is (empty? (query-state-fn)) assertion-message)
-      (is (<= streaming-timeout-ms elapsed-time-ms (+ streaming-timeout-ms 1000)) assertion-message))))
+      (is (empty? id->healthy-instance) assertion-message)
+      (is (<= (- streaming-timeout-ms 1000) elapsed-time-ms (+ streaming-timeout-ms 1000)) assertion-message))))
```
shamsimam, line 472: Why are we subtraction 1 second?

waiter/integration/waiter/token_watch_integration_test.clj
```diff
@@ -390,17 +395,18 @@
...
+          {:keys [initial-event-time-epoch-ms token->index]} (query-state-fn)
           _ (exit-fn)
+          _ (is (some? initial-event-time-epoch-ms))
+          elapsed-time-ms (- end-time-epoch-ms initial-event-time-epoch-ms)
           assertion-message (str {:elapsed-time-ms elapsed-time-ms
                                   :headers headers})]
-      (is (empty? (query-state-fn)) assertion-message)
-      (is (<= streaming-timeout-ms elapsed-time-ms (+ streaming-timeout-ms 1000)) assertion-message))))
+      (is (empty? token->index) assertion-message)
+      (is (<= (- streaming-timeout-ms 1000) elapsed-time-ms (+ streaming-timeout-ms 1000)) assertion-message))))
```
shamsimam, line 412: Why are we subtraction 1 second?

kevo1ution: because its now measuring the time from when the client received the initial event. That time is later than when the server sent the event because of some latency. And it's possible for that delay to be greater than the delay when the server closes the connection making the `elapsed-time-ms` slightly less than the `streaming-timeout-ms`. 

We expect the elapsed time to always be very close to the streaming-timeout.

waiter/integration/waiter/instance_tracker_integration_test.clj
kevo1ution: this has the same reasoning as below

shamsimam (approved):

PR #1438, opened 2021-09-01
Fix test-service-account-injection for default-namespace

DaoWen:

Fix test-service-account-injection for case when `default-namespace` ≠ `*`.

Fixes test broken in PR #1420.

### DIFF
--- a/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
+++ b/waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
@@ -313,9 +313,9 @@
           (is (thrown? Exception #"Cannot use run-as-requester with a specific namespace"
                        (check-pod-namespace waiter-url (assoc star-user-header :x-waiter-namespace not-current-user) current-user))))))))
 
-(defn- get-pod-service-account
+(defn- get-pod-service-account-info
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/kubernetes_scheduler_integration_test.clj
```diff
@@ -337,9 +337,12 @@
   (testing-using-waiter-url
     (when (using-k8s? waiter-url)
       (let [current-user (retrieve-username)]
-        (testing "Has service account with matching default namespace"
+        (testing "No service account for default namespace, or matches user"
           (let [service-account (get-pod-service-account waiter-url nil)]
-            (is (= current-user service-account))))
+            ;; blank when default namespace is set to a specific user
+            ;; matches run-as-user when default-namespace=*
+            (is (or (str/blank? service-account)
```
shamsimam, line 344: Should we have an if-else checking default namespace is `*` and then have separate assertions?

DaoWen: Changed to choose the assertion based on pod's `k8s/namespace` property.

shamsimam (approved):

PR #1443, opened 2021-09-07
adds support for periodically refreshing the statsd host IP

shamsimam:

- adds support for periodically refreshing the statsd host IP

Waiter routers should be resilient to the statsd host IPs changing, the client should regularly re-compute the IP.

### DIFF
--- a/waiter/src/waiter/statsd.clj
+++ b/waiter/src/waiter/statsd.clj
@@ -31,6 +31,11 @@
   [s]
   (str/replace s #"[^a-zA-Z\d\-_]" "_"))
 
+(defn resolve-ip
+  "Determines the IP address of a host, given the host's name."
+  [host-address]
+  (InetAddress/getByName host-address))
+
 (let [config (atom nil)
       socket-agent (agent nil)
       environment (atom nil)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/statsd.clj
```diff
@@ -43,10 +48,22 @@
...
+  (defn refresh-host-ip
+    "Attempts to refresh the host IP in the config atom.
+     Fails silently, with a warning log, if there is an error resolving the IP.
+     Returns the value of the config atom at the end of the operation."
+    [host]
+    (try
+      (when-let [host-ip (resolve-ip host)]
+        (swap! config assoc :host host-ip))
+      (catch Throwable e
+        (log/warn e "Error resolving IP for host" host)))
```
DaoWen, line 64: Nitpick: It would be nice to have a more clear error message here (explaining the context of this host→ip lookup), but the stack trace on the exception should provide that.

DaoWen (approved):

waiter/src/waiter/statsd.clj
```diff
@@ -156,6 +173,7 @@
             config-cluster (:cluster config-map)
             config-server (:server config-map)
             config-publish-interval-ms (:publish-interval-ms config-map)
+            config-refresh-interval-ms (get config-map :refresh-interval-ms 600000)
```
DaoWen, line 176: Nitpick: Prefer using `(-> 10 from-minutes to-millis)` (or whatever those time functions are called).

DaoWen (approved):

PR #1461, opened 2021-10-25
improves logic to determine when exclusive mode is enabled

shamsimam:

- improves logic to determine when exclusive mode is enabled in incoming requests and avoiding unintentional updates to the service description
- enabling exclusive mode also changing the integration tests to include the token in the service description environment

`WAITER_CONFIG_TOKEN` should only be configured when explicit tokens are in use (and not for service-ID tokens) and not be dependent on what the token defaults are (even if no tokens are being used). This resolves a bug that occurs when service-ID tokens (of the form `^SERVICE-ID#<target-service-id>`) get included in the `WAITER_CONFIG_TOKEN` environment, effectively changing the computed service ID and configuration (by removing the old token from the environment) and making the target service unreachable.

### DIFF
--- a/waiter/src/waiter/util/client_tools.clj
+++ b/waiter/src/waiter/util/client_tools.clj
@@ -1319,8 +1319,8 @@
          (-> jwt-config :oidc-authorize-uri str/blank? not)
          (-> jwt-config :oidc-token-uri str/blank? not))))
 
-(defn exclusive-mode?
-  "Get the authenticator that Waiter is configured to use"
+(defn exclusive-mode-by-default?
+  "Returns true if service-mapping is exclusive by default."
   [waiter-url]
   (-> (waiter-settings waiter-url)
     (get-in [:token-config :token-defaults :service-mapping])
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/util/client_tools.clj
```diff
@@ -1318,3 +1318,10 @@
          (map? jwt-config)
          (-> jwt-config :oidc-authorize-uri str/blank? not)
          (-> jwt-config :oidc-token-uri str/blank? not))))
+
+(defn exclusive-mode?
+  "Get the authenticator that Waiter is configured to use"
```
DaoWen, line 1323: Wrong docstring?

waiter/src/waiter/util/client_tools.clj
```diff
@@ -1318,3 +1318,10 @@
          (map? jwt-config)
          (-> jwt-config :oidc-authorize-uri str/blank? not)
          (-> jwt-config :oidc-token-uri str/blank? not))))
+
+(defn exclusive-mode?
```
DaoWen, line 1322: It might be more clear if this was named `exclusive-by-default?` or something like that, since that's what it actually checks (defaults).

waiter/src/waiter/util/client_tools.clj
shamsimam: Oops, fixed.

waiter/src/waiter/util/client_tools.clj
shamsimam: Renamed function.

DaoWen (approved):

PR #1467, opened 2021-11-04
adds support for ping after kill operation

shamsimam:

- adds support for ping after kill operation

We would like the mechanism to restart a service after killing all its current instances. The service owner understand that such an operation can cause in-flight requests to fail.

### DIFF
--- a/cli/waiter/subcommands/kill.py
+++ b/cli/waiter/subcommands/kill.py
@@ -36,7 +36,7 @@ def register(add_parser):
     ping_group.add_argument('--no-ping', action='store_false', dest='ping_token',
                             help='skips pinging the token/service; pinging the token is disabled by default')
     ping_group.add_argument('--ping', action='store_true', dest='ping_token',
-                            help='pings the token after stopping maintenance.')
+                            help='pings the token/service after killing service(s).')
     parser.set_defaults(ping_token=False)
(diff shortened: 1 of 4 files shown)


cli/waiter/subcommands/kill.py
```diff
@@ -22,4 +31,20 @@ def register(add_parser):
                         dest='is-service-id', action='store_true')
     parser.add_argument('--timeout', '-t', help='timeout (in seconds) for kill to complete',
                         type=check_positive, default=30)
+
+    ping_group = parser.add_mutually_exclusive_group(required=False)
+    ping_group.add_argument('--no-ping', action='store_false', dest='ping_token',
+                            help='skips pinging the token/service; pinging the token is disabled by default')
+    ping_group.add_argument('--ping', action='store_true', dest='ping_token',
+                            help='pings the token after stopping maintenance.')
```
kevo1ution, line 39: this isn't related to maintenance mode; we should change this to something like:
```suggestion
                            help='pings the token after killing services.')
```

kevo1ution (changes requested):

cli/waiter/subcommands/kill.py
shamsimam: Updated message to:
```
pings the token/service after killing service(s).
```

kevo1ution (approved):

PR #1468, opened 2021-11-06
Token syncer: sync when user editable token fields are the same even when mismatching metadata fields

kevo1ution:

- Previously, we skipped syncing for the case where same user edited token on different roots (e.g. clusters), when they have the same user editable token fields (cpus, mem, etc.) irrespective of system metadata differences (cluster, last-updated-time, etc.). This change makes it so that we still sync the token in this case and choose the latest token description
- Previously, we errored when syncing  for the case where different users edited the the latest token and current token, when they have the same user editable token fields (cpus, mem, etc.) irrespective of system metadata differences (cluster, last-updated-time, etc.). This change makes it so that we still sync the token in this case and choose the latest token description.

- this decreases the number of token syncer errors on two scenarios we think its safe to choose the latest description even when different users made the changes or when the root is different.
- the previous sync strategy caused errors to propagate later when one of the tokens would be edited causing the sync operation to fail due to cluster/author differences.

### DIFF
--- a/token-syncer/src/token_syncer/commands/syncer.clj
+++ b/token-syncer/src/token_syncer/commands/syncer.clj
@@ -78,75 +78,70 @@
   [{:keys [store-token]} cluster-urls token latest-token-description opt-out-metadata-name cluster-url->token-data]
   (pc/map-from-keys
     (fn [cluster-url]
-      (let [ignored-root-mismatch-equality-comparison-keys ["cluster" "last-update-time" "last-update-user" "previous" "root"]
-            system-metadata-keys (conj ignored-root-mismatch-equality-comparison-keys "deleted")
+      (let [
+            ;; don't include "deleted" system metadata key, this must be considered for determining root-mismatch
+            ignored-root-mismatch-equality-comparison-keys ["cluster" "last-update-time" "last-update-user" "previous" "root"]
             cluster-result
             (try
               (let [{:keys [description error status] :as token-data} (get cluster-url->token-data cluster-url)
                     {latest-root "root" latest-update-user "last-update-user"} latest-token-description
-                    {cluster-root "root" cluster-update-user "last-update-user"} description]
-                (cond
-                  error
-                  {:code :error/token-read
-                   :details {:message (.getMessage error)}}
+                    {cluster-root "root" cluster-update-user "last-update-user"} description
+                    {:keys [code] :as result}
+                    (cond
+                      error
+                      {:code :error/token-read
+                       :details {:message (.getMessage error)}}
 
-                  (nil? status)
-                  {:code :error/token-read
-                   :details {:message "status missing from response"}}
+                      (nil? status)
+                      {:code :error/token-read
+                       :details {:message "status missing from response"}}
 
-                  (nil? latest-root)
-                  {:code :error/token-read
-                   :details {:message "token root missing from latest token description"}}
+                      (nil? latest-root)
+                      {:code :error/token-read
+                       :details {:message "token root missing from latest token description"}}
 
-                  (and latest-root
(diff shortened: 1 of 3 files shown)


token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -78,75 +78,71 @@
   [{:keys [store-token]} cluster-urls token latest-token-description opt-out-metadata-name cluster-url->token-data]
   (pc/map-from-keys
     (fn [cluster-url]
-      (let [ignored-root-mismatch-equality-comparison-keys ["cluster" "last-update-time" "last-update-user" "previous" "root"]
-            system-metadata-keys (conj ignored-root-mismatch-equality-comparison-keys "deleted")
+      (let [
```
shamsimam, line 81: unnecessary new line

token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -78,75 +78,71 @@
...
-                             (if (utils/successful? response) :success/sync-update :error/sync-update))
-                     :details (cond-> {:status status}
-                                      (utils/successful? response) (assoc :etag (get headers "etag")))})
-
-                  :else
-                  {:code :success/token-match}))
+                      :else
+                      {:code :success/token-match})]
+                ;; log full token descriptions when there was a syncing error detected
+                (when (str/starts-with? (name code) "error")
```
shamsimam, line 141:
```suggestion
                (when (= (namespace code) "error")
```

token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -78,75 +78,71 @@
...
-                  (and (seq description)
-                       (not= latest-root cluster-root)
-                       (not (get latest-token-description "deleted"))
-                       (not (get description "deleted"))
-                       (= (apply dissoc latest-token-description system-metadata-keys)
-                          (apply dissoc description system-metadata-keys)))
-                  (if (= latest-update-user cluster-update-user)
-                    {:code :success/skip-token-sync}
-                    {:code :error/token-sync
-                     :details {:message "token contents match, but were edited by different users"}})
```
shamsimam, line 127: We should identify and log this instance as we are wiping out history from the target cluster.

token-syncer/src/token_syncer/commands/syncer.clj
```diff
@@ -78,75 +78,71 @@
...
+                      ;; token user-specified content (including "deleted") is different, and the last update user or root is different
+                      (and (seq description)
+                           (not= latest-root cluster-root)
+                           (not= latest-update-user cluster-update-user)
```
shamsimam, line 122: We are okay if `root` and `cluster` do not match now as long as rest of the content matches. Should the two lines above form an `or` clause?

kevo1ution: This is a different situation where the contents don't match. Later in the `and` call:
```
(not= (apply dissoc description ignored-root-mismatch-equality-comparison-keys)
                                 (apply dissoc latest-token-description ignored-root-mismatch-equality-comparison-keys))
```

token-syncer/src/token_syncer/commands/syncer.clj
kevo1ution: I logged the situation in the portion where we attempt to sync the token

token-syncer/src/token_syncer/commands/syncer.clj
shamsimam: Oops, my bad I read one of the terms as dealing with `cluster` instead of `last-update-user`

shamsimam (approved):

PR #1470, opened 2021-11-09
computes the service-id using compute-descriptor function in consent handler

shamsimam:

- computes the service-id using compute-descriptor function in consent handler

The consent handler incorrectly computes service-IDs when the `compute-descriptor` function dynamically adds parameters to the service description (e.g. when service-mapping is exclusive).

### DIFF
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1067,6 +1067,14 @@
    :refresh-service-descriptions-fn (pc/fnk [[:state kv-store]]
                                       (fn refresh-service-descriptions-fn [service-ids]
                                         (sd/refresh-service-descriptions kv-store service-ids)))
+   :request->consent-service-id (pc/fnk [[:state kv-store service-description-builder service-id-prefix waiter-hostnames]
+                                         attach-service-defaults-fn attach-token-defaults-fn]
+                                  (fn request->consent-service-id [request]
+                                    (let [assoc-run-as-user-approved? (constantly true)]
+                                      (-> (descriptor/compute-descriptor
+                                            attach-service-defaults-fn attach-token-defaults-fn service-id-prefix kv-store waiter-hostnames request
+                                            service-description-builder assoc-run-as-user-approved?)
+                                        :service-id))))
    :request->descriptor-fn (pc/fnk [[:settings [:token-config history-length]]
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -1062,9 +1060,7 @@
       (when-not (seq service-description-template)
         (throw (ex-info "Unable to load description for token" {:log-level :info :status http-404-not-found :token token})))
       (let [auth-user (:authorization/user request)
-            service-id (-> service-description-template
-                           (sd/assoc-run-as-requester-fields auth-user)
-                           service-description->service-id)
+            service-id (request->consent-service-id request)
```
DaoWen, line 1063: Why is service-id computed separately so many places???

waiter/src/waiter/core.clj
```diff
@@ -1067,6 +1067,14 @@
    :refresh-service-descriptions-fn (pc/fnk [[:state kv-store]]
                                       (fn refresh-service-descriptions-fn [service-ids]
                                         (sd/refresh-service-descriptions kv-store service-ids)))
+   :request->consent-service-id (pc/fnk [[:state kv-store service-description-builder service-id-prefix waiter-hostnames]
```
DaoWen, line 1070: Why are we defining a separate function for the *consent* service-id? Shouldn't we just have one consistent way to compute the service-id? It seems like maintaining separate functions is setting us up for more inconsistency in the future.

DaoWen (approved):
I'm merging because this still puts us in a better situation that we're in now (bug fixed) — but we should probably address the underlying problem of computing the service-id in multiple places without a single consistent function.

waiter/src/waiter/core.clj
shamsimam: This gap is reduced in the new way, as now we use the common `compute-descriptor` with an override for allowing run-as-requester without permission (the consent workflow is supposed to give us the permission) to get the service ID.

waiter/src/waiter/handler.clj
shamsimam: The two places it is computed in the consent worflow:

1. Computed to include the HTML form we generate to render the consent form
2. Computed to confirm that the service-ID sent in the form submission matches the service we are trying to launch.

PR #1475, opened 2021-11-24
moves service-mapping logic to ServiceDescriptionBuilder build() function

shamsimam:

- moves service-mapping logic to ServiceDescriptionBuilder build() function

Moving the service-mapping computation into the build() function allows delaying the decision of how the service-mapping ensures a unique service-id per token and it also delays what gets presented as the `core-service-description`. The refactoring also helps implement other service mapping schemes which may or may not affect the service-id computation.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -856,7 +856,7 @@
                   service-id-prefix source-tokens token-sequence token-service-mapping username]}]
     (let [core-service-description
           (cond-> user-service-description
-            ;; each unique token permutation will create a unique service
+            ;; include token name in the service description to enforce unique token->service mappings
             ;; leverage WAITER_CONFIG_ prefixed environment variables being allowed
             (= "exclusive" token-service-mapping)
             (assoc-token-in-env token-sequence)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -831,23 +831,37 @@
...
-            (let [candidate-service-description (assoc-run-as-requester-fields user-service-description username)
-                  candidate-service-id (service-description->service-id service-id-prefix candidate-service-description)]
-              (if (assoc-run-as-user-approved? candidate-service-id)
-                (do
-                  (log/debug "appending run-as-user into pre-approved service" candidate-service-id)
-                  candidate-service-description)
-                user-service-description)))
+          (cond-> user-service-description
+            ;; each unique token permutation will create a unique service
+            ;; leverage WAITER_CONFIG_ prefixed environment variables being allowed
```
DaoWen, line 860: Please rephrase this comment. I don't think that [_permutation_](https://en.wiktionary.org/wiki/permutation#English) is used accurately here.

shamsimam: Updated the comment

DaoWen (approved):

PR #1477, opened 2021-11-27
adds descriptor entry to identify services using a run-as-user different from token's run-as-user

shamsimam:

- adds descriptor entry to identify services using a runs-as-user different from token's run-as-user
- populates the run-as-user-source as an annotation in replicaset specs

We would like to be able to identify services running as a user (e.g. run-as-requester use-cases) other than the `run-as-user` specified on the token. We then populate the value in the descriptor and persist this information as an annotation on Kubernetes pods.

We can only know a service has received its `run-as-user` from the token if it is produced via an exclusive mapping. For legacy use-cases, it is possible for an on-the-fly service to have the same service description as one generated via a token and we cannot know with certainty whether the run-as-user has changed.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -865,7 +865,7 @@
   ServiceDescriptionBuilder
 
   (build [this user-service-description
-          {:keys [assoc-run-as-user-approved? component->last-update-epoch-time component->previous-descriptor-fns reference-type->entry
+          {:keys [assoc-run-as-user-approved? component->last-update-epoch-time component->previous-descriptor-fns reference-type->entry run-as-user-changed?
                   service-id-prefix source-tokens token-sequence token-service-mapping username]}]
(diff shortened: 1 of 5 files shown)


DaoWen (approved):

waiter/src/waiter/service_description.clj
```diff
@@ -884,10 +884,14 @@
           service-description (compute-effective this service-id core-service-description)
           reference-type->entry (cond-> (or reference-type->entry {})
                                   (seq source-tokens)
-                                  (assoc :token {:sources (map walk/keywordize-keys source-tokens)}))]
+                                  (assoc :token {:sources (map walk/keywordize-keys source-tokens)}))
+          ;; only for exclusive mapping can we be sure that the service will always have the run-as-user coming from the token
+          ;; for legacy mapping services, it is possible for the run-as-user to come from the token or the request
```
DaoWen, line 889: Thanks for adding this comment. It was helpful for clarifying the condition below.

napple (approved):

PR #1478, opened 2021-11-29
adds support for default service-mapping that dynamically determines when to use exclusive mode

shamsimam:

- adds support for a new `default` service-mapping option
- adds the `component->last-update-epoch-time` argument to ServiceDescriptionBuilder#build()
- enables exclusive mode for new versions of services beyond a cutoff date (configured via `service-mapping-exclusive-start-time`)
- adds the `exclusive-promotion-start-time` configuration to settings

When we change the default behavior for `service-mapping`, we want existing services to continue to run unaffected (without their service IDs changing to prevent launching bunch of new services). However, we want new services to launch with `exclusive` mode if `service-mapping`.

## How are we affecting the change?

There is a new option for service-mapping: `default` which allows the system to choose whether to use `legacy` mapping or `exclusive` mapping. For "old" versions of services, which we identify using the update time new `component->last-update-epoch-time` dictionary we use `legacy` mode. This allows current services IDs to continue to be valid in the `default` mode. For "new" versions of services (beyond a cut-off date specified by the `exclusive-promotion-start-time` configuration), we switch the service-mapping to `exclusive` mode.

### DIFF
--- a/waiter/src/waiter/config.clj
+++ b/waiter/src/waiter/config.clj
@@ -13,7 +13,9 @@
 ;; See the License for the specific language governing permissions and
 ;; limitations under the License.
 ;;
-(ns waiter.config)
+(ns waiter.config
+  (:require [clj-time.coerce :as tc]
+            [waiter.util.date-utils :as du]))
 
 (def ^:private config-promise (promise))
 
@@ -36,3 +38,11 @@
   []
   {:pre [(realized? config-promise)]}
   (get-in @config-promise [:waiter-principal]))
+
+(defn retrieve-exclusive-promotion-start-epoch-time
+  "Retrieves the configured (or system default) exclusive promotion start epoch time."
+  []
+  {:pre [(realized? config-promise)]}
+  (let [default-exclusive-promotion-start-time "2022-02-01T00:00:00.000Z"
+        exclusive-promotion-start-time (get-in @config-promise [:token-config :exclusive-promotion-start-time] default-exclusive-promotion-start-time)]
+    (-> exclusive-promotion-start-time (du/str-to-date) (tc/to-long))))
--- a/waiter/src/waiter/config.clj
+++ b/waiter/src/waiter/config.clj
@@ -43,6 +43,6 @@
   "Retrieves the configured (or system default) exclusive promotion start epoch time."
   []
   {:pre [(realized? config-promise)]}
-  (let [default-exclusive-promotion-start-time "2022-02-01T00:00:00.000Z"
+  (let [default-exclusive-promotion-start-time "2050-06-01T00:00:00.000Z"
         exclusive-promotion-start-time (get-in @config-promise [:token-config :exclusive-promotion-start-time] default-exclusive-promotion-start-time)]
     (-> exclusive-promotion-start-time (du/str-to-date) (tc/to-long))))
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -26,6 +26,7 @@
             [schema.core :as s]
             [slingshot.slingshot :as sling]
             [waiter.authorization :as authz]
+            [waiter.config :as config]
             [waiter.headers :as headers]
             [waiter.kv :as kv]
             [waiter.metrics :as metrics]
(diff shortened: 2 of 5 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -831,10 +832,26 @@
...
+  (let [exclusive-promotion-start-epoch-time (config/retrieve-exclusive-promotion-start-epoch-time)]
+    (< exclusive-promotion-start-epoch-time reference-update-epoch-time)))
+
+(defn adjust-waiter-config-token
+  "Dissociates the exclusive mapping token from the environment variable if it was added as part of default promotion after the promotion start time."
+  [service-description token-service-mapping reference-update-epoch-time]
+  (cond-> service-description
+    (and (= "default" token-service-mapping)
+         (not (promote-default-to-exclusive? reference-update-epoch-time)))
+    (utils/dissoc-in waiter-config-token-path)))
```
DaoWen, line 854: Is it possible that this will change the service-id? I've seen services where `env: {}`, but I think this would change those service descriptions to delete the `env` entirely.

waiter/src/waiter/config.clj
```diff
@@ -36,3 +38,11 @@
   []
   {:pre [(realized? config-promise)]}
   (get-in @config-promise [:waiter-principal]))
+
+(defn retrieve-exclusive-promotion-start-epoch-time
+  "Retrieves the configured (or system default) exclusive promotion start epoch time."
+  []
+  {:pre [(realized? config-promise)]}
+  (let [default-exclusive-promotion-start-time "2022-02-01T00:00:00.000Z"
```
DaoWen, line 46: This date seems oddly specific for the open-source default... Can we have the default be "never" instead? Maybe set it to the year 3000?

waiter/test/waiter/service_description_test.clj
```diff
@@ -996,6 +1015,193 @@
...
+                :on-the-fly? nil
+                :reference-type->entry {:token {:sources (seq (source-tokens-entry test-token token-data))}}
+                :service-authentication-disabled false
+                :service-description (merge service-description-defaults service-description-template)
+                :service-id (service-description->service-id service-id-prefix service-description-template)
+                :service-preauthorized true
+                :source-tokens (source-tokens-entry test-token token-data)}
+               result-map))))
+
+    (testing "exclusive service-mapping on token with run-as-requester"
```
DaoWen, line 1186: Haven't we determined that this shouldn't be allowed? It seems like we need to address this, and then have a test here specifically verifying that we don't accidentally re-enable `WAITER_CONFIG_TOKEN` when service-mapping=default.

DaoWen: OK, discussed this with Shams. Rather than _disallowing_ this case, we're just going to provide the backend with additional information so that it can make informed decisions (e.g., regarding fetching tls certificates or other things that might be tricky with run-as-requester instances).

waiter/src/waiter/service_description.clj
shamsimam: Nice observation, I will remove the `env` if it becomes empty.

waiter/src/waiter/service_description.clj
```diff
@@ -831,10 +832,26 @@
...
-  (assoc-in service-description ["env" "WAITER_CONFIG_TOKEN"] (str/join "," token-sequence)))
+  (assoc-in service-description waiter-config-token-path (str/join "," token-sequence)))
+
+(defn- promote-default-to-exclusive?
+  "Determines if a service description should be promoted to exclusive mode based on reference last update time."
+  [reference-update-epoch-time]
+  (let [exclusive-promotion-start-epoch-time (config/retrieve-exclusive-promotion-start-epoch-time)]
+    (< exclusive-promotion-start-epoch-time reference-update-epoch-time)))
+
+(defn adjust-waiter-config-token
```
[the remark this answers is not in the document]
shamsimam: Removed the function.

waiter/src/waiter/config.clj
shamsimam: CHanged to `2050-06-01T00:00:00.000Z`

DaoWen (approved):

PR #1485, opened 2021-12-06
adds support for loading parameters in /tokens endpoint

shamsimam:

- adds support for loading parameters in /tokens endpoint

We would like to avoid an N+1 query pattern to perform a search on tokens and then another set of token retrievals to load information of specific fields.

### DIFF
--- a/waiter/src/waiter/token.clj
+++ b/waiter/src/waiter/token.clj
@@ -646,6 +646,28 @@
                          :streaming-timeout streaming-timeout})]
         (utils/exception->response ex request)))))
 
+(defn extract-token-data-param-keys
+  "Extracts the names of parameters from the list of include parameters."
+  [include-params]
+  (->> include-params
+    (filter #(str/starts-with? (str %) "param."))
+    (set)
+    (map #(-> % (str/split #"\.") (rest) (vec)))
+    (filter #(contains? sd/token-data-keys (first %)))
+    (vec)))
+
+(defn extract-token-data-param-vals
+  "Extracts the values of parameters from the provided token description."
+  [token-description display-params-keys]
+  (loop [[entry-keys & remaining-keys] display-params-keys
+         result-map {}]
+    (if entry-keys
+      (recur remaining-keys
+             (let [entry-val (get-in token-description entry-keys)]
+               (cond-> result-map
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/token.clj
```diff
@@ -668,6 +690,12 @@
                           (string? owner-param) #{owner-param}
                           (coll? owner-param) (set owner-param)
                           :else (list-token-owners kv-store))
+                 include-param (get request-params "include")
```
kevo1ution, line 693: It seems that the `include` query parameter will have several meanings. It is used for `metadata`, `deleted`, `service-id`, `cluster`; maybe we can create a new query field name instead? Does `fields` sound okay? so:
`/tokens?fields=metadata` will provide token `{"metadata": {}, "token":"kitchen", ...other default index fields`, and still includes the default fields.

Here is an example `fields` interface: https://developers.facebook.com/docs/graph-api/overview#fields

kevo1ution: The response will not have a nested `parameters` response anymore

shamsimam: We spoke offline and decided to use the `parameters` query parameter. We did not go with a flat map due to the conflict with the `maintenance` field already displayed in the `/tokens` output today.

waiter/src/waiter/token.clj
```diff
@@ -646,6 +646,27 @@
...
+  (->> parameters
+    (set)
+    (map #(-> % (str/split #"\.") (vec)))
+    (filter #(contains? sd/token-data-keys (first %)))
+    (vec)))
+
+(defn extract-token-data-param-vals
+  "Extracts the values of parameters from the provided token description."
+  [token-description display-params-keys]
+  (loop [[entry-keys & remaining-keys] display-params-keys
```
kevo1ution, line 661: nitpick: This can be a `reduce`

kevo1ution (dismissed):

waiter/src/waiter/token.clj
shamsimam: Changed `loop` to `reduce`.

kevo1ution (approved):

PR #1494, opened 2021-12-21
triggers token updates if the cluster field changes

shamsimam:

- triggers token updates if the cluster field changes

We should update the cluster if the values do not match since a user sending a request to a specific cluster is signaling intent to update the cluster parameter.

### DIFF
--- a/waiter/test/waiter/token_test.clj
+++ b/waiter/test/waiter/token_test.clj
@@ -637,9 +637,9 @@
             (doseq [key sd/service-parameter-keys]
               (is (= (get service-description-2 key) (get body-map key))
                   (str {:body-map body-map :service-description service-description-2})))
-            (doseq [key (-> required-metadata-keys (conj "cluster") (disj "deleted"))]
+            (doseq [key (disj required-metadata-keys "deleted")]
               (is (contains? body-map key) (str "Missing entry for " key " in " body-map)))
-            (doseq [key (-> optional-metadata-keys (disj "cluster") (conj "deleted"))]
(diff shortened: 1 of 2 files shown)


waiter/test/waiter/token_test.clj
```diff
@@ -814,19 +851,24 @@
             (let [body-map (-> body str json/read-str)]
               (doseq [key sd/service-parameter-keys]
                 (is (= (get service-description key) (get body-map key))))
-              (doseq [key (disj sd/user-metadata-keys "stale-timeout-mins")]
-                (is (= (get service-description key) (get body-map key))))
+              (doseq [key (disj sd/user-metadata-keys "cluster" "stale-timeout-mins")]
```
kevo1ution, line 854: adding `cluster` here seems unnecessary, because the sd/user-metadata-keys do not contain `cluster`. `cluster` should remain a system metadata key.

kevo1ution (changes requested):

waiter/test/waiter/token_test.clj
shamsimam: The cluster is showing up because it is stored in the `kv-store` from line #521 above.

shamsimam: Actually, you're right. This should not be showing up in the response. I may have added this incorrectly. Checking...

shamsimam: Removed unnecessary operations with the cluster key from the tests.

kevo1ution (approved):

PR #1497, opened 2021-12-23
CLI subcommand delete respects cluster groups

kevo1ution:
- deleting a token in a single cluster group will delete the token on the primary cluster and rely on the token syncer to sync the delete. Subsequent deletes for the same token yield a no-op
```
$ cat .waiter.json
{
  "clusters": [
    {
      "name": "WAITER1",
      "url": "http://127.0.0.1:9091/",
      "disabled": false,
      "sync-group": "production",
      "default-for-create": true
    },
    {
      "name": "WAITER2",
      "url": "http://127.0.0.1:9191/",
      "sync-group": "production",
      "disabled": false
    }
  ],
  "metrics": {
    "disabled": true,
    "host": "localhost",
    "port": 8125,
    "line-formats": {
      "count": "{namespace}.{name}:{value}|c"
    }
  }
$ waiter create token --cpus 1 
Successfully created token.
$ waiter --cluster waiter2 create token --cpus 1 
Attempting to create token on WAITER2...
Successfully created token.
$ waiter delete token
Deleting token token in WAITER2...
Successfully deleted token in WAITER2.
$ waiter delete token
No matching data found in WAITER1 / WAITER2.
Do you need to add another cluster to your configuration?
```
- when deleting token in multiple cluster groups, there is an error that is consistent with updating a token in multiple cluster groups
```
$ cat .waiter.json 
{
  "clusters": [
    {
      "name": "WAITER1",
      "url": "http://127.0.0.1:9091/",
      "disabled": false,
      "sync-group": "production",
      "default-for-create": true
    },
    {
      "name": "WAITER2",
      "url": "http://127.0.0.1:9191/",
      "disabled": false
    }
  ],
  "metrics": {
    "disabled": true,
    "host": "localhost",
    "port": 8125,
    "line-formats": {
      "count": "{namespace}.{name}:{value}|c"
    }
  }
}
$ waiter show token

=== WAITER1 / token ===

Last Updated: 8 minutes ago (kevin)

Owner  kevin
CPUs   1.0

<No command specified>


=== WAITER2 / token ===

Last Updated: 5 minutes ago (kevin)

Owner  kevin
CPUs   1.0

<No command specified>

$ waiter delete token
Could not infer the target cluster for this operation because there are multiple cluster groups that contain a description for this token: groups-{0, 'production'} clusters-{'WAITER2', 'WAITER1'}.
Consider specifying with the --cluster flag which cluster you are targeting.
```
- when using the `--force` flag, the token will be deleted on all clusters regardless of cluster group (this is the old behavior and will prevent backwards breaking change)
- if any of the queried tokens has the `waiter-token-sync-opt-out` metadata configured to `true`, then ask the user if they want to delete the token for each cluster

- handle cluster groups and token syncing to provide a simpler interface when deleting the token

### DIFF
--- a/cli/waiter/subcommands/delete.py
+++ b/cli/waiter/subcommands/delete.py
@@ -86,6 +86,7 @@ def delete(clusters, args, _, enforce_cluster):
         cluster = get_target_cluster_from_token(clusters, token_name, enforce_cluster)
         if cluster is None:
             print_no_data(clusters)
+            return 1
         else:
             token_config = query_result['clusters'][cluster['name']]
             token_etag = token_config['etag']
--- a/cli/waiter/subcommands/delete.py
+++ b/cli/waiter/subcommands/delete.py
@@ -65,13 +65,7 @@ def delete(clusters, args, _, enforce_cluster):
     sync_opt_out = any(get_in(token_config, ['token', 'metadata', 'waiter-token-sync-opt-out']) == 'true'
                        for _, token_config in cluster_data_pairs)
 
-    if num_clusters == 1:
-        cluster_name = cluster_data_pairs[0][0]
-        token_etag = cluster_data_pairs[0][1]['etag']
-        cluster = clusters_by_name[cluster_name]
-        success = delete_token_on_cluster(cluster, token_name, token_etag)
-        return 0 if success else 1
-    elif force_delete or sync_opt_out:
+    if force_delete or sync_opt_out:
         cluster_data_pairs = sorted(query_result['clusters'].items())
         clusters_by_name = {c['name']: c for c in clusters}
         num_clusters = len(cluster_data_pairs)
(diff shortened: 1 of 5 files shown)


cli/waiter/subcommands/delete.py
```diff
@@ -61,26 +62,36 @@ def delete(clusters, args, _, __):
...
         for cluster_name, data in cluster_data_pairs:
-            if args.get('force', False):
+            if force_delete or (num_clusters == 1):
                 should_delete = True
             else:
                 should_delete = str2bool(input(f'Delete token in {terminal.bold(cluster_name)}? '))
             if should_delete:
                 cluster = clusters_by_name[cluster_name]
                 success = delete_token_on_cluster(cluster, token_name, data['etag'])
                 overall_success = overall_success and success
```
shamsimam, line 84: We are missing a return after this `for` loop inside the `if` block.

kevo1ution: good catch. The overall_success should be returned, I'm going to add a test case to confirm this too

shamsimam (approved):

PR #1504, opened 2022-01-12
fixes handling of grpc-status headers

shamsimam:

- fixes propagation of grpc-status headers as trailers in grpc responses

Without propagation of the grpc-status in trailers, gRPC clients complain about gRPC responses.

### DIFF
--- a/waiter/src/waiter/process_request.clj
+++ b/waiter/src/waiter/process_request.clj
@@ -658,7 +658,7 @@
   (if trailers
     (let [correlation-id (cid/get-correlation-id)
           trailers-copy-ch (async/chan 1)
-          {:keys [grpc-status] :as grpc-headers} (select-keys headers ["grpc-message" "grpc-status"])]
+          {:strs [grpc-status] :as grpc-headers} (select-keys headers ["grpc-message" "grpc-status"])]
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/process_request.clj
```diff
@@ -658,7 +658,7 @@
   (if trailers
     (let [correlation-id (cid/get-correlation-id)
           trailers-copy-ch (async/chan 1)
-          {:keys [grpc-status] :as grpc-headers} (select-keys headers ["grpc-message" "grpc-status"])]
+          {:strs [grpc-status] :as grpc-headers} (select-keys headers ["grpc-message" "grpc-status"])]
```
shamsimam, line 661: The `num-messages=0` test fails without this fix.

kevo1ution: the original looks like a bug too. `grpc-status` will be nil because the second argument: `["grpc-message" "grpc-status"]` will cause the resulting map to never have the `:grpc-status` defined

kevo1ution (approved):

waiter/src/waiter/process_request.clj
shamsimam: Yeah, that was the bug :(

PR #1507, opened 2022-01-19
adds deep merge support for default env and metadata

shamsimam:

- adds deep merge support for default env and metadata

The previous shallow merging of `env` and `metadata` parameters caused the service description defaults and profile defaults to not be propagated to service configurations without being explicitly overridden. Instead, we want to _change behavior_ where deep-merge causes the `env` and `metadata` to be propagated and only explicit configuration causing the default values to be overriden.

### DIFF
--- a/waiter/src/waiter/service_description.clj
+++ b/waiter/src/waiter/service_description.clj
@@ -149,6 +149,9 @@
 (def ^:const service-parameter-keys
   (set/union service-override-keys service-non-override-keys))
 
+; keys in service description metadata that need a deep merge
+(def ^:const service-deep-merge-keys #{"env" "metadata"})
+
 ; keys allowed in service description metadata, these need to be distinct from service parameter keys
 (def ^:const service-metadata-keys #{})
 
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -1045,14 +1063,14 @@
...
-(defn- token-sequence->merged-data
+(defn token-sequence->merged-data
   "Computes the merged token-data using the provided token-sequence and token->token-data mapping.
    It removes the metadata keys from the returned result."
   [token->token-data token-sequence]
   (loop [loop-token-data {}
          [token & remaining-tokens] token-sequence]
     (if token
-      (recur (merge loop-token-data (token->token-data token))
+      (recur (merge-parameters loop-token-data (token->token-data token))
```
DaoWen, line 1073: You're just updating the existing loop/recur here, so it's probably better *not* to change this, but...

It makes me sad to see loop/recur in Clojure code when there's a higher-order function that would work great in its place. E.g., in this case, we really should have used `reduce` to merge all of the token data.

DaoWen (approved):

kevo1ution (approved):

PR #1511, opened 2022-02-14
removes OIDC auth cookies from request sent to the backend

shamsimam:

- removes OIDC auth cookies from request sent to the backend

The backend is not expected to process OIDC auth cookies generated by Waiter, we should avoid sending them to the backend.

### DIFF
--- a/waiter/src/waiter/auth/authentication.clj
+++ b/waiter/src/waiter/auth/authentication.clj
@@ -29,7 +29,13 @@
 
 (def ^:const AUTH-COOKIE-EXPIRES-AT "x-auth-expires-at")
 
+(def ^:const AUTH-COOKIE-EXPIRES-AT-ENTRY-PREFIX (str AUTH-COOKIE-EXPIRES-AT "="))
+
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
+
+(def ^:const AUTH-COOKIE-NAME-ENTRY-PREFIX (str AUTH-COOKIE-NAME "="))
+
+(def ^:const OIDC-CHALLENGE-COOKIE-PREFIX "x-waiter-oidc-challenge-")
 
 (def ^:const auth-expires-at-uri "/.well-known/auth/expires-at")
 
@@ -157,9 +163,9 @@
 (defn remove-auth-cookie
   "Removes the auth cookies"
   [cookie-string]
-  (-> cookie-string
-    (cookie-support/remove-cookie AUTH-COOKIE-EXPIRES-AT)
-    (cookie-support/remove-cookie AUTH-COOKIE-NAME)))
+  (when cookie-string
+    (->> [AUTH-COOKIE-EXPIRES-AT-ENTRY-PREFIX AUTH-COOKIE-NAME-ENTRY-PREFIX OIDC-CHALLENGE-COOKIE-PREFIX]
+      (cookie-support/remove-cookies cookie-string))))
 
 (defn select-auth-header
   "Filters and return the first authorization header that passes the predicate."
--- a/waiter/test/waiter/auth/oidc_test.clj
+++ b/waiter/test/waiter/auth/oidc_test.clj
@@ -696,3 +696,33 @@
     (is (false? (too-many-oidc-challenge-cookies? {:headers {"cookie" cookie-str}} 6)))
     (is (true? (too-many-oidc-challenge-cookies? {:headers {"cookie" cookie-str}} 5)))
     (is (true? (too-many-oidc-challenge-cookies? {:headers {"cookie" cookie-str}} 1)))))
+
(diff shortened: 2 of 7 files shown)


waiter/src/waiter/auth/authentication.clj
```diff
@@ -29,8 +29,14 @@
 
 (def ^:const AUTH-COOKIE-EXPIRES-AT "x-auth-expires-at")
 
+(def ^:const AUTH-COOKIE-EXPIRES-AT-ENTRY-PREFIX (str AUTH-COOKIE-EXPIRES-AT "="))
+
 (def ^:const AUTH-COOKIE-NAME "x-waiter-auth")
 
+(def ^:const AUTH-COOKIE-NAME-ENTRY-PREFIX (str AUTH-COOKIE-NAME "="))
+
+(def ^:const OIDC-CHALLENGE-COOKIE-PREFIX "x-waiter-oidc-challenge-")
```
shamsimam, line 38: Unfortunately, we have to share the OIDC cookie name in this file but couldn't come up with a better alternative that would avoid additional computation.

waiter/test/waiter/auth/oidc_test.clj
```diff
@@ -696,3 +696,33 @@
...
+           {:cookie-string "a=b; x-waiter-oidc-challenge-a1b2c3=123; x-waiter-oidc-challenge-3cb2a1=321; c=d" :expected "a=b; c=d"}
+           {:cookie-string "a=b; x-waiter-oidc-challenge-a1b2c3=123; c=d; x-waiter-oidc-challenge-3cb2a1=321" :expected "a=b; c=d"}
+           {:cookie-string "a=b; c=d; x-waiter-oidc-challenge-a1b2c3=123; x-waiter-oidc-challenge-3cb2a1=321" :expected "a=b; c=d"}
+           {:cookie-string "x-waiter-oidc-challenge-a1b2c3=123; x-waiter-oidc-challenge-4d5e6f=456; x-waiter-oidc-challenge-3cb2a1=321; a=b; c=d" :expected "a=b; c=d"}
+           {:cookie-string "x-waiter-oidc-challenge-a1b2c3=123; a=b; x-waiter-oidc-challenge-3cb2a1=321; x-waiter-oidc-challenge-4d5e6f=456; c=d" :expected "a=b; c=d"}
+           {:cookie-string "x-waiter-oidc-challenge-a1b2c3=123; a=b; c=d; x-waiter-oidc-challenge-3cb2a1=321; x-waiter-oidc-challenge-4d5e6f=456" :expected "a=b; c=d"}
+           {:cookie-string "a=b; x-waiter-oidc-challenge-a1b2c3=123; x-waiter-oidc-challenge-3cb2a1=321; c=d; x-waiter-oidc-challenge-4d5e6f=456" :expected "a=b; c=d"}]]
+    (is (= expected (auth/remove-auth-cookie cookie-string)) (str test-case))))
```
napple, line 728: I don’t see existing cases for `remove-auth-cookie` that tea removal of the other prefixes. Maybe that’s tested indirectly? Should they be tested explicitly like these oidc cases?

I might also include cases where the prefix string is present but it’s not used as a prefix, eg `blah-x-waiter-oidc-challenge-abc=123` should not be removed.

napple: Maybe include a case with an empty cookie string, too.

waiter/test/waiter/auth/oidc_test.clj
```diff
@@ -696,3 +696,33 @@
...
+(deftest test-remove-challenge-cookies
+  (doseq [{:keys [cookie-string expected] :as test-case}
+          [{:cookie-string nil :expected nil}
+           {:cookie-string "a=b" :expected "a=b"}
+           {:cookie-string "a=b; c=d" :expected "a=b; c=d"}
+           {:cookie-string "x-waiter-oidc-challenge-A1B2C3=1234; a=b" :expected "a=b"}
+           {:cookie-string "a=b; x-waiter-oidc-challenge-A1B2C3=1234" :expected "a=b"}
+           {:cookie-string "x-waiter-oidc-challenge-a1b2c3=1234; a=b" :expected "a=b"}
+           {:cookie-string "a=b; x-waiter-oidc-challenge-a1b2c3=1234" :expected "a=b"}
+           {:cookie-string "a=b; x-waiter-oidc-challenge-a1b2c3=1234" :expected "a=b"}
```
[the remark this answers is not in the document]
shamsimam: Removed.

waiter/test/waiter/auth/oidc_test.clj
shamsimam: Added additional test cases and a separate unit test function.

napple (approved):

PR #1513, opened 2022-02-16
reports deployment error in get service endpoint

shamsimam:

- reports deployment error in get service endpoint

Reporting the deployment error in the API allows service owners to determine why a service is failing to start.

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -350,7 +350,8 @@
                                   instance-count (reduce + (vals instance-counts-map))
                                   effective-service-description (service-id->service-description-fn service-id :effective? true)
                                   resource-usage (compute-resource-usage scheduler service-id instance-count)
-                                  service-metrics (get service-id->metrics service-id)]
+                                  service-metrics (get service-id->metrics service-id)
+                                  {:keys [deployment-error-message service-status-label]} (service/retrieve-service-status-and-deployment-error service-id global-state)]
                               (cond->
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/handler.clj
```diff
@@ -350,18 +350,21 @@
...
+                                  {:keys [deployment-error-message service-status-label]} (service/retrieve-service-status-and-deployment-error service-id global-state)]
                               (cond->
                                 {:instance-counts instance-counts-map
                                  :last-request-time (get service-metrics "last-request-time")
                                  :request-metrics (select-keys service-metrics ["outstanding" "total"])
                                  :resource-usage resource-usage
                                  :service-id service-id
                                  :service-description core-service-description
-                                 :status (service/retrieve-service-status-label service-id global-state)
+                                 :status service-status-label
```
kevo1ution, line 362: changing from a message to a map is a breaking change. I can't think of any of our clients off the top of my head that use status, but should we try to avoid this? Maybe a new field that is `deployment-error`

shamsimam: This field has not changed, it continues to be a string. The map being returned is destructured above. 
Our integration tests already rely on this field and would have complained if the value had changed.

kevo1ution: oops I misread and thought you were putting the return value for the function `(service/retrieve-service-status-and-deployment-error)` as the `:status`. This looks good

PR #1526, opened 2022-03-11
Add ability to consume external metrics for last request time

kevo1ution:

- ability to watch on external metrics service to update `last-request-time` metric and start new services with new request times.

- we want to be able to trigger various parts of the service lifecycle off of external metrics

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1002,6 +1002,25 @@
                   ;; in order for Waiter to garbage collect its metrics:
                   :transient-metrics-timeout-ms 300000}
 
+ ; Configures how waiter consumes and handles metrics from an external metrics service
+ ; these metrics can be used for waiter to make autoscaling decisions
+ :metrics-consumer {
+                    ; max time to wait for metrics service to establish connection when watch request is made
+                    :connection-timeout-ms 5000
+
+                    ; total allowed time for inactivity on a metrics service watch response stream before terminating
+                    ; and retrying the watch request
+                    :idle-timeout-ms 10000
+
+                    ; metrics services to aggregate metrics from
+                    :metrics-service-urls []
+
+                    ; time before retrying a new watch connection to the metrics service
+                    :retry-delay-ms 1000
+
+                    ; total number of token metrics events that can be buffered before old events are overwritten
+                    :token-metric-chan-buffer-size 4096}
+
  ; ---------- Miscellaneous ----------
 
  ;; Custom components allow sharing of components between the scheduler and service description builder.
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -1013,7 +1013,7 @@
                     :idle-timeout-ms 10000
 
                     ; metrics services to aggregate metrics from
-                    :metrics-service-urls []
+                    :metrics-services []
 
                     ; time before retrying a new watch connection to the metrics service
                     :retry-delay-ms 1000
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -48,6 +48,7 @@
             [waiter.interstitial :as interstitial]
             [waiter.kv :as kv]
             [waiter.metrics :as metrics]
+            [waiter.metrics-consumer :as metrics-consumer]
             [waiter.metrics-sync :as metrics-sync]
             [waiter.password-store :as password-store]
             [waiter.process-request :as pr]
@@ -133,6 +134,7 @@
                               ["/launch-metrics" :state-launch-metrics-handler-fn]
                               ["/leader" :state-leader-handler-fn]
                               ["/local-usage" :state-local-usage-handler-fn]
+                              ["/metrics-consumer" :state-metrics-consumer-handler-fn]
                               ["/maintainer" :state-maintainer-handler-fn]
                               ["/router-metrics" :state-router-metrics-handler-fn]
                               ["/scheduler" :state-scheduler-handler-fn]
@@ -1110,6 +1112,13 @@
    :retrieve-descriptor-fn (pc/fnk [request->descriptor-fn]
                              (fn retrieve-descriptor-fn [run-as-user token]
                                (descriptor/retrieve-descriptor request->descriptor-fn run-as-user token)))
+   :retrieve-latest-descriptor-fn (pc/fnk [[:state kv-store service-description-builder
+                                            service-id-prefix waiter-hostnames]
+                                           assoc-run-as-user-approved? attach-service-defaults-fn attach-token-defaults-fn]
+                                    (fn retrieve-latest-descriptor-fn [run-as-user token]
+                                      (descriptor/retrieve-latest-descriptor
+                                        attach-service-defaults-fn attach-token-defaults-fn service-id-prefix kv-store waiter-hostnames service-description-builder
+                                        assoc-run-as-user-approved? run-as-user token)))
    :retrieve-token-based-fallback-fn (pc/fnk [retrieve-descriptor-fn service-id->service-description-fn]
                                        (fn retrieve-token-based-fallback-fn [service-id current-for-tokens]
                                          (descriptor/retrieve-token-based-fallback
@@ -1324,6 +1333,22 @@
    :messages (pc/fnk [[:settings {messages nil}]]
                (when messages
                  (utils/load-messages messages)))
+   :metrics-consumer-maintainer (pc/fnk
+                                  [[:routines retrieve-descriptor-fn router-metrics-helpers]
+                                   [:settings
+                                    [:metrics-consumer connection-timeout-ms metrics-service-urls idle-timeout-ms retry-delay-ms
+                                     token-metric-chan-buffer-size]]
+                                   [:state clock kv-store local-usage-agent router-id user-agent-version
+                                    token-cluster-calculator]]
+                                  (let [{:keys [service-id->metrics-fn]} router-metrics-helpers
+                                        http-client (hu/http-client-factory {:client-name (str "waiter-metrics-consumer-" user-agent-version)
+                                                                             :conn-timeout connection-timeout-ms
+                                                                             :socket-timeout idle-timeout-ms
+                                                                             :user-agent (str "waiter-metrics-consumer/" user-agent-version)})]
+                                    (metrics-consumer/start-metrics-consumer-maintainer
+                                      http-client clock kv-store token-cluster-calculator retrieve-descriptor-fn service-id->metrics-fn
+                                      metrics-consumer/make-metrics-watch-request local-usage-agent router-id metrics-service-urls
+                                      token-metric-chan-buffer-size retry-delay-ms)))
    ;; This function is defined as a convenience to avoid repeated extraction from daemons/service-chan-maintainer.
    :populate-maintainer-chan! (pc/fnk [service-chan-maintainer]
                                 (let [{:keys [populate-maintainer-chan!]} service-chan-maintainer]
@@ -1420,6 +1445,14 @@
                                 (async/tap router-state-push-mult state-chan)
                                 (maintainer/start-service-chan-maintainer
                                   {} state-chan query-service-maintainer-chan start-service remove-service retrieve-channel)))
+   :start-new-services-handler (pc/fnk
+                                 [[:routines retrieve-latest-descriptor-fn start-new-service-fn]
+                                  [:state kv-store fallback-state-atom leader?-fn]
+                                  metrics-consumer-maintainer]
+                                 (let [{:keys [token-metric-chan-mult]} metrics-consumer-maintainer]
+                                   (scheduler/start-new-services-handler
+                                     retrieve-latest-descriptor-fn kv-store token-metric-chan-mult start-new-service-fn
+                                     leader?-fn fallback-state-atom)))
    :state-sources (pc/fnk [[:scheduler scheduler]
                            [:state query-service-maintainer-chan]
                            autoscaler autoscaling-multiplexer gc-for-transient-metrics interstitial-maintainer
@@ -1843,6 +1876,13 @@
                                     (wrap-secure-request-fn
                                       (fn maintainer-state-handler-fn [request]
                                         (handler/get-chan-latest-state-handler router-id query-state-fn request)))))
+   :state-metrics-consumer-handler-fn (pc/fnk [[:daemons metrics-consumer-maintainer]
+                                               [:state router-id]
+                                               wrap-secure-request-fn]
+                                        (let [{:keys [query-state-fn]} metrics-consumer-maintainer]
+                                          (wrap-secure-request-fn
+                                            (fn maintainer-state-handler-fn [request]
+                                              (handler/get-daemon-state router-id query-state-fn request)))))
    :state-router-metrics-handler-fn (pc/fnk [[:routines router-metrics-helpers]
                                              [:state router-id]
                                              wrap-secure-request-fn]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1449,10 +1449,13 @@
                                 [[:routines retrieve-latest-descriptor-fn start-new-service-fn]
                                  [:state kv-store fallback-state-atom leader?-fn]
                                  metrics-consumer-maintainer]
-                                (let [{:keys [token-metric-chan-mult]} metrics-consumer-maintainer]
+                                (let [{:keys [token-metric-chan-mult]} metrics-consumer-maintainer
+                                      service-exists? (fn service-exists?
+                                                        [service-id]
+                                                        (descriptor/service-exists? @fallback-state-atom service-id))]
                                   (scheduler/start-new-services-daemon
-                                    retrieve-latest-descriptor-fn kv-store token-metric-chan-mult start-new-service-fn
-                                    leader?-fn fallback-state-atom)))
+                                    retrieve-latest-descriptor-fn service-exists? kv-store token-metric-chan-mult
+                                    start-new-service-fn leader?-fn)))
    :state-sources (pc/fnk [[:scheduler scheduler]
                            [:state query-service-maintainer-chan]
                            autoscaler autoscaling-multiplexer gc-for-transient-metrics interstitial-maintainer
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1336,7 +1336,7 @@
    :metrics-consumer-maintainer (pc/fnk
                                   [[:routines retrieve-descriptor-fn router-metrics-helpers]
                                    [:settings
-                                    [:metrics-consumer connection-timeout-ms metrics-service-urls idle-timeout-ms retry-delay-ms
+                                    [:metrics-consumer connection-timeout-ms metrics-services idle-timeout-ms retry-delay-ms
                                      token-metric-chan-buffer-size]]
                                    [:state clock kv-store local-usage-agent router-id user-agent-version
                                     token-cluster-calculator]]
@@ -1347,7 +1347,7 @@
                                                                              :user-agent (str "waiter-metrics-consumer/" user-agent-version)})]
                                     (metrics-consumer/start-metrics-consumer-maintainer
                                       http-client clock kv-store token-cluster-calculator retrieve-descriptor-fn service-id->metrics-fn
-                                      metrics-consumer/make-metrics-watch-request local-usage-agent router-id metrics-service-urls
+                                      metrics-consumer/make-metrics-watch-request local-usage-agent router-id metrics-services
                                       token-metric-chan-buffer-size retry-delay-ms)))
    ;; This function is defined as a convenience to avoid repeated extraction from daemons/service-chan-maintainer.
    :populate-maintainer-chan! (pc/fnk [service-chan-maintainer]
--- a/waiter/src/waiter/core.clj
+++ b/waiter/src/waiter/core.clj
@@ -1445,14 +1445,14 @@
                                 (async/tap router-state-push-mult state-chan)
                                 (maintainer/start-service-chan-maintainer
                                   {} state-chan query-service-maintainer-chan start-service remove-service retrieve-channel)))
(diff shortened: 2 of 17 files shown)


waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+              (cond-> {:last-update-time last-update-time
+                       :supported-include-params ["buffer-state" "watches-state"]}
+                (contains? include-flags "buffer-state")
+                (assoc :buffer-state {:envoy-cluster-event-count (.count token-metric-chan-buffer)})
+                (contains? include-flags "watches-state")
+                (assoc :watches-state (reduce
+                                        (fn [watches-state {:keys [metrics-service-url query-state-fn]}]
+                                          (assoc watches-state metrics-service-url (query-state-fn include-flags)))
+                                        {}
+                                        metrics-service-watches)))))
```
shamsimam, line 163: If `metrics-service-watches` is a map keyed by the `metrics-service-url`, we can avoid the `reduce` and replace with a `pc/map-vals`.

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+                (reset! state-atom current-state)
+                (let [[msg current-chan]
+                      (async/alts! [exit-chan token-metric-chan query-chan])
+                      next-state
+                      (condp = current-chan
+                        exit-chan
+                        (do
+                          (log/warn "Stopping metrics-consumer-maintainer")
+                          (when (not= :exit msg)
+                            (throw (ex-info "Stopping metrics-consumer-maintainer" {:time (clock) :reason msg}))))
```
shamsimam, line 177:
```suggestion
                          (if (not= :exit msg)
                            (throw (ex-info "Stopping metrics-consumer-maintainer" {:time (clock) :reason msg}))
                            (log/warn "Stopping metrics-consumer-maintainer"))
```

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+                                    (log/info "starting" {:service-id latest-service-id})
+                                    (start-new-service-fn latest-descriptor)))))
+                            current-state))
+
+                        query-chan
+                        (let [{:keys [response-chan include-flags]} msg]
+                          (async/put! response-chan (query-state-fn include-flags))
+                          current-state))]
+                  (if next-state
+                    (recur (assoc current-state :last-update-time (clock)))
```
shamsimam, line 212:
```suggestion
                    (recur (assoc next-state :last-update-time (clock)))
```

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+                                is-new-last-request-time? (or (nil? stored-last-request-time)
+                                                              (t/before? stored-last-request-time lastRequestTime))]
+                            (log/info "metrics consumer event" {:token token
+                                                                :stored-last-request-time stored-last-request-time
+                                                                :candidate-last-request-time lastRequestTime
+                                                                :fallback-service-id fallback-service-id
+                                                                :latest-service-id latest-service-id})
+                            (when
+                              (and is-new-last-request-time?
+                                   (leader?-fn))
```
shamsimam, line 195: The leader check can be much earlier to avoid processing in `let` block or even the processing of the event if only the leader is starting new services.

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+                              (timers/start-stop-time!
+                                (metrics/waiter-timer "core" "metrics-consumer" "update-last-request-time")
+                                (log/info "updating last request time for service" {:fallback-service-id fallback-service-id
+                                                                                    :last-request-time lastRequestTime})
+                                (send local-usage-agent metrics/update-last-request-time-usage-metric fallback-service-id lastRequestTime)
```
shamsimam, line 200: Does not feel like local metrics updates should only be triggered on the leader. Any router connecting to the metrics service should be able to update its local metrics.

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+            (timers/start-stop-time!
+              (metrics/waiter-timer "core" "metrics-consumer" "map-events-fn")
+              (when-let [{:strs [run-as-user] :as service-description-template}
+                         (sd/token->service-parameter-template kv-store token :error-on-missing false)]
+                (when (and run-as-user
+                           (not (sd/run-as-requester? service-description-template))
+                           (not (sd/requires-parameters? service-description-template)))
+                  (try
+                    (let [{:keys [descriptor latest-descriptor]} (retrieve-descriptor-fn run-as-user token)]
+                      (assoc event "lastRequestTime" (f/parse lastRequestTime)
```
shamsimam, line 136: Let's change to kebab case for consistency
```suggestion
                      (assoc event "last-request-time" (f/parse lastRequestTime)
```

waiter/config-shell.edn
```diff
@@ -1,4 +1,7 @@
 {
+ :metrics-consumer {
+                    :metrics-service-urls ["http://localhost:9001"]
```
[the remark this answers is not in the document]
kevo1ution: agreed. I left this in because I set up a dummy nodejs metrics service. going to remove this for now, and in the future add a mock service that we can test against

waiter/config-full.edn
```diff
@@ -1003,6 +1003,25 @@
...
+ :metrics-consumer {
+                    ; max time to wait for metrics service to establish connection when watch request is made
+                    :connection-timeout-ms 5000
+
+                    ; total allowed time for inactivity on a metrics service watch response stream before terminating
+                    ; and retrying the watch request
+                    :idle-timeout-ms 10000
+
+                    ; metrics services to aggregate metrics from
+                    :metrics-service-urls ["https://metrics-service.example.com:9000"]
```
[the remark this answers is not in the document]
kevo1ution: I think this is a good idea, but the added latency for service in the different datacenter will probably still be under 1000 ms which seems negligible to me for the connection timeout (5000 ms) and the idle timeout (10000 ms). Do you agree, and maybe this is something we can add as a follow-up in the future?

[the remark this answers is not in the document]
shamsimam: I'm good with whatever you decide here, just wanted to confirm you had thought about it.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,32 @@
...
+  (cid/with-correlation-id
+    "start-new-services-goroutine"
+    (let [correlation-id (cid/get-correlation-id)
+          process-token-event-ch-buffer (async/sliding-buffer 1000)
+          process-token-event-ch
+          (async/chan
+            process-token-event-ch-buffer
+            (comp
+              (filter (fn leader? [_]
+                        (leader?-fn)))
```
shamsimam, line 680: You can pass `leader?-fn` directly to filter. Though this feels strange. Maybe wrap in a when clause and guard execution?

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,32 @@
                         service
                         (assoc instances :active-instances active-instances))))))))
 
+(defn start-new-services-handler
+  [token-metric-chan-mult start-new-service-fn leader?-fn fallback-state-atom]
```
shamsimam, line 670: Feels like this component should be able to work of local-usage-agent state and any differences to last execution time of service IDs. Are there any benefits to passing along the data in the mult?

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+                                                         :metrics-endpoint metrics-endpoint})
+                        (reset! state-atom (assoc @state-atom :last-event-time (clock)))
+                        (let [actual-events (->> object
+                                                 (filter filter-events-fn)
+                                                 (map map-events-fn)
+                                                 (filter some?))]
+                          (meters/mark! (metrics/waiter-meter "core" "metrics-consumer" metrics-service-url "token-metric-event-count")
+                                        (count actual-events))
+                          (doseq [event actual-events]
+                            (async/put! event-chan event))))
```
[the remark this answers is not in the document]
kevo1ution: I thought it was easier to read and manage the channel, especially now that other components may listen to the channel. I think I can mitigate this by increasing the default buffer size for the channel. Does this sound okay?

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,32 @@
...
+    (let [correlation-id (cid/get-correlation-id)
+          process-token-event-ch-buffer (async/sliding-buffer 1000)
+          process-token-event-ch
+          (async/chan
+            process-token-event-ch-buffer
+            (comp
+              (filter (fn leader? [_]
+                        (leader?-fn)))
+              (filter (fn latest-service-does-not-exist?
+                        [{{:keys [service-id]} :latest-descriptor}]
```
shamsimam, line 682: Seems like unnecessarily burdening the metrics service components to retrieve the latest descriptor only to be used here. This component should be able to compute its own latest service ID from the token.

service-id + services-mapping gives us token.
We have existing functions to check if a service is stale.
If the service is stale, token can be used to calculate new service ID.

waiter/src/waiter/scheduler.clj
kevo1ution: It seems simpler/easier to read when using the data in the mult instead of doing a last-request-time diff check on all service-ids on an interval to recalculate the events (that were already calculated)

The diff would need to check if:
- new last request time for a service-id
- service-id maps to a token is actually pointing to a new service-id
- start service for this token

the mult already has the filtered tokens that require a new service to be started

waiter/src/waiter/scheduler.clj
kevo1ution: how do we compute the service-id of the current service handling requests for the token to update the local-usage-agent?
https://github.com/twosigma/waiter/pull/1526/files#diff-38b64ff3d02023614b695361fd800397fe9420a12941e9aaae8f45d509d639e5R155

shamsimam: Via `compute-descriptor`
https://github.com/twosigma/waiter/blob/6d71edddb64e8a588cea31e167ae5bb7e3f2cd9b/waiter/src/waiter/descriptor.clj#L280-L299

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,222 @@
...
+  [url request-options handle-blob-fn]
+  (let [request-options (assoc request-options :as :stream)
+        {:keys [body error status] :as response} (clj-http/get url request-options)]
+    (when error
+      (throw error))
+    (when-not (hu/status-2XX? status)
+      (throw (ex-info "received non 2XX response from metrics service" {:response response})))
+    (with-open
+      [stream (InputStreamReader. body)]
+      (doseq [blob (cheshire/parsed-seq stream)]
```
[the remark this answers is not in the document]
kevo1ution: in the refactor, cheshire/parsed-seq exits eagerly whenever the watch stream closes. This is confirmed by the units tests that make the watch chan send fragments on an interval (the abort-ch causes the body-chan to close, which terminates the lazy sequence).

waiter/src/waiter/core.clj
```diff
@@ -1843,6 +1876,13 @@
                                     (wrap-secure-request-fn
                                       (fn maintainer-state-handler-fn [request]
                                         (handler/get-chan-latest-state-handler router-id query-state-fn request)))))
+   :state-metrics-consumer-handler-fn (pc/fnk [[:daemons metrics-consumer-maintainer]
```
shamsimam, line 1879: Do we need a start new services state endpoint?

waiter/src/waiter/descriptor.clj
```diff
@@ -502,3 +502,16 @@
                         :headers {"x-waiter-token" token}
                         :request-time (t/now)}]
     (request->descriptor-fn pseudo-request)))
+
+(defn retrieve-latest-descriptor
+  "Retrieves the latest descriptor for run-as-user and token."
+  [attach-service-defaults-fn attach-token-defaults-fn service-id-prefix kv-store waiter-hostnames service-description-builder
+   assoc-run-as-user-approved? run-as-user token]
+  (let [pseudo-request {:authorization/user run-as-user
+                        ;; we do not know env from request headers and cannot support parameterized services
```
shamsimam, line 511: Can we detect and throw an error if the function is called with such a token?

waiter/src/waiter/descriptor.clj
```diff
@@ -502,3 +502,16 @@
...
+                        :headers {"x-waiter-token" token}
+                        :request-time (t/now)}
+        service-approved? (fn service-approved? [service-id] (assoc-run-as-user-approved? pseudo-request service-id))]
```
shamsimam, line 514: There is already an `fn service-approved?` in this file. We should rename both to have unique names.

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,193 @@
...
+                  (throw res))
+                (swap! state-atom assoc :last-watch-time (clock))
+                (doseq [{:strs [object type]} (utils/chan-to-json-seq!! body-chan)]
+                  (meters/mark! (metrics/waiter-meter "core" "metrics-consumer" cluster "event-rate"))
+                  (if (contains? #{"initial" "update"} type)
+                    (do
+                      (log/info "received event payload" {:count (count object) :type type})
+                      (swap! state-atom assoc :last-event-time (clock))
+                      (doseq [event object]
+                        (async/put! event-chan event)))
```
shamsimam, line 86: Can we avoid a loop to put each individual event into the channel? We should be able to send the collections of events and get it processed donwstream.

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,193 @@
...
+          token-metric-chan-buffer (async/buffer token-metric-chan-buffer-size)
+          token-metric-chan
+          (async/chan
+            token-metric-chan-buffer
+            (comp
+              (map (fn format-event
+                     [raw-event]
+                     (let [event (walk/keywordize-keys raw-event)
+                           kabob-event (set/rename-keys event {:lastRequestTime :last-request-time})]
+                       (update kabob-event :last-request-time f/parse))))
```
shamsimam, line 132: Nitpick: this feels wasteful. We can `dissoc` the bad keys and `assoc` the formatted value.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,40 @@
...
                         (assoc instances :active-instances active-instances))))))))
 
+(defn start-new-services-handler
+  "Listens on token-metric-chan-mult and determines whether to trigger start new services for a token. Only the leader
+  router starts new services. Tokens with new last-request-time pointing to a service descriptor that does not exist will
+  have the latest service started.
+
+  Returns a map with keys:
+  :process-token-event-ch is a channel where messages are pushed to it when a service was started for a token event."
+  [retrieve-latest-descriptor-fn kv-store token-metric-chan-mult start-new-service-fn leader?-fn fallback-state-atom]
```
shamsimam, line 676: Can we pass in a curried `service-exists?` function, then we do not pass in fallback state atom here? Then we to worry about the atom being changed by `start-new-services-handler` function.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,40 @@
                         service
                         (assoc instances :active-instances active-instances))))))))
 
+(defn start-new-services-handler
```
shamsimam, line 669: Can we rename to `daemon` instead of handler as it is not handling a request.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,40 @@
...
+                                service-does-not-exist? (not (descriptor/service-exists? @fallback-state-atom service-id))]
+                            (when service-does-not-exist?
+                              (cid/cinfo correlation-id "starting" {:service-id (get latest-descriptor :service-id)})
+                              (start-new-service-fn latest-descriptor))
+                            service-does-not-exist?)))))
+            (fn process-token-event-ch-ex-handler
+              [e]
+              (cid/cerror correlation-id e "unexpected error when processing new token metric")))]
+      (async/tap token-metric-chan-mult process-token-event-ch)
+      {:process-token-event-ch process-token-event-ch})))
```
shamsimam, line 701: Should this component expose some state, e.g. recently started services and the metric that triggered it?

waiter/src/waiter/settings.clj
```diff
@@ -89,6 +89,12 @@
                                                                                       s/Any s/Any}}
                                      (s/required-key :router-update-interval-ms) schema/positive-int
                                      (s/required-key :transient-metrics-timeout-ms) schema/positive-int}
+   (s/required-key :metrics-consumer) {(s/required-key :connection-timeout-ms) schema/non-negative-num
+                                       (s/required-key :idle-timeout-ms) schema/non-negative-num
+                                       (s/required-key :metrics-services) [{(s/required-key :url) schema/non-empty-string
+                                                                            (s/required-key :cluster) schema/non-empty-string}]
```
shamsimam, line 95: lexicographic order

waiter/src/waiter/util/utils.clj
```diff
@@ -30,13 +31,13 @@
             [waiter.util.http-utils :as hu])
   (:import (clojure.core.async.impl.channels ManyToManyChannel)
            (clojure.lang ExceptionInfo)
-           (java.io OutputStreamWriter)
+           (java.io OutputStreamWriter ByteArrayInputStream SequenceInputStream InputStreamReader)
```
shamsimam, line 34: Lexicographic order

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,196 @@
...
+                      http-client hu/http-streaming-request-async router-id correlation-id metrics-service-conf
+                      token-metric-chan clock retry-delay-ms)]
+                (async/tap exit-mult exit-chan)
+                watch))
+            metrics-service-url->conf)
+          query-state-fn
+          (fn query-metrics-consumer-maintainer-fn
+            [include-flags]
+            (let [{:keys [last-token-event-time]} @state-atom]
+              (cond-> {:last-token-event-time last-token-event-time
```
shamsimam, line 183: Nitpick: lexicographic order

waiter/src/waiter/scheduler.clj
```diff
@@ -27,6 +27,7 @@
             [slingshot.slingshot :as ss]
             [waiter.config :as config]
             [waiter.correlation-id :as cid]
+            [waiter.descriptor :as descriptor]
```
shamsimam, line 30: This import is unused.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,46 @@
...
+                    (when service-does-not-exist?
+                      (cid/cinfo correlation-id "starting" {:service-id (get latest-descriptor :service-id)})
+                      (start-new-service-fn latest-descriptor))
+                    service-does-not-exist?)))
+              events))
+          process-token-event-ch
+          (async/chan
+            process-token-event-ch-buffer
+            (comp
+              (map process-events-fn)
```
shamsimam, line 701: Typically the transform does not have side-effects. We should refactor this later into a go-block to read data from the channel and invoke `process-events`.

waiter/src/waiter/scheduler.clj
```diff
@@ -665,6 +666,46 @@
...
+
+  Returns a map with keys:
+  :process-token-event-ch is a channel where messages are pushed to it when a service was started for a token event."
+  [retrieve-latest-descriptor-fn service-exists? kv-store token-metric-chan-mult start-new-service-fn leader?-fn]
+  (cid/with-correlation-id
+    "start-new-services-goroutine"
+    (let [correlation-id (cid/get-correlation-id)
+          process-token-event-ch-buffer (async/sliding-buffer 1000)
+          process-events-fn
+          (fn process-events [events]
```
shamsimam, line 682: This function has side-effects, maybe call it `process-events!`

waiter/src/waiter/metrics_consumer.clj
```diff
@@ -0,0 +1,196 @@
...
+                    {:keys [body-chan error-chan] :as res}
+                    (async/<! (http-streaming-request-async-fn http-client metrics-endpoint
+                                                               :abort-ch abort-ch
+                                                               :query-string query-string))]
+                ; when async chan is triggered, we want to also trigger the abort channel on the watch connection
+                (async/tap exit-chan-mult abort-ch)
+                (when (instance? Throwable res)
+                  (throw res))
+                (swap! state-atom assoc :last-watch-time (clock))
+                (doseq [{:strs [object type]} (utils/chan-to-json-seq!! body-chan)]
```
shamsimam, line 77: Is the blocking call `utils/chan-to-json-seq!!` inside a go-block intentional?

waiter/src/waiter/scheduler.clj
kevo1ution: sgtm

waiter/src/waiter/metrics_consumer.clj
kevo1ution: its because its calling chan-to-seq!!, I couldn't find a nonblocking way of creating an InputStreamReader. from a channel. But this chan-to-seq!! should not block the caller as it is all lazy-sequences

kevo1ution: I think likely we'll need to override the InputStreamReader. with `proxy` and implement the `read` functions to use the channel directly. I can do this in a followup pr as well if that is okay with you

shamsimam (approved):

waiter/src/waiter/metrics_consumer.clj
shamsimam: The caller is blocked in the doseq when trying to read the next element.

PR #1531, opened 2022-03-17
requires owner to be able to run as run-as-user

shamsimam:

- requires owner to be able to run as run-as-user

We would like to check that the `owner` is able to own tokens configured to run as `run-as-user`.

### DIFF
--- a/waiter/src/waiter/token_validator.clj
+++ b/waiter/src/waiter/token_validator.clj
@@ -144,6 +144,15 @@
               (throw (ex-info (str "Cannot modify " parameter-name " token metadata")
                               {:status http-400-bad-request
                                :token-metadata new-token-metadata
+                               :log-level :warn}))))
+          ;; owner must be able to run as run-as-user
+          (when (and run-as-user (not= "*" run-as-user))
+            (when-not (authz/own? entitlement-manager owner run-as-user)
+              (throw (ex-info (str "Owner: " owner " cannot run as user: " run-as-user)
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/token_validator.clj
```diff
@@ -144,6 +144,15 @@
...
+                               :log-level :warn}))))
+          ;; owner must be able to run as run-as-user
+          (when (and run-as-user (not= "*" run-as-user))
+            (when-not (authz/own? entitlement-manager owner run-as-user)
+              (throw (ex-info (str "Owner: " owner " cannot run as user: " run-as-user)
+                              {:authenticated-user authenticated-user
+                               :owner owner
+                               :run-as-user run-as-user
+                               :status http-403-forbidden
                                :log-level :warn})))))
```
DaoWen, line 156: Nitpick: We shouldn't be logging user errors with `warn`-level messages. We should only use `warn` and `error` logging for things that we would actually be concerned about when filtering our logs (i.e., when Waiter is in some bad state or doing something it shouldn't be doing). However, since we probably need to make a pass over all of our logs to make this more consistent, I don't feel strongly about changing this now.

DaoWen (approved):

waiter/src/waiter/token_validator.clj
shamsimam: Agree, will make the change for this file in a subsequent PR.

DaoWen (approved):

PR #1551, opened 2022-04-14
avoids marking killed instances as failed instances by tracking kill candidates

shamsimam:

- avoids marking killed instances as failed instances by tracking kill candidates

The kubernetes scheduler, since it uses watches, encounters a race between treating deleted instances as killed or failed. By tracking kill candidates we can err on the side of caution and only track non-kill candidates as failed instances.

**Note**: A data race is still possible where our kill candidate actually fails and hence the subsequent but the kill operation returns success.

### DIFF
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -862,7 +862,7 @@
     "Tracks which instances are being used as kill candidates for the provided duration."
     [instance-id reason duration-ms]
     (log/info "tracking instance as kill candidate" {:duration-sm duration-ms :id instance-id :reason reason})
-    (swap! kill-candidates-state-atom update instance-id update :tracked-states set/union #{reason})
+    (swap! kill-candidates-state-atom update-in [instance-id :tracked-states] set/union #{reason})
     (when (= :prepare-to-kill reason)
       (when (pos? duration-ms)
         (async/go
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -842,8 +842,54 @@
                                  ([service-id] (retrieve-syncer-state @syncer-state-atom service-id)))}))
 
 ;;
-;; Support for tracking killed instances
+;; Support for tracking killed instance candidates
 ;;
+(defn- cleanup-kill-candidate
+  "Cleans up state for a kill candidate.
+   If candidate was not killed but still failed, we trigger the failed instance callback."
+  [kill-candidates-state instance-id]
+  (let [{:keys [track-failed-instance tracked-states]} (get kill-candidates-state instance-id)]
+    (when (and track-failed-instance
+               (contains? tracked-states :failed)
+               (not (contains? tracked-states :killed)))
+      (log/info "tracking kill candidate as failed instance" {:id instance-id})
+      (track-failed-instance))
+    (dissoc kill-candidates-state instance-id)))
+
+(let [kill-candidates-state-atom (atom {})]
+
+  (defn track-kill-candidate!
+    "Tracks which instances are being used as kill candidates for the provided duration."
+    [instance-id reason duration-ms]
+    (log/info "tracking instance as kill candidate" {:duration-sm duration-ms :id instance-id :reason reason})
+    (swap! kill-candidates-state-atom update instance-id update :tracked-states set/union #{reason})
+    (when (= :prepare-to-kill reason)
+      (when (pos? duration-ms)
+        (async/go
+          (async/<! (async/timeout duration-ms))
+          (log/info "end tracking instance as kill candidate" {:id instance-id})
+          (swap! kill-candidates-state-atom cleanup-kill-candidate instance-id)))))
+
+  (defn track-failed-kill-candidate!
+    "Tags a kill candidate as a failed instance for processing during cleanup."
+    [instance-id track-failed-instance-callback]
+    (let [kill-candidate?-atom (atom false)]
+      (swap! kill-candidates-state-atom
+             (fn update-failed-kill-candidate-state-1 [kill-candidates-state]
+               (cond-> kill-candidates-state
+                 (contains? kill-candidates-state instance-id)
+                 (update instance-id (fn update-failed-kill-candidate-state-2 [state]
+                                       (reset! kill-candidate?-atom true)
+                                       (-> state
+                                         (update :tracked-states set/union #{:failed})
+                                         (assoc :track-failed-instance track-failed-instance-callback)))))))
+      (when @kill-candidate?-atom
+        (log/info "kill candidate marked as follow-up for failed instance" {:id instance-id}))))
+
+  (defn is-kill-candidate?
+    "Returns true if the queried instance is currently being tracked as a kill candidate."
+    [instance-id]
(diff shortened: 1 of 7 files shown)


waiter/src/waiter/scheduler.clj
```diff
@@ -856,10 +902,16 @@
...
-  (log-service-instance instance :fail :info)
-  (add-instance-to-buffered-collection! transient-store max-instances-to-keep service-id instance
-                                        (fn [] #{}) (fn [instances] (-> (sort-instances instances) (rest) (set)))))
+  [transient-store max-instances-to-keep service-id {:keys [id] :as instance}]
+  (let [track-failed-instance-helper
+        (fn track-failed-instance-helper []
+          (do
+            (log-service-instance instance :fail :info)
+            (add-instance-to-buffered-collection! transient-store max-instances-to-keep service-id instance
+                                                  (fn [] #{}) (fn [instances] (-> (sort-instances instances) (rest) (set))))))]
```
DaoWen, line 911: `(fn [] #{})` → `set`

DaoWen: Maybe we should just use a `sorted-set` instead? That would also make the `remove-fn` simpler:
```suggestion
                                                  sorted-set #(disj % (first %)))))]
```

shamsimam: This didn't necessarily change in this PR. I only pulled in existing code into a function. Changing to a set may cause downstream effects on other consumers of this value who may be expecting lists.

waiter/src/waiter/scheduler.clj
```diff
@@ -842,8 +842,54 @@
...
+      (track-failed-instance))
+    (dissoc kill-candidates-state instance-id)))
+
+(let [kill-candidates-state-atom (atom {})]
+
+  (defn track-kill-candidate!
+    "Tracks which instances are being used as kill candidates for the provided duration."
+    [instance-id reason duration-ms]
+    (log/info "tracking instance as kill candidate" {:duration-sm duration-ms :id instance-id :reason reason})
+    (swap! kill-candidates-state-atom update instance-id update :tracked-states set/union #{reason})
```
DaoWen, line 865:
```suggestion
    (swap! kill-candidates-state-atom update-in [instance-id :tracked-states] conj reason)
```

shamsimam: `reason` is initially `nil`, `conj` doesn't get me a set :(

waiter/src/waiter/scheduler.clj
DaoWen: It was already a set — but you're right, this change is unrelated to your PR. I didn't realize this was just whitespace change. I can make a follow-on patch later to make this change separately.

waiter/src/waiter/scheduler.clj
```diff
@@ -842,8 +842,54 @@
...
+    [instance-id track-failed-instance-callback]
+    (let [kill-candidate?-atom (atom false)]
+      (swap! kill-candidates-state-atom
+             (fn update-failed-kill-candidate-state-1 [kill-candidates-state]
+               (cond-> kill-candidates-state
+                 (contains? kill-candidates-state instance-id)
+                 (update instance-id (fn update-failed-kill-candidate-state-2 [state]
+                                       (reset! kill-candidate?-atom true)
+                                       (-> state
+                                         (update :tracked-states set/union #{:failed})
```
DaoWen, line 884:
```suggestion
                                         (update :tracked-states conj :failed)
```

waiter/src/waiter/scheduler.clj
DaoWen: 😢

waiter/src/waiter/scheduler.clj
DaoWen: I guess this is initially `nil` as well, so never mind...

waiter/src/waiter/scheduler.clj
DaoWen: Just the `update-in` change then...

```suggestion
    (swap! kill-candidates-state-atom update-in [instance-id :tracked-states] set/union #{reason})
```

shamsimam: Committed suggestion

waiter/src/waiter/scheduler.clj
```diff
@@ -842,8 +842,54 @@
...
+  (defn track-failed-kill-candidate!
+    "Tags a kill candidate as a failed instance for processing during cleanup."
+    [instance-id track-failed-instance-callback]
+    (let [kill-candidate?-atom (atom false)]
+      (swap! kill-candidates-state-atom
+             (fn update-failed-kill-candidate-state-1 [kill-candidates-state]
+               (cond-> kill-candidates-state
+                 (contains? kill-candidates-state instance-id)
+                 (update instance-id (fn update-failed-kill-candidate-state-2 [state]
+                                       (reset! kill-candidate?-atom true)
```
DaoWen, line 882: From the [docs for `swap!`](https://clojuredocs.org/clojure.core/swap!):

> Note that `f` may be called multiple times, and thus should be free of side effects.  Returns the value that was swapped in.

Rather than using the atom, should we just look at the value returned from `swap!` to see if it has `instance-id :tracked-states :failed`?

waiter/src/waiter/scheduler.clj
```diff
@@ -842,8 +842,54 @@
...
 ;;
+(defn- cleanup-kill-candidate
+  "Cleans up state for a kill candidate.
+   If candidate was not killed but still failed, we trigger the failed instance callback."
+  [kill-candidates-state instance-id]
+  (let [{:keys [track-failed-instance tracked-states]} (get kill-candidates-state instance-id)]
+    (when (and track-failed-instance
+               (contains? tracked-states :failed)
+               (not (contains? tracked-states :killed)))
+      (log/info "tracking kill candidate as failed instance" {:id instance-id})
```
DaoWen, line 855: We probably don't want to log inside `swap!` since we could end up calling this multiple times if the swap fails.

From the [docs for `swap!`](https://clojuredocs.org/clojure.core/swap!):

> Note that `f` may be called multiple times, and thus should be free of side effects.  Returns the value that was swapped in.

Can we refactor this to pass in the atom, do the `swap!` inside this function, and then log based on the result?

waiter/src/waiter/scheduler.clj
```diff
@@ -842,8 +842,54 @@
...
+
+(let [kill-candidates-state-atom (atom {})]
+
+  (defn track-kill-candidate!
+    "Tracks which instances are being used as kill candidates for the provided duration."
+    [instance-id reason duration-ms]
+    (log/info "tracking instance as kill candidate" {:duration-sm duration-ms :id instance-id :reason reason})
+    (swap! kill-candidates-state-atom update-in [instance-id :tracked-states] set/union #{reason})
+    (when (= :prepare-to-kill reason)
+      (when (pos? duration-ms)
```
DaoWen, line 867: We should use `and` instead of double-`when`

waiter/src/waiter/scheduler.clj
shamsimam: Removed this function.

DaoWen (approved):

PR #1554, opened 2022-04-18
deflake minmax-instance scaling test

DaoWen:

deflake minmax-instance scaling test

failing because it's seeing more than max instances


waiter/integration/waiter/autoscaling_test.clj
```diff
@@ -241,18 +242,16 @@
           _ (log/info "making canary request...")
           {:keys [cookies service-id]} (request-fn)
           get-target-instances
-          (fn []
-            (let [instances
-                  (loop [routers (routers waiter-url)
```
DaoWen, line 246: This was just a stylistic change I made while I was editing the test anyway. `loop` feels very imperative, making the code feel less functional.

PR #1576, opened 2022-05-22
Add deployment errors to services based on k8s events

napple:

- When a service has no running pods and k8s events show an inability to create pods, add a deployment error to the service map.

Certain k8s events, like `FailedCreate`, indicate that a pod cannot be created. When no pods are running, this should be reflected as an error that is visible to the user.


napple:
Closed in favor of https://github.com/twosigma/waiter/pull/1577

PR #1577, opened 2022-05-22
Add deployment errors to services when k8s events indicate an inabili…

napple:

- When a service has no running pods and k8s events show an inability to create pods, add a deployment error to the service map.

Certain k8s events, like `FailedCreate`, indicate that a pod cannot be created. When no pods are running, this should be reflected as an error that is visible to the user.

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -612,12 +612,13 @@
   [{:keys [instances k8s/events k8s/namespace task-stats]}]
   (let [{:keys [message reason]} (last events)
         running-count (:running task-stats)]
-    (when (and (pos? instances)
+    (when (and (number? instances)
+               (pos? instances)
                (= 0 running-count)
                (= reason "FailedCreate"))
       (prettify-quota-error message namespace))))
 
-(defn- get-services
+(defn get-services
   "Get all Waiter Services (reified as ReplicaSets) running in this Kubernetes cluster."
   [{:keys [service-id->deployment-error-cache watch-state]}]
   (let [service-id->service (-> watch-state deref :service-id->service)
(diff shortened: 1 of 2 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -596,7 +596,29 @@
...
+  "Extract a deployment error if the given service includes a k8s event that is preventing
+   deployment."
+  [{:keys [instances k8s/events k8s/namespace task-stats]}]
+  (let [{:keys [message reason]} (last events)
+        running-count (:running task-stats)]
+    (when (and (number? instances)
+               (pos? instances)
+               (= 0 running-count)
+               (= reason "FailedCreate"))
+      (prettify-quota-error message namespace))))
```
shamsimam, line 619: How do we know for sure that it is a quota error message at this point? Should the method name be more generic, `prettify-failed-create-error`?

waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -605,8 +627,10 @@
     (map
       (fn [service-id]
         (let [deployment-error (get-in service-id->deployment-error [service-id :data])
-              service (or (get service-id->service service-id) (create-empty-service service-id))]
+              service (or (get service-id->service service-id) (create-empty-service service-id))
+              k8s-event-deployment-error (get-k8s-event-deployment-error service)]
           (cond-> service
+            k8s-event-deployment-error (assoc :deployment-error k8s-event-deployment-error)
```
shamsimam, line 633: Can we add a comment on the order in which the deployment errors will show up, e.g. now, failed create errors will be ignored if there are other errors. Alternatively, should we be appending to the string?

shamsimam (dismissed):

waiter/src/waiter/scheduler/kubernetes.clj
napple: I decided to prefer the deployment error from the cache, which currently exclusively represents errors from the k8s API in response to a replicaset create request. I would actually only expect one type of deployment error or the other to be present at any time. I'll add a comment touching on these things.

PR #1582, opened 2022-05-31
add annotation to exclude repicasets and pods

DaoWen:

-


DaoWen:
Shams pointed out that it would be better to handle the `excluded` pods more like `expired` pods (i.e., add a flag, and then use that flag to decide how the pod is treated). I also realized that handling Service/ReplicaSet deletion and Pod deletion (scaling) across two schedulers (during scheduler migration) has a bunch of nasty corner cases, and using these annotations wouldn't be sufficient to work around those.

PR #1584, opened 2022-05-31
Structure k8s event deployment errors as detailed maps

napple:

- Restructure k8s event-sourced deployment errors as detailed maps

These detailed maps are consumed by various Waiter components to provide more detailed information to users.


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -609,14 +609,18 @@
...
+  (let [{:keys [message reason] :as latest-event} (last events)
         running-count (:running task-stats)]
     (when (and (number? instances)
                (pos? instances)
                (= 0 running-count)
                (= reason "FailedCreate"))
-      (prettify-failed-create-error message namespace))))
+      (let [details-fn (constantly {:k8s-event latest-event})
+            error-fn (fn [msg] (prettify-failed-create-error msg namespace))
+            deployment-error (create-service-deployment-error "k8s-event" message error-fn details-fn)]
```
shamsimam, line 621: Does this need to be a `when-let`?

napple: I don't think so. All the items in this `let` block are unconditional; they're just formatting. The condition happens earlier, but that `when`'s expression doesn't return a value I would want to assign conditionally.

shamsimam (approved):

PR #1585, opened 2022-06-03
Expose router details, including fqdn and ssl port

napple:

- Introduce a dedicated setting for a router's fqdn
- Include fqdn and ssl-port in service instances exposed for service discovery
- Include router details, including fqdn and ssl-port, in visible state

By exposing each router's fqdn and ssl port, direct requests to routers can be issued securely.

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -163,6 +163,9 @@
                   ;; However, larger headers will also consume more memory.
                   :response-header-size 8192
 
+                  ;; The fully qualified domain name for this router
+                  :router-fqdn "route1.waiter.example.com"
+
                   ;; Whether to send the Date header in responses
                   :send-date-header? false
 
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -107,9 +107,15 @@
 
  ; ---------- Node ----------
 
+ ; The fully qualified domain name for this router
+ :router-fqdn "route1.waiter.example.com"
+
  ; Use a descriptive name for this router; this is useful in multi-router
  ; scenarios for identifying which router handled a request:
  :router-id-prefix "router-1"
+
+ ; The ssl port configured with certs for the name given by :router-fqdn
+ :router-ssl-port 12323
 
  ; ---------- Network ----------
 
@@ -162,9 +168,6 @@
                   ;; Larger headers will allow for more and/or larger cookies and longer HTTP headers (eg for redirection).
                   ;; However, larger headers will also consume more memory.
                   :response-header-size 8192
-
-                  ;; The fully qualified domain name for this router
-                  :router-fqdn "route1.waiter.example.com"
 
                   ;; Whether to send the Date header in responses
                   :send-date-header? false
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -107,15 +107,9 @@
 
  ; ---------- Node ----------
 
- ; The fully qualified domain name for this router
- :router-fqdn "route1.waiter.example.com"
-
  ; Use a descriptive name for this router; this is useful in multi-router
  ; scenarios for identifying which router handled a request:
  :router-id-prefix "router-1"
-
- ; The ssl port configured with certs for the name given by :router-fqdn
- :router-ssl-port 12323
 
  ; ---------- Network ----------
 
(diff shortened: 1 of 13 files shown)


waiter/config-full.edn
```diff
@@ -107,10 +107,16 @@
...
 
+ ; The fully qualified domain name for this router
+ :router-fqdn "route1.waiter.example.com"
+
  ; Use a descriptive name for this router; this is useful in multi-router
  ; scenarios for identifying which router handled a request:
  :router-id-prefix "router-1"
 
+ ; The ssl port configured with certs for the name given by :router-fqdn
+ :router-ssl-port 12323
```
shamsimam, line 118: Let's move this to `server-options` or `zookeeper`?

waiter/config-full.edn
```diff
@@ -107,10 +107,16 @@
 
  ; ---------- Node ----------
 
+ ; The fully qualified domain name for this router
+ :router-fqdn "route1.waiter.example.com"
```
shamsimam, line 111: Let's move this to `server-options` or `zookeeper`?

napple: I originally included these keys in `server-options`, but I moved them out because they aren't used by the Jetty server (which is the primary purpose of `server-options`), and I thought they confused things since the Jetty server won't actually respond to requests sent to that fqdn/port combination. We could include them in `zookeeper`, since they're currently only leveraged in service discovery. However, if we wanted to allow for them to be more generally used, then we could introduce some new sub-map with a key like `fqdn-config`?

shamsimam: Maybe `router-config` ?

waiter/config-full.edn
napple: Will move inside `:router-config`

waiter/src/waiter/core.clj
```diff
@@ -645,10 +645,14 @@
                                 (resolved-factory-fn (merge context component-config))))
                             custom-components)))
    :discovery (pc/fnk [[:curator curator]
-                       [:settings [:cluster-config name] [:zookeeper base-path discovery-relative-path] host port]
+                       [:settings
+                        [:cluster-config name]
+                        [:zookeeper base-path discovery-relative-path]
+                        host port {router-config {}}]
```
shamsimam, line 651: Why is the syntax necessary for `router-config`?

waiter/src/waiter/discovery.clj
```diff
@@ -89,14 +104,39 @@
...
+  [instance]
+  (let [custom-details (pc/map-keys keyword (.getPayload instance))
+        details {:address (.getAddress instance)
+                 :custom-details custom-details
+                 :id (.getId instance)
+                 :name (.getName instance)
+                 :port (.getPort instance)}]
+    details))
+
+(defn router-id->details
```
shamsimam, line 118: Nitpick: rename fucntion:
```suggestion
(defn retrieve-router-id->details
```

waiter/src/waiter/discovery.clj
```diff
@@ -89,14 +104,39 @@
...
+   and discovery paths."
+  [discovery-path]
+  (str discovery-path "-v2"))
+
 (defn register
-  [router-id curator service-name discovery-path {:keys [host port]}]
-  (let [instance (->service-instance router-id service-name {:host host :port port})
+  [router-id curator service-name discovery-path-base {:keys [host port router-fqdn router-ssl-port]}]
+  (let [discovery-path (versioned-discovery-path discovery-path-base)
+        instance (->service-instance router-id service-name {:host host :port port :router-fqdn router-fqdn :router-ssl-port router-ssl-port})
```
shamsimam, line 139: For backward compatibility should the router register to both the old and `v2` paths?

waiter/src/waiter/core.clj
napple: Because `router-config` is optional and if it's not present in settings, `pc/fnk` fails and Waiter won't start.

waiter/src/waiter/discovery.clj
napple: I'm happy to rename - I just followed the pattern established by `router-id->endpoint-url` above.

waiter/src/waiter/discovery.clj
napple: No, I don't think so. The reason "-v2" is introduced is that the existing service instances and the new one are incompatible. If this new service, using a HashMap payload, also registered with the old path, then on startup of a new router, the existing routers would discover it and fail trying to deserialize its definition.

shamsimam: We discussed offline nd decided to avoid registering to both locations as the peers not discovering each other would be expected only in a temporary state.

shamsimam (approved):

PR #1587, opened 2022-06-07
Add waiting-for-available-instance metric  (aka queued-request-count) To Metric Syncer

kevo1ution:

- Adds `waiting-for-available-instance` counter metric to be synced by the routers and available in aggregate form across all routers

- When calculating for total outstanding requests of a service who may have requests that bypass the waiter routers, we need to take the `queued_req_count` (`waiting-for-available-instance` counter) and add it to the `active-request-count` from the external metrics for all instances in that service.

### DIFF
--- a/waiter/integration/waiter/metrics_sync_integration_test.clj
+++ b/waiter/integration/waiter/metrics_sync_integration_test.clj
@@ -324,7 +324,6 @@
     (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
           routers (routers waiter-url)
           extra-headers {:content-type "application/json"
-                         :x-waiter-env-raven_export_metrics "false"
                          ; service should take at least 5 seconds to start so that we can guarantee requests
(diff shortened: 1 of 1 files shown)


waiter/integration/waiter/metrics_sync_integration_test.clj
```diff
@@ -285,3 +285,72 @@
...
+                (= expected-queued-requests actual-queued-requests)))
+            routers))
+        :interval 1 :timeout 5)))
+
+(deftest ^:parallel ^:integration-slow test-waiting-for-available-instance-metrics-syncer
+  (testing-using-waiter-url
+    (let [{:keys [cookies]} (make-request waiter-url "/waiter-auth")
+          routers (routers waiter-url)
+          extra-headers {:content-type "application/json"
+                         :x-waiter-env-raven_export_metrics "false"
```
shamsimam, line 327: Why do we need to configure this?

shamsimam (dismissed):

PR #1588, opened 2022-06-08
refactor start new service maintainer to support last-request-time from any source

kevo1ution:

- `start-new-service-maintainer` now maintains a current map `service-id->last-request-time` and finds new `last-request-time` metrics for services to determine which tokens to start services for instead of relying on a source channel.

- We want the `start-new-service-maintainer` to work with external metrics changes
- Doing diffs on an interval is a more reliable way to handle starting new services as the buffer may fill up faster than we can handle starting new services.
- Starting new services when a token point to a new service-id is not a sub-second urgent task.

### DIFF
--- a/waiter/src/waiter/scheduler.clj
+++ b/waiter/src/waiter/scheduler.clj
@@ -1258,7 +1258,7 @@
     [{:keys [service-id source-tokens]}]
     (let [many-source-tokens? (> 1 (count source-tokens))]
       (when many-source-tokens?
-        (cid/cinfo correlation-id "Skipping service-id because it maps to many source tokens."
+        (cid/cinfo correlation-id "skipping service-id because it maps to many source tokens."
                    {:service-id service-id
                     :source-tokens source-tokens}))
       (= 1 (count source-tokens)))))
@@ -1278,7 +1278,7 @@
               (not (sd/run-as-requester? service-description-template))
               (not (sd/requires-parameters? service-description-template)))]
         (when-not supported-token?
-          (cid/cinfo correlation-id "Skipping token because it is either run-as-requester or parameterized."
+          (cid/cinfo correlation-id "skipping token because it is either run-as-requester or parameterized."
                      {:token token}))
         supported-token?))))
 


waiter/src/waiter/metrics_sync.clj
```diff
@@ -469,9 +473,21 @@
...
+                                                  (get external-metrics)
+                                                  (pc/map-vals
+                                                    (fn [{:strs [metrics]}]
+                                                      ; if there are external-metrics, then last-request-time will always be
+                                                      ; a valid ISO-8601 string
+                                                      (update metrics "last-request-time" du/str-to-date)))
+                                                  aggregate-service-metrics)]
+                                     (if (some? last-request-time)
+                                       (update router-metrics "last-request-time" t/max-date last-request-time)
+                                       router-metrics))
```
shamsimam, line 490:
```suggestion
                                     (cond-> router-metrics
                                       (some? last-request-time)
                                       (update "last-request-time" t/max-date last-request-time))
```

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
           (log/error e "Error in launch-metrics-maintainer. Instance launch metrics will not be collected."))))
     {:exit-chan exit-chan
      :query-chan query-chan}))
+
+
+(defn start-new-services-maintainer
+  "This ensures that tokens receiving requests will have the latest service corresponding to it running. Whenever there
```
shamsimam, line 1218:
```suggestion
  "This daemon ensures that tokens receiving requests will have the latest service corresponding to it running. Whenever there
```

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
...
+                          (metrics/waiter-timer "core" "start-new-services-maintainer" "process-services")
+                          (try
+                            (let [count-atom (atom 0)
+                                  fallback-state @fallback-state-atom
+                                  new-service-id->last-request-time
+                                  (->> (service-id->metrics-fn)
+                                    seq
+                                    (map
+                                      (fn [[service-id {:strs [last-request-time]}]]
+                                        [service-id last-request-time]))
```
shamsimam, line 1268: Seems like `pc/map-keys` would work better here.

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
...
+                                    (filter
+                                      (fn [[_ last-request-time]]
+                                        (some? last-request-time)))
+                                    (into {}))
```
shamsimam, line 1272: `utils/filterm` would read better here.

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
...
+                                  service-ids-with-new-last-request-times
+                                  (->> (seq new-service-id->last-request-time)
```
shamsimam, line 1274: Nitpick: indentation is off for subsequent forms.

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
...
+                                    (filter
+                                      (fn [[service-id new-last-request-time]]
+                                        (let [old-last-request-time (get service-id->last-request-time service-id)]
+                                          (or (nil? old-last-request-time)
+                                              (t/after? new-last-request-time old-last-request-time)))))
+                                    (map first))
+                                  tokens-with-new-last-request-times
+                                  (->> service-ids-with-new-last-request-times
```
shamsimam, line 1282: Nitpick: indentation is off for subsequent forms.

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,121 @@
...
+                                    (into {}))
+                                  service-ids-with-new-last-request-times
+                                  (->> (seq new-service-id->last-request-time)
+                                    (filter
+                                      (fn [[service-id new-last-request-time]]
+                                        (let [old-last-request-time (get service-id->last-request-time service-id)]
+                                          (or (nil? old-last-request-time)
+                                              (t/after? new-last-request-time old-last-request-time)))))
+                                    (map first))
+                                  tokens-with-new-last-request-times
```
shamsimam, line 1281: Feels like we should query if a service is stale and receiving requests past the time it went stale. That should be the trigger to determine the token and start a new version of the service. It should work for any service including run-as-requester and parameterized.

waiter/src/waiter/metrics_sync.clj
```diff
@@ -469,9 +473,21 @@
...
+                                         ; TODO: we will also need to aggregate active-request-counts and queued-requests
+                                         ; and choose between external vs router metrics if service has bypass configured
+                                         {:strs [last-request-time]}
+                                         (some->> service-id
+                                                  (get external-metrics)
+                                                  (pc/map-vals
+                                                    (fn [{:strs [metrics]}]
+                                                      ; if there are external-metrics, then last-request-time will always be
+                                                      ; a valid ISO-8601 string
+                                                      (update metrics "last-request-time" du/str-to-date)))
```
shamsimam, line 486: Should we use a safe parse method and not error out if the parsing fails?

waiter/src/waiter/scheduler.clj
```diff
@@ -1251,3 +1212,168 @@
...
+                          (metrics/waiter-timer "core" "start-new-services-maintainer" "process-services")
+                          (try
+                            (let [fallback-state @fallback-state-atom
+                                  new-service-id->last-request-time
+                                  (create-service-id->last-request-time-from-metrics service-id->metrics-fn)
+                                  tokens-with-new-last-request-times
+                                  (get-tokens-to-start-services-for
+                                    correlation-id service-id->stale-info service-id->source-tokens-entries-fn kv-store
+                                    new-service-id->last-request-time service-id->last-request-time)]
+                              (doseq [token tokens-with-new-last-request-times]
```
shamsimam, line 1349: Let's move this to a helper function. It makes the go-block smaller and also allows unit testing the function.

waiter/src/waiter/util/date_utils.clj
```diff
@@ -52,6 +52,18 @@
   (^DateTime [date-str formatter]
     (when date-str (str-to-date date-str formatter))))
 
+(defn str-to-date-ignore-error
+  "Ignores errors when parsing date-str. If there is an error, it will log it and return nil."
+  (^DateTime [date-str]
+   (str-to-date-safe date-str formatter-iso8601))
```
shamsimam, line 58:
```suggestion
   (str-to-date-ignore-error date-str formatter-iso8601))
```

waiter/src/waiter/util/date_utils.clj
```diff
@@ -52,6 +52,18 @@
...
+  (^DateTime [date-str formatter]
+   (try
+     (f/parse
+       (f/with-zone formatter t/utc)
+       date-str)
```
shamsimam, line 63:
```suggestion
     (str-to-date date-str formatter)
```

shamsimam (approved):

PR #1593, opened 2022-06-17
Auto scale services in bypass using external metrics

kevo1ution:

- Calculate "outstanding" request metric using the active-request-count for each instance and waiting-on-available-instance (aka queued request count) for services that have requests bypassing the waiter routers.

- We want to autoscale services based on external metrics for services that have it enabled.

### DIFF
--- a/waiter/integration/waiter/metrics_sync_integration_test.clj
+++ b/waiter/integration/waiter/metrics_sync_integration_test.clj
@@ -477,12 +477,13 @@
 
 (defn assert-scale-to-instances
   "Assert number of instances for a service-id."
-  [routers service-id expected-num-instances]
+  [routers cookies service-id expected-num-instances]
   (is (wait-for
         (fn []
           (every?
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/service_description.clj
```diff
@@ -1772,3 +1772,9 @@
     (fn [service-description]
       (or (empty? param-predicates)
           (every? #(%1 service-description) param-predicates)))))
+
+(defn service-description-bypass-enabled?
+  "Returns true if the provided service description has bypass enabled and false otherwise."
+  [service-desc]
+  (= "true"
+     (get-in service-desc ["metadata" "waiter-proxy-bypass-opt-in"])))
```
DaoWen, line 1780:
```suggestion
  (utils/match-yes-like (get-in service-desc ["metadata" "waiter-proxy-bypass-opt-in"])))
```

PR #1600, opened 2022-07-01
retrieves instance start time from main container status

shamsimam:

- retrieves instance start time from main container status

Reporting incorrect instance start time is confusing when the application container on a pod restarts.


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -469,7 +472,7 @@
                               :log-directory (log-dir-path run-as-user primary-container-restart-count)
                               :port port0
                               :service-id service-id
-                              :started-at pod-started-at
+                              :started-at (or main-container-started-at main-container-last-terminated-at pod-started-at)
```
shamsimam, line 475: A slight improvement over the previous PR to include the last termination time if the container hasn't transitioned to running yet.

DaoWen (approved):

PR #1601, opened 2022-07-04
use waiter labels for pod anti-affinities

DaoWen:

Use `waiter/...` labels for the pod anti-affinity selectors.

The `app` label isn't really something we use in other parts of the api. We actually clobber it with a different value internally. It's better to use the `waiter/cluster` and `waiter/service-hash` fields since they're more likely to be unique and very unlikely to be overwritten in other code.


DaoWen:
Turns out that anti-affinities don't work well when they need to preempt low-priority pods. Looking into node topology constraints instead.

PR #1608, opened 2022-07-15
Remove the metrics consumer in favor of the external metrics endpoint

kevo1ution:

- We no longer support the metrics consumer watches on an external metrics service

- We opt for a push model where metrics are pushed directly to waiter routers with the external metrics endpoint

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -44,9 +44,8 @@
  ;; 2. Have the same :zookeeper :base-path and :discovery-relative-path to allow computing router endpoints
  ;; 3. Have the same :zookeeper :base-path and :leader-latch-relative-path to participate in leadership election
 
- :cluster-config {
-                  ;; Determines if Waiter supports requests bypassing the waiter routers.
-                  :bypass-supported? "false"
+ :cluster-config {;; Determines if Waiter supports requests bypassing the waiter routers.
+                  :bypass-supported? false
 
                   ;; Number of routers required in order for leader election to take place;
                   ;; it's a good idea to set this to an odd number:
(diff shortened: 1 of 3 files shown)


waiter/config-full.edn
```diff
@@ -1124,6 +1105,9 @@
                ;; response headers to include in the request log prefixed by 'response-header-'
                :response-headers #{"content-encoding" "transfer-encoding"}}
 
+ ;; Configure set of strings to tag the Waiter Cluster
+ :tags #{}
```
shamsimam, line 1109: Should this be a map? Also, can we put this under `cluster-config`?

waiter/config-full.edn
```diff
@@ -45,6 +45,9 @@
  ;; 3. Have the same :zookeeper :base-path and :leader-latch-relative-path to participate in leadership election
 
  :cluster-config {
+                  ;; Determines if Waiter supports requests bypassing the waiter routers.
+                  :bypass-supported? "false"
```
shamsimam, line 49: Nitpick: Why is the value not a boolean?

waiter/src/waiter/settings.clj
```diff
@@ -37,7 +37,8 @@
                                     (s/optional-key :ttl) schema/positive-int
                                     s/Keyword schema/require-symbol-factory-fn}
                                    schema/contains-kind-sub-map?)
-   (s/required-key :cluster-config) {(s/required-key :min-routers) schema/positive-int
+   (s/required-key :cluster-config) {(s/required-key :bypass-supported?) schema/non-empty-string
```
shamsimam, line 40: Nitpick: This should be a boolean value

shamsimam (dismissed):

shamsimam (approved):

PR #1612, opened 2022-07-27
Scaling down in two phases for bypass services

kevo1ution:

- Two-phase scale down for pods that are part of services in bypass mode
- Kill instance function for k8 will add `waiter/prepared-to-scale-down-at` annotation to the pod that will be scaled down.
- responder will only returns instances with :prepared-to-scale-down-at if they are ready (otherwise return no instances, as the autoscaler must repick the correct instance to fully scale down).
- Added pod cleaner daemon that hard deletes pods after they have reached a certain timeout

- Allows for safe scale down of instances that may be receiving requests from external load balancers.
- External load balancers need to get updated that the instance is no longer able to receive new traffic. This is done by respecting the `ejected` flag for the instance

## Followups
- Use the external metrics to confirm that an instance has no in-flight requests when it is chosen for killing
- Eagerly kill instances with no outstanding requests
- Sort which instances are chosen to be killed by the least amount of total outstanding requests (right now we rely on current algorithm in the responder for picking an instance to be killed)
- Optimize the scaling up function to attempt to relabel pods that are preparing to scale down so another pod doesn't need to be started

### DIFF
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -815,13 +815,13 @@
  ; instances. Such instances are guaranteed not to be served up as an available instance until sufficient time has
  ; elapsed since their last use. This is referred to as ejecting an instance:
  :ejection-config {
-                   ;; The bypass instance cannot be killed during this period right after being ejected because the instance may still be
-                   ;; handling requests, and those request metrics will have some latency before waiter routers are up to date.
-                   :bypass-grace-buffer-ms 15000
-
                    ;; The minimum amount of time that needs to pass before the instance of a bypass service will be fully deleted after
                    ;; being marked as ejected.
-                   :bypass-max-eject-time-secs 120
+                   :bypass-force-kill-time-ms 120000
+
+                   ;; The bypass instance cannot be killed during this period right after being ejected because the instance may still be
+                   ;; handling requests, and those request metrics will have some latency before waiter routers are up to date.
+                   :bypass-grace-kill-time-ms 15000
 
                    ;; Erroneous instances are ejected using an exponential delay based on the number of successive
                    ;; failures and eject-backoff-base-time-ms (milliseconds):
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -526,15 +526,15 @@
                                  :pod-base-port 31000
 
                                  ;; Determines the interval the pod cleanup daemon should check for pods in phase 1 of safe scale down
-                                 ;; to phase 2 where the pod is deleted.
+                                 ;; and hard delete any killable pods.
                                  :pod-cleanup-interval-ms 5000
 
-                                 ;; The minimum amount of time that needs to pass before the pod cleanup daemon can considering deleting
+                                 ;; The minimum amount of time that needs to pass before the pod cleanup daemon can consider deleting
                                  ;; the pod. This allows for some lingering metrics from raven to be tracked by waiter routers, before knowing
                                  ;; if a pod is no longer serving any active requests.
                                  :pod-cleanup-grace-buffer-ms 15000
 
-                                 ;; The pod daemon will delete the pod after this timeout is reached even if it is still serving requests
+                                 ;; The pod daemon will delete the pod after this timeout is reached even if it is still serving requests.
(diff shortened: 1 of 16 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -948,16 +967,34 @@
...
+            desc (service-id->service-description-fn service-id)
+            bypass-enabled? (sd/service-description-bypass-enabled? desc)
+            first-phase? (and bypass-enabled? (nil? prepared-to-scale-down-at))
+            second-phase? (and bypass-enabled? (some? prepared-to-scale-down-at))
+            kill-result (cond first-phase?
+                              (do
+                                (mark-pod-for-scale-down this instance)
+                                {:marked-prepared-for-scale-down true
+                                 :message "Successfully annotated pod to be prepared for scale down"})
+                              second-phase?
```
shamsimam, line 982: Don't we also need to check whether time has elapsed?

waiter/src/waiter/state/maintainer.clj
```diff
@@ -163,7 +164,8 @@
                                    update-state-chan (retrieve-channel (get service-id->channel-map'' service-id) :update-state)]
                                (if (or healthy-instances unhealthy-instances)
                                  (async/put! update-state-chan
-                                             (let [update-state {:healthy-instances healthy-instances
+                                             (let [update-state {:all-healthy-instances all-healthy-instances
+                                                                 :healthy-instances healthy-instances
```
shamsimam, line 168: Why would all healthy instances be different from healthy instances?

shamsimam:
Where do we use the external metrics to confirm that an instance has no in-flight requests when it is chosen for killing?

waiter/src/waiter/scaling.clj
```diff
@@ -178,20 +188,21 @@
...
+            (or (async/poll! response-chan) {:message :no-instance-killed, :status http-404-not-found})]
+        (cond
+          killed? (cid/cinfo correlation-id "killed instance" instance-id)
+          marked-prepared-for-scale-down (cid/cinfo correlation-id "marked instance for scale down" instance-id)
+          :else (cid/cinfo correlation-id "unable to kill instance" kill-response))
         (-> (utils/clj->json-response {:kill-response kill-response
                                        :service-id service-id
                                        :source-router-id src-router-id
-                                       :success instance-killed?}
+                                       :success success?}
```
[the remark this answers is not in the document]
kevo1ution: I didn't add because it is included in the nested object `kill-response`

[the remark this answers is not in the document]
kevo1ution: ie. `.kill-response.marked-prepared-for-scale-down`

waiter/src/waiter/scheduler/kubernetes.clj
kevo1ution: This is done in the `find-killable-instance` in the `responder`. I did it this way because it didn't make sense to return an instance that was not killable based the timeout not getting reached in that fn.

kevo1ution:
> Where do we use the external metrics to confirm that an instance has no in-flight requests when it is chosen for killing?

I did not do that here yet, it will be in a follow-up PR to keep this PR more concise.

shamsimam:
> > Where do we use the external metrics to confirm that an instance has no in-flight requests when it is chosen for killing?
> 
> I did not do that here yet, it will be in a follow-up PR to keep this PR more concise.

Okay adding that to follow-ups above.

waiter/config-full.edn
```diff
@@ -815,6 +815,14 @@
  ; instances. Such instances are guaranteed not to be served up as an available instance until sufficient time has
  ; elapsed since their last use. This is referred to as ejecting an instance:
  :ejection-config {
+                   ;; The minimum amount of time that needs to pass before the instance of a bypass service will be fully deleted after
```
DaoWen, line 818: Is this supposed to say "maximum" rather than "minimum"?

kevo1ution:
Closing in favor of https://github.com/twosigma/waiter/pull/1653 where we use the k8s `preStop` command instead for delaying termination of the pod and allowing requests to drain.

PR #1616, opened 2022-08-04
Add helper options to make-request for testing

kevo1ution:

- adding helper options to make-request and other client_tools functions

- we want to use these options in integration tests

### DIFF
--- a/waiter/src/waiter/util/client_tools.clj
+++ b/waiter/src/waiter/util/client_tools.clj
@@ -1200,11 +1200,11 @@
 (defn post-token
   "Sends a POST request with the given token definition"
   [waiter-url {:keys [token] :as token-map} &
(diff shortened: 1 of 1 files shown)


waiter/src/waiter/util/client_tools.clj
```diff
@@ -1196,11 +1200,11 @@
 (defn post-token
   "Sends a POST request with the given token definition"
   [waiter-url {:keys [token] :as token-map} &
-   {:keys [cookies headers query-params] :or {cookies [] headers {} query-params {}}}]
+   {:keys [cookies headers query-params use-host-header] :or {cookies [] headers {} query-params {} use-host-header true}}]
```
shamsimam, line 1203: Instead of the `use-host-header` flag, can we not require the client to pass in the headers? I think Waiter will prefer `x-waiter-token` header over host header.

shamsimam (approved):

PR #1617, opened 2022-08-08
Drain mode on Waiter Router

kevo1ution:

- Adds new endpoints for managing the drain process of a waiter router
- `POST /drain?drain-timeout-ms=1000&crash=true` control whether the router should crash at the end of the drain and how long to report unhealthy on the `/status` endpoint for the waiter router to let it drain.
- `GET /drain` will provide the current drain status (whether drain mode is enabled and if a crash will be triggered at the end)

- We may need to drain the waiter router. Load balancers external to the router should assume that if the router is not reporting 200 on `/status`, the router should not be sent any new traffic.

### DIFF
--- a/waiter/src/waiter/handler.clj
+++ b/waiter/src/waiter/handler.clj
@@ -1200,7 +1200,7 @@
                 (when crash-process?
                   (log/warn "Going to attempt to kill router process after timeout." {:drain-timeout-secs drain-timeout-secs})
                   (async/go
-                    (async/<! (async/timeout drain-timeout-secs))
+                    (async/<! (async/timeout (* 1000 drain-timeout-secs)))
                     (if (:crash-process? @drain-atom)
                       (crash-fn)
                       (log/warn "Cancelled attempt to kill waiter process!"))))


waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
...
-   Includes representation of request if requested using the include=request-info query param."
-  [{:keys [body trailers-fn] :as request}]
+   Includes representation of request if requested using the include=request-info query param.
+   If the router is in drain mode then return 503 server error."
+  [drain-mode?-fn {:keys [body trailers-fn] :as request}]
   (try
     (when (instance? InputStream body)
       (log/info "consuming request body before rendering response")
       (slurp body))
+    (when (drain-mode?-fn)
```
shamsimam, line 1221: We should report [503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)

shamsimam: This will be a frequently used endpoint.

waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
   [request]
   (utils/exception->response (ex-info (utils/message :not-found) {:log-level :info :status http-404-not-found}) request))
 
+(defn drain-handler
+  "Handles Get /drain requests by providing the current drain state.
+  Handles Post /drain requests and sets the drain-until for the router based on the provided 'drain-timeout-ms' query
```
DaoWen, line 1179: I feel like this timeout should be in seconds rather than milliseconds.

I can't think of a situation where we'd want milliseconds here, and it seems easy to accidentally set this thinking it's seconds instead, and interrupt all of the active connections.

waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
...
+                                      :drain-mode? (drain-mode?-fn)})
+      :post (let [{:strs [drain-timeout-ms] :or {drain-timeout-ms "120000"} :as request-params} (-> request ru/query-params-request :query-params)
+                  auth-user (get request :authorization/user)
+                  drain-timeout-ms (utils/parse-int drain-timeout-ms)
+                  crash-process? (utils/request-flag request-params "crash")
+                  now (clock)
+                  drain-until (t/plus now (t/millis drain-timeout-ms))]
+              (when (not (admin-user?-fn auth-user))
+                (throw (ex-info "Must be an admin to use this endpoint."
+                                {:auth-user auth-user :log-level :info :status http-401-unauthorized})))
```
DaoWen, line 1197: Should use `403 Forbidden` (indicates that authorization failed), not `401` (indicates that authentication failed).

waiter/src/waiter/core.clj
```diff
@@ -1575,6 +1582,15 @@
...
                                     (fn display-settings-handler-fn [_]
                                       (settings/display-settings settings))))
+   :drain-handler-fn (pc/fnk [[:routines admin-user?-fn]
+                              [:state clock drain-atom drain-mode?-fn]
+                              wrap-secure-request-fn]
+                       (wrap-secure-request-fn
+                         (fn drain-handler-fn [request]
+                           (let [crash-fn (fn crash-fn []
+                                            (log/fatal "Drain timeout finished. Kill waiter process now!")
+                                            (System/exit 1))]
```
DaoWen, line 1592: This isn't really an error. Should this be `exit 0` instead?

waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
...
+              (log/warn "Putting router into drain mode! It should begin failing health checks.")
+              (swap! drain-atom assoc :drain-until drain-until :crash-process? crash-process?)
```
DaoWen, line 1199: What happens when there's already a drain in process?

waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
...
+              (when crash-process?
+                (log/warn "Going to attempt to kill router process after timeout." {:drain-timeout-ms drain-timeout-ms})
+                (async/go
+                  (async/<! (async/timeout drain-timeout-ms))
+                  (if (:crash-process? @drain-atom)
```
DaoWen, line 1204: This logic means that if the `:crash-process?` is changed between when this async thread is set up and when the timeout finishes, then the behavior changes. I don't think this is what we want.

waiter/src/waiter/handler.clj
```diff
@@ -1174,14 +1174,52 @@
...
+                    (crash-fn)
+                    (log/warn "Cancelled attempt to kill waiter process!"))))
+              (utils/clj->json-response {:result @drain-atom}))
```
DaoWen, line 1207: Similar to above. We don't want to return a different result here if the drain-atom changed (racing). We should store the swapped-in value in a `let` or something, and use that result here.

waiter/src/waiter/handler.clj
DaoWen: You even said `503` in the comment... 😞

waiter/test/waiter/core_test.clj
```diff
@@ -913,13 +913,22 @@
...
       (is (= {"status" "ok"} (json/read-str body))))))
 
+(deftest test-health-check-handler-handler-drain-mode
+  (testing "health-check-handler:status-ok"
+    (let [request {:request-method :get, :uri "/status"}
+          waiter-request?-fn (fn [_] true)
+          handlers {:status-handler-fn ((:status-handler-fn request-handlers) {:state {:drain-mode?-fn (constantly true)}})
+                    :waiter-request?-fn (constantly true)}
+          {:keys [status]} ((ring-handler-factory waiter-request?-fn handlers) request)]
+      (is (= http-500-internal-server-error status)))))
```
DaoWen, line 930: Should be `503`

DaoWen (changes requested):

waiter/src/waiter/handler.clj
shamsimam: We should reflect the unit in the query parameter name.

waiter/src/waiter/handler.clj
DaoWen: That makes sense. You explained that in the docstring too. I should have read that more closely.

DaoWen (approved):

PR #1634, opened 2022-08-25
adds support for configuring the byte buffer pool used in http clients

shamsimam:

- adds support for configuring the byte buffer pool used in http clients

We wish to limit the size of the underlying MappedByteBufferPool in the jetty http clients.

Uses changes from https://github.com/twosigma/jet/pull/60

### DIFF
--- a/waiter/src/waiter/settings.clj
+++ b/waiter/src/waiter/settings.clj
@@ -71,7 +71,7 @@
                                                   (s/required-key :async-request-max-timeout-ms) schema/positive-int
                                                   (s/required-key :async-request-timeout-ms) schema/positive-int
                                                   (s/required-key :byte-buffer-capacity-factor) (s/maybe schema/non-negative-int)
-                                                  (s/required-key :byte-buffer-heap-percent) (s/maybe schema/positive-num)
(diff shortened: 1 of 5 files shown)


waiter/src/waiter/settings.clj
```diff
@@ -70,6 +70,9 @@
                                                   (s/required-key :async-request-max-status-checks) schema/positive-int
                                                   (s/required-key :async-request-max-timeout-ms) schema/positive-int
                                                   (s/required-key :async-request-timeout-ms) schema/positive-int
+                                                  (s/required-key :byte-buffer-capacity-factor) (s/maybe schema/non-negative-int)
+                                                  (s/required-key :byte-buffer-heap-percent) (s/maybe schema/positive-num)
```
napple, line 74: Should this be limited to (0,100]? Or some other hard max less than 100?

shamsimam: sgtm, will change. Actually need to be clearer that this is in the range (0, 1]

shamsimam: Changed to enforce byte-buffer-heap-percent is a positive fraction less than or equal to 1

napple (approved):

DaoWen (approved):

PR #1651, opened 2022-09-06
retry replicaset create on 5XX error from k8s

DaoWen:

retry replicaset create on 5XX error from k8s

k8s can return a 5XX error for some transient issues (e.g., etcd is busy)

adding a couple of retries will make waiter more resilient to that type of temporary k8s api flakiness

### DIFF
--- a/waiter/src/waiter/scheduler/kubernetes.clj
+++ b/waiter/src/waiter/scheduler/kubernetes.clj
@@ -814,6 +814,12 @@
   [{:keys [replicaset-spec-builder-fn] :as scheduler} service-id service-description context]
   (replicaset-spec-builder-fn scheduler service-id service-description context))
 
+(defn- server-error?
+  "Checks if the given exception object corresponds to a 5XX response indicating a server error."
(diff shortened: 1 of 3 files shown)


waiter/src/waiter/scheduler/kubernetes.clj
```diff
@@ -814,6 +814,12 @@
   [{:keys [replicaset-spec-builder-fn] :as scheduler} service-id service-description context]
   (replicaset-spec-builder-fn scheduler service-id service-description context))
 
+(defn- server-error?
+  "Checks if the given exception object corresponds to a 5XX response indicating a server error."
+  [maybe-exception]
+  (when-let [status (-> maybe-exception ex-data :status)]
+    (<= 500 status 599)))
```
shamsimam, line 821: Can we use `status-codes.http-500-internal-server-error`?
Nitpick: can we move this function to http-utils.clj?

shamsimam (approved):
Should retries limited only to create-service?

PR #1652, opened 2022-09-07
promotes common request timeout headers to token parameters

shamsimam:

- promotes common request timeout headers to token parameters

In some scenarios, services want different defaults for request timeouts than those provided by Waiter. Promoting these timeouts to such configurable token parameters makes such use cases possible.

### Implementation notes

- Implementation continues to be backward compatible as service ID does not change when timeouts are configured
- Values provided in request headers continue to be used with the highest priority.

### DIFF
--- a/waiter/src/waiter/util/http_utils.clj
+++ b/waiter/src/waiter/util/http_utils.clj
@@ -286,3 +286,10 @@
   [user-agent-products request]
   (when-let [user-agent (some-> request (get-in [:headers "user-agent"]) str/lower-case)]
     (some #(str/includes? user-agent %) user-agent-products)))
+
+(defn merge-response-headers
+  "Attaches provided headers into the response."
+  [response kvm]
+  (->> kvm
+       (pc/map-vals str)
+       (update response :headers merge)))
--- a/waiter/config-full.edn
+++ b/waiter/config-full.edn
@@ -884,8 +884,7 @@
                                :output-buffer-size 4096
 
                                ;; The default amount of time (milliseconds) each request will wait in
-                               ;; the Waiter queue before an instance is available to process it. This
-                               ;; can be overriden in the service description:
+                               ;; the Waiter queue before an instance is available to process it.
                                :queue-timeout-ms 300000
 
                                ;; Configures the idle timeout (milliseconds) in the response output stream:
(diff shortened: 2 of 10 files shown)


waiter/src/waiter/util/http_utils.clj
```diff
@@ -286,3 +286,12 @@
   [user-agent-products request]
   (when-let [user-agent (some-> request (get-in [:headers "user-agent"]) str/lower-case)]
     (some #(str/includes? user-agent %) user-agent-products)))
+
+(defn assoc-response-headers
+  "Attaches provided headers into the response."
+  [response & kvs]
+  (loop [headers (get response :headers)
+         [k v & rem-hvs] kvs]
+    (if (str/blank? k)
```
DaoWen, line 295: Is this a typo? Did you mean to filter out blank values rather than blank keys? Only asking because the way you use this, it looks like you might end up with blank values, but there's no way the literal strings you pass in could be blank.

```suggestion
    (if (str/blank? v)
```

waiter/src/waiter/util/http_utils.clj
```diff
@@ -286,3 +286,12 @@
...
+      (assoc response :headers headers)
+      (recur (assoc headers k (str v)) rem-hvs))))
```
DaoWen, line 297: Whenever you're tempted to use `loop` in Clojure, there's almost always a more idiomatic/declarative/functional way to do it instead. Unless you're implementing something really low-level — e.g., a new type of collection transformation — then I think it's always preferable to use higher-level functions instead.

```suggestion
(defn assoc-response-headers
  "Attaches provided headers into the response.
   Headers with blank values are excluded."
  [response & kvs]
  (let [xf (comp (map vec)
                 (remove (fn [[k v]] (str/blank? v))))
        kv-pairs (partition 2 kvs)]
    (update response :headers into xf kv-pairs)))
```

DaoWen (changes requested):

waiter/src/waiter/util/http_utils.clj
shamsimam: This is intentional, blank keys trigger termination of the loop, perhaps it is better to check as `nil?`

waiter/src/waiter/util/http_utils.clj
shamsimam: This lgtm, will rewrite.

DaoWen: I totally misunderstood the blank thing. This should be much simpler:

```
(defn assoc-response-headers
  "Attaches provided headers into the response."
  [response & kvs]
  (let [kv-pairs (partition 2 kvs)]
    (update response :headers into (map vec) kv-pairs)))
```

shamsimam: Rewritten to use merge.

DaoWen (approved):

PR #1660, opened 2022-09-15
reports 403 error on OIDC too many cookies

shamsimam:

- reports 403 error on OIDC too many cookies

Makes the error for an unauthenticated OIDC request clearer.

### DIFF
--- a/waiter/src/waiter/auth/oidc.clj
+++ b/waiter/src/waiter/auth/oidc.clj
@@ -326,29 +326,37 @@
     (log/info "request has" num-challenge-cookies "oidc challenge cookies")
     (> num-challenge-cookies num-allowed)))
 
+(defn- promote-401-to-403?
+  "Determines whether the response represents a 401 from a downstream handler which attempted no authentication."
+  [response]
+  ;; when downstream auth is disabled, the 401 response indicates OIDC auth handler is authoritative in
+  ;; terms of the 4XX authentication response that is generated. We can rewrite the 401 response to a
+  ;; 403 response to indicate an authentication error due to too many cookies.
+  (and (map? response)
+       (let [{:keys [waiter/auth-disabled? status]} response]
+         (and auth-disabled?
+              (= status http-401-unauthorized)
+              (utils/waiter-generated-response? response)))))
+
 (defn make-oidc-auth-too-many-cookies-response-updater
   "Returns a response updater that rewrites 401 waiter responses to 302 redirects."
   [request num-allowed]
   (fn update-oidc-auth-too-many-cookies-response
-    [{:keys [waiter/auth-disabled? status] :as response}]
-    (let [num-challenge-cookies (count-challenge-cookies request)]
-      (cond-> response
-        ;; when downstream auth is disabled, the 401 response indicates OIDC auth handler is authoritative in
-        ;; terms of the 4XX authentication response that is generated. We can rewrite the 401 response to a
-        ;; 403 response to indicate an authentication error due to too many cookies.
-        (and auth-disabled?
-             (= status http-401-unauthorized)
-             (utils/waiter-generated-response? response))
-        (merge (-> {:message (str "Too many OIDC challenge cookies (allowed: "
-                                  num-allowed ", provided: " num-challenge-cookies ")")
-                    :status http-403-forbidden}
-                   (utils/data->error-response request)
-                   (cookies/cookies-response)))))))
+    [response]
+    (cond-> response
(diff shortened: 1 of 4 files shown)


waiter/src/waiter/auth/oidc.clj
```diff
@@ -308,18 +309,41 @@
...
+(defn make-oidc-auth-too-many-cookies-response-updater
+  "Returns a response updater that rewrites 401 waiter responses to 302 redirects."
+  [request num-allowed]
+  (fn update-oidc-auth-too-many-cookies-response
+    [{:keys [waiter/auth-disabled? status] :as response}]
+    (let [num-challenge-cookies (count-challenge-cookies request)]
+      (cond-> response
+        (and auth-disabled?
+             (= status http-401-unauthorized)
+             (utils/waiter-generated-response? response))
```
DaoWen, line 338: I don't understand what's going on here. We probably need a comment explaining why we need these checks.

shamsimam: Discussed offline, updated with comments.

waiter/src/waiter/auth/oidc.clj
```diff
@@ -328,15 +352,21 @@
    request-handler]
   (let [oidc-authority (utils/uri-string->host oidc-authorize-uri)]
     (fn oidc-auth-handler [request]
-      (let [oidc-mode-delay (delay (request->oidc-mode allow-oidc-auth-api? allow-oidc-auth-services? oidc-default-mode request))]
+      (let [oidc-mode-delay (delay (request->oidc-mode allow-oidc-auth-api? allow-oidc-auth-services? oidc-default-mode request))
+            skip-oidc? (or (auth/request-authenticated? request)
+                           (= :disabled @oidc-mode-delay)
+                           ;; OIDC auth is no-op when request cannot be redirected
+                           (not (supports-redirect? oidc-authority oidc-redirect-user-agent-products request)))
+            many-oidc-challenge-cookies? (delay (too-many-oidc-challenge-cookies? request oidc-num-challenge-cookies-allowed-in-request))]
```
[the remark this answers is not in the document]
shamsimam: Will do.

waiter/src/waiter/auth/oidc.clj
```diff
@@ -328,20 +358,28 @@
...
 
-          :else
-          (ru/update-response
-            (request-handler request)
-            (make-oidc-auth-response-updater jwt-auth-server @oidc-mode-delay oidc-same-site-attribute password request)))))))
+              :else
+              (ru/update-response
+                (request-handler request)
+                (make-oidc-auth-response-updater jwt-auth-server @oidc-mode-delay oidc-same-site-attribute password request))))
+          ;; do not let the downstream waiter/auth-disabled? on the response flow out of the OIDC auth handler
```
DaoWen, line 381: I believe that "comments should describe *why* rather than *what*" is a good philosophy. Code should be self-documenting as to _what_ it's doing, as I think is the case here. What is unclear to me is _why_ we don't want this property to flow out of the OIDC auth handler.

waiter/src/waiter/auth/oidc.clj
```diff
@@ -328,20 +358,28 @@
...
+                               (= :disabled @oidc-mode-delay)
+                               ;; OIDC auth is no-op when request cannot be redirected
+                               (not (supports-redirect? oidc-authority oidc-redirect-user-agent-products request)))
+                many-oidc-challenge-cookies? (delay (too-many-oidc-challenge-cookies? request oidc-num-challenge-cookies-allowed-in-request))]
+            (cond
+              (or skip-oidc?
+                  ;; OIDC auth is avoided if client already has too many challenge cookies
+                  @many-oidc-challenge-cookies?)
+              (cond-> (request-handler request)
+                ;; update a 401 response to 403 if there were too many cookies and downstream auth options are disabled
```
DaoWen, line 372: Probably good to document here _why_ we return 403 even though 401 is technically a better fit.

DaoWen (approved):

waiter/src/waiter/auth/oidc.clj
shamsimam: Improved the comment with why we want it removed - to remove confusion in debugging in other handlers.

waiter/src/waiter/auth/oidc.clj
shamsimam: Improved comment on why 403 is a more appropriate response.

DaoWen (approved):

PR #1665, opened 2022-09-24
fixes too many cookies response code

shamsimam:

- fixes too many cookies response code

The test is failing after the recent 401/403 too many cookies change


blevz (approved):
Lgtm

PR #1667, opened 2022-10-03
cli tests: ping_token should check the nested response status

DaoWen:

`ping_token` util function should check the nested response status

`/waiter-ping` is supposed to always return 200 ok, even when the backend fails.

what we actually want to do here is make sure that the nested ping request was also successful.


DaoWen:
I thought this code was hitting the `/waiter-ping` endpoint. It is not. This patch doesn't work since the response is actually the `Hello world` coming directly from kitchen.

PR #1687, opened 2022-11-28
logs and metrics for k8s api 5XX responses

DaoWen:

Add logs and metrics for k8s api 5XX responses

Improve visibility into k8s api errors


blevz (approved):
lgtm

PR #1690, opened 2022-12-10
Add ensure-ssl cli command

blevz:

- Adds a command to check SSL for a provided token

Provide a common method for clients to ensure that SSL has been set up for newly created tokens.  Necessary for CI/CD improvements

### DIFF
--- a/cli/waiter/action.py
+++ b/cli/waiter/action.py
@@ -90,6 +90,18 @@ def ping_on_cluster(cluster, timeout, wait_for_request, token_name, service_exis
             print(f'Service is currently {format_status(service_status)}.')
     return succeeded
 
+def check_ssl(token_name, timeout_seconds):
+    """Returns true if a request to the token's DNS name doesn't encounter an SSL error"""
+    timeout_millis = timeout_seconds * 1000
+    try:
+        http_util.__get(f"https://{token_name}", read_timeout=timeout_millis)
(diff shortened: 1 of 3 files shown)


cli/waiter/action.py
```diff
@@ -90,6 +90,18 @@ def check_service_status():
             print(f'Service is currently {format_status(service_status)}.')
     return succeeded
 
+def check_ssl(token_name, timeout_seconds):
+    """Returns true if a request to the token's DNS name doesn't encounter an SSL error"""
+    timeout_millis = timeout_seconds * 1000
+    try:
+        http_util.__get(f"https://{token_name}", read_timeout=timeout_millis)
```
shamsimam, line 97: Should we log the response for debugging purposes?

blevz: What would the contents of the response help with?  We only care about whether the connection can be established and a ssl handshake completed
Logging the response would only happen when the command exits successfully
When the command fails it never receives a response (because it throws trying to get it)

cli/waiter/action.py
```diff
@@ -90,6 +90,15 @@ def check_service_status():
...
+    except requests.exceptions.SSLError:
```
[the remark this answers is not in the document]
blevz: Modified to catch any exception, if we cannot receive a response then ssl is not ready

shamsimam (approved):

PR #1691, opened 2022-12-15
reference updates support on tokens endpoint

shamsimam:

- adds a new `reference-update-chan` which reference tracking components can send update notifications on
- adds a new `/reference/notify` endpoint to receive and process reference update notifications from peers
- updates the service description builder to calculate references from the token definition
- updates the token index (can be triggered by `/tokens/reindex` endpoint) to include references in the token metadata/index
- adds support for handling reference update notifications by extending builder interface
- updates the token watch component to listen for reference updates and trigger events on the tokens watches when reference updates for tokens.

We would like to be able to propagate new service IDs for tokens when underlying references for a token update.

### DIFF
--- a/waiter/test/waiter/scheduler_test.clj
+++ b/waiter/test/waiter/scheduler_test.clj
@@ -53,6 +53,10 @@
     (locking lock
       (f))))
 
+(defn- extract-reference-type->reference-name-helper
+  [_]
+  {})
+
 (deftest test-record-Service
   (let [test-instance-1 (->Service "service1-id" 100 100 {:running 0, :healthy 0, :unhealthy 0, :staged 0})
         test-instance-2 (make-Service {:id "service2-id" :instances 200 :task-count 200})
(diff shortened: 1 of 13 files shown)


DaoWen (approved):

waiter/test/waiter/scheduler_test.clj
```diff
@@ -53,6 +53,10 @@
     (locking lock
       (f))))
 
+(defn- extract-reference-type->reference-name-fn
```
DaoWen, line 56: Nitpick: we don't usually give `defn` symbols names ending in `-fn`.

waiter/test/waiter/scheduler_test.clj
```diff
@@ -53,6 +53,10 @@
...
+  [_]
+  {})
```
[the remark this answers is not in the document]
shamsimam: Renamed.

waiter/test/waiter/scheduler_test.clj
shamsimam: Renamed.

DaoWen (approved):

PR #1693, opened 2022-12-30
refactors instance selection logic in CLI ssh

shamsimam:

- moves instance selection logic in CLI to helper functions

We are planning CLI subcommands that will perform management operations on service instances. Making the instance selection logic reusable via helper functions will simplify implementing those subcommands.


cli/waiter/instance_select.py
```diff
@@ -0,0 +1,122 @@
+from enum import Enum
```
blevz, line 1: The change in this file appear to be a move + file changes, any chance we can do those changes separately?  Will make reviewing the changes much easier

cli/waiter/cli.py
```diff
@@ -3,7 +3,7 @@
 from urllib.parse import urlparse
 
 from waiter import configuration, http_util, metrics, version
-from waiter.subcommands import create, delete, ssl, init, kill, maintenance, ping, show, ssh, tokens, update
+from waiter.subcommands import create, delete, init, kill, maintenance, ping, show, ssh, ssl, tokens, update
```
blevz, line 6: We should separate formatting changes from functional changes

blevz (changes requested):

cli/waiter/cli.py
shamsimam: Happy to move this to a separate PR if you prefer, it is a leftover from my attempt at cleaning the capstone branch and this showed up as a merge conflict.

cli/waiter/instance_select.py
shamsimam: The PR moves some common functionality from ssh.py to this new file o make them reusable and as a result, includes some changes to the ssh function calls.
Can you point them out? it will make it easier for me to fix them.

blevz (approved):

cli/waiter/instance_select.py
shamsimam: Moved the import change to a different PR: https://github.com/twosigma/waiter/pull/1697

cli/waiter/cli.py
shamsimam: Moved the import change to a different PR: #1697

PR #1701, opened 2023-01-11
makes the reserve instance channel buffer size dynamically configurable

shamsimam:

- adds error class to back pressure response
- makes the reserve instance channel buffer size dynamically configurable

Services that run with high concurrency or high max instances can end up inadvertently triggering the back-pressure mechanism when there is an influx of requests.


waiter/src/waiter/service.clj
```diff
@@ -239,9 +239,10 @@
            (metrics/service-timer ~service-id "reserve-instance")
            (when-not (au/offer! service-chan# [~reason-map instance-resp-chan# ~exclude-ids-set ~timeout-in-millis])
              (throw (ex-info "Unable to request an instance."
-                             {:status http-503-service-unavailable
+                             {:error-class error-class-reserve-back-pressure
+                              :reason-map ~reason-map
                               :service-id ~service-id
-                              :reason-map ~reason-map})))
+                              :status http-503-service-unavailable})))
```
shamsimam, line 245: This change is not directly related to the PR, but helps us easily identify when the back pressure mechanism is being triggered.

PR #1711, opened 2023-01-26
adds api support for expiring an instance

shamsimam:

- adds api support for expiring an instance

We have had the need to expire instances, e.g. due to the node/container running the pod going bad. Exposing the ability to expire such instances in the API enables users to expire their instances when they detect such bad instances.

### DIFF
--- a/waiter/test/waiter/core_test.clj
+++ b/waiter/test/waiter/core_test.clj
@@ -191,16 +191,13 @@
                         (if (instance? Throwable result)
                           (throw result)
                           result))))
-        configuration {:daemons {:populate-maintainer-chan! {:query-state-fn (constantly {})}
-                                 :router-state-maintainer {:maintainer {:notify-instance-killed-fn (constantly {})}}}
-                       :routines {:peers-acknowledged-eject-requests-fn (constantly true)
-                                  :allowed-to-manage-service?-fn (constantly true)
-                                  :service-id->service-description-fn (constantly {})}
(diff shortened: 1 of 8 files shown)


waiter/test/waiter/core_test.clj
```diff
@@ -191,16 +191,13 @@
...
-                                 :router-state-maintainer {:maintainer {:notify-instance-killed-fn (constantly {})}}}
-                       :routines {:peers-acknowledged-eject-requests-fn (constantly true)
-                                  :allowed-to-manage-service?-fn (constantly true)
-                                  :service-id->service-description-fn (constantly {})}
-                       :scheduler {:scheduler scheduler}
-                       :state {:scaling-timeout-config scaling-timeout-config
-                               :scheduler-interactions-thread-pool scheduler-interactions-thread-pool}
-                       :wrap-ignore-disabled-auth-fn utils/wrap-identity
-                       :wrap-secure-request-fn utils/wrap-identity}
+        notify-instance-killed-fn (constantly {})
```
juandipg, line 194: do we normally name functions `-fn`?

juandipg (approved):

waiter/test/waiter/core_test.clj
shamsimam: Yes, we generally add the `-fn` suffix for local variables representing function, you can also it above with `wrap-secure-request-fn`. 
But there isn't a lot consistency around it in the codebase and there are other occurrences without that suffix :(

